PR #2, opened 2018-05-31
Nrfjprog: add --erase

mbolivar:
This is based on #1 to make merging easier. Just the final commit is what matters.

Straightfoward port of https://github.com/zephyrproject-rtos/zephyr/pull/6809 to west.

CC: @SebastianBoe, @nvlsianpu

### DIFF
--- a/runner/nrfjprog.py
+++ b/runner/nrfjprog.py
@@ -13,11 +13,12 @@ from .core import ZephyrBinaryRunner, RunnerCaps
 class NrfJprogBinaryRunner(ZephyrBinaryRunner):
     '''Runner front-end for nrfjprog.'''
 
-    def __init__(self, cfg, family, softreset):
+    def __init__(self, cfg, family, softreset, erase=False):
         super(NrfJprogBinaryRunner, self).__init__(cfg)
         self.hex_ = cfg.kernel_hex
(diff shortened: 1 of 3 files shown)


mbolivar:
@jhedberg this restores symmetry in the flashing behavior which you've previously reported to boards using nrjfprog.

runner/nrfjprog.py
```diff
@@ -76,14 +80,26 @@ def get_board_snr_from_user(self):
...
+                 '--eraseall',
+                 '-f', self.family,
+                 '--snr', board_snr],
+                program_cmd
+                ])
+        else:
+            if self.family == 'NRF51':
+                commands.append(program_cmd + ['--sectorerase'])
+            else:
+                commands.append(program_cmd + ['--sectoranduicrerase'])
```
carlescufi: I was going to say that we shouldn't erase the UICR here. But looking at our current soc.c, it actually writes to the UICR during startup, so this could lead to that failing. Until we change that soc.c this is correct.

mbolivar: I think this fragment may have come from you actually :)

https://github.com/zephyrproject-rtos/zephyr/pull/6809#issuecomment-376461275

PR #5, opened 2018-06-09
Support for bootstrapping west

mbolivar:
This pull request adds a bootstrap script, and makes that what would get uploaded to PyPI.

Not done:

1. allow reinitializing an already initialized directory (to get another manifest, manifest branch, or reinitialize the west repository)
2. macOS, Windows testing
3. actually push to PyPI (I tested this using `pip install -e` as described in the README)
4. auto-updating west itself

I think that 1. and 2. should block 3., but that 4. ought to be done within west itself, since it's more complicated. Once 1. is done and we're happy with the cross platform support I'm hoping (fingers crossed here!) that we won't have to touch the bootstrap script hardly ever.

One (I think) nice behavior of this script is that it works like "git init" rather than "repo init" in that you can optionally specify the directory to bootstrap into, and the tool will make it if it doesn't exist. This saves a mkdir in the directions.

CC: @ulfalizer
CC: @tejlmand

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,13 +1,72 @@
 This package contains the Zephyr RTOS meta tool, 'west'.
 
-To install west from this package into your user directory, run:
+WARNING
+-------
 
-$ python3 setup.py install --user
+DO NOT INSTALL WITH "python3 setup.py install".
 
-You can also try it out by just running the package as a module from
-this directory:
+Please use pip to install in development mode as documented below.
+
+Important Note
+--------------
+
+West is distributed in two pieces:
+
+1. A bootstrap/wrapper script, which is distributed via PyPI.
+2. The "main" west package and entry points, which are fetched by the
+   bootstrap script.
+
+This somewhat unusual arrangement is because:
+
+- One of West's jobs is to manage interaction with Zephyr's multiple
+  Git repositories, including its own.
+- West is in its experimental stages and is moving quickly, meaning
+  users need to stay on HEAD.
+
+The default setup.py installs the **wrapper script only**.
+
+Installation
+------------
+
+To install the West bootstrapping/wrapper script from this package in
+development mode, clone this repository and run this from the top
+level directory:
+
+$ pip3 install -e .
+
+Then use the wrapper script to initialize a Zephyr installation with::
+
+  $ west init your/zephyr/install-dir
+  $ cd your/zephyr/install-dir
+  $ west <command>
+
+The ``west init`` call will:
+
+- create your/zephyr/install-dir
+- clone the Zephyr manifest repository (whose URL can be overridden
+  with the ``-u`` option, and branch with ``--mr`` / ``--manifest-rev``)
+- clone the latest West repository (the URL override is ``-w``, and
+  revision/branch override is ``--wr`` / ``--west-rev``)
+
+Running ``west <command>`` from :file:`your/zephyr/install-dir` or
+underneath it will invoke west in "wrapper" mode: any commands other
+than ``init`` will be delegated to the West tree pulled by ``west
+init``.
+
+This arrangement may seem familiar to Android (platform, not app)
+developers. The source management features of West were indeed
+inspired by the Android Repo tool's features, but West makes
+significant departures from Repo's behavior.
+
+Alternative Usage
+-----------------
+
+If you don't want to change your system outside of cloning this
+repository, you can also clone West's Git repository and run the
+package as a module:
 
 $ python3 -m west
 
-If you go this route, make sure to install the requirements as
-specified by the install_requires line in setup.py.
+Only do this if you know what you're doing (and make sure to install
+the requirements as specified by the install_requires line in
+setup.py).
--- a/bootstrap/main.py
+++ b/bootstrap/main.py
@@ -1,0 +1,195 @@
+# Copyright 2018 Open Source Foundries Limited.
+#
+# SPDX-License-Identifier: Apache-2.0
+
+'''West's bootstrap/wrapper script.
+'''
+
+import argparse
+import os
+from os.path import join
+import importlib
+import platform
+import subprocess
+import sys
+
+MANIFEST = 'manifest'
+MANIFEST_DEFAULT = 'https://github.com/zephyrproject-rtos/manifest'
+MANIFEST_REV_DEFAULT = 'master'
+WEST = 'west'
+WEST_DEFAULT = 'https://github.com/zephyrproject-rtos/west'
+WEST_REV_DEFAULT = 'master'
+# NB: on Windows, this will be hidden manually, using attrib.
+WEST_DIR = '.west'
+
+#
+# Helpers shared between init and wrapper mode
+#
+
+
+class WestError(RuntimeError):
+    pass
+
+
+class WestNotFound(WestError):
+    '''Neither the current directory nor any parent has a West installation.'''
+
+
+def find_west_topdir(start):
+    '''Find the top-level installation directory, starting at ``start``.
+
+    If none is found, raises WestNotFound.'''
+    if not os.path.isdir(start):
+        raise WestNotFound()
+
+    if WEST_DIR in os.listdir(start):
+        return start
+
+    dirname = os.path.dirname(start)
+    if start == dirname:
+        # / on POSIX, top level drive name on Windows.
+        raise WestNotFound()
+    else:
+        return find_west_topdir(dirname)
+
+
+def make_topdir_sentinel(directory):
+    '''Make the hidden sentinel that marks a West installation.
+
+    This is similar to .git or .repo directories: it marks the top
+    level directory of the installation. This allows runtime building
+    other paths, etc., which is how this script delegates to the
+    cloned west in wrapper mode (i.e. after 'west init' is run and the
+    "real" west repository has been cloned).'''
+    # The sentinel is a directory in case we ever need to stash
+    # anything inside. It could just as well be an empty file for now,
+    # but this is more future-proof.
+    sentinel = join(directory, WEST_DIR)
+    os.mkdir(sentinel)
+    # On Windows, we have to manually mark a file as hidden.
+    if platform.system() == 'Windows':
+        subprocess.check_call(['attrib', '+H', sentinel])
+
+
+def clone(url, rev, dest):
+    def repository_type(url):
+        if url.startswith(('http:', 'https:', 'git:', 'git+ssh:', 'file:')):
+            return 'GIT'
+        else:
+            return 'UNKNOWN'
+
+    if os.path.exists(dest):
+        msg = 'refusing to clone into existing location {}'.format(dest)
+        raise WestError(msg)
+
+    if repository_type(url) == 'GIT':
+        subprocess.check_call(['git', 'clone', url, '-b', rev, dest])
+    else:
+        raise WestError('Unknown URL scheme for repository: {}'.format(url))
+
+
+#
+# west init
+#
+
+
+def init(argv):
+    '''Command line handler for ``west init`` invocations.
+
+    This exits the program with a nonzero exit code if fatal errors occur.'''
+    init_parser = argparse.ArgumentParser(
+        prog='west init',
+        description='Bootstrap initialize a Zephyr installation')
+    init_parser.add_argument(
+        '-u', '--manifest-url', default=MANIFEST_DEFAULT,
+        help='Zephyr manifest fetch URL, default ' + MANIFEST_DEFAULT)
+    init_parser.add_argument(
+        '--mr', '--manifest-rev', default=MANIFEST_REV_DEFAULT,
+        dest='manifest_rev',
+        help='Manifest revision to fetch, default ' + MANIFEST_REV_DEFAULT)
+    init_parser.add_argument(
+        '-w', '--west-url', default=WEST_DEFAULT,
+        help='West fetch URL, default ' + WEST_DEFAULT)
+    init_parser.add_argument(
+        '--wr', '--west-rev', default=WEST_REV_DEFAULT, dest='west_rev',
+        help='West revision to fetch, default ' + WEST_REV_DEFAULT)
+    init_parser.add_argument(
+        'directory', nargs='?', default=None,
+        help='Initializes in this directory, creating it if necessary')
+
+    args = init_parser.parse_args(args=argv)
+    directory = args.directory or os.getcwd()
+
+    try:
+        topdir = find_west_topdir(directory)
+        init_reinit(topdir, args)
+    except WestNotFound:
+        init_bootstrap(directory, args)
+
+
+def init_bootstrap(directory, args):
+    '''Bootstrap a new manifest + West installation in the given directory.'''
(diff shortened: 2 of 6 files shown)


bootstrap/main.py
```diff
@@ -0,0 +1,194 @@
...
+MANIFEST = 'manifest'
+MANIFEST_DEFAULT = 'https://github.com/zephyrproject-rtos/manifest'
+MANIFEST_REV_DEFAULT = 'master'
+WEST = 'west'
+WEST_DEFAULT = 'https://github.com/zephyrproject-rtos/west'
+WEST_REV_DEFAULT = 'master'
+if platform.system() == 'Windows':
+    # On Windows, the sentinel is hidden with attributes, rather than a
+    # leading '.', and thus doesn't contain one.
+    WEST_HIDDEN_SENTINEL = 'west-top'
```
mbolivar: We need to be confident that this filename won't show up by accident in any of the repositories we manage, so it's worth picking its name carefully; any other ideas for what it should be named welcome.

bootstrap/main.py
```diff
@@ -0,0 +1,194 @@
...
+    If none is found, raises WestNotFound.'''
+    if not os.path.isdir(start):
+        raise WestNotFound()
+
+    if WEST_HIDDEN_SENTINEL in os.listdir(start):
+        return start
+
+    dirname = os.path.dirname(start)
+    if start == dirname:
+        # / on POSIX, top level drive name on Windows.
```
mbolivar: I confirmed this behavior -- that the dirname of a drive is the drive itself -- in a windows 10 VM.

bootstrap/main.py
```diff
@@ -0,0 +1,194 @@
...
+        if (url.startswith('http:') or url.startswith('https:') or
+                url.startswith('git:') or url.startswith('git+ssh:') or
+                url.startswith('file:')):
+            return 'GIT'
+        else:
+            return 'UNKNOWN'
+
+    if os.path.exists(dest):
+        msg = 'refusing to clone into existing location {}'.format(dest)
+        raise WestError(msg)
```
mbolivar: There should be more consistency around when to throw an exception and when it's OK to abort.

The helpers should probably only throw exceptions on error, as is done here. But this routine calls sys.exit below as well, which is sloppy and inconsistent.

bootstrap/main.py
```diff
@@ -0,0 +1,194 @@
...
+def make_topdir_sentinel(directory):
+    '''Make the hidden sentinel that marks a West installation.'''
+    # The sentinel is a directory in case we ever need to stash
+    # anything inside. It could just as well be an empty file for now,
+    # but this is more future-proof.
+    sentinel = join(directory, WEST_HIDDEN_SENTINEL)
+    os.mkdir(sentinel)
+    # On Windows, we have to manually mark a file as hidden.
+    if platform.system() == 'Windows':
+        subprocess.check_call(['attrib', '+H', sentinel])
```
mbolivar: I verified this works as expected with files earlier today, need to get back into the Windows environment and retest this all with the sentinel being a directory.

bootstrap/main.py
```diff
@@ -0,0 +1,194 @@
...
+            sys.exit('Cannot initialize in {}: permission denied'.format(
+                directory))
+        except FileExistsError:
+            sys.exit('Something else created {} concurrently; quitting'.format(
+                directory))
+        except Exception as e:
+            sys.exit("Can't create directory {}: {}".format(
+                directory, e.args))
+
+    make_topdir_sentinel(directory)
```
mbolivar: Unwinding this sequence on error would need more careful handling if we want to get worried about it failing. I'm hoping to keep it simple here, though, since the whole idea is that the bootstrap directory isn't in use by anything else.

carlescufi:
@mbolivar I am getting this while trying to install on macOS using Hombrew's Python3:

```
carles@zweck ~/src/west (pr/5)
$ python3 setup.py install --user
/usr/local/Cellar/python/3.6.5/Frameworks/Python.framework/Versions/3.6/lib/python3.6/distutils/dist.py:261: UserWarning: Unknown distribution option: 'long_description_content_type'
  warnings.warn(msg)
running install
error: can't combine user with prefix, exec_prefix/home, or install_(plat)base
```

carlescufi:
Once solved the installation hurdle, now I get:

```
carles@zweck ~ 
$ west
  File "/usr/local/bin/west", line 11, in <module>
    load_entry_point('west==0.2.0', 'console_scripts', 'west')()
  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 480, in load_entry_point
    return get_distribution(dist).load_entry_point(group, name)
  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2693, in load_entry_point
    return ep.load()
  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2324, in load
    return self.resolve()
  File "/usr/local/lib/python3.6/site-packages/pkg_resources/__init__.py", line 2330, in resolve
    module = __import__(self.module_name, fromlist=['__name__'], level=0)
ModuleNotFoundError: No module named 'bootstrap'
```

bootstrap/main.py
```diff
@@ -0,0 +1,194 @@
...
+    if platform.system() == 'Windows':
+        subprocess.check_call(['attrib', '+H', sentinel])
+
+
+def clone(url, rev, dest):
+    def repository_type(url):
+        if (url.startswith('http:') or url.startswith('https:') or
+                url.startswith('git:') or url.startswith('git+ssh:') or
+                url.startswith('file:')):
+            return 'GIT'
```
ulfalizer: `startswith()` accepts a tuple, so could do `if url.startswith(('http:', 'https:', 'git:', ...)): ...`

bootstrap/main.py
```diff
@@ -0,0 +1,194 @@
...
+        else:
+            return 'UNKNOWN'
+
+    if os.path.exists(dest):
+        msg = 'refusing to clone into existing location {}'.format(dest)
+        raise WestError(msg)
+
+    rtype = repository_type(url)
+    if rtype == 'GIT':
```
ulfalizer: `rtype` doesn't seem to be used elsewhere, so could do `if repository_type(url) == 'GIT':'.

bootstrap/main.py
```diff
@@ -0,0 +1,194 @@
...
+
+    dirname = os.path.dirname(start)
+    if start == dirname:
+        # / on POSIX, top level drive name on Windows.
+        raise WestNotFound()
+    else:
+        return find_west_topdir(dirname)
+
+
+def make_topdir_sentinel(directory):
```
ulfalizer: Would be nice to have a description of why the sentinel is needed, when it's checked, etc. here.

bootstrap/main.py
mbolivar: I was just planning ahead here for adding schemas for p4, svn, etc. as discussed at the kickoff. This way allows adding those as additional elif blocks, minimizing diff. But it'd be a trivial patch either way, so I'm fine with that.

bootstrap/main.py
mbolivar: I thought it was obvious why by analogy to git and repo, but sure, I can add those.

bootstrap/main.py
mbolivar: I had no idea! Thanks for teaching me all these nice idioms.

bootstrap/main.py
mbolivar: Works as expected.

mbolivar:
@carlescufi @ulfalizer comments addressed.

(Carles's issue was addressed by only supporting `pip3 install -e .` as an installation method, which seems to be current best practice: https://packaging.python.org/tutorials/installing-packages/#installing-from-a-local-src-tree).

bootstrap/main.py
ulfalizer: So it's like a `.git` for West, except still empty, and only used to flag a West installation?

Maybe it could be explained in those terms:

"foo/ is similar to a .git/ directory. Currently, it is only used to flag a directory as having a West installation in it. Files might be added to it in the future."

Could have `WEST_DIR` to take the analogy further maybe (similar to `GIT_DIR`). Not sure if that would get confusing though...

bootstrap/main.py
```diff
@@ -0,0 +1,195 @@
...
+import importlib
+import platform
+import subprocess
+import sys
+
+MANIFEST = 'manifest'
+MANIFEST_DEFAULT = 'https://github.com/zephyrproject-rtos/manifest'
+MANIFEST_REV_DEFAULT = 'master'
+WEST = 'west'
+WEST_DEFAULT = 'https://github.com/zephyrproject-rtos/west'
```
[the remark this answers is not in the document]
carlescufi: got an answer from @mbolivar on IRC. Bootstrapping requires this to avoid having the manifest-parsing code in PyPI

README.rst
```diff
@@ -1,13 +1,65 @@
...
+
+  $ west init your/zephyr/install-dir
+  $ cd your/zephyr/install-dir
+  $ west <command>
+
+The ``west init`` call will:
+
+- create your/zephyr/install-dir
+- clone the Zephyr manifest repository (whose URL can be overridden
+  with the ``-u`` option, and branch with ``--manifest-branch``)
```
carlescufi: why the inconsistency? `--manifest-branch` and `--west-revision`. Wouldn't it be better to use `--manifest-rev` (I also propose to shorten to `rev`

carlescufi: Alternatively, some sort of compound URL@rev scheme to use a single param?

bootstrap/main.py
carlescufi: I would actually call it `.west` in all platforms and use the attributes on Windows to hide it while maintaining the same name

bootstrap/main.py
```diff
@@ -0,0 +1,198 @@
...
+        init_reinit(topdir, args)
+    except WestNotFound:
+        init_bootstrap(directory, args)
+
+
+def init_bootstrap(directory, args):
+    '''Bootstrap a new manifest + West installation in the given directory.'''
+    if not os.path.isdir(directory):
+        try:
+            os.makedirs(directory, exist_ok=False)
```
carlescufi: This creates a folder specified by the user if it doesn't exist, correct? Should we output some text explaining it to the user?

README.rst
mbolivar: Crap, that's a bug in the README. It's consistent in the code: `--manifest-rev` and `--west-rev`. They both have shorthands: `--mr` and `--wr` respectively. README now fixed.

bootstrap/main.py
carlescufi: Should we clone west into the sentinel folder? I don't think so but just checking

README.rst
```diff
@@ -1,13 +1,72 @@
...
+  $ cd your/zephyr/install-dir
+  $ west <command>
+
+The ``west init`` call will:
+
+- create your/zephyr/install-dir
+- clone the Zephyr manifest repository (whose URL can be overridden
+  with the ``-u`` option, and branch with ``--mr`` / ``--manifest-rev``)
+- clone the latest West repository (the URL override is ``-w``, and
+  revision/branch override is ``--wr`` / ``--west-revision``)
```
carlescufi: still `--west-revision` here, should it be `--west-rev`?

bootstrap/main.py
```diff
@@ -0,0 +1,195 @@
...
+import argparse
+import os
+from os.path import join
+import importlib
+import platform
+import subprocess
+import sys
+
+MANIFEST = 'manifest'
+MANIFEST_DEFAULT = 'https://github.com/zephyrproject-rtos/manifest'
```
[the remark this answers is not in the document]
mbolivar: Note we can always update the bootstrap script if necessary to change this default. I personally think the project leaving GitHub would imply such big changes for developers and users that having to upgrade their west installation to get the new location (or just run `west init -u NEW_URL` if that's impossible) would be pretty low down on the list of issues.

README.rst
mbolivar: -ETOOMUCHCOFFEE, thanks, fixed.

bootstrap/main.py
```diff
@@ -0,0 +1,198 @@
...
+                directory))
+        except FileExistsError:
+            sys.exit('Something else created {} concurrently; quitting'.format(
+                directory))
+        except Exception as e:
+            sys.exit("Can't create directory {}: {}".format(
+                directory, e.args))
+
+    make_topdir_sentinel(directory)
+    clone(args.manifest_url, args.manifest_rev, join(directory, MANIFEST))
```
carlescufi: Should we use `--recurse-submodules` when cloning? In the (very unlikely) case that we ever include submodules in west itself, since west cannot depend on manifest-provided repos so it might use submodules one day?

mbolivar:
@carlescufi 

> Should we clone west into the sentinel folder? I don't think so but just checking

I think one of your points of criticism about repo is that its putting itself into a hidden directory was a bit too magical, so I didn't do that here. I'm happy to if we want to, but agree with you that we should not.

> Should we use --recurse-submodules when cloning? In the (very unlikely) case that we ever include submodules in west itself, since west cannot depend on manifest-provided repos so it might use submodules one day?

I think not. The only things the bootstrap script clones are the manifest and west itself, neither of which have submodules, and both of which are under our control.

If we ever need to handle submodules in repositories managed by the manifest, that'd be a "west sync" feature to be dealt with in west itself, not the wrapper.

README.rst
mbolivar: > Alternatively, some sort of compound URL@rev scheme to use a single param?

I think that's a bit too tricky personally. We don't know if `@` will ever be significant within a URL schema. What if I want to say `-u git+ssh://user@host/manifest`? Same could be said of any other special character.

mbolivar: Let's also keep in mind that 99.9% of the time, users want master.

bootstrap/main.py
mbolivar: Missed this comment at first, sorry. All comments now addressed.

carlescufi:
>I think one of your points of criticism about repo is that its putting itself into a hidden directory was a bit too magical, so I didn't do that here. I'm happy to if we want to, but agree with you that we should not.

Indeed, but since this is going to be relatively immutable I just wanted to raise the question before someone comes and asks why we didn't do it like repo.

>I think not. The only things the bootstrap script clones are the manifest and west itself, neither of which have submodules, and both of which are under our control.

I can imagine uses of submodules in west, but then again I have a very fertile imagination. Let's stick to PyPi for dependencies and, if the upstream repo doesn't want to push to PyPi, have a copy directly in the west repo.

README.rst
carlescufi: Yes, I agree. I much prefer the current set of command-line options to a compound scheme. I just like to raise these on a review in case someone can think of a way of actually making it useful.

mbolivar:
> Indeed, but since this is going to be relatively immutable I just wanted to raise the question before someone comes and asks why we didn't do it like repo.

Smart.

> I can imagine uses of submodules in west, but then again I have a very fertile imagination.

Please not for the bootstrap script :(

PR #6, opened 2018-07-27
Sync with zephyr

mbolivar:
I can't seem to request review from @tejlmand or @ulfalizer on this PR, so putting them here.

Getting back to work on West now after a break and some other tasks. Some commits went into Zephyr without going through this tree first, so step 1 of synchronizing is applying them here.

I'll send a PR to Zephyr pulling in the changes from the West repository next.

PR #7, opened 2018-08-02
Add testing infrastructure, start adding test suites for runner

mbolivar:
CC: @tejlmand, @Ulfalizer

I keep breaking runners (often in boards I don't have or use cases I don't know much about, like the Arduino 101 BLE core) when refactoring and making other changes.

This is justifiably annoying people (including me). A proper test suite is needed.

In this pull request, west is restructured to support proper test cases with py.test. I've added test suites for the pyocd and nrfjprog runners for starters. As I fix bugs in other runners, I'll add regression tests. These tests use unittest.mock to monkey-patch methods which do system calls and check that those calls match what we expect.


mbolivar:
Note this depends on #6

tests/west/runner/test_nrfjprog.py
```diff
@@ -0,0 +1,239 @@
...
+#
+
+TEST_DEF_SNR = 'test-default-serial-number'  # for mocking user input
+TEST_OVR_SNR = 'test-override-serial-number'
+
+#
+# Expected results.
+#
+# These record lists of system calls made by the nrfjprog runner. Each
+# system call is a list of arguments passed to the check_call() method.
```
ulfalizer: Had to stare at this for a while to realize that it's testing that those particular commands are executed. Maybe the variable could be renamed to something like `EXPECTED_COMMANDS`, and the comment rewritten to something like this:

```
This dictionary maps different configurations to the commands we expect to be
executed for them. Verification is done by patching the check_call() method,
which is used to run the commands.
```

tests/west/runner/test_nrfjprog.py
```diff
@@ -0,0 +1,239 @@
...
+# - SR: soft reset, Y for yes, N for pin reset
+# - SNR: serial number override, Y for yes, N for 'use default'
+# - E: full chip erase, Y for yes, N for sector / sector and UICR only
+#
+
+EXPECTED_RESULTS = {
+    # NRF51:
+    '1NNN':
+    [['nrfjprog', '--program', RC_KERNEL_HEX, '-f', 'NRF51', '--snr', TEST_DEF_SNR, '--sectorerase'],  # noqa: E501
+     ['nrfjprog', '--pinreset', '-f', 'NRF51', '--snr', TEST_DEF_SNR]],
```
ulfalizer: Subjective, but tuples instead of lists makes it obvious that these aren't getting modified later (already pretty clear here though...)

tests/west/runner/test_nrfjprog.py
```diff
@@ -0,0 +1,239 @@
...
+@pytest.fixture
+def nrfjprog_init():
+    '''NrfJprogBinaryRunner constructed with __init__ args/kwargs.
+
+    Returns a factory that consumes a runner configuration and
+    nrfjprog runner __init__ arguments.'''
+    def _factory(cfg, family, softreset, snr, erase):
+        args = [cfg, family, softreset, snr]
+        return NrfJprogBinaryRunner(*args, erase=erase)
+    return _factory
```
ulfalizer: I might be missing something, but what would you lose by having these `_factory` functions as ordinary helper functions instead? It looks like the `@fixture` functions themselves could never fail.

tests/west/runner/test_nrfjprog.py
```diff
@@ -0,0 +1,239 @@
...
+
+@pytest.fixture
+def nrfjprog_init():
+    '''NrfJprogBinaryRunner constructed with __init__ args/kwargs.
+
+    Returns a factory that consumes a runner configuration and
+    nrfjprog runner __init__ arguments.'''
+    def _factory(cfg, family, softreset, snr, erase):
+        args = [cfg, family, softreset, snr]
+        return NrfJprogBinaryRunner(*args, erase=erase)
```
ulfalizer: Could this be just `return NrfJprogBinaryRunner(cfg, family, softreset, snr, erase=erase)`?

tests/west/runner/test_nrfjprog.py
mbolivar: It was done this way to match the data types provided by the `@patch('...check_call')` decorators below.

In particular, looking here:

```
    assert cc.call_args_list == [call(x) for x in expected]
```

The `call_args_list` attributes use lists:

```
(Pdb) pp cc.call_args_list
[call(['nrfjprog', '--program', 'test-zephyr.hex', '-f', 'NRF51', '--snr', 'test-override-serial-number', '--sectorerase']),
 call(['nrfjprog', '--pinreset', '-f', 'NRF51', '--snr', 'test-override-serial-number'])]
```

We could convert them into tuples of tuples, but why waste cycles?

tests/west/runner/test_nrfjprog.py
mbolivar: I am new to py.test, but it seems to prefer using fixtures, and I'm just trying to look cool.

There's nothing stopping us from using an ordinary helper here, though; I'll make the change.

mbolivar:
@ulfalizer review comments addressed, thanks

tests/west/runner/test_nrfjprog.py
ulfalizer: Hmm... should work without conversion though, since a new list is built, and `for x in ...` can iterate through tuples as well as lists.

Not the most important thing in the world though.

Maybe it could be shortened to `assert cc.call_args_list == [call(x) for x in expected_commands(*test_case)]`, since `expected` is only used in one place. Could move `expected` to right before it's used otherwise as well.

Just some personal OCD though...

tests/west/runner/test_nrfjprog.py
ulfalizer: Forgot to commit them? :)

tests/west/runner/test_nrfjprog.py
ulfalizer: Hmm... thought `call()` would act the same whether you called it with a list or a tuple. Never mind then.

mbolivar: `call` is recording a sequence of arguments; the sequence has length 1, and is a list passed to `check_call`. Since `[1,2,3] != (1,2,3)`, I think it's correct for the test to fail when passed a tuple with equal items.

mbolivar:
> Looks like nrfjprog_init() and nrfjprog_create() could be ordinary functions still.

I originally had them being fixtures depending on `runner_config`. Now that they are passed the configuration instead (and `runner_config` itself can be parametrized), they can actually be eliminated entirely.

tests/west/runner/test_nrfjprog.py
ulfalizer: Oh, right, the mock just gets the argument list as-is...

mbolivar:
@nashif @carlescufi similar to https://github.com/zephyrproject-rtos/west/pull/6, this has been reviewed by @ulfalizer for 5 days now -- can we make him an approving reviewer on West to unblock the merge?

nashif:
> @nashif @carlescufi similar to #6, this has been reviewed by @ulfalizer for 5 days now -- can we make him an approving reviewer on West to unblock the merge?

@ulfalizer needs to accept the invite :)

mbolivar:
> @ulfalizer needs to accept the invite :)

Hah!

PR #8, opened 2018-08-09
Manifest commands RFC

ulfalizer:
This RFC adds manifest parsing and three basic commands (sync, diff, and
status).

More error checking needs to be added. This is mostly to get some
feedback on the approach. There are some cases that turn tricky if you
always keep a local branch to avoid a detached HEAD.

I'm wondering if 'revision' is supposed to always point to a branch (as
opposed to e.g. a SHA). SHAs would be more flexible, but make it even
trickier to keep a local branch.

I've written a bit in
https://github.com/zephyrproject-rtos/zephyr/issues/6770 as well.
This is a quick-and-dirty version of the local branch strategy.

### DIFF
--- a/west/cmd/git.py
+++ b/west/cmd/git.py
@@ -1,0 +1,280 @@
+# Copyright (c) 2018, Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: Apache-2.0
+
+'''west git commands'''
+
+import collections
+import os
+import shutil
+import subprocess
+
+import pykwalify.core
+import yaml
+
+from . import WestCommand
+from .. import log
+
+
+class Sync(WestCommand):
+    def __init__(self):
+        super().__init__(
+            'sync',
+            'Clone/update all Git repositories specified in the manifest file')
+
+    def do_add_parser(self, parser_adder):
+        return _add_common_git_flags(parser_adder, self)
+
+    def do_run(self, args, user_args):
+        projects = _all_projects(args)
+
+        # TODO: Error checking (if _git(...).returncode != 0: ...)
+
+        for project in projects:
+            if not os.path.exists(project.path):
+                _git_top(project, 'clone -b (branch) (url)/(name) (path)')
+            else:
+                # Fetch first to make sure the project's branch exists. It
+                # might not if 'git clone' was aborted, for example.
+                _git(project, 'fetch')
+
+                # TODO: The local branch might not exist if the manifest was
+                # updated. Maybe local branches could be stored in a separate
+                # namespace...
+
+                # Check out the main branch and update it
+                _git(project, 'checkout (branch)')
+                _git(project, 'rebase FETCH_HEAD')
+
+                # Switch back to previous branch the user was on
+                _git(project, 'checkout -')
+
+
+class Diff(WestCommand):
+    def __init__(self):
+        super().__init__(
+            'diff',
+            "Run 'git diff' for each project. Extra arguments are passed "
+            "as-is to 'git diff'.",
+            accepts_unknown_args=True)
+
+    def do_add_parser(self, parser_adder):
+        return _add_common_git_flags(parser_adder, self)
+
+    def do_run(self, args, user_args):
+        for project in _all_projects(args):
+            if _check_repo(project):
+                # Use paths that are relative to the base directory to show
+                # which repo any changes are in
+                _git(project, 'diff --src-prefix=(path)/ --dst-prefix=(path)/',
+                     extra_args=user_args)
+
+
+class Status(WestCommand):
+    def __init__(self):
+        super().__init__(
+            'status',
+            "Run 'git status' for each project. Extra arguments are passed "
+            "as-is to 'git status'.",
+            accepts_unknown_args=True)
+
+    def do_add_parser(self, parser_adder):
+        return _add_common_git_flags(parser_adder, self)
+
+    def do_run(self, args, user_args):
+        for project in _all_projects(args):
+            if _check_repo(project):
+                _git(project, 'fetch')
+
+                log.inf("=== 'git status' for {} (in {}) ==="
+                        .format(project.name, project.path))
+                _git(project, 'status', extra_args=user_args)
+
+
+def _add_common_git_flags(parser_adder, command):
+    # Adds common command-line flags for the Git-related commands. The manifest
+    # file contains repository information.
+
+    parser = parser_adder.add_parser(
+        command.name,
+        description=command.description)
+
+    parser.add_argument(
+        '-m', '--manifest',
+        dest='manifest',
+        help='path to manifest file (default: <zephyr-base>/manifest/default.yml)')
+
+    # TODO: Make schema optional?
+
+    parser.add_argument(
+        '-s', '--schema',
+        dest='schema',
+        help='path to pykwalify schema for manifest (default: <zephyr-base>/manifest/default.yml)')
+
+    return parser
+
+
+# Holds information about a project, taken from the manifest file
+Project = collections.namedtuple('Project', 'name url revision path')
+
+
+def _all_projects(args):
+    # Parses the manifest file, returning a list with Project instances for all
+    # projects. Also verifies the manifest against a pykwalify schema.
+
+    if (not args.manifest or not args.schema) and not args.zephyr_base:
+        log.die('Zephyr base directory not specified (via --zephyr-base or ZEPHYR_BASE)')
+
+    if args.schema:
+        schema_filename = args.schema
+    else:
+        schema_filename = os.path.join(
+            args.zephyr_base, 'manifest', 'schema.yml')
+
+    if args.manifest:
+        manifest_filename = args.manifest
+    else:
+        manifest_filename = os.path.join(
+            args.zephyr_base, 'manifest', 'manifest.yml')
+
+
+    # Validate the manifest with pykwalify
+
+    try:
+        pykwalify.core.Core(
+            source_file=manifest_filename, schema_files=[schema_filename]
+        ).validate()
+    except pykwalify.errors.SchemaError as e:
+        log.die('{} malformed (schema: {}):\n{}'
+                .format(manifest_filename, schema_filename, e))
+
+
+    # Load project information from manifest
+
+    with open(manifest_filename) as f:
+        manifest = yaml.safe_load(f)['manifest']
+
+    projects = []
+
+    # mp = manifest project (dictionary with values parsed from the manifest)
+    for mp in manifest['projects']:
+        # Fill in any missing fields in 'mp' with values from the 'defaults'
+        # dictionary
+        for key, val in manifest['defaults'].items():
+            mp.setdefault(key, val)
+
+        # Add the repository URL to 'mp'
+        for remote in manifest['remotes']:
+            if remote['name'] == mp['remote']:
+                mp['url'] = remote['url']
+                break
+        else:
+            log.die('Remote {} not defined in {}'
+                    .format(mp['remote'], manifest_filename))
+
+        # If 'mp' doesn't specify a clone path, the project's name is used
+        if 'path' not in mp:
+            mp['path'] = mp['name']
+
+        # If 'mp' doesn't specify a branch, 'master' is used
+        if 'revision' not in mp:
+            mp['revision'] = 'master'
+
+        # Use named tuples to store project information. That gives nicer
+        # syntax compared to a dict (project.name instead of project['name'],
+        # etc.)
+        projects.append(Project(mp['name'], mp['url'], mp['revision'],
+                                mp['path']))
+
+    return projects
+
+
+def _check_repo(project):
+    # Returns True if the project's repository exists and looks like a Git
+    # repository. Otherwise, returns False and prints a message about the
+    # repistory being skipped.
+
+    if not os.path.exists(project.path):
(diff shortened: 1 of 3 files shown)


mbolivar:
> I'm wondering if 'revision' is supposed to always point to a branch (as
> opposed to e.g. a SHA). SHAs would be more flexible, but make it even
> trickier to keep a local branch.

Repo supports branches (this is what the tip manifest of the AOSP uses), tags (via revision="refs/tags/xxx", this is what all releases use), SHAs (this is what pinned manifests generated by `repo manifest -r` contain), and mixes of the three.

The ability to support SHAs is important: this is how you can use a manifest to freeze and distribute an entire configuration of all the Git trees to others, including users.

west/cmd/git.py
```diff
@@ -0,0 +1,255 @@
...
+
+    def do_add_parser(self, parser_adder):
+        return _add_common_git_flags(parser_adder, self)
+
+    def do_run(self, args, user_args):
+        projects = _all_projects(args)
+
+        # TODO: Error checking (if _git(...).returncode != 0: ...)
+
+        for project in projects:
```
mbolivar: Needs a TODO to handle projects that have been removed as well.

(I guess this means persisting a representation of the current projects somewhere and comparing it to the updated manifest that `west sync` pulls down.)

west/cmd/git.py
```diff
@@ -0,0 +1,255 @@
...
+            if not os.path.exists(project.path):
+                _git_top(project, 'clone -b (branch) (url)/(name) (path)')
+            else:
+                # Fetch first to make sure the project's branch exists. It
```
mbolivar: So the thinking here seems to be that we can assume that local branches with the same names track remote branches named in the manifest. As you've noticed, this doesn't play well if the manifest names SHAs (or I guess tags) in its revision fields.

That kind of makes the way repo behaves (as I understand it) seem at least not crazy:

- if detached head, synchronize to latest manifest version
- if there is a local branch checked out, rebase on top of the new manifest branch

Updating a detached HEAD to another detached HEAD seems fine.

I've said before that I think rebase by default is counter-intuitive and not a good idea for West, though -- in my experience teaching people Repo, this is one of the most common frustrations and stumbling blocks, to the point that I usually teach `repo sync -d` first.

If we want to avoid that for West, I think these are the potentially reasonable options if there is a local branch:

1. act like git pull (fetch + merge)
2. act like git fetch (fetch only)
3. scream and die unless a `--rebase` or `--local-branch-strategy={rebase,detach,merge}` or so flag was given
4. always fetch and detach HEAD (unless overridden by a flag)

Am I missing any?

I'm leaning against 1, because it encourages people to submit bad pull requests with upstream merges inside, and 2, because `west sync` ought to, well, synchronize with upstream whenever possible.

The other two seem potentially OK to me.

Thoughts?

west/cmd/git.py
```diff
@@ -0,0 +1,255 @@
...
+        manifest_filename = os.path.join(
+            args.zephyr_base, 'manifest', 'manifest.yml')
+
+
+    # Validate the manifest with pykwalify
+
+    try:
+        pykwalify.core.Core(
+            source_file=manifest_filename, schema_files=[schema_filename]
+        ).validate()
```
mbolivar: I checked the pykwalify sources: they use safe_load, so this should be OK.

west/cmd/git.py
```diff
@@ -0,0 +1,255 @@
...
+    with open(manifest_filename) as f:
+        manifest = yaml.load(f)['manifest']
+
+    projects = []
+
+    # mp = manifest project (dictionary with values parsed from the manifest)
+    for mp in manifest['projects']:
+        # Fill in any missing fields in 'mp' with values from the 'defaults'
+        # dictionary
+        for key, val in manifest['defaults'].items():
```
mbolivar: This seems safe for now, but note that the manifest-wide defaults might contain keys which do not apply to projects.

west/cmd/git.py
```diff
@@ -0,0 +1,255 @@
...
+        for project in _all_projects(args):
+            if _check_repo(project):
+                _git(project, 'fetch')
+
+                print("=== 'git status' for {} (in {}) ==="
+                      .format(project.name, project.path))
+                _git(project, 'status', extra_args=user_args)
+
+
+def _add_common_git_flags(parser_adder, command):
```
mbolivar: Ah, also: it would be nice to be able to take a `nargs='*'` argument specifying the list of projects to synchronize, diff against, or get status for.

Supporting either the name of the project in the manifest or its locally checked out path would be most convenient.

west/cmd/git.py
```diff
@@ -0,0 +1,255 @@
+# Copyright (c) 2018 Open Source Foundries Limited.
```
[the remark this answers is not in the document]
ulfalizer: Processed it as copy-paste. Glad I didn't go into law. :D

I'll put Nordic Semiconductor there instead.

west/cmd/git.py
```diff
@@ -0,0 +1,255 @@
...
+        ).validate()
+    except pykwalify.errors.SchemaError as e:
+        log.die('{} malformed (schema: {}):\n{}'
+                .format(manifest_filename, schema_filename, e))
+
+
+    # Load project information from manifest
+
+    with open(manifest_filename) as f:
+        manifest = yaml.load(f)['manifest']
```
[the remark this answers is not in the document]
ulfalizer: Hadn't noticed it. Will fix.

west/cmd/git.py
```diff
@@ -0,0 +1,255 @@
...
+    #
+    # Returns a subprocess.CompletedProcess instance.
+
+    return _git_helper(project, cmd, extra_args, project.path, capture_output)
+
+
+def _git_helper(project, cmd, extra_args, cwd, capture_output):
+    # Runs a git command.
+    #
+    #   cwd: Directory to switch to first (None = current directory)
```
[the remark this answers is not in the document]
ulfalizer: Oops... will fix

west/cmd/git.py
ulfalizer: Yeah, would be nice to have later

west/cmd/git.py
ulfalizer: Thought it was defined as defaults for projects.

west/cmd/git.py
mbolivar: Projects removed from manifests should be removed from working copies. Otherwise there's old cruft around which no longer matters.

What to do with any local branches, though? Perhaps repo keeps git repositories and objects under `.repo`, and symlinks to those from working trees, precisely to handle this case without deleting local work. No symlinks on Windows, though.

Maybe error out and ask for a re-run with a `--force` option if there are repositories with local branches that would be removed, as a starting point?

west/cmd/git.py
mbolivar: It's defaults for the manifest

west/cmd/git.py
ulfalizer: I'd also be mad if it removed any non-checked-in files I happened to have in there. "Lemme remove this stuff you no longer need for you" is a bit dangerous...

Personally, I'd be fine if they were just left behind, but we'll see.

ulfalizer:
Fixed most of the code complaints.

west/cmd/git.py
ulfalizer: I thought the manifest was a project description. :)

west/cmd/git.py
ulfalizer: > 1. act like git pull (fetch + merge)

Would need to check out the branch first in that case (and there would have to
be a local branch). The current version does that, and then switches back to
the previous position. That feels a bit wonky though, but maybe it could work
alright in practice.

> 2. act like git fetch (fetch only)

Would want some way to rebase local work on top of new upstream stuff, so would
probably need additional commands in that case. Wonder if that might be too
complex.

> 3. scream and die unless a --rebase or --local-branch-strategy={rebase,detach,merge} or so flag was given
> 4. always fetch and detach HEAD (unless overridden by a flag)

IMO, it's nice if sync (by default at least) brings in upstream work while also
preserving local changes. I suspect that's what most people would expect.

Keeping a local branch gets a bit tricky when you can have SHAs. I wonder if
there might be gotchas related to changing the revision in the manifest file
too (e.g. to a different branch).

Always just doing a 'git rebase <revision>' is conceptually simplest at least
(and will work with a detached HEAD as well).

The implementation gets a bit tricky though, because e.g. 'git status' won't
know what the upstream is, and so won't tell you stuff like how many commits
behind you are. Gotta implement that stuff manually I think.

ulfalizer: The only "robust" way I can think of to do it with a non-detached HEAD would be to have a single local branch (e.g. `master`), and updating it against the revision from manifest (ignoring the branch name there).

That'd give you e.g. `git status` for free, but only if the revision from the manifest is a branch name (plus you'd have to update which branch the local branch tracks if the manifest is changed to use a different branch), so might have to implement that stuff manually anyway.

When the revision is an SHA, I don't think the local branch would give you much, beyond avoiding people having to see "detached HEAD" when they run `git status` manually.

ulfalizer: Hmm... thought of another thing that might work. Wonder if it's too magic though.

Could keep a ref (branch) that's always updated to point to the revision specified in the manifest, and that the user isn't expected to update (could put it in `upstream/revision` or the like, to make it clearer that it's "internal", though it'd be nice to explain it still).

That branch could then be set as the upstream of the local branch, and `git status` and the like would automatically give sensible information (might get confusing for `git push` and the like though).

Might be a horrible idea for reasons I haven't thought of yet though. It's something that's not a detached HEAD at least. :P

ulfalizer: Seems kinda neat in that you will always have an easy reference for the upstream revision at least, without having to go to the manifest.

Could have `west fetch` update the `upstream/revision` refs, and `west sync` rebase the local branch on top of it, if you wanted to split it up.

west/cmd/git.py
mbolivar: Repo lets you manage default values that are related to the behavior of the tool itself there. For example: `<default sync-j="4"/>` uses 4 concurrent TCP connections while syncing. I imagine we may end up with use cases that are similar eventually.

west/cmd/git.py
mbolivar: >> 1. act like git pull (fetch + merge)

> Would need to check out the branch first in that case (and there would have to
> be a local branch).

Right -- note that this list was prefixed with (emphasis added): "I
think these are the potentially reasonable options **if there is a
local branch**". Since we're discussing what to do in case of a local
branch (instead of detached HEAD), it's a safe assumption that there
is one :).

> The current version does that, and then switches back to
> the previous position. That feels a bit wonky though, but maybe it could work
> alright in practice.

Sorry, I don't follow what you're saying. I'm thinking this option 1
would be something like:

```
git fetch <manifest-remote>/<manifest-revision>
git merge FETCH_HEAD
```

That doesn't require switching back to a previous position.

Note that I'm against this option as it encourages invalid pull
request submissions (i.e. PRs which include "merge upstream into my
feature branch" commits).

>> 2. act like git fetch (fetch only)
> Would want some way to rebase local work on top of new upstream stuff, so would
> probably need additional commands in that case. Wonder if that might be too
> complex.

Sorry, I don't follow this either. Why would we *need* a separate
rebase command, when there is already `git rebase`?

We haven't discussed the "submit a pull request" or "fetch a pull
request" commands, but updating local branches to rebase against
upstream might better belong as part of those.

>> 3. scream and die unless a --rebase or --local-branch-strategy={rebase,detach,merge} or so flag was given
>> 4. always fetch and detach HEAD (unless overridden by a flag)
> IMO, it's nice if sync (by default at least) brings in upstream work while also
> preserving local changes. I suspect that's what most people would expect.

> Keeping a local branch gets a bit tricky when you can have SHAs. I wonder if
> there might be gotchas related to changing the revision in the manifest file
> too (e.g. to a different branch).

> Always just doing a git rebase <revision> is conceptually simplest at least
> (and will work with a detached HEAD as well).

I'm confused about what you're saying here. Do think I've missed any
options? Do you lean towards or against any of the four I've listed
above?

> The implementation gets a bit tricky though, because e.g. git status won't
> know what the upstream is, and so won't tell you stuff like how many commits
> behind you are. Gotta implement that stuff manually I think. Would have that
> problem with an SHA regardless though...

Do you want to play around with `repo start`, `repo sync`, and `repo
status` to get a baseline idea of how it works there? That might help
as a starting point, just seeing how at least one other tool works.

The way `repo start` works is that it creates a local branch that
tracks `<manifest-remote>/<manifest-revision>`. So `repo start test
<project>` will create a `test` branch with the following in the
project's .git/config:

```
[branch "test"]
	remote = osf-dev
	merge = refs/heads/master
```

Then `repo sync` will rebase on top of there. And `status` always
knows the remote, so you can do things like:

```
$ repo info -o <project>
Manifest branch: master
Manifest merge branch: refs/heads/master
Manifest groups: all,-notdefault
----------------------------

Projects Overview
<project>
* test                              ( 1 commit, Mon Aug 13 13:13:35 2018 -0500)
                                      - 5136aa26 add test commit
```

west/cmd/git.py
ulfalizer: Ah, yeah, might have to change the approach at that point, if they get added to the `defaults` dict.

ulfalizer:
Messed-up the branch name when I created this one. New PR is at https://github.com/zephyrproject-rtos/west/pull/11.

PR #10, opened 2018-08-16
Add build cmd

mbolivar:
CC @tejlmand, @ulfalizer

Only the last two commits in this PR matter. This is written on top of the series which adds tests since that moves all the sources around in the tree.

Add a build command.


src/west/cmd/build.py
```diff
@@ -0,0 +1,255 @@
...
+
+    def do_run(self, args, ignored):
+        self.args = args        # Avoid having to pass them around
+        log.dbg('args:', args, level=log.VERBOSE_EXTREME)
+        self._sanity_precheck()
+        self._setup_build_dir()
+        if is_zephyr_build(self.build_dir):
+            self._update_cache()
+        else:
+            self.args.cmake = True
```
ulfalizer: Wonder if changing the arguments in the code could get confusing. Might expect them to always match what the user passed in.

Maybe this could be called `force_cmake` or the like if it's changed.

src/west/cmd/build.py
```diff
@@ -0,0 +1,255 @@
...
+                        else '')))
+        # FIXME needs some extra logic in here
+        if self.cmake_cache:
+            board = self.cmake_cache.get('CACHED_BOARD')
+        else:
+            board = 'UNKNOWN'   # shouldn't happen
+        log.inf('BOARD:', board)
+
+        self._run_cmake(self.args.cmake_opts)
+        self._sanity_check()
```
ulfalizer: Just to check, is it intentional that `_sanity_check()` is called twice in this function?

mbolivar:
@ulfalizer thanks for the review, comments addressed.

mbolivar:
recheck

nashif:
> recheck

you need to merge the other PR first, this one does not see .shippable.yml

mbolivar:
> you need to merge the other PR first, this one does not see .shippable.yml

Ah. Can I get your +1 on the other one (https://github.com/zephyrproject-rtos/west/pull/12) then? Needs review.

nashif:
recheck

PR #11, opened 2018-08-20
Add project commands for dealing with multiple repositories

ulfalizer:
Add simple Google repo-like functionality, for dealing with multiple Git
repositories.

The command set is mirrored after Git. All commands except
'list-projects' accept an optional list of projects, and default to all
(cloned) projects.

  - west list-projects

    Lists projects and their repositories

  - west fetch

    Clone/fetch projects. Supports 'clone-depth', for making shallow
    clones.

  - west rebase

    Rebase local branches to the revision specified in the manifest

  - west pull

    'west fetch' + 'west rebase' (similar to 'sync' in Google repo)

  - west branch

    Create a new branch in one or more repositories (for working on some
    issue)

  - west checkout

    Check out a branch in each repository that has it

  - west diff

    Run 'git diff' in each repository

  - west status

    Run 'git status' in reach repository

There's no way to submit a multi-repository change for review yet.

Currently, a convenience branch 'manifest-rev' is created in each
project, which points to the revision specified for the project in the
manifest. 'manifest-rev' is updated by 'west fetch' and 'west pull'.

Local branches created with 'west branch' are set to track
'manifest-rev'. This makes e.g. a plain 'git status' or 'git pull' work
sensibly even when the manifest revision is an SHA.

We'll see if 'manifest-rev' is too magic later. It's explained in the
help texts of all the relevant commands at least.

It might be nicer to create 'manifest-rev' in e.g. refs/remotes instead
of refs/heads. Git doesn't seem to like trying to create a branch that
tracks a branch in refs/remotes unless it's a "proper" upstream branch
though.

### DIFF
--- a/src/west/cmd/project.py
+++ b/src/west/cmd/project.py
@@ -1,0 +1,646 @@
+# Copyright (c) 2018, Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: Apache-2.0
+
+'''West project commands'''
+
+import argparse
+import collections
+import os
+import shlex
+import shutil
+import subprocess
+import textwrap
+
+import pykwalify.core
+import yaml
+
+from . import WestCommand
+from .. import log
+from .. import util
+
+
+# Branch that points to the revision specified in the manifest (which might be
+# an SHA). Local branches created with 'west branch' are set to track this
+# branch.
+_MANIFEST_REV_BRANCH = 'manifest-rev'
+
+
+class ListProjects(WestCommand):
+    def __init__(self):
+        super().__init__(
+            'list-projects',
+            _wrap('''
+            List projects.
+
+            Prints the path to the manifest file and lists all projects along
+            with their clone paths and manifest revisions. Also includes
+            information on which projects are currently cloned.
+            '''))
+
+    def do_add_parser(self, parser_adder):
+        return _add_parser(parser_adder, self)
+
+    def do_run(self, args, user_args):
+        log.inf("Manifest path: {}\n".format(_manifest_path(args)))
+
+        for project in _all_projects(args):
+            log.inf('{:15} {:30} {:15} {}'.format(
+                project.name,
+                os.path.join(project.path, ''),  # Add final '/' if missing
+                project.revision,
+                "(cloned)" if _cloned(project) else "(not cloned)"))
+
+
+class Fetch(WestCommand):
+    def __init__(self):
+        super().__init__(
+            'fetch',
+            _wrap('''
+            Clone/fetch projects.
+
+            Fetches upstream changes in each of the specified projects
+            (default: all projects). Repositories that do not already exist are
+            cloned.
+
+            ''' + _MANIFEST_REV_HELP))
+
+    def do_add_parser(self, parser_adder):
+        return _add_parser(parser_adder, self, _project_list_arg)
+
+    def do_run(self, args, user_args):
+        for project in _projects(args, listed_must_be_cloned=False):
+            _fetch(project)
+
+
+class Pull(WestCommand):
+    def __init__(self):
+        super().__init__(
+            'pull',
+            _wrap('''
+            Clone/fetch and rebase projects.
+
+            Fetches upstream changes in each of the specified projects
+            (default: all projects) and rebases the checked-out branch (or
+            detached HEAD state) on top of '{}', effectively bringing the
+            branch up to date. Repositories that do not already exist are
+            cloned.
+
+            '''.format(_MANIFEST_REV_BRANCH) + _MANIFEST_REV_HELP))
+
+    def do_add_parser(self, parser_adder):
+        return _add_parser(parser_adder, self, _project_list_arg)
+
+    def do_run(self, args, user_args):
+        for project in _projects(args, listed_must_be_cloned=False):
+            if _fetch(project):
+                _rebase(project)
+
+
+class Rebase(WestCommand):
+    def __init__(self):
+        super().__init__(
+            'rebase',
+            _wrap('''
+            Rebase projects.
+
+            Rebases the checked-out branch (or detached HEAD) on top of '{}' in
+            each of the specified projects (default: all cloned projects),
+            effectively bringing the branch up to date.
+
+            '''.format(_MANIFEST_REV_BRANCH) + _MANIFEST_REV_HELP))
+
+    def do_add_parser(self, parser_adder):
+        return _add_parser(parser_adder, self, _project_list_arg)
+
+    def do_run(self, args, user_args):
+        for project in _projects(args):
+            if _cloned(project):
+                _rebase(project)
+
+
+class Branch(WestCommand):
+    def __init__(self):
+        super().__init__(
+            'branch',
+            _wrap('''
+            Create topic branch.
+
+            Creates a branch in each of the specified projects (default: all
+            cloned projects).
+
+            The new branches are set to track '{}'.
+
+            '''.format(_MANIFEST_REV_BRANCH) + _MANIFEST_REV_HELP))
+
+    def do_add_parser(self, parser_adder):
+        return _add_parser(parser_adder, self, _branch_arg, _project_list_arg)
+
+    def do_run(self, args, user_args):
+        for project in _projects(args):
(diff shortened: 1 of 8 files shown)


ulfalizer:
This is still RFC'ish, though I think it could be merged at this point to make it easier to try out.

Feedback on the command set or anything else welcome. I don't have experience working with Google repo.

ulfalizer:
Suggestions for backend Git commands are welcome too. I just with a plain spammy `git status` for now for example. Could polish it as we go.

west/cmd/project.py
```diff
@@ -0,0 +1,604 @@
...
+    try:
+        int(s, 16)
+    except ValueError:
+        return False
+
+    return len(s) == 40
+
+
+def _git_base(project, cmd, *, extra_args=(), capture_stdout=False, check=True):
+    # Runs a git command in the base directory. See _git_helper() for parameter
```
mbolivar: Still not sure what "base directory" means here, but still I think you need to be going back up the tree until you find WEST_DIR. Am I missing something?

ulfalizer: Means the current directory (as opposed to one of the project directories). I was assuming the project commands would just do their thing in whatever directory you happen to be in.

Does `repo` do a search?

mbolivar: I need to try this out, but wouldn't fetch need to go back to the top level directory to check paths properly?

ulfalizer: Was thinking plain git commands. Maybe it works differently in repo though.

mbolivar: So I tried this out, and I think it's looking for ZEPHYR_BASE when it need not:

```
$ west init -w https://github.com/ulfalizer/west --wr project-rfc test
$ cd test/
$ ls
manifest  west
$ west fetch
warning: --zephyr-base missing and no ZEPHYR_BASE in the environment
fatal error: Explicit path to manifest file not given, and Zephyr base directory not specified (via --zephyr-base or $ZEPHYR_BASE)
```

If ZEPHYR_BASE continues to mean the same thing as always, that's not where the manifest is: the manifest is the 'manifest' directory relative to `.west` (which could be in $PWD or somewhere above it).

mbolivar: Note that we  put the west repo in the top level directory (where `.west` is, called WEST_DIR in the bootstrap script), but that's something we could move around if we wanted since nobody's using it yet. The marker of the top level is the `.west` dir, analogous to `.git`.

src/west/cmd/project.py
```diff
@@ -0,0 +1,601 @@
...
+    # Returns a subprocess.CompletedProcess instance.
+
+    # TODO: Run once somewhere?
+    if shutil.which('git') is None:
+        log.die('Git is not installed or cannot be found')
+
+    args = [_expand_shorthands(project, arg) for arg in cmd.split()]
+
+    popen = subprocess.Popen(
+        ('git', *args, *extra_args),
```
mbolivar: not python 3.4 -- mind adding some basic tests since we seem to have a pattern of using too-recent language features?

src/west/cmd/project.py
```diff
@@ -0,0 +1,601 @@
...
+
+    if os.path.exists(os.path.join(util.west_topdir(), project.path)):
+        existed = True
+        _verify_repo(project)
+        _inf(project, '=== Fetching changes for (name-and-path) ===')
+        _git(project, 'remote update')
+    else:
+        existed = False
+        _inf(project, '=== Cloning (name-and-path) ===')
+        _git_base(project, 'clone (url)/(name) (path)')
```
mbolivar: needs clone-depth manifest support as discussed on irc

src/west/cmd/project.py
```diff
@@ -0,0 +1,601 @@
...
+_project_list_arg = argparse.ArgumentParser(add_help=False)
+_project_list_arg.add_argument('projects', metavar='PROJECT', nargs='*')
+
+# Common branch argument
+_branch_arg = argparse.ArgumentParser(add_help=False)
+_branch_arg.add_argument('branch', metavar='BRANCH_NAME')
+
+
+# Holds information about a project, taken from the manifest file
+Project = collections.namedtuple('Project', 'name url revision path')
```
mbolivar: clone-depth could go here

src/west/cmd/project.py
```diff
@@ -0,0 +1,601 @@
...
+def _add_parser(parser_adder, cmd, *extra_args):
+    # Adds and returns a subparser for the project-related WestCommand 'cmd'.
+    # All of these commands (currently) take the manifest path flag, so it's
+    # hardcoded here.
+
+    return parser_adder.add_parser(
+        cmd.name,
+        description=cmd.description,
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        parents=(_manifest_arg, *extra_args))
```
mbolivar: not python 3.4

ulfalizer:
Fixed issues and added `clone-depth` support.

`git clone --depth n` implies `--single-branch`, so `--branch` needs to be passed when doing a shallow clone. I made it so that `--branch` is always passed to `git clone` for now. Maybe we'd want a detached head or something for the initial repository state though.

ulfalizer:
Some more changes:

 - `west checkout` now supports a `-b` flag to create the branch before checking it out (if the branch already exists, it works like `west checkout`). This is probably similar to `repo start`.

 - `_inf()` now always adds `===` to the beginning of messages, to clearly separate `west` output from `git' output.

Another thing I've been thinking of: I wonder if the `manifest-rev` branch could be skipped whenever `revision` points to a branch, instead tracking the upstream branch directly. That would give more useful information when you run `git` commands. Only tricky part is if a revision is changed from a branch to a SHA or vice versa.

ulfalizer:
@mbolivar 
Re. tests, I still think the command set is being hashed out. Might be better to add them once things stabilize a bit, to avoid having to constantly rewrite tests.

mbolivar:
> Re. tests, I still think the command set is being hashed out. Might be better to add them once things stabilize a bit, to avoid having to constantly rewrite tests.

Perhaps just running the commands in a temporary directory and making sure no exceptions are raised? That should be enough to catch missing language features. Not sure how much work that would be.

ulfalizer:
@mbolivar 
Should be doable. I'll have a look.

ulfalizer:
@mbolivar 
Added some basic tests.

The way they hook into the project command classes is a bit hacky. I wonder if there could be some top-level function in West that just parses arguments and runs commands, without any tricky side effects. It would be handy to use in tests too.

ulfalizer:
Maybe we'd want to dynamically create some Git repositories for the tests later as well, to avoid a bunch of network traffic.

tests/west/project/test_project.py
```diff
@@ -0,0 +1,200 @@
...
+            command_object.do_add_parser(parser.add_subparsers())
+
+            # Pass the parsed arguments and unknown arguments to run it
+            command_object.do_run(*parser.parse_known_args(shlex.split(cmd)))
+            break
+    else:
+        assert False, "unknown command " + command_name
+
+
+# Example taken from https://docs.pytest.org/en/latest/fixture.html
```
mbolivar: More idiomatic would be to just use https://docs.pytest.org/en/latest/tmpdir.html

ulfalizer: Ok, will fix. Don't merge just yet.

mbolivar: OK

ulfalizer: Fixed

src/west/util.py
```diff
@@ -20,3 +21,26 @@ def wrap(text, indent):
...
+    '''
+    Returns the absolute path of the first directory containing a .west/
+    directory, searching the current directory and its parents.
+
+    Raises WestNotFound if no .west/ directory is found.
+    '''
+    search_dir = os.getcwd()
+
+    # While the directory is not the root directory...
+    while search_dir != os.path.dirname(search_dir):
```
ulfalizer: Noticed a subtle bug here. Won't find a `.west/` in the root directory.

Guessing it's no panic though. :P

mbolivar: Heh, whoops :), nice catch!

PR #12, opened 2018-08-21
Shippable integration

mbolivar:
This ought to provide shippable integration for the recently merged test suite. It's my first time though, so I expect it might have some issues (and I haven't set up an account on shippable itself, just tried to do the local integration here).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,3 +4,4 @@ west.egg-info/
 __pycache__/
 .pytest_cache/
 .eggs/
+shippable/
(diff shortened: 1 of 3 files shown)


mbolivar:
@nashif I've followed the shippable tutorials for the repository side. Do you need anything else from me, or does Zephyr already have an account etc. with Shippable this can be hooked into?

nashif:
lets see if shippable works...

mbolivar:
As discussed on IRC, let's postpone 3.7 testing for now. Looks like it's working for other versions.

PR #13, opened 2018-08-23
cmd: build: fix force_cmake calculation

mbolivar:
The logic to decide when to re-run CMake is wrong.


src/west/cmd/build.py
```diff
@@ -101,7 +101,9 @@ def do_run(self, args, ignored):
         self._setup_build_dir()
         if is_zephyr_build(self.build_dir):
             self._update_cache()
-        self.force_cmake = self.args.cmake or self.args.cmake_opts
+            self.force_cmake |= bool(self.args.cmake or self.args.cmake_opts)
```
ulfalizer: Can `force_cmake`get set before this point? Could just make it a plain `= self.args.cmake or self.args.cmake_opts` otherwise I guess.

Not sure how bitwise operators work on `bool` in Python either, to be honest. They seem to preserve the boolness somehow at least...

ulfalizer: Looks like `x |= y` with `x` and `y` works due to the result of arithmetic operators having the "narrowest" type out of `x` and `y`, which is `bool`.

Maybe `self.use_cmake = self.use_cmake or bool(self.args.cmake or self.args.cmake_opts)` or something would be more readable.

mbolivar: All done! Thanks. I've split the rename into a separate patch.

PR #14, opened 2018-08-24
Add 'forall' command

ulfalizer:
Passes its command as-is to the shell, within the repositories of each
of the specified projects (or all cloned projects by default).

I was thinking of making the shell quoting for long commands optional,
by joining all the words with a space or the like, but it might get
tricky with argparse. Could look into it more later.

Piggyback some cleanup: Make the absolute path of a project available in
project.abspath, to avoid having to do a bunch of join()s with
util.west_topdir().

Fix some copy-paste comment mess-ups in the tests too.

### DIFF
--- a/src/west/cmd/project.py
+++ b/src/west/cmd/project.py
@@ -225,6 +225,39 @@ class Status(WestCommand):
                 _git(project, 'status', extra_args=user_args)
 
 
+class ForAll(WestCommand):
+    def __init__(self):
+        super().__init__(
+            'forall',
+            _wrap('''
+            Runs a shell (Linux) or batch (Windows) command within the
+            repository of each of the specified projects (default: all cloned
+            projects). Note that you have to quote the command if it consists
+            of more than one word, to prevent the shell you use to run 'west'
+            from splitting it up.
+
+            Since the command is run through the shell, you can use wildcards
+            and the like.
+
+            For example, the following command will list the contents of
+            proj-1's and proj-2's repositories on Linux, in long form:
+
+              west forall -c 'ls -l' proj-1 proj-2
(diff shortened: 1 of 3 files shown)


src/west/cmd/project.py
```diff
@@ -225,6 +225,39 @@ def do_run(self, args, user_args):
...
 
 
+class ForAll(WestCommand):
+    def __init__(self):
+        super().__init__(
+            'forall',
+            _wrap('''
+            Runs a shell (Linux) or batch (Windows) command within the
+            repository of each of the specified projects (default: all cloned
+            projects). Note that you have to quote the command if it consists
```
mbolivar: FWIW, I think requiring the quoting is a fine thing. Much less ambiguity in the presence of multiple shells we need to support.

src/west/cmd/project.py
```diff
@@ -225,6 +225,39 @@ def do_run(self, args, user_args):
...
+
+              west forall -c 'ls -l' proj-1 proj-2
+            '''))
+
+    def do_add_parser(self, parser_adder):
+        return _add_parser(parser_adder, self, _command_arg, _project_list_arg)
+
+    def do_run(self, args, user_args):
+        for project in _projects(args):
+            if _cloned(project):
```
mbolivar: This idiom of "loop over cloned projects" is showing up in multiple places. At some point, it might be worth having a `_cloned_projects()` helper that returns an iterator for that, or adding an `if_cloned` kwarg to `_projects()`.

ulfalizer: Yep, was starting to notice that repetition too. I'll take care of it in the next cleanup pass. :)

PR #17, opened 2018-08-27
Make 'west branch' by itself list all local branches

ulfalizer:
Analogous to running 'git branch', except for all projects. Also lists
the project(s) each branch appears in.

Example output:

    $ python3 -m west branch
    FIX-FOO            zephyr, Kconfiglib
    manifest-rev       zephyr, net-tools, Kconfiglib
    master             zephyr, net-tools
    zephyr             Kconfiglib

I'm planning to change 'git clone' into 'git init' + 'git remote add' +
'git fetch' later, which avoids creating the last two local branches.
Maybe manifest-rev should be hidden as well.

### DIFF
--- a/src/west/cmd/project.py
+++ b/src/west/cmd/project.py
@@ -124,22 +124,39 @@ class Branch(WestCommand):
         super().__init__(
             'branch',
             _wrap('''
-            Create topic branch.
+            Create a branch or list branches, in multiple projects.
 
             Creates a branch in each of the specified projects (default: all
-            cloned projects).
+            cloned projects). The new branches are set to track '{}'.
 
-            The new branches are set to track '{}'.
+            With no arguments, lists all local branches along with the
+            repositories they appear in.
 
             '''.format(_MANIFEST_REV_BRANCH) + _MANIFEST_REV_HELP))
 
     def do_add_parser(self, parser_adder):
-        return _add_parser(parser_adder, self, _branch_arg, _project_list_arg)
+        return _add_parser(parser_adder, self, _opt_branch_arg,
+                           _project_list_arg)
 
     def do_run(self, args, user_args):
-        for project in _projects(args):
-            if _cloned(project):
+        # Generator
+        projects = (project for project in _projects(args) if _cloned(project))
+
+        if args.branch:
+            # Create a branch in the specified projects
+            for project in projects:
                 _create_branch(project, args.branch)
+        else:
+            # No arguments. List local branches from all projects along with
+            # the projects they appear in.
+
+            branch2projs = collections.defaultdict(list)
+            for project in projects:
+                for branch in _branches(project):
+                    branch2projs[branch].append(project.name)
+
+            for branch, projs in sorted(branch2projs.items()):
+                log.inf('{:18} {}'.format(branch, ", ".join(projs)))
 
 
 class Checkout(WestCommand):
@@ -275,7 +292,10 @@ _b_flag = _arg(
     action='store_true',
     help='create the branch before checking it out')
 
-# Common branch argument
+# Optional branch argument
+_opt_branch_arg = _arg('branch', nargs='?', metavar='BRANCH_NAME')
+
+# Mandatory branch argument
 _branch_arg = _arg('branch', metavar='BRANCH_NAME')
 
 # Common project list argument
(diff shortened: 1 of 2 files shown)


ulfalizer:
https://github.com/zephyrproject-rtos/west/pull/16 needs to go in before the clone-related changes.

src/west/cmd/project.py
```diff
@@ -124,22 +124,38 @@ def __init__(self):
...
 
     def do_add_parser(self, parser_adder):
-        return _add_parser(parser_adder, self, _branch_arg, _project_list_arg)
+        return _add_parser(parser_adder, self, _opt_branch_arg,
+                           _project_list_arg)
 
     def do_run(self, args, user_args):
-        for project in _projects(args):
-            if _cloned(project):
+        projects = [project for project in _projects(args) if _cloned(project)]
```
mbolivar: I'm not asking for a change here, but I'm curious why you chose a list rather than a tuple comprehension here, since projects isn't mutated.

src/west/cmd/project.py
```diff
@@ -124,22 +124,38 @@ def __init__(self):
...
+            # No arguments. List local branches from all projects along with
+            # the projects they appear in.
+
+            branch2projs = collections.defaultdict(list)
+            for project in projects:
+                for branch in _branches(project):
+                    branch2projs[branch].append(project.name)
+
+            for branch, projs in sorted(branch2projs.items()):
+                log.inf('{:18} {}'.format(branch, " ".join(projs)))
```
mbolivar: Would you mind comma-separating the projects? I know we don't really support paths with whitespaces in Zephyr now, but it'd be nice not to rely on that.

src/west/cmd/project.py
```diff
@@ -493,6 +512,17 @@ def _verify_repo(project):
                       "'git rev-parse --show-cdup'")
 
 
+def _branches(project):
+    # Returns a sorted list of all local branches in 'project'
```
mbolivar: Is this really sorted?

src/west/cmd/project.py
ulfalizer: Mostly handier syntax to be honest, though it's inconsistent. There are no tuple comprehension in Python.

The code below would turn `projects` into a generator instead, that can only be iterated over once (the parentheses are redundant, and the thing on the right is a *generator expression*):

    projects = (project for project in _projects(args) if _cloned(project))

Would have to make it something like this instead to get a tuple:

    projects = tuple(project for project in _projects(args) if _cloned(project))

In that case, the tuple constructor runs the generator to get the elements.

src/west/cmd/project.py
ulfalizer: Sure, will do. Was thinking space-separation might be nicer for scripts and the like, but that's kinda obscure...

src/west/cmd/project.py
ulfalizer: Yep, because `git for-each-ref` returns them sorted. Will add a clarifying comment.

src/west/cmd/project.py
mbolivar: > In that case, the tuple constructor runs the generator to get the elements.

Sorry for being sloppy; I did mean `tuple(generator expression)`.

Though since the variable is iterated over only once in the routine, the "bare" generator would work as well and not use any extra characters, no?

src/west/cmd/project.py
mbolivar: Thanks for explaining.

src/west/cmd/project.py
ulfalizer: Oh, right, forgot that it's only iterated once, just in separate paths.

Yeah, could use a bare generator then. Wonder if people would find that confusing though, since generator expressions are kinda obscure...

src/west/cmd/project.py
ulfalizer: Done

src/west/cmd/project.py
ulfalizer: Turns out you need the parentheses anyway.

Feels a bit risky in case someone thinks they can iterate it twice in the future, but could give it a try. ;)

src/west/cmd/project.py
ulfalizer: Except if you have spaces in repository names at least... seems like a recipe for pain.

mbolivar: Fair point -- let's keep this as is, then?

src/west/cmd/project.py
mbolivar: > Turns out you need the parentheses anyway.

Yeah, generator expressions are always parenethesized -- it's just that the same parentheses used for a function call etc. can do double duty for the generator.

PR #20, opened 2018-08-31
Pypi prep

mbolivar:
A couple of patches I would like to merge before tagging v0.1.0 and uploading to PyPI.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -1,81 +1,110 @@
-This package contains the Zephyr RTOS meta tool, 'west'.
-
-WARNING
--------
-
-DO NOT INSTALL WITH "python3 setup.py install".
-
-Please use pip to install in development mode as documented below.
-
-Important Note
---------------
-
-West is distributed in two pieces:
-
-1. A bootstrap/wrapper script, which is distributed via PyPI.
-2. The "main" west package and entry points, which are fetched by the
-   bootstrap script.
-
-This somewhat unusual arrangement is because:
-
-- One of West's jobs is to manage interaction with Zephyr's multiple
-  Git repositories, including its own.
-- West is in its experimental stages and is moving quickly, meaning
-  users need to stay on HEAD.
-
-The default setup.py installs the **wrapper script only**.
+This is the Zephyr RTOS meta tool, ``west``.
 
 Installation
 ------------
 
-To install the West bootstrapping/wrapper script from this package in
-development mode, clone this repository and run this from the top
-level directory:
+Install west's bootstrapper with pip::
 
-$ pip3 install -e .
+  pip3 install west
 
-Then use the wrapper script to initialize a Zephyr installation with::
+Then install the rest of west and a Zephyr development environment in
+a directory of your choosing::
 
-  $ west init your/zephyr/install-dir
-  $ cd your/zephyr/install-dir
-  $ west <command>
+  mkdir zephyrproject && cd zephyrproject
+  west init
+  west fetch
 
-The ``west init`` call will:
+What just happened:
 
-- create your/zephyr/install-dir
-- clone the Zephyr manifest repository (whose URL can be overridden
-  with the ``-u`` option, and branch with ``--mr`` / ``--manifest-rev``)
-- clone the latest West repository (the URL override is ``-w``, and
-  revision/branch override is ``--wr`` / ``--west-rev``)
+- ``west init`` runs the bootstrapper, which clones the west source
+  repository and a *west manifest* repository. The manifest contains a
+  YAML description of the Zephyr installation, including Git
+  repositories and other metadata. The ``init`` command is the only
+  one supported by the bootstrapper itself; all other commands are
+  implemented in the west source repository it clones.
 
-Running ``west <command>`` from :file:`your/zephyr/install-dir` or
-underneath it will invoke west in "wrapper" mode: any commands other
-than ``init`` will be delegated to the West tree pulled by ``west
-init``.
+- ``west fetch`` clones the repositories in the manifest, creating
+  working trees in the installation directory. In this case, the
+  bootstrapper notices the command (``fetch``) is not ``init``, and
+  delegates handling to the "main" west implementation in the source
+  repository it cloned in the previous step.
 
-This arrangement may seem familiar to Android (platform, not app)
-developers. The source management features of West were indeed
-inspired by the Android Repo tool's features, but West makes
-significant departures from Repo's behavior.
+(For those familiar with it, this is similar to how Android's Repo
+tool works.)
 
-Alternative Usage
------------------
+Usage
+-----
 
-If you don't want to change your system outside of cloning this
-repository, you can also clone West's Git repository and run the
-package as a module:
+West has multiple sub-commands. After running ``west init``, you can
+run them from anywhere under ``zephyrproject``.
 
-$ python3 -m west
+For a list of available commands, run ``west -h``. Get help on a
+command with ``west <command> -h``. For example::
 
-Only do this if you know what you're doing (and make sure to install
-the requirements as specified by the install_requires line in
-setup.py).
+  $ west -h
+  usage: west [-h] [-z ZEPHYR_BASE] [-v]
+              {build,flash,debug,debugserver,attach,list-projects,fetch,pull,rebase,branch,checkout,diff,status,forall}
+              ...
+  [snip]
+  $ west flash -h
+  usage: west flash [-h] [-H] [-d BUILD_DIR] ...
+  [snip]
 
 Test Suite
 ----------
 
-To run the test suite, use:
+To run the test suite, run this from the west repository::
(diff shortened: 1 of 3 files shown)


README.rst
```diff
@@ -1,81 +1,99 @@
...
+To install everything in development mode instead, clone the West
+repository and run this from the top level directory::
+
+  pip3 install -e .
+
+If you don't want to change your system outside of cloning this
+repository, you can run the ``west`` package as a module::
 
-To run the test suite, use:
+  PYTHONPATH=/path/to/west/repository/src python3 -m west
```
carlescufi: I guess this only works on *nix. On Windows one would have to resort to things like [these](https://superuser.com/a/1049433). In particular the easiest is probaby `cmd /C "set PYTHONPATH=/path/to/west/repo && python3 -m west"`

README.rst
```diff
@@ -1,81 +1,98 @@
...
-----------
+  west init -w https://example.com/your-west-repository.git
+
+You can also add ``--west-rev some-branch`` to use ``some-branch``
+instead of ``master``.
+
+To use another manifest repository (optionally with ``--mr
+some-manifest-branch``)::
+
+  west init -u https://example.com/your-manifest-repository.git
```
carlescufi: This made me realize that a common usecase would be to specify both a non-default west location and also a non-default manifest repo. Maybe we could add a combined option like:
`west init -uw https://github.com/myuser`, and west would automatically append `/west.git` and `/manifest.git` to the URL in order to fetch those.

mbolivar: Eew :).

Sure, we could do something like that. Maybe `-b` or `--url-base`? I'm assuming this is post-0.1.0.

README.rst
```diff
@@ -1,81 +1,98 @@
...
-Please use pip to install in development mode as documented below.
+  pip3 install west
 
-Important Note
---------------
+Then install the rest of West and a Zephyr development environment in
+a directory of your choosing::
 
-West is distributed in two pieces:
+  mkdir zephyr-dir && cd zephyr-dir
```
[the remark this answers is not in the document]
mbolivar: We could, for sure, but I didn't want this to be ambiguous with the actual Zephyr tree that gets cloned by `west fetch`.

Do you want me to make this change?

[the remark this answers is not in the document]
carlescufi: Right, in that case how about `mkdir zephyrproject && cd zephyrproject`?

README.rst
carlescufi: >Eew :)
You mean eew `-uw`? `-b` is fine with me :)

README.rst
```diff
@@ -1,81 +1,98 @@
-This package contains the Zephyr RTOS meta tool, 'west'.
+This is the Zephyr RTOS meta tool, ``west``.
 
-WARNING
--------
+Installation
+------------
 
-DO NOT INSTALL WITH "python3 setup.py install".
+Install West's bootstrapper with pip::
```
carlescufi: I wonder if we should capitalize West, or leave it lowercase `west`.

README.rst
```diff
@@ -1,81 +1,98 @@
...
+- ``west init`` runs the bootstrapper, which clones the West source
+  repository and a *West manifest* repository. The manifest contains a
+  YAML description of the Zephyr installation, including Git
+  repositories and other metadata.
 
-- One of West's jobs is to manage interaction with Zephyr's multiple
-  Git repositories, including its own.
-- West is in its experimental stages and is moving quickly, meaning
-  users need to stay on HEAD.
+- ``west fetch`` clones the repositories in the manifest, creating
```
carlescufi: maybe comment here that `west fetch` is run by the freshly cloned `west` code

README.rst
```diff
@@ -1,81 +1,98 @@
...
-Installation
-------------
+Usage
+-----
 
-To install the West bootstrapping/wrapper script from this package in
-development mode, clone this repository and run this from the top
-level directory:
+West has multiple sub-commands. After running ``west init``, you can
+run them from anywhere under ``zephyr-dir``.
```
carlescufi: `zephyrproject`

README.rst
```diff
@@ -1,81 +1,98 @@
...
-WARNING
--------
+Installation
+------------
 
-DO NOT INSTALL WITH "python3 setup.py install".
+Install West's bootstrapper with pip::
 
-Please use pip to install in development mode as documented below.
+  pip3 install west
```
carlescufi: One of the things I feel are extremely useful and sorely miss on Windows cmd line is git bash autocomplete. I wonder if we could reuse their script somehow and actually install it during bootstrapping.

README.rst
```diff
@@ -1,81 +1,98 @@
...
+
+  west init -u https://example.com/your-manifest-repository.git
 
-To run the test suite, use:
+After ``init`` time, you can hack on the West tree in ``zephyr-dir``.
+
+To install everything in development mode instead, clone the West
+repository and run this from the top level directory::
+
+  pip3 install -e .
```
ulfalizer: Would be nice to have a note on how you run `west` after running this command (`python -m west`). A plain `west` doesn't work.

README.rst
mbolivar: Oh, just seems a bit weird to me personally, but I see the value for short command lines.

README.rst
mbolivar: Sure, I'll lowercase it.

README.rst
mbolivar: I'm not sure what this means ... ``west fetch`` isn't run by the code, not sure how code can run something. Sorry to be dense, could you please clarify?

README.rst
mbolivar: Taking a serious look at autocomplete on all the platforms is on my wishlist as well -- should we start backlogging this in the issues for this repo?

README.rst
mbolivar: It should... what goes wrong?

mbolivar:
Most changes addressed, some questions still remain though, @ulfalizer @carlescufi

README.rst
carlescufi: Just after installing `west` from PyPi with `pip`, you can run `west fetch`. The Python script that actually executes all the required `git fetch` operations is the one in the `west/` folder, not the one installed by pip, isn't it? This is no big deal, just a small clarification that might or might not be applicable.

README.rst
carlescufi: I think it makes sense, but this can wait if you don't want to add it now.

README.rst
carlescufi: Yes actually. I will create one right now.

README.rst
ulfalizer: Never mind. It's working.

Maybe a note could be added about needing `.local/bin/` in PATH.

mbolivar:
> Approving as-is, although there are improvements that can be postponed. For now the biggest priority is to get this up on PyPi.

I've added a --base-url option to the bootstrap code and addressed README comments.

Should be good to go -- @ulfalizer?

mbolivar:
> Should be good to go

I've also added a Python 2 version check and reordered the commits so the one resetting to 0.1.0 can be tagged properly.

carlescufi:
Is this ready for merging now?

mbolivar:
> do a point release when that's done

Sorry, I am wrong, no need to do another release for self-update, as that's non-bootstrap and won't get sent to pypi.

Moving west and the manifest would require changes to what's in pypi if we do it after making the first 0.1.0 upload.

README.rst
carlescufi: Nice, thank you

carlescufi:
@ulfalizer please re-review so we can merge this and then rebase your PR on top

carlescufi:
>The change to having west and manifest under a single top level directory would be pretty significant as well, since that would break existing installations and affect the bootstrapper. We should decide on that before asking for wider use.

That one is part of the bootstrap but in a separate PR, I suggest that we merge this first then rebase that PR on top.

As per publishing to pypi, I would publish as soon as this one is merged just because I'm nervous someone may take the `west` name otherwise.

mbolivar:
> That one is part of the bootstrap but in a separate PR, I suggest that we merge this first then rebase that PR on top.
> 
> As per publishing to pypi, I would publish as soon as this one is merged just because I'm nervous someone may take the west name otherwise

+1 to both.

README.rst
```diff
@@ -1,81 +1,110 @@
...
+
+After ``init`` time, you can hack on the west tree in ``zephyrproject``.
+
+To install everything in development mode instead, clone the west
+repository and run this from the top level directory::
+
+  pip3 install -e .
+
+(On Linux, make sure ``~/.local/bin`` is in your ``PATH`` if you go
+this route.)
```
ulfalizer: Maybe this could be added at the end, within the parenthesis: "Alternatively, you can run ``west`` with ``python -m west``."

PR #26, opened 2018-09-05
Store West source code and the manifest in west/

ulfalizer:
I was confused by the other design, because I intuitively expected West
metadata to go in .west/. To me the other design feels a little bit like
having an empty /home/foo/user/ directory that marks the /home/foo/
directory as containing the user's files.

Use west/ instead of .west/ to avoid "hiding" the implementation of the
building/flashing commands. I would've preferred to split them from the
repo-like functionality instead, but if they're going to be in the same
repository, then it's a decent comprise.

Remove the code that hides the west directory on Windows (via 'attrib')
as well, since it doesn't make much sense anymore.

This commit also fixes a small bug in west_topdir(): A west/ in the root
directory wasn't found.

### DIFF
--- a/src/bootstrap/main.py
+++ b/src/bootstrap/main.py
@@ -7,7 +7,6 @@
 
 import argparse
 import os
-from os.path import join
 import importlib
 import platform
 import subprocess
@@ -19,8 +18,7 @@ MANIFEST_REV_DEFAULT = 'master'
 WEST = 'west'
 WEST_DEFAULT = 'https://github.com/zephyrproject-rtos/west'
 WEST_REV_DEFAULT = 'master'
-# NB: on Windows, this will be hidden manually, using attrib.
-WEST_DIR = '.west'
+WEST_DIR = 'west'
 
 #
 # Helpers shared between init and wrapper mode
@@ -51,24 +49,6 @@ def find_west_topdir(start):
         raise WestNotFound()
     else:
         return find_west_topdir(dirname)
-
-
-def make_topdir_sentinel(directory):
-    '''Make the hidden sentinel that marks a West installation.
-
-    This is similar to .git or .repo directories: it marks the top
-    level directory of the installation. This allows runtime building
-    other paths, etc., which is how this script delegates to the
-    cloned west in wrapper mode (i.e. after 'west init' is run and the
-    "real" west repository has been cloned).'''
-    # The sentinel is a directory in case we ever need to stash
-    # anything inside. It could just as well be an empty file for now,
-    # but this is more future-proof.
-    sentinel = join(directory, WEST_DIR)
-    os.mkdir(sentinel)
-    # On Windows, we have to manually mark a file as hidden.
-    if platform.system() == 'Windows':
-        subprocess.check_call(['attrib', '+H', sentinel])
 
 
 def clone(url, rev, dest):
@@ -145,9 +125,14 @@ def init_bootstrap(directory, args):
     else:
         print('Initializing in', directory)
 
-    make_topdir_sentinel(directory)
-    clone(args.manifest_url, args.manifest_rev, join(directory, MANIFEST))
-    clone(args.west_url, args.west_rev, join(directory, WEST))
+    # Clone the west source code and the manifest into west/. Git will create
+    # the west/ directory if it does not exist.
+
+    clone(args.west_url, args.west_rev,
+          os.path.join(directory, WEST_DIR, WEST))
+
+    clone(args.manifest_url, args.manifest_rev,
+          os.path.join(directory, WEST_DIR, MANIFEST))
 
 
 def init_reinit(directory, args):
@@ -169,7 +154,7 @@ def wrap(argv):
                  'Use "west init" to install Zephyr here'.format(start))
     # Put the top-level west source directory at the highest priority
     # except for the script directory / current working directory.
-    sys.path.insert(1, join(topdir, WEST, 'src'))
+    sys.path.insert(1, os.path.join(topdir, WEST, 'src'))
     main_module = importlib.import_module('west.main')
     main_module.main(argv=argv)
 
--- a/src/west/cmd/project.py
+++ b/src/west/cmd/project.py
@@ -316,7 +316,7 @@ def _arg(*args, **kwargs):
 # Common manifest file argument
 _manifest_arg = _arg(
     '-m', '--manifest',
-    help='path to manifest file (default: <west-topdir>/manifest/default.yml)')
+    help='path to manifest file (default: west/manifest/default.yml)')
 
 # Optional -b flag for 'west checkout'
 _b_flag = _arg(
@@ -470,13 +470,11 @@ def _validate_manifest(manifest_path):
 
 
 def _manifest_path(args):
-    # Returns the path to the manifest file. Unless explicitly specified by the
-    # user, it defaults to manifest/default.yml within the West top directory.
+    # Returns the path to the manifest file. Defaults to
+    # .west/manifest/default.yml if the user didn't specify a manifest.
 
-    if args.manifest:
-        return args.manifest
-
-    return os.path.join(util.west_topdir(), 'manifest', 'default.yml')
+    return args.manifest or os.path.join(util.west_dir(), 'manifest',
+                                         'default.yml')
 
 
 def _fetch(project):
--- a/src/west/util.py
+++ b/src/west/util.py
@@ -27,20 +27,30 @@ class WestNotFound(RuntimeError):
     '''Neither the current directory nor any parent has a West installation.'''
 
 
+def west_dir():
+    '''
+    Returns the absolute path of the first west/ directory found, searching the
+    current directory and its parents.
+
+    Raises WestNotFound if no west/ directory is found.
+    '''
+    return os.path.join(west_topdir(), 'west')
+
+
 def west_topdir():
     '''
-    Returns the absolute path of the first directory containing a .west/
-    directory, searching the current directory and its parents.
-
-    Raises WestNotFound if no .west/ directory is found.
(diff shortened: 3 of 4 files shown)


ulfalizer:
CC @carlescufi @mbolivar

mbolivar:
I'm definitely interested in a compromise solution that works for everyone and what you have proposed seems a reasonable way forward.

One thing I would ask is that the directory not be hidden on Unix or Windows. There are lots of tools (ack, ripgrep, the silver searcher) that ignore hidden directories. Since users often need to know where the runner source code lives, hiding it in this way makes their lives harder.

carlescufi:
I completely agree with the proposal combined with @mbolivar's comment about `.west` not being hidden on Unix. Can we make it visible even though it begins with `.`?
I believe this will not harm usability for Unix users, since they already know that everything that begins with `.` is metadata, even if they can see it.

mbolivar:
> Can we make it visible even though it begins with .?

Doesn't work that way, I'm afraid, for ls or the other tools I have mentioned.


https://github.com/BurntSushi/ripgrep/issues/623

ulfalizer:
Yeah, the filesystem itself has no concept of hidden files on *nix. It's just a convention of utilities that files starting with `.` aren't shown.

Personally, I think it's enough to not hide it on Windows. I removed the `attrib` stuff there.

ulfalizer:
If we make `west` too "weird" there, I think we risk scaring other groups of developers away.

mbolivar:
> I think straying from conventions to be helpful can make things more confusing instead, and might scare away other devs.
> 

Other Linux devs like us, you mean?

>> Edit: and beyond that, having a directory that starts with a "." on Windows is "weird" in its own right.
> 
> Makes it consistent at least, and probably isn't horribly confusing. Think e.g. Vim uses _vimrc instead on Windows, but that might be some DOS compatibility thing, since it's ancient.

It would also be consistent just to name it "west". And that "probably isn't horribly confusing" is interesting -- no problem straying from Windows conventions, as long as it's consistent with what Linux users expect, seems to be what you're saying.

Given that Zephyr is for firmware developers, I can't say I really agree. I can't back it up, but I strongly suspect the large majority of FW developers use Windows, not Linux.

mbolivar:
> The . marks it as metadata rather than part of the projects.

Yup, I'm aware of that, and it is wrong in this case: the runner source code **is** part of the projects, not metadata, which is what inspired my concern in the first place.

Would you want `zephyr/scripts/.kconfig`? Why is this different?

ulfalizer:
> Yup, I'm aware of that, and it is wrong in this case: the runner source code is part of the projects, not metadata, which is what inspired my concern in the first place.

Hmm... I wonder if the real crux is that we're mixing two unrelated things into the same util.

There's the helpers for building and flashing, and then there's the repo-like.

For the repo-like, I think it makes sense to put its source code in `.repo-like` (`.west` in this case). That's just a mechanism for it to update itself, and pretty "internal" (though it's still nice to explain it).

For the building and flashing helpers, it makes more sense to put them together with the projects.

Thoughts?

carlescufi:
I believe we are overthinking this.
In my opinion west covers 2 different aspects:

* building, flashing, debugging
* multirepo

Those are both covered by west and its functionality, so for me the folder structure consistent with this would be:

```
.
`-- zephyrproject
    |-- crypto
    |   `-- mbedtls
    |-- hal
    |-- west
    |   |-- manifest
    |   `-- west
    `-- zephyr
```

I agree with @mbolivar that hiding the `west` folder is not desirable in our particular case. I also understand what @ulfalizer's concerns with having metadata in a visible folder, since that's not how tools like `git` do that. But this is a bit different from `git`, and I do think we want a single tool (and not `west-repo` and `west-build-tool`) so I am in favor of the tree structure I posted, nothing hidden.

carlescufi:
Adding @nashif for his take on this

ulfalizer:
@carlescufi 
What are the advantages of having a single command that combines repo-like functionality with building and flashing? I slept on it a bit, and now I'm pretty sure that the root cause is that we're putting unrelated things into the same command.

The `repo`-like command wants to self-update (just to save people having to `pip install --upgrade my-repo` I guess) . The building and flashing stuff is random helper scripts.

I think we might end up with a nicer design with a few small simple unglamorous utilities that do one thing each. At the code level, it's just moving some stuff around.

We could still say that the repo command is part of the "west suite", if we wanted to keep the branding. ;)

carlescufi:
>What are the advantages of having a single command that combines repo-like functionality with building and flashing?

That's a good question. I would argue that the main advantage for users is a single entry point. When you work with Zephyr you use `west` for everything related to dealing with the source code, retrieving it, building it and running it.
This might seem contrary to the Unix philosophy, and in some ways it is, but this is much easier to explain to newcomers than 2 tools: You can manage things manually, or you can use `west`, which does this for you. You can think of it a little bit like a command-line IDE like Visual Studio or Xcode. Those tools allow you to branch, push, commit, build, debug and flash.

nashif:
dont really care were the west .py code resides, .west seems reasonable to me, however we want to allow modification to manifest so having it in .west does not work. Obviously one can pull it just like any other repo into the top tree and have it alongside the other repos, in this case we should be able to use that copy when building instead of the one in .west.

In short, if we expect something to be modified by the user, it should not be in .west.

ulfalizer:
> > What are the advantages of having a single command that combines repo-like functionality with building and flashing?
> That's a good question. I would argue that the main advantage for users is a single entry point. When you work with Zephyr you use west for everything related to dealing with the source code, retrieving it, building it and running it.

I wonder if that really makes things easier. I don't expect the `git` command to build things for me, and I think I'd be confused if it did.

Could do a "reversal test". If there already was a `repo` (too bad they took the name...) command and a `build` helper, would you want to merge them into the same command?

> This might seem contrary to the Unix philosophy, and in some ways it is, but this is much easier to explain to newcomers than 2 tools: You can manage things manually, or you can use west, which does this for you. You can think of it a little bit like a command-line IDE like Visual Studio or Xcode. Those tools allow you to branch, push, commit, build, debug and flash.

Could be explained like "use the `repo` command to clone down repositories, and use the `build` command to build projects and flash them onto boards." I think people are less confused by small, simple utilities that do one thing each.

The `build` helper belongs with the projects though, not with the multi-git helper, so maybe it'd be explained in the building section.

carlescufi:
> wonder if that really makes things easier. I don't expect the git command to build things for me, and I think I'd be confused if it did.

Sure but then again `git` doesn't point to a static location by default on the network, and it is not tied to a particular project. I do think that a single tool makes things easier for the user.

>Could do a "reversal test". If there already was a repo (too bad they took the name...) command and a build helper, would you want to merge them into the same command?

not sure I follow. If there was one in `git`? or in Zephyr? in what form?

>Could be explained like "use the repo command to clone down repositories, and use the build command to build projects and flash them onto boards." I think people are less confused by small, simple utilities that do one thing each.

Except they wouldn't be called `repo` or `build`. They would be called `west-repo` or `zephyr-repo` and `west-build` or `zephyr-build`, and that is IMHO not as user friendly as `west`.

ulfalizer:
> Sure but then again git doesn't point to a static location by default on the network, and it is not tied to a particular project. I do think that a single tool makes things easier for the user.

Still think it's confusing to arbitrarily put unrelated stuff into the same command.

> Could do a "reversal test". If there already was a repo (too bad they took the name...) command and a build helper, would you want to merge them into the same command?

If there were already helpers around CMake integrated into Zephyr, where you could run `flash ...`, etc. (the naming and the parameters aren't important), and they weren't called "West", and we decided that we need multi-repo support, would you then want to move `flash ...`, etc. into the multi-repo tool?

> Except they wouldn't be called repo or build. They would be called west-repo or zephyr-repo and west-build or zephyr-build, and that is IMHO not as user friendly as west.

I wouldn't give them a  `west-` prefix (would need to come up with a different name than `repo` though, obviously :).

I think talking about "West" hurts, because it adds an implicit assumption about what belongs together. What we're really adding is

 - Building helpers
 - Flashing helpers
 - A `repo`-like command for managing multiple repositories

It's not clear to me that all of those belong together (though the first two probably do) and should all be implemented in the same place.

What we have is good enough for getting the technical bits done though, and it could easily be changed later.

ulfalizer:
Just to get things moving:

If it's decided that the building/flashing helpers and the ``repo``-like live in the same repository, then I don't have any strong opinions on directory layout and could go with whatever.

I was thinking for a moment that you could clone the building/flashing code separately and have ``.west`` be like ``.repo``, but then you'd get the repo code in the building/flashing repository as well, and it'd look weird.

carlescufi:
@ulfalizer thanks. So according to my IRC logs there is no support for splitting the tools into more than one.

It might not be ideal, but I checked with @mbolivar and he is fine with my proposal [here](https://github.com/zephyrproject-rtos/west/pull/26#issuecomment-419430267). If you don't have any objections we can go with that, but I am open to variants if you think another layout would be better. Ideas?

ulfalizer:
@carlescufi 
Could go with that. With the building/flashing tools in the same place as the repo tool, it's probably less pain to put it all out in the open.

ulfalizer:
@carlescufi 
Changed

mbolivar:
> @mbolivar can you take another look? I think this plus #20, which is already merged, leave `west` ready to submit to PyPi.

Ah whoops, just noticed that you wanted this in before 0.1.0... sorry, I wasn't expecting that because the final commit in #20 explicitly set the version to 0.1.0. I'm having another look now, but I've already tagged and uploaded it to PyPI. Since we're the only users I expect there's no issue bumping the release again.

carlescufi:
@mbolivar no  issues at all, can you please update PyPi with a new version including these changes once they are approved and merged by you?

carlescufi:
Removed DNM/RFC since this has now been discussed at length.

PR #36, opened 2018-09-20
[RFC] Packaging reorg

mbolivar:
This series makes various fixes to our packaging strategy, and reorganizes the bootstrap script so it lives in the west namespace, as discussed on IRC. I know we haven't decided to do this yet but I wanted to put the code up just for reference.

Posting as an RFC as while tests pass, I want to do some more testing in parallel to make sure I didn't break real use cases with the boards I have access to.


tejlmand:
Small comment so far.

I agree that we need to remove the __init__.py in top-level west, to make it namespace.
But are we sure we want to do that for all sub-packages as well ?

Keeping __init__.py on sub-packages keeps the import shorter for those who will be using `west` as a lib, while at the same time allow for the split of modules and placement of _bootloader in the `west` namespace.

mbolivar:
> I agree that we need to remove the `__init__.py` in top-level west, to make it namespace.
> But are we sure we want to do that for all sub-packages as well ?
>
> Keeping `__init__.py` on sub-packages keeps the import shorter for those who will be using `west` as a lib, while at the same time allow for the split of modules and placement of _bootloader in the `west` namespace.

My opinion is that forcing all the imports to be "west.foo.bar" by making everything a namespace package is more consistent, and makes it more clear when we're importing  west code. But I'm not married to the idea and would be fine with keeping some of the subpackages as "legacy portions" (in PEP 420 terms) to save typing on imports.

I'm going to move most of the unrelated patches into a cleanup pull request for 0.2.0 and reopen this as a non-RFC with that change as soon as I can.

PR #43, opened 2018-09-21
Remove unused logic related to ZEPHYR_BASE

ulfalizer:
`validate_context()` is only used to set `ZEPHYR_BASE` from `args.zephyr_base`
and vice versa, and `args.zephyr_base` is never read despite being set.
This made it look like the code did more than it actually does to me,
and confused me about the logic.

Simplify the code to just the core `ZEPHYR_BASE = args.zephyr_base`
logic.

If `ZEPHYR_BASE` needs to be read somewhere later in the code, it might be
cleaner to either read it directly from `os.environ` or from a separate
variable, to avoid modifying the argument list.

### DIFF
--- a/src/west/main.py
+++ b/src/west/main.py
@@ -44,24 +44,8 @@ COMMANDS = (
 '''Built-in West commands.'''
 
 
-class InvalidWestContext(RuntimeError):
-    pass
-
-
 def command_handler(command, known_args, unknown_args):
     command.run(known_args, unknown_args)
-
-
-def validate_context(args, unknown):
-    '''Validate the run-time context expected by west.'''
-    if args.zephyr_base:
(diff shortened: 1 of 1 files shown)


mbolivar:
Just needs a rebase

mbolivar:
> Just needs a rebase

Though note that I did originally write this code with solving #46 in mind, so we're just going to have to put something like it right back.

I am sympathetic to YAGNI type concerns but sometimes we are gonna need it...

mbolivar:
@ulfalizer do you want to update this now that #55 is merged, or would you like to drop this one?

mbolivar:
> @ulfalizer do you want to update this now that #55 is merged, or would you like to drop this one?

ping

mbolivar:
Dropping this one for now to keep the pull request list tidy, please reopen if you wish

PR #47, opened 2018-09-24
Simplify west/.west_toplevel existence checks

ulfalizer:
Directly check whether west/.west_toplevel exists instead of checking if
'west' is a directory whose contents include '.west_toplevel', which is
also a bit more efficient.


mbolivar:
Crap, the rebase broke the repository, will post a fix

Issue #46, opened 2018-09-23
west should set ZEPHYR_BASE, not look for it


mbolivar:
West knows where Zephyr is, since it parses the manifest.

This issue tracks extracting the manifest parsing code in the project command implementations into a helper module, where it can be used in the context validation code to find zephyr and set ZEPHYR_BASE in the environment before doing anything else.

carlescufi:
I assume this is the cause of the issue I'm seeing:

1. Install 0.2.0rc1 with pip
2. west init -u https://github.com/mbolivar/manifest --mr multirepo-test-manifest zephyrproject
3. source zephyr\zephyr-env.cmd
4. Build, flash successfully
5. cd %userprofile%
6. west --version:
```
fatal error: you must specify a command
usage: west [-h] [-z ZEPHYR_BASE] [-v] {flash,debug,debugserver} ...
```

mbolivar:
That's actually because the experimental west launchers in zephyr don't support the --version flag. This is being addressed in https://github.com/zephyrproject-rtos/zephyr/pull/10167.

tejlmand:
note, in case another manifest is used, which uses a forked version of zephyr it should be considered how west discovers which of the repo's in the manifest that is the correct zephyr.

Should it be required that:
```
path: zephyr
```
under the project containing Zephyr is used, if it is not Zephyr itself.

carlescufi:
As I see it we have the following issues to solve as part of this:

1. Make sure that users of older monorepo installations can continue to use `make/ninja flash`.
2. Make sure that users who do not want to use `west` directly can still build using CMake directly after the multirepo transition (provided they have cloned the correct repositories first).
3. Document that out-of-tree code (samples, drivers, etc) need to live within the current active wworkspace or ww (the folder F passed to `west init F`).

carlescufi:
@tejlmand 

Good point. Either fix the path to `zephyr` or add metadata to the manifest format.

mbolivar:
> Good point. Either fix the path to `zephyr` or add metadata to the manifest format.

Agree. We can have both though, if we want.

Proposal:

- the project named "zephyr" is ZEPHYR_BASE by default, if it exists.
- if your zephyr has another name, you need to add `is-zephyr-base: True` to the project definition in the manifest. At most one project shall have this flag set, or the manifest is invalid.
- a manifest without a ZEPHYR_BASE is invalid.

The diff to the manifest schema would be:

```
--- a/src/west/commands/schema.yml
[...]

PR #55, opened 2018-09-27
Manifest fixes and refactor

mbolivar:
Factor the manifest parsing to its own module to prepare to handle #46, and fix some issues in the manifest handling along the way. Add regression tests too.

Once we agree on the #46 strategy I'll send a follow-up implementing it using the new module.


mbolivar:
Thanks for the review and comment!

> Would it make sense to make a Manifest class, with methods instead ?

I thought about that, but then figured @ulfalizer would tell me we didn't need it yet and send a patch removing it ;).

I do think that's what will be needed to implement the #46 series, but was putting it off until then.

mbolivar:
Marking DNM for now since:

- we have agreement on the final implementation strategy
- it would be better to get this in rebased on top of #58

mbolivar:
@carlescufi @tejlmand @ulfalizer: this now:

Fixes #46 

I've rebased it on top of #58 so the west.xxx import style is used.

> Would it make sense to make a Manifest class, with methods instead ?

This is present as well.

src/west/manifest-schema.yml
```diff
@@ -133,3 +142,6 @@ mapping:
               clone-depth:
                 required: false
                 type: int
+              is-zephyr-base:
```
carlescufi: I think we could generalize this. I am thinking how best to do this but how about something like:

```
env-map:
   required: false
   type: str
```

So then you could do things like:

```
  projects:
    - name: zephyr
      env-map: ZEPHYR_BASE
    - name: socs
      env-map: SOC_DIR
    - name: boards
      env-map: BOARDS_DIR
```

src/west/manifest.py
```diff
@@ -115,9 +115,19 @@ def __init__(self, source_file=None, source_data=None):
         Each element's values are fully initialized; there is no need
         to consult the defaults field to supply missing values.'''
 
+        self.zephyr_base = None
+        '''Absolute path to the ZEPHYR_BASE directory implied by this
+        manifest, if any. By default this is the path to the project whose
+        name is "zephyr", if one exists. The user may also override
```
carlescufi: If we do switch to `env-map` then I would remove the guesswork completely and require that in the manifest.

src/west/main.py
```diff
@@ -65,16 +67,32 @@ def command_handler(command, known_args, unknown_args):
...
-        if 'ZEPHYR_BASE' not in os.environ:
-            log.wrn('--zephyr-base missing and no ZEPHYR_BASE',
-                    'in the environment')
-        else:
-            args.zephyr_base = os.environ['ZEPHYR_BASE']
+        manifest = Manifest.from_file()
+        zb = manifest.zephyr_base
+        zb_src = 'the manifest file {}'.format(manifest.path)
+
+    zb_env = os.environ.get('ZEPHYR_BASE')
```
carlescufi: Should you set the source here as well? That way we could later query where the `ZEPHYR_BASE` value came from regardless of where it did.

src/west/manifest-schema.yml
mbolivar: I tried to think about a general solution but I wasn't sure I had a "good enough" answer. Coming up with an ad-hoc hack we can deprecate later feels better than coming up with a general solution that doesn't quite fit.

I can see issues that make this potentially overly limiting:

- what if i wanted to set two or more environment variables based on some attribute of the project? (having the key be a string like ZEPHYR_BASE instead of a sequence or other compound data type means we can only get one value)
- what if i want to set a relative path instead of an absolute path?
- what if i wanted to append onto an existing environment variable (like PATH, PYTHONPATH) instead? (perhaps with a or platform-dependent configurable separator)
- what if i want to use a subpath inside a project? ("append this-project-abspath/bin to PATH")

mbolivar: > Coming up with an ad-hoc hack we can deprecate later feels better than coming up with a general solution that doesn't quite fit.

... so one potential solution is "use the project named zephyr if it exists, or the project with path zephyr if that fails" and we can defer trying to be clever.

src/west/main.py
mbolivar: That's a great idea.

src/west/manifest-schema.yml
carlescufi: > I tried to think about a general solution but I wasn't sure I had a "good enough" answer. Coming up with an ad-hoc hack we can deprecate later feels better than coming up with a general solution that doesn't quite fit.
> 
> I can see issues that make this potentially overly limiting:
> 
> * what if i wanted to set two or more environment variables based on some attribute of the project? (having the key^W edit: value be a string like ZEPHYR_BASE instead of a sequence or other compound data type means we can only get one value)
> * what if i want to set a relative path instead of an absolute path?
> * what if i wanted to append onto an existing environment variable (like PATH, PYTHONPATH) instead? (perhaps with a or platform-dependent configurable separator)
> * what if i want to use a subpath inside a project? ("append this-project-abspath/bin to PATH")

I see your point, but wouldn't all of those be solvable by adding further sub-options in the manifest later on? i.e. instead of going from the bottom up, we could go top-down by adding a generic environment variable setting in the manifest that can later be refined by adding to the schema in a way that we cover all the variants that you mention.

carlescufi: > > Coming up with an ad-hoc hack we can deprecate later feels better than coming up with a general solution that doesn't quite fit.
> 
> ... so one potential solution is "use the project named zephyr if it exists, or the project with path zephyr if that fails" and we can defer trying to be clever.

Actually going back to it now, I think we don't know enough at this point about the requirements and about the exact way we're gonna split Zephyr to take a decision, so I agree with your assessment above and would like to propose deferring the decision as well.

mbolivar: > I see your point, but wouldn't all of those be solvable by adding further sub-options in the manifest later on?

Not unless you want to deprecate the env-map proposal, or keep it but add other keys on the side which do "env-map-and-more". The pykwalify schema wants the values to be of a certain type, which we would have to change incompatibly to make the additions to `env-map` I am asking about.

src/west/main.py
carlescufi: I think I am ready to approve if you make this change.

mbolivar:
@carlescufi addressed your review comments, thanks!

This should be ready to review now.

src/west/main.py
mbolivar: > I think I am ready to approve if you make this change.

I've included this in the verbose output, let me know if you're looking for something else (like a ZEPHYR_BASE_ORIGIN variable which was also set in the environment).

Issue #62, opened 2018-10-02
West update can leave manifest and west in unusable states


mbolivar:
As a user, I expect that if `west update` fails (either run directly or as implied by `west fetch` or `pull`), that my installation is still in a usable state. However, the current implementation can leave either of these repositories in a broken (half-rebased) state if the update fails, with git merge conflict markers in source files.

To reproduce, just commit any changes that won't rebase cleanly and run west update.

The behavior instead should be to see if the rebase is possible to complete cleanly *without* damaging the existing installation, reporting an error if it fails with additional information on how to proceed.

ulfalizer:
Think some variant of https://github.com/zephyrproject-rtos/west/pull/81#discussion_r226635054 would solve it in a nicer and simpler way.

PR #64, opened 2018-10-04
Check if projects are up-to-date when rebasing

ulfalizer:
Commit messages say it all:

```
commands: project: Check if projects are up-to-date when rebasing

_update() updates the special west/ and manifest/ repositories and has
code for checking if HEAD contains all the commits from a particular
branch (is up-to-date re. the branch). Move that code into a separate
_up_to_date_with() function, and also use it when rebasing normal
projects.

This gives a less spammy message when running 'pull' and 'rebase' on
already up-to-date projects:

  $ west rebase

A function for checking how many commits a commit is behind/ahead
another commit would be nice to have in the future as well.

This commit also removes the special error message when self-updating
fails, returning to plain _git() with check=True (which points out the
repository path at least). I think it might be nicer to handle "context
hints" by raising an exception from _git_helper() on errors and catching
it, but I thought I'd add it separately.
```

```
commands: project: Rename 'exists' to 'cloned'
    
More explicit.
```

```
commands: project: Fix comment nits in _fetch()
    
 - Missing single quote after 'clone-depth'
    
 - Not just many server disallow fetching arbitrary commits by SHA, but
   most. I think 'almost all' might be accurate too.
```


ulfalizer:
Re. this part, I could add that to this PR later too, if you think the regressions's too bad to have in there for a few days.

```
This commit also removes the special error message when self-updating
fails, returning to plain _git() with check=True (which points out the
repository path at least). I think it might be nicer to handle "context
hints" by raising an exception from _git_helper() on errors and catching
it, but I thought I'd add it separately.
```

mbolivar:
> - Not just many server disallow fetching arbitrary commits by SHA, but most. I think 'almost all' might be accurate too.

FYI, it works fine if you stand up a private GitLab instance.

> This commit also removes the special error message when self-updating fails,

Given that we've tagged 0.2.0 and area asking for wider testing, and the presence of issue #62, I think keeping this error message is very important -- is it possible to address #62 with this series?

carlescufi:
@ulfalizer will you address #62 in this PR as per the [above comment](https://github.com/zephyrproject-rtos/west/pull/64#issuecomment-427421481)?

PR #66, opened 2018-10-04
Rename 'list-projects' to 'list'

ulfalizer:
I got tired of typing 'list-projects' all the time.

'list' might be a bit less obviously related to projects, but I don't
think it would be that bad. Feel free to complain.

### DIFF
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -24,10 +24,10 @@ from west.manifest import default_path, Remote, Project, \
 _MANIFEST_REV_BRANCH = 'manifest-rev'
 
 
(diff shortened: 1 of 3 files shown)


mbolivar:
Seems fine to me. Feel free to take the RFC off. I think if we want to add `--projects`, `--remotes`, etc. flags later for what to list, we can. I'm already concerned we have a bunch of commands so keeping the count down this way LGTM.

ulfalizer:
Alright, removed RFC.

carlescufi:
I like the idea, if there's no objections I will merge this

PR #74, opened 2018-10-09
west should be able to list boards

sathishscripts:
Signed-off-by: Sathishkumar Duraisamy <sathishkumar.duraisamy@yahoo.com>

### DIFF
--- a/src/west/commands/build.py
+++ b/src/west/commands/build.py
@@ -190,11 +190,15 @@ class Build(WestCommand):
         self.source_dir = os.path.abspath(source_dir)
 
     def _get_available_boards_str(self):
+        """
+        Creates a string of available boards ready to pass via _check_force
+        :return: str
+        """
         brd_path = os.getenv("ZEPHYR_BASE") + "/boards"
         # Creates list of available architectures but also includes folder like common
         archs = [arch for arch in os.listdir(brd_path) if os.path.isdir(os.path.join(brd_path,arch))]
-        # Creates list of all board files
-        brd_files = glob.glob(brd_path + "/**/*.yaml", recursive=True)
+        # Creates list of all board files. The pattern "/**/*.yaml" not supported in python 3.4
+        brd_files = glob.iglob(brd_path + "/*/*/*.yaml") # /arch/soc/board.yaml
         # Lets create boards files database
         brds_db = {}
         for brd in brd_files:
--- a/src/west/commands/build.py
+++ b/src/west/commands/build.py
@@ -4,7 +4,7 @@
 
 import argparse
 import os
-from pathlib import Path
+import glob
 
 from west import log
 from west import cmake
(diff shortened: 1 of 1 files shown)


sathishscripts:
@mbolivar Can you please review the code? Is it possible/recommended to use kconfiglib?

src/west/commands/build.py
```diff
@@ -188,6 +189,20 @@ def _setup_source_dir(self):
...
+        brds_paths = path.glob("**/Kconfig.board")
+        brds_db = {}
+        for brd in brds_paths:
+            try:
+                brds_db[brd.parent.parent.name].append(brd.parent.name)
+            except KeyError:
+                brds_db[brd.parent.parent.name] = [brd.parent.name, ]
+        brds_str = ""
+        for arch in brds_db.keys():
+            brds_str += "\r\n" + arch + ":\r\n\t\t" + "\r\n\t\t".join(brds_db[arch])
```
mbolivar: Can you use print() instead of manually adding Windows newlines? West is a cross-platform program and needs to work on macOS and Linux, where just newlines are used.

src/west/commands/build.py
```diff
@@ -188,6 +189,20 @@ def _setup_source_dir(self):
             source_dir = os.getcwd()
         self.source_dir = os.path.abspath(source_dir)
 
+    def _get_available_boards_str(self):
+        path = Path(os.getenv("ZEPHYR_BASE")+"/boards")
```
mbolivar: I've never used pathlib. It looks really nice and convenient. Maybe we'll switch to it someday in west! But for now, we just use os.path -- can you please keep that consistent and use that too?

src/west/commands/build.py
```diff
@@ -188,6 +189,20 @@ def _setup_source_dir(self):
...
+        brds_paths = path.glob("**/Kconfig.board")
```
mbolivar: I think you should take a look at usage.cmake in the Zephyr repository to see how they do it. Maybe you can use the same approach here, or even better, try to extract that component as a script you can run with `cmake -P`, so there is just one implementation. Either way will keep west consistent with cmake (the latter way of running cmake directly is preferred of course).

sathishscripts:
Shortly I update the changes required.

sathishscripts:
I have tried writing test cases, but environment variable "ZEPHYR_BASE" not available in test_environment. So skipped!!!

mbolivar:
Hi @sathishscripts, thanks a lot and sorry for the slow response. I took a look at your updates but they appear to have been done as additional commits. The Zephyr contribution workflow requires that we rebase our changes into an existing patch series.

You can see the full details here: https://docs.zephyrproject.org/latest/contribute/contribute_guidelines.html#contribution-workflow

Especially see step 15. Could you take a look at that and have a go?

> I have tried writing test cases, but environment variable "ZEPHYR_BASE" not available in test_environment. So skipped!!! Any recommendations?

For a smaller change like this I think just testing locally is OK given the test case limitations. We need to think about adding a test fixture that provides a "real-ish" Zephyr tree in the test environment without having to clone a full one.

sathishscripts:
HI @mbolivar, I was also away with few personal works. Today, I shall work on this. I thought it would be better to have test case that checks against cmake, so that it keeps in sync with cmake. For the moment now, may be skip test case.

PR #80, opened 2018-10-17
west: main: don't swallow Exception

### LINKED ISSUE #79: West should not hide backtraces for internal errors
West currently catches all exceptions in `main.py` and prints just the exception message, telling users to pass `-v` to see the full backtrace.

The issue with this is that if an exception is raised due to a problem with the code, then the backtrace is useful information. Removing it is counterproductive, especially if the problem can't be easily reproduced.

Another issue is that it mixes up expected and unexpected exceptions. For expected exceptions, they ought to be turned into friendly output closer to the point where they're generated, instead of throwing an exception all the way up to `main.py`. We already do this with `log.die()` and the like.

Note that I'm not saying that all exceptions should generate backtraces. Just the unexpected exceptions should. Expected errors should be handled where they're generated, and turned into friendlier output there.

mbolivar:
Fixes #79.

### DIFF
--- a/src/west/main.py
+++ b/src/west/main.py
@@ -219,12 +219,7 @@ def main(argv=None):
     except CommandContextError as cce:
         log.die('command', args.command, 'cannot be run in this context:',
                 *cce.args)
-    except Exception as exc:
-        log.err(*exc.args, fatal=True)
-        if args.verbose:
-            raise
-        else:
-            log.inf(for_stack_trace)
+
 
 if __name__ == "__main__":
     main()


mbolivar:
> Why propagate the exception up to the user without `-v`? I've always thought it's ugly to see a stack trace instead of the exception string description when something unexpected happens.

We still swallow the stack trace for expected exceptions (like CalledProcessError) and allow the user to pass -v in those cases to get the full stack trace. We can add more expected exception types to this list for sure.

I believe @ulfalizer's point as he described in #79  is that unexpected exceptions (the blanket "Exception" catchall at the bottom) should propagate to the user. I think he has a good point that if an error is hard to reproduce and unexpected, the stack trace is valuable.

mbolivar:
@ulfalizer I'll wait for your +1 before merging to confirm this fixes the issue you reported as desired.

Issue #59, opened 2018-09-28
West update should be skipped if west or manifest is on a feature branch


mbolivar:
The current auto-update behavior could use a bit of tweaking to make it more intuitive to those users who are editing their manifest or west repository and don't want to type --no-update all the time.

The issue is that it's not very convenient to switch to a manifest branch with custom changes and then issue 'west fetch', 'west pull', etc. commands, as that would try to update the manifest itself as well, which isn't what the user wants IMO. We could emulate the repo behavior instead, which is only to update the manifest itself to incorporate remote changes if the manifest repository is on the default branch tracking the remote which is created by west init.

This applies at least to the manifest, though for consistency it could be applied to west as well.

mbolivar:
@carlescufi @ulfalizer @tejlmand I actually lean towards calling this a bug -- I don't think it should block 0.2.0 but I do think we should fix it as soon as possible. Thoughts?

carlescufi:
>The issue is that it's not very convenient to switch to a manifest branch with custom changes and then issue 'west fetch', 'west pull', etc. commands, as that would try to update the manifest itself as well, which isn't what the user wants IMO

If you are switching to a manifest branch, you do want to pick up the new manifest from that branch, don't you? Or maybe I'm not following here

mbolivar:
> If you are switching to a manifest branch, you do want to pick up the new manifest from that branch, don't you? Or maybe I'm not following here

Here's the workflow I'm thinking of:

```
$ cd west/manifest
$ git checkout -b my-manifest-branch
$ $EDITOR default.yml   # add some projects, etc.; this is exactly the manifest I want now
$ west fetch
```

As a user, I expect `west fetch` not to rebase my manifest or otherwise touch it **in this case**, because it's checked out on a feature branch, not the original branch which tracks the remote I gave to `west init`.

This is how repo behaves, and I think it is the least surprising thing. If OTOH my manifest were on the 'default' branch which just tracks upstream, an update would make sense.

I think similar arguments apply to west itself.

carlescufi:
@mbolivar now I understand the issue. Absolutely. I concur with your assessment.

tejlmand:
How about the situation, where a user follows a given branch during development.
as example:
```
$ cd west/manifest
$ git checkout dev-branch-to-follow
$ west fetch
```
in case updates are made to the manifest on the remote branch, then you do want to fetch those changes.

I think getting back to the proposal here:
[...]

PR #81, opened 2018-10-19
commands: project: Try to undo merge conflicts on west/manifest updates

### LINKED ISSUE #62: West update can leave manifest and west in unusable states

ulfalizer:
By request. If something that looks like a merge conflict is detected,
'git rebase --abort' is run automatically.

Guess that there's a merge conflict if no rebase is in progress before
'git rebase', but a rebase is in progress after it. This avoids
'git --reset'ing away deliberate rebases.

Fixes #62

### DIFF
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -595,11 +595,23 @@ def _up_to_date_with(project, revision):
                         capture_stdout=True).stdout
 
     # SHA of the most recent commit shared between HEAD and 'revision'
-    merge_base = _git(project, 'merge-base HEAD ' + revision,
+    merge_base = _git(project, 'merge-base HEAD -- ' + revision,
                       capture_stdout=True).stdout
 
     # If they're equal, HEAD has all the commits in 'revision'
     return revision_sha == merge_base
+
+
+def _rebase_in_progress(project):
+    # Returns True if a rebase is currently in progress. Based on ideas from
+    # https://stackoverflow.com/questions/3921409/how-to-know-if-there-is-a-git-rebase-in-progress.
+
+    # _git() changes the working directory, so we need an absolute path
+    git_dir = _git(project, 'rev-parse --absolute-git-dir',
+                   capture_stdout=True).stdout
+
+    return os.path.isdir(os.path.join(git_dir, 'rebase-merge')) or \
+           os.path.isdir(os.path.join(git_dir, 'rebase-apply'))
 
 
 def _cloned(project):
@@ -697,15 +709,40 @@ def _update_special(name):
 
         if not _up_to_date_with(project, "FETCH_HEAD^{commit}"):
             # New upstream changes
-            _git(project, 'rebase FETCH_HEAD^{commit}')
-            _inf(project, 'Updated (rebased) (name-and-path) to the '
-                          'latest version')
 
-            if project.name == 'west':
-                # Signal self-update, which will cause a restart. This is a
-                # bit nicer than doing the restart here, as callers will
-                # have a chance to flush file buffers, etc.
-                raise WestUpdated()
+            # Check if the user is currently rebasing. That would cause the
+            # rebase below to fail, and we don't want to wipe out their work.
+            if _rebase_in_progress(project):
+                _wrn(project, "Couldn't auto-update (name-and-path), because "
+                              "it's in the middle of a rebase")
+
+            elif _git(project, 'rebase FETCH_HEAD^{commit}', check=False) \
+                 .returncode:
+
+                _wrn(project, 'Failed to auto-update (rebase) (name-and-path) '
+                              'to the latest version')
+
+                # A rebase can fail for many reasons, including uncommited
+                # changes
+                if _rebase_in_progress(project):
+                    # Looks like we have a failed rebase, caused by us. Try to
+                    # undo it, and keep going. This feels a bit iffy, but was
+                    # requested.
+                    _wrn(project,
+                         "Running 'git rebase --abort' to undo what looks "
+                         'like a failed auto-update (rebase) of '
+                         '(name-and-path)')
+                    _git(project, 'rebase --abort')
+
+            else:
+                _inf(project, 'Updated (rebased) (name-and-path) to the '
+                              'latest version')
+
+                if project.name == 'west':
+                    # Signal self-update, which will cause a restart. This is a
+                    # bit nicer than doing the restart here, as callers will
+                    # have a chance to flush file buffers, etc.
+                    raise WestUpdated()
 
 
 _FAILED_UPDATE_MSG = """


mbolivar:
Hey @ulfalizer, which request was this? There's no "fixes" tag attached to the PR, so I'm not sure which issue this is about.

On the off chance that it's #62, I wonder if there isn't a way to reset the tree to a freshly named branch, attempt the rebase, commit it on success, and abort on failure (deleting the fresh name either way).

mbolivar:
> There's no "fixes" tag attached to the PR, so I'm not sure which issue this is about.

Ah, whoops, I see it is #62 in the commit message. Above question stands, then :).

ulfalizer:
@mbolivar 
IMO, this is already too magic, but pretty understandable still. I think I managed to cover most of the corner cases that get hidden in something as simple as "just `git rebase --abort`".

Can we compromise and use this version?

ulfalizer:
By the way, a neat way to prevent many merge failures would be to move the non-repo-related commands out of the `west` command. That way, people who work on the flashing tools and the like will never run into them.

I could open an issue for it later.

ulfalizer:
> On the off chance that it's #62, I wonder if there isn't a way to reset the tree to a freshly named branch, attempt the rebase, commit it on success, and abort on failure (deleting the fresh name either way).

Might've misunderstood this. Thought you were talking about leaving the user's work in some other branch, to "force" the update to succeed.

Maybe you just meant it is an alternative way to do it. It should be able to deal with the user already being in a rebase, and with uncommitted local changes (I wonder if there's other stuff I haven't thought of). If it can do that while being simpler than the `git rebase --abort` version, then I'm all for.

src/west/commands/project.py
```diff
@@ -697,15 +709,40 @@ def _update_special(name):
...
+                # changes
+                if _rebase_in_progress(project):
+                    # Looks like we have a failed rebase, caused by us. Try to
+                    # undo it, and keep going. This feels a bit iffy, but was
+                    # requested.
+                    _wrn(project,
+                         "Running 'git rebase --abort' to undo what looks "
+                         'like a failed auto-update (rebase) of '
+                         '(name-and-path)')
+                    _git(project, 'rebase --abort')
```
tejlmand: I assume just aborting here and not raising an error means the update will continue with the manifest repo (in case that `west` itself failed, line 90 will continue with `_update_manifest()`)

I'm don't believe that was the intention, as I assume that in worst case you can end-up in a situation where an updated manifest and an old `west` version can happen to be incompatible, leaving the project in bad shape.

I think the request by @mbolivar and @carlescufi  was to completely abort, and provide info on how top proceed.

Also the other way around, the manifest fails, and `west` is updated, do we want to roll-back `west` ?
(This may not be as severe, as I will assume `west` to be back-ward compatible with old manifest formats)

ulfalizer: The old behavior was to just raise an error telling the user to fix the state of the repository, and not do anything about the merge conflict.

I liked it that way, but people requested some kind of automatic roll-back (or "only update if it's possible", but you have to actually try to rebase in Git to figure out that it isn't possible).

If we turned this into an error instead, then it would be the same as the old behavior, except that the merge conflict would be removed. Rerunning west would try to update, get a merge conflict, remove the merge conflict, and then error out.

That might be the worst of both worlds for the user, because they'd both have to manually fix the state of the repository, and also wouldn't have a merge conflict to go on.

Maybe you were thinking about going back to the original behavior though.

ulfalizer: Hmm... another option here would be to simply refuse to update west and/or the manifest if the current commit is not an ancestor of the upstream revision (so that it can't be fast-forwarded). Should probably also check if the working tree or staging area is dirty in that case (and probably other stuff I haven't thought of).

IMO, the original behavior was nice and simple. Combining it with https://github.com/zephyrproject-rtos/west/issues/59 would remove some of the gotchas.

tejlmand: My understanding of what was requested, was to leave the state as it was before the rebase, ie. rolling back everything, and at the same time issue a error.

At least, that's how I interpret this:
> The behavior instead should be to see if the rebase is possible to complete cleanly without damaging the existing installation, reporting an error if it fails with additional information on how to proceed.

but guess @mbolivar  can clarify

The old behavior was to just abandon the rebase and ask the user to clean up.
In my opinion, that's not so nice.

We don't know if he prefers a rebase, and solving the conflicts or if it is just temp work, which he wants to discard.
Or if he might even want to force push his work, and then redo the fetch.

So rolling back to where he was prior to the command, as if we never touched his work is the nicest thing, we can do.

If branches are used as suggested
> On the off chance that it's #62, I wonder if there isn't a way to reset the tree to a freshly named branch, attempt the rebase, commit it on success, and abort on failure (deleting the fresh name either way).

then you create the branch on all projects, and in case any of them fail, we can abort and remove all branches without committing, and user's project is left untouched.

Rolling back on e.g. `west` is harder if the rebase has completed, and the manifest update fails.

ulfalizer: > The old behavior was to just abandon the rebase and ask the user to clean up.
> In my opinion, that's not so nice.

Problem is that the next time you run west, it will try to update again, and fail again. In practice, that's the same as a repository that needs to be fixed manually, only you don't have the merge conflict to go on anymore.

tejlmand: So to make it clear.
My understanding is of requested behavior (to which I agree):
- Try to rebase

If rebase fails:
- Rollback to previous state, i.e. untouched by `west` tool
- Tell user what went wrong, and maybe suggest a solution (much harder)

Where old behaviour was:
- Try to rebase

If rebase fails:
- leave repo in dangling rebase state
- Tell user what went wrong

ulfalizer: In general, I think we should be careful about trying to "fix" Git rather than just augmenting it for the multi-repo stuff.

A lot of thought has gone into how it works, and we probably can't do better if we try. Its behavior makes sense ("hey, I couldn't update this thing, but I'll give you what I could do and let you fix the rest"). It's what people used to Git expect. We shouldn't surprise them too much, because losing them would be worse than losing the people who can't be bothered to learn a tiny bit of Git.

That said, we should write awesome docs, because people way too often blame the users when it's their own fault for not documenting things well. Assume your users are awesome and do the technically right thing, and write good docs. :)

For https://github.com/zephyrproject-rtos/west/issues/59, I agree that it's the right thing. If you're working on something on a separate branch, you're often expecting that it won't rebase cleanly, so west shouldn't try to rebase it for you in that case.

ulfalizer: Maybe the right thing here is to never automatically update unless it's a clean fast-forward update though. Not sure how useful the automatic `rebase` behavior is when working on stuff.

I just see some work and corner-cases there (SHAs, tags, unclean working tree, yada yada), so it makes me a bit cranky. It might not be that bad in practice. :P

ulfalizer: @carlescufi @mbolivar @tejlmand 
Awesome! The Git guys already thought of this, with `git pull --ff-only`. I'm pretty sure that that's what we really should be using for the self-update. I probably messed up by going for `git rebase` there.

I think that would solve a whole bunch of problems, while making the code simpler at the same time.

See any gotchas?

ulfalizer: @tejlmand 
Didn't see your most recent comment. Sorry about it.

See the comment above for what I suspect is the actual right thing to do here, unrelated to the bad manual rebase stuff I've been doing.

tejlmand: > In general, I think we should be careful about trying to "fix" Git rather than just augmenting it for the multi-repo stuff.

We should not try to fix `git`, and we should not try to assume what the user wants.
When I do a `get fetch`, `git pull`, `git rebase`, `git merge`, and `git merge --no-ff`, I know exactly what they'll do to my repo, and where those commands needs help.

As a starter to `west`, I don't know that `west` will do a rebase and abort, and thereafter I need to use `git` manually.
And we don't know if user expect their project to end in a non-working state.
We don't know if he wants to preserve / revert his changes.
Therefore I think the best is to assume nothing, and leave the repo's as-is. (ie. rolling back)

> A lot of thought has gone into how it works, and we probably can't do better if we try. Its behavior makes sense ("hey, I couldn't update this thing, but I'll give you what I could do and let you fix the rest"). It's what people used to Git expect. We shouldn't surprise them too much, because losing them would be worse than losing the people who can't be bothered to learn a tiny bit of Git.

Is it what people expects ?
I would say that we should try  not to touch peoples local work, unless users request so.
If there later comes a request, where people would like `west` to leave the repo in partially rebased state, then we can try this.
Until then, I think we should only rebase if it can be done cleanly.

> The Git guys already thought of this, with git pull --ff-only. It even aborts if the working tree is unclean, etc.

This sounds nicer, except that if user has local commits, he is forced to merge / rebase those changes himself, even if a rebase could be applied cleanly.
But in that case, the user is working with west or manifest development, and I think it is fine that he has to use git directly for that.
Most important feature for `west` is to make it easier for those who develops on Zephyr multi repo itself.
Those maintaining west or manifest repos may be considered little more advanced users ;-)

Only question left then is:
- What to do if `west` updates successfully but `manifest` fails ?
Rollback both ?

ulfalizer:
Simpler approach in https://github.com/zephyrproject-rtos/west/pull/82.

src/west/commands/project.py
> 
> @tejlmand
> Didn't see your most recent comment. Sorry about that.
> 
> See the comment above for what I suspect is the actual right thing to do here, unrelated to the bad manual rebase stuff I've been doing.

I've seen it.
And currently I believe the `git pull --only-ff` to be the best approach.

Any thoughts on this:
> What to do if west updates successfully but manifest fails ?
Rollback both ?

ulfalizer: Yeah, the assumption was that developers would want their changes forward-ported on updates (at least when working on the main branch), unless suppressed with `--no-update`. A merge conflict is a much more informative thing to give them than a rolled-back repository in that case, and doesn't touch their work. It felt like "hey, we know you want to update, but we're gonna remove the standard Git mechanism for dealing with merge conflicts".

It's all kinda moot with the `--ff-only` though. Now we assume that people never want to automatically forward-port changes instead. That solves the problem with work branches and is simple to get right, so it's probably better.

I agree that we should try to make it so that most users never have to touch the west/manifest repositories.

> * What to do if `west` updates successfully but `manifest` fails ?
>   Rollback both ?

Could do this later maybe, if we ever need it. Theoretically, you might need to skip the west update if the manifest update fails too, in case backwards-incompatible changes are made to the manifest format.

It's simple to solve manually if you assume that people tinkering with the west and manifest repositories know what they're doing.

ulfalizer: Yeah, the assumption was that developers would want their changes forward-ported on updates (at least when working on the main branch), unless suppressed with `--no-update`. A merge conflict is a much more informative thing to give them than a rolled-back repository in that case, and doesn't touch their work. It felt like "hey, we know you want to update, but we're gonna remove the standard Git mechanism for dealing with merge conflicts".

It's all kinda moot with the `--ff-only` though. Now we assume that people never want to automatically forward-port changes instead. That solves the problem with work branches and is simple to get right, so it's probably better.

I agree that we should try to make it so that most users never have to touch the west/manifest repositories.

> * What to do if `west` updates successfully but `manifest` fails ?
>   Rollback both ?

Could do this later maybe, if we ever need it. Theoretically, you might need to skip the west update if the manifest update fails too, in case backwards-incompatible changes are made to the manifest format.

It's simple to solve manually if you assume that people tinkering with the west and manifest repositories know what they're doing.

ulfalizer: Yeah, the assumption was that developers would want their changes forward-ported on updates (at least when working on the main branch), unless suppressed with `--no-update`. A merge conflict is a much more informative thing to give them than a rolled-back repository in that case, and doesn't touch their work. It felt like "hey, we know you want to update, but we're gonna remove the standard Git mechanism for dealing with merge conflicts".

It's all kinda moot with the `--ff-only` though. Now we assume that people never want to automatically forward-port changes instead. That solves the problem with work branches and is simple to get right, so it's probably better.

I agree that we should try to make it so that most users never have to touch the west/manifest repositories.

> * What to do if `west` updates successfully but `manifest` fails ?
>   Rollback both ?

Could do this later maybe, if we ever need it. Theoretically, you might need to skip the west update if the manifest update fails too, in case backwards-incompatible changes are made to the manifest format.

It's simple to solve manually if you assume that people tinkering with the west and manifest repositories know what they're doing.

Issue #77, opened 2018-10-12
West should support reinitializing an existing installation


mbolivar:
We've always left a hole in west's `init_reinit()` routine, which is what gets run when a user attempts to `west init` an existing west installation. Right now, it simply reports an error.

With the new configuration file infrastructure (https://github.com/zephyrproject-rtos/west/pull/75), we should have all the pieces to reinitialize now:

1. west repository
2. manifest repository
3. configuration directory with new west and manifest revisions
4. west marker file

Once those are reinitialized, the next `west fetch` should pull down the updated configuration specified by the above without any extra work or changes.

ulfalizer:
Looked into the `west reset` thing, but it's starting to feel icky to me. I don't think we should create our own less general Git command wrappers in cases where you could easily just use `west forall ...`.

On the upside, `git reset --keep` is even better than I expected it to be (see the docstring below). If reiniting must reset stuff, then I think it's the best choice.

Icky WIP below, which I don't think we should go for. I was thinking of only supporting the mode flags (`--soft`, `--hard`, etc.), but once you start duplicating existing flags like that just to pass them through as-is, it's starting to stink, IMO.

```python
class Reset(WestCommand):
    def __init__(self):
        super().__init__(
            'reset',
            _wrap('''
            Reset projects to '{}'.

            Runs 'git reset --keep {0}' for the specified projects (default:
            all projects), resetting them back to the revision specified in the
            manifest (as of the most recent fetch of project data).

            'git reset --keep' works like 'git reset --hard', but aborts the
            reset if any file modified between HEAD and '{0}' has local
            changes, carrying over any local changes otherwise.

            Extra arguments are passed as-is to 'git reset', and override
            '--keep {0}'. For example, 'west reset --soft HEAD~' runs 'git
            reset --soft HEAD~' in all projects.
            '''.format(_MANIFEST_REV_BRANCH)),
            accepts_unknown_args=True)

    def do_add_parser(self, parser_adder):
        return _add_parser(parser_adder, self, _project_list_arg)

    def do_run(self, args, user_args):
        for project in _all_projects(args):
            _git(project, 'reset',
                 extra_args=user_args or ('--keep', _MANIFEST_REV_BRANCH))
```

PR #84, opened 2018-10-24
commands: project: Fetch directly from URLs instead of remotes + support reinit

ulfalizer:
Three commits:

```
commands: project: Fetch directly from URLs instead of remotes

Instead of fetching from remotes, fetch from URLs. This bypasses the
entire remote system, which is nice in that we no longer rely on how
remotes are configured. Users could delete remotes or point them
somewhere else, and things will still work.

I had somehow managed to miss all along is that Git allows URLs wherever
names of remotes are allowed.

Also put URLs into the initial west/config configuration file. Keep the
name 'remote' for the configuration value, as putting the name of a
remote there will work as well.

IOW, both of the following will work:

    [west]
    remote = https://github.com/zephyrproject-rtos/west
    revision = master

    [west]
    remote = my-custom-remote
    revision = master

For projects, create a remote named after the remote in the manifest
(like before), just for the user's convenience. We never use it
ourselves.

Fixes: #83
```

```
commands: project: Allow changing configuration options with 'west init'

Support rerunning 'west init' in an already initialized West
installation to update configuration values. Currently, this just
modifies west/config.

Only arguments that are explicitly passed are changed. For example,
'west init --mr foo' will update the manifest revision to 'foo', leaving
all other settings unchanged.

Passing no arguments when reinitializing is a no-op, and prints an
error.

Treat the initial west/config creation as an update that happens to set
all configuration values, to reuse some code.

Piggyback some small cleanup:

 - Rename WEST/MANIFEST_DEFAULT to WEST/MANIFEST_URL_DEFAULT. There's
   already WEST/MANIFEST_REV_DEFAULT.

 - Rename find_west_topdir() to west_topdir(). This is what the version
   in the main West code is called (would be nice to reuse things
   directly from there in the future).

   Also make west_topdir() implicitly default to the working directory.

Fixes: #77
```

```
bootstrap: Add more 'west init' documentation

Add some more information to 'west init --help':

 - Explain what the bootstrap does, in some detail

 - Fix some reference to "Zephyr" that should say "West"

 - Mention that reinit just updates west/config, and that the updated
   settings won't take effect until they're read

 - Fix some documentation style consistency nits
```

### DIFF
--- a/src/west/_bootstrap/main.py
+++ b/src/west/_bootstrap/main.py
@@ -134,9 +134,10 @@ configuration settings. This is just an alternative to manually editing
 west/config. Only explicitly passed configuration values (e.g.
 --mr NEW_MANIFEST_REVISION) are updated.
 
-Note: If you update the west/manifest revision or URL, the change will take
-effect whenever the configuration setting is next read, e.g. in `west fetch`.
-'west init' only updates configuration settings.
+If the URL/revision of the manifest or west repositories is updated, then
+'west update' is automatically run afterwards to update the repositories. Note
+that this will fail if the new revision can't be fast-forwarded to. In that
+case, you will need to e.g. stash away your changes in a separate branch.
 '''.format(WEST_MARKER))
 
     init_parser.add_argument(
@@ -257,6 +258,15 @@ def reinit(config_path, args):
                 west_url, args.west_rev, manifest_url, args.manifest_rev)
 
     print('=== Updated configuration written to {} ==='.format(config_path))
+
+    cmd = ['update']
+    if west_url or args.west_rev:
+        cmd.append('--update-west')
+    if manifest_url or args.manifest_rev:
+        cmd.append('--update-manifest')
+    print("=== Running 'west {}' to update repositories ==="
+          .format(' '.join(cmd)))
+    wrap(cmd)
 
 
 def update_conf(config_path, west_url, west_rev, manifest_url, manifest_rev):
--- a/src/west/_bootstrap/main.py
+++ b/src/west/_bootstrap/main.py
@@ -115,26 +115,59 @@ def init(argv):
     This exits the program with a nonzero exit code if fatal errors occur.'''
     init_parser = argparse.ArgumentParser(
         prog='west init',
-        description='Bootstrap initialize a Zephyr installation')
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        description=
+'''
+Initializes a Zephyr installation.
+
+In more detail, does the following:
+
+  1. Clones the manifest repository to west/manifest, and the west repository
+     to west/west
+
+  2. Creates a marker file west/{}
+
+  3. Creates an initial configuration file west/config
+
+'west init' can be rerun on an already initialized West instance to update
+configuration settings. This is just an alternative to manually editing
+west/config. Only explicitly passed configuration values (e.g.
+--mr NEW_MANIFEST_REVISION) are updated.
+
+Note: If you update the west/manifest revision or URL, the change will take
+effect whenever the configuration setting is next read, e.g. in `west fetch`.
+'west init' only updates configuration settings.
+'''.format(WEST_MARKER))
+
     init_parser.add_argument(
         '-b', '--base-url',
-        help='''Base URL for both 'manifest' and 'zephyr' repositories; cannot
-        be given if either -m or -w are''')
+        help='''Base URL for the 'manifest' and 'west' repositories. Cannot
+             be given if either -m or -w are.''')
+
     init_parser.add_argument(
         '-m', '--manifest-url',
-        help='Zephyr manifest fetch URL, default ' + MANIFEST_DEFAULT)
+        help='Manifest repository URL (or remote name) (default: {})'
+             .format(MANIFEST_URL_DEFAULT))
+
     init_parser.add_argument(
         '--mr', '--manifest-rev', dest='manifest_rev',
-        help='Manifest revision to fetch, default ' + MANIFEST_REV_DEFAULT)
+        help='Manifest revision to fetch (default: {})'
+             .format(MANIFEST_REV_DEFAULT))
+
     init_parser.add_argument(
         '-w', '--west-url',
-        help='West fetch URL, default ' + WEST_URL_DEFAULT)
+        help='West repository URL (or remote name) (default: {})'
+             .format(WEST_URL_DEFAULT))
+
     init_parser.add_argument(
         '--wr', '--west-rev', dest='west_rev',
-        help='West revision to fetch, default ' + WEST_REV_DEFAULT)
+        help='West revision to fetch (default: {})'
+             .format(WEST_REV_DEFAULT))
+
     init_parser.add_argument(
         'directory', nargs='?',
-        help='Initializes in this directory, creating it if necessary')
+        help='''Directory to initialize West in. Missing directories will be
+             created automatically. (default: current directory)''')
 
     args = init_parser.parse_args(args=argv)
 
--- a/src/west/_bootstrap/main.py
+++ b/src/west/_bootstrap/main.py
@@ -202,12 +202,12 @@ def init_bootstrap(directory, args):
     config = configparser.ConfigParser()
 
     config['west'] = {
-        'remote': 'origin',
+        'remote': args.west_url,
         'revision': args.west_rev
     }
 
     config['manifest'] = {
-        'remote': 'origin',
+        'remote': args.manifest_url,
         'revision': args.manifest_rev
     }
 
@@ -256,7 +256,8 @@ def wrap(argv):
             sys.exit(0)         # run outside of an installation directory
         else:
             sys.exit('Error: not a Zephyr directory (or any parent): {}\n'
-                     'Use "west init" to install Zephyr here'.format(start))
+                     'Use "west init" to install Zephyr here'
+                     .format(os.getcwd()))
 
     west_git_repo = os.path.join(topdir, WEST_DIR, WEST)
     if printing_version:
--- a/src/west/_bootstrap/main.py
+++ b/src/west/_bootstrap/main.py
@@ -30,7 +30,7 @@ WEST_DIR = 'west'
 # Subdirectory to check out the west source repository into.
 WEST = 'west'
 # Default west repository URL.
-WEST_DEFAULT = 'https://github.com/zephyrproject-rtos/west'
+WEST_URL_DEFAULT = 'https://github.com/zephyrproject-rtos/west'
 # Default revision to check out of the west repository.
 WEST_REV_DEFAULT = 'master'
 # File inside of WEST_DIR which marks it as the top level of the
@@ -44,7 +44,7 @@ WEST_MARKER = '.west_topdir'
 # Manifest repository directory under WEST_DIR.
 MANIFEST = 'manifest'
 # Default manifest repository URL.
-MANIFEST_DEFAULT = 'https://github.com/zephyrproject-rtos/manifest'
+MANIFEST_URL_DEFAULT = 'https://github.com/zephyrproject-rtos/manifest'
 # Default revision to check out of the manifest repository.
 MANIFEST_REV_DEFAULT = 'master'
 
@@ -62,13 +62,24 @@ class WestNotFound(WestError):
     '''Neither the current directory nor any parent has a West installation.'''
 
 
-def find_west_topdir(start):
-    '''Find the top-level installation directory, starting at ``start``.
+def west_dir(start=None):
+    '''
+    Returns the path to the west/ directory, searching ``start`` and its
+    parents.
 
-    If none is found, raises WestNotFound.'''
+    Raises WestNotFound if no west directory is found.
+    '''
+    return os.path.join(west_topdir(start), WEST_DIR)
+
+
+def west_topdir(start=None):
+    '''
+    Like west_dir(), but returns the path to the parent directory of the west/
+    directory instead, where project repositories are stored
+    '''
     # If you change this function, make sure to update west.util.west_topdir().
 
-    cur_dir = start
+    cur_dir = start or os.getcwd()
 
     while True:
         if os.path.isfile(os.path.join(cur_dir, WEST_DIR, WEST_MARKER)):
@@ -113,38 +124,141 @@ def init(argv):
         '-m', '--manifest-url',
         help='Zephyr manifest fetch URL, default ' + MANIFEST_DEFAULT)
     init_parser.add_argument(
-        '--mr', '--manifest-rev', default=MANIFEST_REV_DEFAULT,
-        dest='manifest_rev',
+        '--mr', '--manifest-rev', dest='manifest_rev',
         help='Manifest revision to fetch, default ' + MANIFEST_REV_DEFAULT)
     init_parser.add_argument(
         '-w', '--west-url',
-        help='West fetch URL, default ' + WEST_DEFAULT)
+        help='West fetch URL, default ' + WEST_URL_DEFAULT)
     init_parser.add_argument(
-        '--wr', '--west-rev', default=WEST_REV_DEFAULT, dest='west_rev',
+        '--wr', '--west-rev', dest='west_rev',
         help='West revision to fetch, default ' + WEST_REV_DEFAULT)
     init_parser.add_argument(
-        'directory', nargs='?', default=None,
+        'directory', nargs='?',
         help='Initializes in this directory, creating it if necessary')
 
     args = init_parser.parse_args(args=argv)
-    directory = args.directory or os.getcwd()
-
-    if args.base_url:
-        if args.manifest_url or args.west_url:
-            sys.exit('fatal error: -b is incompatible with -m and -w')
-        args.manifest_url = args.base_url.rstrip('/') + '/manifest'
-        args.west_url = args.base_url.rstrip('/') + '/west'
-    else:
-        if not args.manifest_url:
-            args.manifest_url = MANIFEST_DEFAULT
-        if not args.west_url:
-            args.west_url = WEST_DEFAULT
 
     try:
-        topdir = find_west_topdir(directory)
-        init_reinit(topdir, args)
+        reinit(os.path.join(west_dir(args.directory), 'config'), args)
     except WestNotFound:
-        init_bootstrap(directory, args)
+        bootstrap(args)
+
+
+def bootstrap(args):
+    '''Bootstrap a new manifest + West installation.'''
+
+    if args.base_url:
+        if args.west_url or args.manifest_url:
+            sys.exit('fatal error: -b is incompatible with -m and -w')
+        west_url = args.base_url.rstrip('/') + '/west'
+        manifest_url = args.base_url.rstrip('/') + '/manifest'
+    else:
+        west_url = args.west_url or WEST_URL_DEFAULT
+        manifest_url = args.manifest_url or MANIFEST_URL_DEFAULT
+
+    west_rev = args.west_rev or WEST_REV_DEFAULT
+    manifest_rev = args.manifest_rev or MANIFEST_REV_DEFAULT
+
+    directory = args.directory or os.getcwd()
+
+    if not os.path.isdir(directory):
+        try:
+            print('Initializing in new directory', directory)
+            os.makedirs(directory, exist_ok=False)
+        except PermissionError:
+            sys.exit('Cannot initialize in {}: permission denied'.format(
+                directory))
+        except FileExistsError:
+            sys.exit('Something else created {} concurrently; quitting'.format(
+                directory))
+        except Exception as e:
+            sys.exit("Can't create directory {}: {}".format(
+                directory, e.args))
+    else:
+        print('Initializing in', directory)
+
+    # Clone the west source code and the manifest into west/. Git will create
+    # the west/ directory if it does not exist.
+
+    clone('west repository', west_url, west_rev,
+          os.path.join(directory, WEST_DIR, WEST))
+
+    clone('manifest repository', manifest_url, manifest_rev,
+          os.path.join(directory, WEST_DIR, MANIFEST))
+
+    # Create an initial configuration file
+
+    config_path = os.path.join(directory, WEST_DIR, 'config')
+    update_conf(config_path, west_url, west_rev, manifest_url, manifest_rev)
+    print('=== Initial configuration written to {} ==='.format(config_path))
+
+    # Create a dotfile to mark the installation. Hide it on Windows.
+
+    with open(os.path.join(directory, WEST_DIR, WEST_MARKER), 'w') as f:
+        hide_file(f.name)
+
+    print('=== West initialized ===')
+
+
+def reinit(config_path, args):
+    '''
+    Reinitialize an existing installation.
+
+    Currently, this is just a shorthand for updating the west/config
+    configuration file.
+    '''
+    if args.base_url:
+        if args.west_url or args.manifest_url:
+            sys.exit('fatal error: -b is incompatible with -m and -w')
+        west_url = args.base_url.rstrip('/') + '/west'
+        manifest_url = args.base_url.rstrip('/') + '/manifest'
+    else:
+        west_url = args.west_url
+        manifest_url = args.manifest_url
+
+    if not (west_url or args.west_rev or manifest_url or args.manifest_rev):
+        sys.exit('West already initialized. Please pass any settings you '
+                 'want to change.')
+
+    update_conf(config_path,
+                west_url, args.west_rev, manifest_url, args.manifest_rev)
+
+    print('=== Updated configuration written to {} ==='.format(config_path))
+
+
+def update_conf(config_path, west_url, west_rev, manifest_url, manifest_rev):
+    '''
+    Creates or updates the configuration file at 'config_path' with the
+    specified values. Values that are None/empty are ignored.
+    '''
+    config = configparser.ConfigParser()
+
+    # This is a no-op if the file doesn't exist, so no need to check
+    config.read(config_path)
+
+    update_key(config, 'west', 'remote', west_url)
+    update_key(config, 'west', 'revision', west_rev)
+    update_key(config, 'manifest', 'remote', manifest_url)
+    update_key(config, 'manifest', 'revision', manifest_rev)
+
+    with open(config_path, 'w') as f:
+        config.write(f)
+
+
+def update_key(config, section, key, value):
+    '''
+    Updates 'key' in section 'section' in ConfigParser 'config', creating
+    'section' if it does not exist.
+
+    If value is None/empty, 'key' is left as-is.
+    '''
+    if not value:
+        return
+
+    if section not in config:
+        config[section] = {}
+
+    config[section][key] = value
 
 
 def hide_file(path):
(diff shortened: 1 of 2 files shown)


ulfalizer:
CC @carlescufi 

Could put a fix for https://github.com/zephyrproject-rtos/west/issues/77 into this PR as well.

carlescufi:
@ulfalizer great! please, yes if you don't mind add the fix for #77 and then I am hoping to get together with @mbolivar before the end of the week and perhaps take a look at it together.

ulfalizer:
Fix for #77 added. `west init` can now be rerun to change configuration values. Only explicitly specified configuration values are updated.

src/west/_bootstrap/main.py
```diff
@@ -191,44 +180,108 @@ def init_bootstrap(directory, args):
...
+
+
+def reinit(config_path, args):
+    '''
+    Reinitialize an existing installation.
+
+    Currently, this is just a shorthand for updating the west/config
+    configuration file.
+    '''
+    if args.base_url:
```
ulfalizer: Wonder if we could get rid of `-b`. It's basically just a shorthand to save a little bit of copy-paste now.

Might be possible to factor out something here anyway though.

carlescufi: I think as a convenience to the user, but not something I'd defend to death.
Being able to do:

`west init -b https://github.com/mycompany`

is nice and short.

mbolivar: As a user, I find this option valuable even if it is redundant. I originally did not want to implement it but I found myself using it all the time when pulling down wworkdirs from non-mainline places.

ulfalizer: Adunno... still feels a bit weird to me to have a special command just for when you happen to have two repositories with particular names and with the same URL prefix.

Not really worth arguing though.

carlescufi:
@ulfalizer @tejlmand @mbolivar 

I briefly tested this to switch from the upstream manifest to a different one that is *not* a fork of the upstream one.

```
C:\Users\Carles\src\wncs>west init -w https://github.com/ulfalizer/west.git --wr fetch-from-url wncs -u https://github.com/NordicPlayground/<redacted>.git
=== Updated configuration written to west\config ===

C:\Users\Carles\src\wncs>west fetch
warning: no common commits
fatal: refusing to merge unrelated histories
WARNING: Skipping automatic update of manifest (west\manifest\). Can't be fast-forwarded to master (from https://github.com/NordicPlayground/<redacted>.git).
```

I thought that we would be able to simply fetch using the new manifest and switch to it. How would we do that? (i.e. switch from one manifest to another)

ulfalizer:
@carlescufi 
I might be wrong, but I think the way `repo` does it is that it requires you to have all the manifests in the same repository. See the `-m` and `-b` options in the [`repo` reference](https://source.android.com/setup/develop/repo#init).

Otherwise, you might have to `git reset --hard` or the like (or delete the manifest repository and clone down a new one). That feels a bit magic and brittle to me.

ulfalizer:
Hmm... maybe a plain `git reset <new revision>` wouldn't be so bad though, reporting an error if it fails.

What corner cases can you see with that?

ulfalizer:
Or `git reset --keep` maybe. Would need to refresh exactly how all of those work.

Might work okay as long as it aborts if there are changes to the working tree or index.

carlescufi:
> I might be wrong, but I think the way `repo` does it is that it requires you to have all the manifests in the same repository. See the `-m` and `-b` options in the [`repo` reference](https://source.android.com/setup/develop/repo#init).

Well, having the repos in the same repository is obviously not possible, since this is often private/confidential.

ulfalizer:
@carlescufi 
You mean the manifest files?

I guess you could just dump them there, so that you have a repository with all the manifests needed by team x.

I'll look into the `reset` business though. Would need to be done in the bootstrap script I think.

carlescufi:
> @carlescufi
> You mean the manifest files?
> 
> I guess you could just dump them there, so that you have a repository with all the manifests needed by team x.
> 
> I'll look into the `reset` business though. Would need to be done in the bootstrap script I think.

@ulfalizer 
Typically how I think it's gonna work is that each org will have their `/manifest.git` with a single `default.yaml`. It would be good to get the `reset` stuff working so that we can switch from the upstream repo having initialized with `west init .` to the org one.

ulfalizer:
Hmm... still a bit of a mess though. If someone has made local commits to the manifest, they will get orphaned away for example.

I wonder if this might be encouraging a bad workflow. Switching manifest repositories feels pretty weird to me. It's still possible to do too, if you know what you're doing.

The bootstrapper would get more complex too. Would need to check if the URL or the revision changed, `git fetch` the revision in that case, and then `git reset`, with some flag that errors out if the working tree or index has been changed.

Not horrible, but it feels like there has to be a better workflow. Changing the entire manifest repository feels like something that warrants a new `west init` to me.

carlescufi:
> Changing the entire manifest repository feels like something that warrants a new `west init` to me.

Right, so that I think is the key question we have to solve. Do we want to be able to switch repos or only branches?

mbolivar:
> > Changing the entire manifest repository feels like something that warrants a new `west init` to me.
> 
> Right, so that I think is the key question we have to solve. Do we want to be able to switch repos or only branches?

If we support re-running west init and it having some effect, why not allow switching repositories if the user asks for it explicitly?

We can always put it behind a `--force` flag if it turns out to have problems, but orphaning local commits doesn't seem like a problem to me since it's not being done behind the user's back. There's always the reflog if people make mistakes.

ulfalizer:
Think both might boil down to whether we want to do any `git reset` shenanigans, or always just update via fast-forward.

Switching repositories should work fine too, as long as they have some common history, and West/manifest can be updated via fast-forward after switching. If it doesn't, then something is broken.

carlescufi:
> Think both might boil down to whether we want to do any `git reset` shenanigans, or always just update via fast-forward.
> 
> Switching repositories should work fine too, as long as they have some common history, and West/manifest can be updated via fast-forward after switching. If it doesn't, then something is broken.

I really don't think we should enforce common history in the case of a manifest. So perhaps the `--force` flag that @mbolivar mentions could be an option. But where do we add that flag? in `west init` or in `west fetch/clone`?

mbolivar:
> But where do we add that flag? in `west init` or in `west fetch/clone`?

I was thinking `init`, since that's where the commits would get orphaned.

mbolivar:
> I was thinking `init`, since that's where the commits would get orphaned.

Though to be explicit I am proposing we don't add `--force` now, just have it in our back pockets if it causes real problems with users.

carlescufi:
> > But where do we add that flag? in `west init` or in `west fetch/clone`?
> 
> I was thinking `init`, since that's where the commits would get orphaned.

Yeah sorry, I forgot that this is the manifest repo and as such gets pulled during `init`.

ulfalizer:
> I really don't think we should enforce common history in the case of a manifest. So perhaps the `--force` flag that @mbolivar mentions could be an option. But where do we add that flag? in `west init` or in `west fetch/clone`?

I'd skip the `--force` flag and just do the `reset` thing, if we really need it. I think we're starting to get too many flags already.

That said, it seems you often bring up the 1% things I never in my wildest fantasies imagined people would want to do. ;)

Do people really switch manifest repositories a lot? The `repo` model where you just dump some manifests into a team repository seems a lot neater and simpler.

carlescufi:
> Do people really switch manifest repositories a lot? The `repo` model where you just dump some manifests into a team repository seems a lot neater and simpler.

I think this will happen a lot, but they might be forks with a common ancestor. Now I will try if that works with the PR today.

ulfalizer:
With `repo`, what I suspect is the "intended" workflow is that whoever manages the manifest repository creates a new branch (which could have commits from wherever). People could then use `-b` to switch to it (which could do a `git checkout` or the like internally).

Would be interesting to see how `repo` deals with changing the manifest repository URL though. Maybe it supports it and is clever about it, since it's pretty old and mature.

mbolivar:
> Would be interesting to see how `repo` deals with changing the manifest repository URL though.

It switches the remote URL in the manifest repository. Just tested.

Ran into an issue because the default branch had its merge config option set to a tag which didn't exist in the other manifest repository, but after removing that `sync -n` worked fine.

ulfalizer:
> It switches the remote URL in the manifest repository. Just tested.

Hmm... sounds like what we're doing, except we currently store the remote URL in the configuration file (after this PR anyway), and only update via fast-forward.

`repo` seems to do some fancier checks for dirtiness internally.

ulfalizer:
So no changes to the working tree or the checked-out commit or the like with `repo` either?

mbolivar:
> So no changes to the working tree or the checked-out commit or the like with `repo` either?

It does change the manifest checkout to the commit in the updated tree.

BTW, you can always try stuff out using this manifest: https://github.com/mbolivar/zephyr-repo-manifest

ulfalizer:
> It does change the manifest checkout to the commit in the updated tree.

Immediately after the `repo init` is run, or once another command is run? Just trying to do some reverse-engineering.

> BTW, you can always try stuff out using this manifest: https://github.com/mbolivar/zephyr-repo-manifest

Yeah, I'll try it out soon. I have like five of those directories lying around that I forgot the names of. ;)

ulfalizer:
@mbolivar 
Played it around with it a bit.

`repo init` updates immediately, but does it via rebasing. You get a merge conflict that you have to resolve manually if it can't be rebased (like `west` used to work, before the ff-only thing).

In practice, that works out to the same behavior that we have now, with some small differences:

1. The update in `west` happens when you run the next fetch command, rather than immediately.

2. We refuse to update if it can't be done via fast-forward.

3. `repo` instead refuses to update if you're on a work branch in the manifest repository.

mbolivar:
> Immediately after the `repo init` is run, or once another command is run?

Immediately

ulfalizer:
> `repo init` updates immediately, but does it via rebasing

Another way you can see that is that if you remove the tip commit in the upstream manifest repo and rerun `repo -u`, it will still be there afterwards.

`repo` never seems to do any resetting, just rebasing.

carlescufi:
The following worked, but still not in line with my expectations:

```
west init -u https://github.com/carlescufi/manifest.git -w https://github.com/ulfalizer/west.git --wr fetch-from-url wncs
west clone 
west init --mr other
west fetch
```

The `default.yaml` was updated *after* `west fetch`, but I would have expected to be updated after `west init --mr other`.

ulfalizer:
@carlescufi 
The thinking behind it is that you're just updating the configuration setting that tells `west` where to fetch self-updates from. All it does is rewrite the configuration file.

That might not be obvious though. It could be documented. I might be biased, but I think it makes sense at some level at least. :)

We could do it then and there as well, but it would be nice if there was a way to share code between the bootstrapper script and the rest of West before doing that, to avoid a bunch of code duplication for a minor thing like that.

mbolivar:
> `repo init` updates immediately, but does it via rebasing.

That's not the behavior I see when I re-run `repo init` in an internal project. I switched from URL X to Y and back to X, and what I got back at the end in the manifest repository was the exact tagged commmit I had before. What a I missing? Is this something that happens when there are local manifest commits only?

ulfalizer:
@mbolivar 
Try `repo init -u https://github.com/ulfalizer/zephyr-repo-manifest` followed by `repo init -u https://github.com/mbolivar/zephyr-repo-manifest`.

The extra commit I added in the fork is still there afterwards:

```
$ gl
commit 2bc0aa2dd1b836772acfdb37deeab3fabc70497b (HEAD -> default)
Author: Ulf Magnusson <ulfalizer@gmail.com>
Date:   Fri Oct 26 16:15:41 2018 +0200

    foobar

commit 06dee588e984db20f79c37c51774327bc2685a47 (origin/master)
Author: Marti Bolivar <marti@foundries.io>
Date:   Thu Aug 16 13:27:13 2018 -0500

    add default.xml
    
    Signed-off-by: Marti Bolivar <marti@foundries.io>
```

There might be other behaviors I'm missing, but looks like a plain rebase there at least.

ulfalizer:
You can also get merge conflicts via `repo init -u`, so at least in some cases it must be doing a rebase.

mbolivar:
> The extra commit I added in the fork is still there afterwards:

> You can also get merge conflicts via `repo init -u`, so at least in some cases it must be doing a rebase.

OK, got it, thanks, so it is about when there's extra commits in the branch, right?

ulfalizer:
@mbolivar 
Yep. Unless the old revision is an ancestor of the new revision, some commits will be rebased over, and the new `HEAD` won't be the new revision.

ulfalizer:
I'll add some documentation for the reinit in this PR too.

Doesn't mean it's set in stone, but should have been there already.

carlescufi:
@ulfalizer @mbolivar so what is the conclusion on this? is this staying as-is or are we moving towards a `git reset --keep` model at `init` time?

ulfalizer:
We could probably tweak it later if a need comes up. In the meantime, this is nice and simple to understand, IMO (especially with some more docs for `west init`).

carlescufi:
> We could probably tweak it later if a need comes up. In the meantime, this is nice and simple to understand, IMO (especially with some more docs for `west init`).

Since @mbolivar requested the feature in the first place, I will defer to him to decide whether this suffices for him. I do not think this is a blocker for me or Nordic.

ulfalizer:
Hmm.... is it deliberate that the bootstrapper talks about "Zephyr" instead of "West" by the way?

Got stuff like "Base URL for both 'manifest' and 'zephyr' repositories". That should probably be changed at least.

carlescufi:
> Hmm.... is it deliberate that the bootstrapper talks about "Zephyr" instead of "West" by the way?
> 
> Got stuff like "Base URL for both 'manifest' and 'zephyr' repositories". That should probably be changed at least.

Nope, that must be a mistake.

ulfalizer:
Wonder what CI's up to...

```
git merge -q origin/master
Auto-merging src/west/commands/project.py
Auto-merging src/west/_bootstrap/main.py
CONFLICT (content): Merge conflict in src/west/_bootstrap/main.py
Automatic merge failed; fix conflicts and then commit the result.
```

Haven't added the new docs yet though.

ulfalizer:
Added some more information to `west init --help`. It now goes through what the bootstrapper does in some details, and mentions the gotcha related to reinit that @carlescufi ran into (only configuration settings are updated, for now).

src/west/_bootstrap/main.py
```diff
@@ -115,26 +115,59 @@ def init(argv):
     This exits the program with a nonzero exit code if fatal errors occur.'''
     init_parser = argparse.ArgumentParser(
         prog='west init',
-        description='Bootstrap initialize a Zephyr installation')
+        formatter_class=argparse.RawDescriptionHelpFormatter,
+        description=
+'''
+Initializes a West instance.
```
mbolivar: What do you think about "Initializes a Zephyr installation."?

I think it's important to keep the focus on this being the command to install *zephyr*, but using west.

src/west/_bootstrap/main.py
```diff
@@ -115,26 +115,59 @@ def init(argv):
...
+  3. Creates an initial configuration file west/config
+
+'west init' can be rerun on an already initialized West instance to update
+configuration settings. This is just an alternative to manually editing
+west/config. Only explicitly passed configuration values (e.g.
+--mr NEW_MANIFEST_REVISION) are updated.
+
+Note: If you update the west/manifest revision or URL, the change will take
+effect whenever the configuration setting is next read, e.g. in `west fetch`.
+'west init' only updates configuration settings.
```
mbolivar: I think that running west init again should result in the same repository state for the manifest and west repositories as if you had run it for the first time whenever possible, rather than this kind of halfway change. I understand that this may not be possible due to a failed rebase if there are out of tree patches sometimes, but that seems like a separate issue.

src/west/_bootstrap/main.py
```diff
@@ -115,26 +115,59 @@ def init(argv):
...
+'''.format(WEST_MARKER))
+
     init_parser.add_argument(
         '-b', '--base-url',
-        help='''Base URL for both 'manifest' and 'zephyr' repositories; cannot
```
mbolivar: Can we "fix" this so that multiline strings are used throughout? Seems cleaner (and is consistent with the main west).

src/west/_bootstrap/main.py
ulfalizer: I think making it too magic would get messy. If you want to initialize another installation with completely different settings, then it's probably better to start from scratch.

Making e.g. `git clone` act as if you had cloned a different repository instead of erroring out wouldn't be the greatest design, to make a parallel. :)

src/west/_bootstrap/main.py
ulfalizer: Alright, I'll change it back.

src/west/_bootstrap/main.py
ulfalizer: Seems a bit overkill. It's consistent with the other strings, just breaking it across lines.

src/west/_bootstrap/main.py
ulfalizer: Done

src/west/_bootstrap/main.py
ulfalizer: Or do you mean using a single-line multiline string for short descriptions that fit on a single line? What'd be the point of that? Just adds a bunch of line noise, IMO.

src/west/_bootstrap/main.py
mbolivar: > Making e.g. `git clone` act as if you had cloned a different repository instead of erroring out wouldn't be the greatest design, to make a parallel. :)

The parallel here is with `repo init`, not `git clone`, I think.

> We could do it then and there as well, but it would be nice if there was a way to share code between the bootstrapper script and the rest of West before doing that, to avoid a bunch of code duplication for a minor thing like that.

I don't see much of an obstacle to doing that. Here are a couple alternatives:

1. put it in the bootstrapper and import the code from the rest of west (importing internal APIs should always be fine
2. make a `west.common` or so and add that package to what gets uploaded to pypi

src/west/_bootstrap/main.py
mbolivar: > Seems a bit overkill. It's consistent with the other strings, just breaking it across lines.

No, it's not consistent with the general pattern, which is to use a multiline string whenever the help string has to occupy multiple lines. Is this needed because of the change to a RawDescriptionHelpFormatter? The argparse internals should be reflowing the help normally otherwise.



> Or do you mean using a single-line multiline string for short descriptions that fit on a single line? What'd be the point of that? Just adds a bunch of line noise, IMO.

I mean using a multiline string for long descriptions that span multiple lines.

ulfalizer: Phew, thought you meant all of them.

Changed.

ulfalizer:
`west update` is now run automatically after updating settings. This could still fail if the new revision can't be fast-forwarded to, like for other updates of the special repositories.

src/west/_bootstrap/main.py
ulfalizer: @mbolivar 
`west update` is now run automatically if you rerun `west init` to change settings. That means the special repositories get updated right away (if they can be fast-forwarded).

Do you think the self-update should be special-cased when you change settings via `west init`? What command should be run instead in that case? I guess the reasons we added the fast-forward-only behavior might still apply here.

Personally, I prefer the `repo` behavior (a plain and super simple `git rebase`, with a nice merge conflict that tells you what conflicts), but we already settled on the fast-forward-only behavior I think. `repo` so far has always worked just like I expected it to work. :)

carlescufi:
@ulfalizer have you rebased this incorrectly? I still see `-u` instead of `-m` in here

tejlmand:
I noticed a strange behavior when doing a re-initialization, which I don't believe is intended.
To easier test yourself, I forked the manifest repo here:
`https://github.com/tejlmand/manifest`
and created `work_branch_1`

I now do:
`west init -u https://github.com/tejlmand/manifest`
which gives me the manifest from `master`, as expected.
I do a
`west clone`
which clones everything.

Now I want to switch to use the manifest, in `work_branch_1` and thus I re-init:
`west init --mr work_branch_1` but if I look at the manifest, it has fast-forwarded the manifest in `master` with the changes from `work_branch_1`, which is not what I would expect.

See:
```
$ west init -u https://github.com/tejlmand/manifest
$ cd west/manifest
$ git status
On branch master
Your branch is up-to-date with 'origin/master'.
nothing to commit, working directory clean
$ cd ../..
$ west init --mr work_branch_1
=== Updated configuration written to /projects/github/west_test/west/config ===
=== Running 'west update --update-manifest' to update repositories ===
=== Updated manifest (west/manifest/) to work_branch_1 (from https://github.com/tejlmand/manifest).
$ cd west/manifest
$ git status
On branch master
Your branch is ahead of 'origin/master' by 1 commit.
  (use "git push" to publish your local commits)
nothing to commit, working directory clean
```

So I got my local `master` fast-forwarded with the changes from work_branch_1, which probably isn't what people expect.

tejlmand:
and a related observation.

If I start out with the `work_branch_1` and then switches to `master`
Note that: `work_branch_1` contains both `openocd` and `zephyr` as repo, whereas `master` only contains `zephyr`.

Doing:
```
$ west init -u https://github.com/tejlmand/manifest --mr work_branch_1
$ west clone
$ west init --mr master
=== Updated configuration written to /projects/github/west_test/west/config ===
=== Running 'west update --update-manifest' to update repositories ===
```
which updates the `west/config` file as expected, however, running
```
$ west pull
=== Fetching changes for zephyr (zephyr/)
<SNIP>
=== Fetching changes for openocd (openocd/)
From https://github.com/zephyrproject-rtos/openocd
 * branch            master     -> FETCH_HEAD
<SNIP>
=== openocd (openocd/) is up-to-date with manifest-rev
```
but `openocd` is *not* part of the manifest file in `master`
and thus I would not expect it to be fetched.

Additional question: If user switch manifest revision, to a revision where a repo has been removed, should we then remove that folder in the repo structure created by `west` ?

Note:
If I do similar from `work_branch_1` to `work_branch_2`, manifest update fails, as it cannot be fast forwarded, as expected:
```
$ west init  --mr work_branch_1
=== Updated configuration written to /projects/github/west_test/west/config ===
=== Running 'west update --update-manifest' to update repositories ===
$ west init  --mr work_branch_2
=== Updated configuration written to /projects/github/west_test/west/config ===
=== Running 'west update --update-manifest' to update repositories ===
fatal: Not possible to fast-forward, aborting.
WARNING: Skipping automatic update of manifest (west/manifest/). Can't be fast-forwarded to work_branch_2 (from https://github.com/tejlmand/manifest).
```

From `work_branch_1` to `master`, no error is reported, as `work_branch_1` is branched from `master`. But manifest file is still pointing to revision `work_branch_1`.
```
$ west init  --mr work_branch_1
=== Updated configuration written to /projects/github/west_test/west/config ===
=== Running 'west update --update-manifest' to update repositories ===
$ west init  --mr master
=== Updated configuration written to /projects/github/west_test/west/config ===
=== Running 'west update --update-manifest' to update repositories ===
$ cd west/manifest
$ git branch
* work_branch_1
```

ulfalizer:
@tejlmand 
That's expected behavior. The `west/` and `manifest/` repositories are only ever updated via fast-forward, so you can never "get rid" of commits. The `repo` behavior is the same, except it does a `rebase` (which can never remove local work either, just move it over).

The old behavior we had was exactly like `repo`, before we switched to the fast-forward-only behavior.

ulfalizer:
Basically, the entire behavior can be summed up as `git pull --ff-only <remote from configuration file> <revision from configuration file>`. The `repo` behavior is a bit simpler: `git fetch ...` + `git rebase <revision from configuration file>`

ulfalizer:
By the way, instead of a workflow where people regularly switch between repositories, it might be neater to have them switch between branches. That's the way `repo` does it (`git checkout <branch>` in the manifest repository).

Needing to switch to an entirely different repository feels pretty exceptional to me, especially if it's unrelated to the first one.

carlescufi:
@ulfalizer ok so here is my issue with this, because there's something we are not in synch with. Someone comes and says "I found a bug, here's my manifest". So I want to go and reproduce this issue, am I forced to then clone from scratch? I'd like to be able to switch to that manifest. Maybe @mbolivar can comment on his workflow, but I think that's where @tejlmand is coming from. That said, this would require a `git reset` of some form, and so I do not think we can decide on this for 0.3.0 since we need to move forward. I propose we defer this since we are not going to be able to decide on this PR.

mbolivar:
> Someone comes and says "I found a bug, here's my manifest". So I want to go and reproduce this issue, am I forced to then clone from scratch? I'd like to be able to switch to that manifest. Maybe @mbolivar can comment on his workflow, but I think that's where @tejlmand is coming from. 

I am sure @tejlmand will provide his own details, but I agree that the use case of switching manifests to reproduce and test another developer's wworkdir is one where "re-init" working in an intuitive way makes sense.

tejlmand:
First, as @carlescufi  points out, one problem with this, is that I cannot change manifest in the same way as I can change branches in git, and test out other peoples work and jump back to my own.
I'm back to the subversion days, where I had 100+ folders that I constantly forgot to remove.
And I need to always do a clone.

Second: 
Doing this
```
$ west init  --mr work_branch_1
<SNIP>
$ west init  --mr work_branch_2
fatal: Not possible to fast-forward, aborting.
```
resulted in an error, which is good, cause it tells the user that it didn't work.

But this:
```
$ west init  --mr work_branch_1
<SNIP>
$ west init  --mr master
=== Updated configuration written to /projects/github/west_test/west/config ===
=== Running 'west update --update-manifest' to update repositories ===
```
gave no error, even though i'm not at `master`.
At the very least, user should be notified that things might not be as he expects.

Third: Re-init is not a re-init, cause it's outcome is not consistent, order matters.
```
$ west init -u https://github.com/tejlmand/manifest --mr work_branch_1
$ west clone
```
and
```
$ west init -u https://github.com/tejlmand/manifest --mr master
$ west init  --mr work_branch_1
$ west clone
```
Give the same end-result, `result A`

Whereas, doing
```
$ west init -u https://github.com/tejlmand/manifest --mr master
$ west clone
```
and
```
$ west init -u https://github.com/tejlmand/manifest --mr work_branch_1
$ west init  --mr master
$ west clone
```
give two different results.
First give you the master manifest, `result B` the second give you again `result A`, where expected would be `result B`.

This is not what I would expect from a re-init.
Can we agree on that ?

Whether to solve this now, or defer the fix till later is another question.

Btw.
> updated via fast-forward, so you can never "get rid" of commits

I know how fast forwards work. What i'm questioning is whether the behavior, seen from the user users point of view when using `west` is what he would expect. Basically I don't believe the users of `west` cares whether we use fast forward or other mechanisms.
It is the end-result seen from the user that should be easy to understand.

ulfalizer:
Yes, can we please get this in already. I constantly go back and tweak stuff, and like small, incremental improvements. This is not set in stone.

By the way, instead of talking about "switching manifests", " and the like, it would be helpful if we used terms such as "rebase", "fast-forward merge", "reset --x", etc., and also discussed details like how a dirty working tree or index should be handled. Right now, I constantly have to guess what is really meant.

If there's something you're wondering about there, then just ask me. I like explaining stuff.

I.e., "when switching manifests, I think we should do a `git reset --hard` to the new revision. That might have bad effect X if the user is doing Y, but Z might make up for it. What do you guys think?" That'd make it a more concrete.

What do you mean by "wworkdir" by the way? The project repositories?

mbolivar:
Is this PR still needed? Seems like #99 supersedes it, and the "fetch from URLs" part was split into a separate PR and merged already.

carlescufi:
@mbolivar I don't think so, no. @ulfalizer can you close this?

mbolivar:
Closed for now, can be reopened if needed

Issue #37, opened 2018-09-21
print remotes with `list-projects`


carlescufi:
The `list-projects` command currently doesn't print the remotes. It would be good to be able to list the remotes, even if it's with an additional param.

Issue #52, opened 2018-09-25
Should "west init" run "west fetch"? (Answer: no)


mbolivar:
Right now, installing multi-repo Zephyr takes the following steps:

```
$ pip install west
$ west init <west-workspace>
$ cd <west-workspace>
$ west fetch
```

This issue tracks whether "west init" should run "west fetch" as well after cloning west and the manifest, to eliminate the last step before a complete installation is cloned.

I am ambivalent about this idea (generally speaking I don't like to add features to the bootstrapper) but thought it was worth recording the question (and its resolution along with rationale).

Reasons for:

- one less step during installation

Reasons against:

- adds a feature to the bootstrapper which is not necessary
- users might want to fetch with different options than can be passed to init, e.g. to fetch a subset of available projects
- fetch might require different credentials than init, which might not be available at init time

tejlmand:
I don't think so.

Let's keep them separate, as one might do a `west init`, but want certain arguments to the `fetch` command.
As example, manifest to use or revision of manifest.

carlescufi:
@myrdyr 
https://github.com/zephyrproject-rtos/west/issues/37

mbolivar:
@tejlmand @myrdyr thanks for the feedback! I've summarized your comments in the issue description.

I think at this point it's clear that it's not worth adding, so I'm going to close this with the answer "no". We can reopen if we have a compelling reason.

mbolivar:
Reopening with the new mental model. Should west init run west clone? (We moved away from west fetch because it created detached HEADS, which are gross).

tejlmand:
I think we should keep `west init` and `west update` seperate.

Reason: 
It allows users to do `west init` and then actually adjust `west.yml` to their needs if they wish so, before running `west update`

If users wants a one step approach, they can do:
`west init; west update`
or
`west init && west update`
depending on their platform.

PR #85, opened 2018-10-24
commands: project: Show remote URLs for projects in 'west list'

ulfalizer:
This came up in #52.

I'm not sure what the appropriate field widths are ({:n}). We could
tweak it later if some fields turn out to be too short/long.

Fixes: #37


ulfalizer:
CC @carlescufi

PR #88, opened 2018-10-30
bootstrap: Don't error out for 'git@' URLs

ulfalizer:
Add 'git@' to the list of accepted prefixes. Alternatively, any username
could be accepted, but keep it simple while we have the URL check.

IMO, the check should be removed instead. It artificially restricts us to
a subset of what Git really supports, for no good reason.

Also fix a typo: s/git-shh:/git-ssh:/


ulfalizer:
CC @carlescufi

Issue #24, opened 2018-09-04
bash autocomplete for `west`


carlescufi:
Including bash autocompletion for `west` commands would immensely help the average user. This can be achieved in a very similar manner to how [git does it](https://github.com/git/git/blob/master/contrib/completion/git-completion.bash).

galak:
Would be nice if autocomplete would be board name / list aware.

PR #89, opened 2018-10-30
Add Bash auto-completion script

ulfalizer:
Currently completes flags and subcommand names.

Fixes: #24

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -39,6 +39,12 @@ What just happened:
 (For those familiar with it, this is similar to how Android's Repo
 tool works.)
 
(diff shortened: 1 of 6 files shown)


ulfalizer:
CC @carlescufi

ulfalizer:
This initial version just has command name and flag completion.

We need to think about installation too, unless we want users to manually `source` it.

carlescufi:
Thanks! Can you please add installation instructions as well?

ulfalizer:
Added some reminders to the West source code and installation instructions to the README.

Issue #83, opened 2018-10-22
Possibly neater and simpler way of handling remotes


ulfalizer:
One thing I had somehow managed to miss all along is that Git allows URLs wherever names of remotes are allowed. You can do `git fetch https://github.com/zephyrproject-rtos/zephyr.git`, for example. That feature might even predate remotes, which are more like bookmarks (with some associated configuration settings, like where to store the refs that are fetched).

By always fetching directly from the URL, we could bypass the entire remote system. The nice thing about that is that we no longer rely on how remotes are configured. It's also really close to what @carlescufi wanted to see, only I thought it'd be a lot messier to implement due to remote management.

Another neat thing is that both of the following will just work (maybe they already do), with no extra work required:

```
[west]
remote = my-custom-remote
revision = master
```

```
[west]
remote = https://github.com/zephyrproject-rtos/west
revision = master
```

Any remotes we create would simply be for the user's convenience. They could even delete them or point them somewhere else, and things would still work.

ulfalizer:
This came up while looking at https://github.com/zephyrproject-rtos/west/issues/77 by the way. I'll see how well this works first, as it would help with that issue.

carlescufi:
That is a really good proposal and will solve a lot of issues that might arise from users that do not know how to manage Git remotes.
+1 to this.

carlescufi:
CC @mbolivar @tejlmand

tejlmand:
Agree, this looks very promising.
+1.

carlescufi:
> I also really like this idea. I don't understand what "bypass the entire remote system" means. Surely you aren't suggesting removal of that concept from the manifest schema?

No, I think @ulfalizer means not  using Git's indirection of `remote` -> `url`

mbolivar:
> No, I think @ulfalizer means not using Git's indirection of `remote` -> url

If that's the case that sounds great.

ulfalizer:
Yeah, it's just switching to specifying repositories via URLs instead of Git remotes. That removes remote management and gives less ways for things to break.

After the change, we never use Git remotes ourselves, except we create one for the user's sake.

PR #90, opened 2018-11-01
commands: project: Fetch directly from URLs instead of remotes

carlescufi:
Instead of fetching from remotes, fetch from URLs. This bypasses the
entire remote system, which is nice in that we no longer rely on how
remotes are configured. Users could delete remotes or point them
somewhere else, and things will still work.

I had somehow managed to miss all along is that Git allows URLs wherever
names of remotes are allowed.

Also put URLs into the initial west/config configuration file. Keep the
name 'remote' for the configuration value, as putting the name of a
remote there will work as well.

IOW, both of the following will work:

    [west]
    remote = https://github.com/zephyrproject-rtos/west
    revision = master

    [west]
    remote = my-custom-remote
    revision = master

For projects, create a remote named after the remote in the manifest
(like before), just for the user's convenience. We never use it
ourselves.

Fixes: #83

### DIFF
--- a/src/west/_bootstrap/main.py
+++ b/src/west/_bootstrap/main.py
@@ -203,12 +203,12 @@ def init_bootstrap(directory, args):
     config = configparser.ConfigParser()
 
     config['west'] = {
-        'remote': 'origin',
+        'remote': args.west_url,
         'revision': args.west_rev
     }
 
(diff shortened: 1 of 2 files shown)


carlescufi:
@ulfalizer's first commit from #84

PR #93, opened 2018-11-01
Multiple fixes for tests on Windows

carlescufi:


src/west/manifest.py
```diff
@@ -278,7 +278,7 @@ def __init__(self, name, remote, defaults, path=None, clone_depth=None,
         self.remote = remote
         self.url = remote.url + '/' + name
         self.path = path or name
-        self.abspath = os.path.join(util.west_topdir(), self.path)
+        self.abspath = os.path.join(util.west_topdir(), os.path.normpath(self.path))
```
mbolivar: Is this right? Why isn't it `os.path.normpath(os.path.join(util.west_topdir(), self.path))`?

tests/west/manifest/test_manifest.py
```diff
@@ -61,8 +61,9 @@ def test_no_defaults():
     # Check the projects are as expected.
     for p, e in zip(manifest.projects, expected):
         deep_eq_check(p, e)
-    assert all(p.abspath == '/west_top/' + p.path for p in manifest.projects)
-
+    assert all(os.path.normpath(p.abspath) ==
```
mbolivar: I thought we guaranteed that p.abspath is a normpath in the previous patch?

tests/west/manifest/test_manifest.py
```diff
@@ -107,8 +108,9 @@ def test_default_clone_depth():
     # Check that the projects are as expected.
     for p, e in zip(manifest.projects, expected):
         deep_eq_check(p, e)
-    assert all(p.abspath == '/west_top/' + p.path for p in manifest.projects)
-
+    assert all(os.path.normpath(p.abspath) ==
```
mbolivar: Same comments apply here

src/west/manifest.py
carlescufi: Will fix

carlescufi:
@mbolivar comments addressed, thanks for the review.

tests/west/project/test_project.py
```diff
@@ -39,7 +40,10 @@ def cmd(cmd):
 
     # cmd() takes the command as a string, which is less clunky to work with.
     # Split it according to shell rules.
-    split_cmd = shlex.split(cmd)
+    if sys.platform == 'win32':
+        split_cmd = shlex.split(cmd, posix=False)
```
ulfalizer: A comment explaining why this is here would be nice:

```python
# posix=False prevents backslashes from getting stripped from paths on Windows
```

Wonder if we could just pass `posix=False` for both. `shlex.split()` it works the same regardless of platform (checked the code), and we never pass backslashes other than in the manifest path.

ulfalizer: Wait... there's a simpler solution here:

    cmd += ' -m "{}"'.format(manifest_path)

Quoting the path before splitting prevents the backslashes from getting removed.

carlescufi: OK, nice suggestion. I will go for that one.

carlescufi: @ulfalizer done

ulfalizer:
Maybe the commit message could be updated too:

    Quote the path to the manifest before splitting so that
    backslash path separators don't get removed on
    Windows.

carlescufi:
@ulfalizer updated the commit message

PR #96, opened 2018-11-02
List enhancements

mbolivar:
Add two options to `west list` will be useful when processing its output in programs:

- `-p` / `--path`: just print project paths
- `-a` / `--abspath`: just print project absolute paths (takes precedence)

Also add a `-s` / `--special` option to include the "special" projects (west and the manifest) in the output.

The command `west list -as west` will be needed by the Zephyr documentation system to find the West source code for use in autodoc Sphinx directives.

The command `west list -a` will be needed by the Zephyr build system to obtain the paths of all West projects to consider for building.

### DIFF
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -38,12 +38,10 @@ class List(WestCommand):
             '''))
 
     def do_add_parser(self, parser_adder):
-        return _add_parser(parser_adder, self)
+        return _add_parser(parser_adder, self, _project_list_arg)
 
     def do_run(self, args, user_args):
-        log.inf("Manifest path: {}\n".format(_manifest_path(args)))
-
-        for project in _all_projects(args):
+        for project in _projects(args):
             log.inf('{:14}  {:18}  {:13}  {}  {}'.format(
                 project.name,
                 project.path,
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -24,6 +24,9 @@ from west.manifest import default_path, Remote, Project, \
 # branch.
 _MANIFEST_REV_BRANCH = 'manifest-rev'
 
+# Names of the "meta" projects in the west directory.
+_META_NAMES = ['west', 'manifest']
+
 
 class List(WestCommand):
     def __init__(self):
@@ -32,22 +35,81 @@ class List(WestCommand):
             _wrap('''
             List projects.
 
-            Prints the path to the manifest file and lists all projects along
-            with their clone paths and manifest revisions. Also includes
-            information on which projects are currently cloned.
-            '''))
+            Individual projects can be specified by name.
+
+            By default, lists all project names in the manifest, along with
+            each project's path, revision, URL, and whether it has been cloned.
+
+            The west and manifest repositories in the top-level west directory
+            are not included by default. Use --all or the special project
+            names "west" and "manifest" to include them.'''))
 
     def do_add_parser(self, parser_adder):
-        return _add_parser(parser_adder, self, _project_list_arg)
+        default_fmt = '{name:14} {path:18} {revision:13} {url} {cloned}'
+        return _add_parser(
+            parser_adder, self,
+            _arg('-a', '--all', action='store_true',
+                 help='''Do not ignore repositories in west/ (i.e. west and the
+                 manifest) in the output. Since these are not part of
+                 the manifest, some of their format values (like "revision")
+                 come from other sources. The behavior of this option is
+                 modeled after the Unix ls -a option.'''),
+            _arg('-f', '--format', default=default_fmt,
+                 help='''Format string to use to list each project; see
+                 FORMAT STRINGS below.'''),
+            _project_list_arg,
+            epilog=textwrap.dedent('''\
+            FORMAT STRINGS
+
+            Projects are listed using a Python 3 format string. Arguments
+            to the format string are accessed by name.
+
+            The default format string is:
+
+            "{}"
+
+            The following arguments are available:
+
+            - name: project name in the manifest
+            - url: full remote URL as specified by the manifest
+            - path: the relative path to the project from the top level,
+              as specified in the manifest where applicable
+            - abspath: absolute and normalized path to the project
+            - revision: project's manifest revision
+            - cloned: "(cloned)" if the project has been cloned, "(not cloned)"
+              otherwise
+            - clone_depth: project clone depth if specified, "None" otherwise
+            '''.format(default_fmt)))
 
     def do_run(self, args, user_args):
-        for project in _projects(args):
-            log.inf('{:14}  {:18}  {:13}  {}  {}'.format(
-                project.name,
-                project.path,
-                project.revision,
-                project.url,
-                "(cloned)" if _cloned(project) else "(not cloned)"))
+        # We should only list the meta projects if they were explicitly
+        # given by name, or --all was given.
+        list_meta = bool(args.projects) or args.all
+
+        for project in _projects(args, include_meta=True):
+            if project.name in _META_NAMES and not list_meta:
+                continue
+
+            # Spelling out the format keys explicitly here gives us
+            # future-proofing if the internal Project representation
+            # ever changes.
+            try:
+                result = args.format.format(
+                    name=project.name,
+                    url=project.url,
+                    path=project.path,
+                    abspath=project.abspath,
+                    revision=project.revision,
+                    cloned="(cloned)" if _cloned(project) else "(not cloned)",
+                    clone_depth=project.clone_depth or "None")
+            except KeyError as e:
+                # The raised KeyError seems to just put the first
+                # invalid argument in the args tuple, regardless of
+                # how many unrecognizable keys there were.
+                log.die('unknown key "{}" in format string "{}"'.
+                        format(e.args[0], args.format))
+
+            log.inf(result)
 
 
 class Clone(WestCommand):
@@ -423,16 +485,19 @@ _no_update_arg = _arg(
 _project_list_arg = _arg('projects', metavar='PROJECT', nargs='*')
 
 
-def _add_parser(parser_adder, cmd, *extra_args):
+def _add_parser(parser_adder, cmd, *extra_args, **kwargs):
     # Adds and returns a subparser for the project-related WestCommand 'cmd'.
     # All of these commands (currently) take the manifest path flag, so it's
-    # hardcoded here.
+    # provided by default here, but any defaults can be overridden with kwargs.
 
-    return parser_adder.add_parser(
-        cmd.name,
-        description=cmd.description,
-        formatter_class=argparse.RawDescriptionHelpFormatter,
-        parents=(_manifest_arg,) + extra_args)
+    if 'description' not in kwargs:
+        kwargs['description'] = cmd.description
+    if 'formatter_class' not in kwargs:
+        kwargs['formatter_class'] = argparse.RawDescriptionHelpFormatter
+    if 'parents' not in kwargs:
+        kwargs['parents'] = (_manifest_arg,) + extra_args
+
+    return parser_adder.add_parser(cmd.name, **kwargs)
 
 
 def _wrap(s):
@@ -469,7 +534,7 @@ def _cloned_projects(args):
         [project for project in _all_projects(args) if _cloned(project)]
 
 
-def _projects(args, listed_must_be_cloned=True):
+def _projects(args, listed_must_be_cloned=True, include_meta=False):
     # Returns a list of project instances for the projects requested in 'args'
     # (the command-line arguments), in the same order that they were listed by
     # the user. If args.projects is empty, no projects were listed, and all
@@ -482,17 +547,25 @@ def _projects(args, listed_must_be_cloned=True):
     # listed_must_be_cloned (default: True):
     #   If True, an error is raised if an uncloned project was listed. This
     #   only applies to projects listed explicitly on the command line.
+    #
+    # include_meta (default: False):
+    #   If True, "meta" projects (i.e. west and the manifest) may be given
+    #   in args.projects without raising errors, and are also included in the
+    #   return value if args.projects is empty.
 
     projects = _all_projects(args)
+
+    if include_meta:
+        projects += [_special_project(name) for name in _META_NAMES]
 
     if not args.projects:
         # No projects specified. Return all projects.
         return projects
 
-    # Got a list of projects on the command line. First, check that they exist
-    # in the manifest.
+    # Got a list of projects on the command line. First, check that they exist.
 
-    project_names = [project.name for project in projects]
+    name_to_project = {project.name: project for project in projects}
+    project_names = name_to_project.keys()
     nonexistent = set(args.projects) - set(project_names)
     if nonexistent:
         log.die('Unknown project{} {} (available projects: {})'
@@ -501,20 +574,28 @@ def _projects(args, listed_must_be_cloned=True):
                         ', '.join(project_names)))
 
     # Return the projects in the order they were listed
-    res = []
-    for name in args.projects:
-        for project in projects:
-            if project.name == name:
-                res.append(project)
-                break
+    res = [name_to_project[name] for name in args.projects]
 
     # Check that all listed repositories are cloned, if requested
     if listed_must_be_cloned:
-        uncloned = [prj.name for prj in res if not _cloned(prj)]
+        # We could still get here with a missing manifest repository if the
+        # user gave a --manifest argument.
+        uncloned_meta = [prj.name for prj in res if not _cloned(prj)
+                         and prj.name in _META_NAMES]
+        if uncloned_meta:
+            log.die('Missing meta project{}: {}.'.
+                    format('s' if len(uncloned_meta) > 1 else '',
+                           ', '.join(uncloned_meta)),
+                    'The Zephyr installation has been corrupted.')
+
+        uncloned = [prj.name for prj in res
+                    if not _cloned(prj) and prj.name not in _META_NAMES]
         if uncloned:
-            log.die('The following projects are not cloned: {}. Please clone '
-                    "them first (with 'west fetch')."
-                    .format(", ".join(uncloned)))
+            log.die('Uncloned project{}: {}.'.
+                    format('s' if len(uncloned) > 1 else '',
+                           ", ".join(uncloned)),
+                    'Please run "west clone {}"'.format(' '.join(uncloned)),
+                    'to clone.')
 
     return res
 
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -46,7 +46,7 @@ class List(WestCommand):
         for project in _all_projects(args):
             log.inf('{:14}  {:18}  {:13}  {}  {}'.format(
                 project.name,
-                os.path.join(project.path, ''),  # Add final '/' if missing
+                project.path,
                 project.revision,
                 project.url,
                 "(cloned)" if _cloned(project) else "(not cloned)"))
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -16,16 +16,13 @@ from west import log
 from west import util
 from west.commands import WestCommand
 from west.manifest import default_path, Remote, Project, \
-                          Manifest, MalformedManifest
+                          Manifest, MalformedManifest, META_NAMES
 
 
 # Branch that points to the revision specified in the manifest (which might be
 # an SHA). Local branches created with 'west branch' are set to track this
 # branch.
 _MANIFEST_REV_BRANCH = 'manifest-rev'
-
-# Names of the "meta" projects in the west directory.
-_META_NAMES = ['west', 'manifest']
 
 
 class List(WestCommand):
@@ -87,7 +84,7 @@ class List(WestCommand):
         list_meta = bool(args.projects) or args.all
 
         for project in _projects(args, include_meta=True):
-            if project.name in _META_NAMES and not list_meta:
+            if project.name in META_NAMES and not list_meta:
                 continue
 
             # Spelling out the format keys explicitly here gives us
@@ -556,7 +553,7 @@ def _projects(args, listed_must_be_cloned=True, include_meta=False):
     projects = _all_projects(args)
 
     if include_meta:
-        projects += [_special_project(name) for name in _META_NAMES]
+        projects += [_special_project(name) for name in META_NAMES]
 
     if not args.projects:
         # No projects specified. Return all projects.
@@ -581,7 +578,7 @@ def _projects(args, listed_must_be_cloned=True, include_meta=False):
         # We could still get here with a missing manifest repository if the
         # user gave a --manifest argument.
         uncloned_meta = [prj.name for prj in res if not _cloned(prj)
-                         and prj.name in _META_NAMES]
+                         and prj.name in META_NAMES]
         if uncloned_meta:
             log.die('Missing meta project{}: {}.'.
                     format('s' if len(uncloned_meta) > 1 else '',
@@ -589,7 +586,7 @@ def _projects(args, listed_must_be_cloned=True, include_meta=False):
                     'The Zephyr installation has been corrupted.')
 
         uncloned = [prj.name for prj in res
-                    if not _cloned(prj) and prj.name not in _META_NAMES]
+                    if not _cloned(prj) and prj.name not in META_NAMES]
         if uncloned:
             log.die('Uncloned project{}: {}.'.
                     format('s' if len(uncloned) > 1 else '',
--- a/src/west/manifest-schema.yml
+++ b/src/west/manifest-schema.yml
@@ -82,7 +82,8 @@ mapping:
       # Each project is a map with the following keys:
       #
       # - name: Mandatory, the name of the git repository. The clone
-      #   URL is formed by remote + '/' + name
+      #   URL is formed by remote + '/' + name. The name cannot
+      #   be one of the reserved values "west" and "manifest".
       # - remote: Optional, the name of the remote to pull it from.
       #   If the remote is missing, the remote'key in the top-level
       #   defaults key is used instead. If both are missing, it's an error.
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -24,6 +24,11 @@ import yaml
 from west import util, log
 
 
+META_NAMES = ['west', 'manifest']
+'''Names of the special "meta-projects", which are reserved and cannot
+be used to name a project in the manifest file.'''
+
+
 def default_path():
     '''Return the path to the default manifest in the west directory.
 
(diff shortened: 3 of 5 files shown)


mbolivar:
Whoops, the commit message on the third patch is out of date. Fixing now.

src/west/commands/project.py
```diff
@@ -46,7 +46,7 @@ def do_run(self, args, user_args):
         for project in _all_projects(args):
             log.inf('{:14}  {:18}  {:13}  {}  {}'.format(
                 project.name,
-                os.path.join(project.path, ''),  # Add final '/' if missing
+                project.path,
```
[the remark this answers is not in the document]
ulfalizer: This was just a small nice-to-have, to make it clear in the output that it's a directory. You can remove it if you think it helps though.

Note that `foo//bar` works the same as `foo/bar`, so redundant slashes won't break anything when opening files. This is POSIX behavior and built right into the kernel.

src/west/commands/project.py
```diff
@@ -38,18 +38,30 @@ def __init__(self):
...
-                "(cloned)" if _cloned(project) else "(not cloned)"))
+        if not (args.path or args.abspath):
+            log.inf("Manifest path: {}\n".format(_manifest_path(args)))
+
+        for project in _projects(args):
+            if args.abspath:  # This must come before the args.path test.
+                log.inf(project.abspath)
+            elif args.path:
+                log.inf(project.path)
+            else:
```
ulfalizer: This feels a bit overly specialized to me, just to get rid of some parsing of fixed-format strings.

I wonder if we could have something like this:

    $ west list --format="{name} {path}"

That would print the project name and the relative path (could have `{abspath}` or the like for absolute paths), and would be super simple to implement:

    format_str.format(name=project.name, path=project.path, ...)

The default behavior with no flags could just be a particular format string.

src/west/commands/project.py
```diff
@@ -44,13 +44,16 @@ def do_add_parser(self, parser_adder):
                            _arg('-a', '--abspath', action='store_true',
                                 help='''just print absolute paths (if given
                                 with -p, this option takes precedence)'''),
+                           _arg('-s', '--special', action='store_true',
+                                help='''include output for "special" projects
+                                (west itself and the manifest repository)'''),
```
ulfalizer: I think it would be more general to have `--projects`, `--west`, and `--manifest` flags. Then you could just list `west` if you wanted to, for example, without getting the other projects in there.

To list just the absolute path to `west`, you'd do

    west list --west --format="{abspath}"

The default would be something like

    west list --projects --west --manifest --format="<default format string>"

carlescufi: I agree that @ulfalizer's proposal feels cleaner. I would have the following:

* `west list` by default lists `--projects`
* `west list -p/--projects`
* `west list -w/--west`
* `west list -m/--manifest`
* `west list -a/--all`

I don't really think we need the `--format` instead of the `--abspath` but I don't mind

src/west/commands/project.py
carlescufi: No preference here for myself, I don't really mind.

tejlmand: No special preference either.
But a small followup question.

Using `--format="{name} {path}"` we get stuff, such as width formatting, for free from built-in python.
But we still need to parse all format names, i.e. `name`, `path`, `abspath`, `...`, or do you have something special lib in mind, which would reduce the parsing we need to do ?

src/west/commands/project.py
tejlmand: `--format` could be a cleaner way, especially if that is a general way to format output from `west`.
That way, user doesn't have to think, should I use `-p` or `-a` or `--something-else` in the various `west <command>` options.

There will always be a `--format`, although what and  how to format may vary.

@carlescufi `--porcelain` is normally intended for humans and have higher risk of changing formatting. Whereas for a script I guess a low-level output command with fixed output is more suited.

carlescufi: @tejlmand from `git help status`:

```
       --porcelain[=<version>]
           Give the output in an easy-to-parse format for scripts. This is similar to the short output, but will remain stable across Git versions and regardless of user configuration. See below for details.

           The version parameter is used to specify the format version. This is optional and defaults to the original version v1 format.
```

src/west/commands/project.py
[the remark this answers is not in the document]
tejlmand: Personally I have no special feelings on this, although the version without `/` (@mbolivar suggestion) at the end is more in-line what you will see with `ls` or `dir` in windows.

Although in both windows you'll see `<DIR>` and in linux, you most often has coloring of the output, which help to identify folders when printed on screen (or `ls -l` shows a `d---------` at start).
Such identification we don't have in `west list`.

But what I think is most important is that we stay consistent.
Either a trailing `/` at the end everywhere we print folders, or not at all.

src/west/commands/project.py
carlescufi: Or actually to simplify another option would be to have:

`west list` (lists all projects)
and
`west list --all` (lists all projects + west + manifest)

src/west/commands/project.py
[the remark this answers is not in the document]
mbolivar: > In windows you'll see <DIR> and in linux, you most often has coloring of the output, which help > to identify folders when printed on screen (or ls -l shows a d--------- at start).
> Such identification we don't have in west list.
> But what I think is most important is that we stay consistent.
> Either a trailing `/` at the end everywhere we print folders, or not at all.

I played around with the "dir" options on windows, and I couldn't find any one that appends a trailing '\' to the output, which is what we get with "west list" right now. So this patch makes west consistent with both `ls` on POSIX operating systems and `dir` on Windows. I'm going to keep it as-is since we all seem OK with that.



As @tejlmand points out, the default `dir` output prefixes directories with <DIR>:

```
C:\Users\mbolivar\west>dir
 Volume in drive C is Windows
 Volume Serial Number is 6072-12F7

 Directory of C:\Users\mbolivar\west

11/01/2018  05:50 AM    <DIR>          .
11/01/2018  05:50 AM    <DIR>          ..
11/01/2018  05:49 AM                81 .gitignore
11/01/2018  05:50 AM    <DIR>          .pytest_cache
11/01/2018  05:49 AM               728 .shippable.yml
11/01/2018  05:49 AM            11,558 LICENSE
11/01/2018  05:49 AM             4,812 README.rst
11/01/2018  05:49 AM    <DIR>          scripts
11/01/2018  05:49 AM             1,348 setup.py
11/01/2018  05:49 AM    <DIR>          src
11/01/2018  05:49 AM    <DIR>          tests
11/01/2018  05:49 AM                36 tests_requirements.txt
               6 File(s)         18,563 bytes
               6 Dir(s)  78,789,935,104 bytes free
```

The "wide" option surrounds directories with square brackets:

```
C:\Users\mbolivar\west>dir /w
 Volume in drive C is Windows
 Volume Serial Number is 6072-12F7

 Directory of C:\Users\mbolivar\west

[.]                      [..]                     .gitignore               [.pytest_cache]          .shippable.yml           LICENSE
README.rst               [scripts]                setup.py                 [src]                    [tests]                  tests_requirements.txt
               6 File(s)         18,563 bytes
               6 Dir(s)  78,789,935,104 bytes free
```

The "bare" format just prints the names:

```
C:\Users\mbolivar\west>dir /b
.gitignore
.pytest_cache
.shippable.yml
LICENSE
README.rst
scripts
setup.py
src
tests
tests_requirements.txt
```

Nothing appends a backslash though.

mbolivar:
Thanks for all the feedback!

For context, I did it this quick and specific way because I thought we were in a hurry, this feature is a blocker for creating the `topic-west` branch in Zephyr, and `repo list` has the same style of options.

I'm fine with spending more time to develop a general solution since that seems to be the consensus among reviewers -- though my personal vote would be to merge simple options that do what we need now and add a more general solution later, deprecating the old way if we need to. 

Doing so would be easy and, from what I can see, harmless. Consider roughly equivalent commands like `git log --oneline` and `git log --pretty='%h %s'` for instance.

Anyway, I will review the feedback and try to synthesize a more general solution from it.

src/west/commands/project.py
ulfalizer: Personally, I would get rid of `--all` and have `west list` act as an alias for `west list --west --manifest --projects`.

That's easy to explain and understand, and everything else can be implemented in terms of it.

Similarly, leaving out `--format` would just substitute some default hardcoded `--format` string in the code, so that everything can go through the same path.

I'm not sure we need some kind of `--porcelain` command if we have the `--format` parameter, at least not yet. If something turns out to be tricky to parse, we could always add more replacement fields that do porcelainish stuff.

ulfalizer: `--format` is better than having extra flags for particular hardcoded formats I think, especially if we need to add more stuff in the future.

I think it's a good idea to see if one thing can be implemented in terms of another thing, and to try to go for the most "powerful" construct in those terms.

ulfalizer:
@mbolivar 
I probably wouldn't have suggested this one if it had been trickier to implement. Note that you get almost all of it for free from `format()` (see https://github.com/zephyrproject-rtos/west/pull/96#discussion_r230936343).

mbolivar:
> I probably wouldn't have suggested this one if it had been trickier to implement. Note that you get almost all of it for free from `format()` (see [#96 (comment)](https://github.com/zephyrproject-rtos/west/pull/96#discussion_r230936343)).

I'm not concerned about the difficulty of implementation. That's easy.

My point is that inviting the development of a general mechanism takes longer. So far we have the following on the table:

- `--path` and `--abspath` are too specific, and could be replaced with a more general `--format` option that uses python format strings to access Project object attributes (@ulfalizer)
- we could have `--west` and `--manifest` flags in addition to `--projects` to list those things too (@ulfalizer)
- or add an `--all` option to those (@carlescufi)
- or just have `--all`, eliminating `--projects`, `--west`, and `--manifest` (@carlescufi)
- or add a `--porcelain` option that produces "machine-readable" output (@carlescufi)
- `west list --special zephyr` is not intuitive because it prints output for zephyr, but zephyr is not a special project (@tejlmand)
- perhaps a  `-p / --projects <all, [west, manifest,zephyr,...]>` (@tejlmand)

We now need to settle on specifics, which will take another day or two with all the round trips between Europe and the states.

The cost here is time to design, not complexity. Obviously if you have infinite time to pick the most powerful abstraction then you define everything in terms of category theory, but that ends up taking a while ;).

ulfalizer:
FWIW, I like ``--west``, ``--manifest``, ``--projects`` and ``--format``. Everything else can be implemented in terms of those, and they are super simple to explain and implement. You just list the stuff you want to print, and how you want it printed.

That's probably less abstracted than some of the other things too, while being more powerful.

src/west/commands/project.py
tejlmand: @ulfalizer ahh, right.
And `name`, `path`, etc. being set to the internal variable value in project.name, project.path, variables in `west` in `format_str.format(name=project.name, path=project.path, ...)` we only need to keep the local names in the formatting string fixed. Other parts of `west` can change as we find fit.

I actually like that, as it gives freedom to users to fix their formatting and keep it fixed, and even if additional output data/possibilities is provided later.
Such additions will just be silently dropped, if user don't print them.

Used in general throughout west, could gives the script output stability we would like.

src/west/commands/project.py
> Personally, I would get rid of `--all` and have `west list` act as an alias for `west list --west --manifest --projects`.
> 

Not so fond of having specific projects directly in the arguments, such as: `--manifest`, `--west`.
I would prefer those to be values to an argument, such as `--special west` or `--projects west`, depending on what we settle upon.

> Similarly, leaving out `--format` would just substitute some default hardcoded `--format` string in the code, so that everything can go through the same path.
> 
> I'm not sure we need some kind of `--porcelain` flag if we have the `--format` flag, at least not yet. If something turns out to be tricky to parse, we could always add more replacement fields that do porcelainish stuff.

Currently, I think the `--format` is the nicest solution, but we should ensure good description / examples in the help text, so people don't have to look in source code to find out values to print.

`--porcelain` could be thought of as just one particular fixed formatting string, if we would like to make easy to use, for majority.

tejlmand:
Think i'm aligned with @ulfalizer on `--format`

as there will be no need for a special `--abspath` argument, which main purpose is aiding Sphinx.

> The command west list -as west will be needed by the Zephyr documentation system to find the West source code for use in autodoc Sphinx directives.

which would mean the above could be implemented as: `west list --projects=west --format="{fullpath}"`

Two notes:
I used `fullpath` instead of `abspath` as fullpath is not abbreviated, which I think is a little nicer
I used `=` as separator between argument and value, as that was used by @ulfalizer in the `--format`

ulfalizer:
@tejlmand 
With that, you'd get `west list --projects=west,projects ...` or the like, which seems a bit weird. It's also more typing than `west list --west --projects`.

"Absolute path" is widespread terminology (got e.g. `os.path.abspath()`), so I think we ought to go for that. "Full path" would make me suspect that a "non-full" path is just the basename or something.

carlescufi:
@mbolivar yes, you are right, this is a bottleneck for such a trivial issue.
I have tried to gather the feedback from everybody in a way that makes everyone happy.

* Use `--format=` to specify the output
* `west list` by default lists `--projects` (I insist on this one, we should not by default show `west` and `manifest` as projects)
* `west list -p/--projects`
* `west list -w/--west`
* `west list -m/--manifest`

This will allow you to do `west list -pwm` to see them all, and it has a nice ring to it ;)

This is in line with what @ulfalizer pointed out for the most part, and I think @tejlmand is fine with it too.

tejlmand:
@carlescufi nice summary.

I can agree to that proposal.

ulfalizer:
@carlescufi 
Looks good to me. Could probably have a plain `west list` print whatever "friendly" format as well, but just printing the projects seems fine in the beginning.

ulfalizer:
Accepting a list of projects makes sense to me, but `--meta` seems like overkill for what will probably only ever be two projects (and it'd still be overkill with ten of them, IMO).

Is there a reason why we can't allow allow both regular projects and special projects at the same time? For example, `west list --west foo bar` could list the special west project, the `foo` project, and the `bar` project.

That seems simpler to me, while also being more powerful.

carlescufi:
> Accepting a list of projects makes sense to me, but `--meta` seems like overkill for what will probably only ever be two projects (and it'd still be overkill with ten of them, IMO).
> 
> Is there a reason why we can't allow both regular projects and special projects at the same time? For example, `west list --west foo bar` could list the special west project, the `foo` project, and the `bar` project.
> 
> That seems simpler to me, while also being more powerful.

Thinking back at this, I think this would work as well, yes. I have no objections to using this instead of `--meta`, unless I am missing a usecase here that we haven't covered. It would be good to hear from @mbolivar and @tejlmand. I think both approaches have merit, and they both cover more or less the same functionality, with a couple of drawbacks for each:

* If we use meta, then we cannot list a "meta" repo alongside a regular one
* If we use `--west` and `--manifest` then we need to specify both and adding a new meta-repo would require a new cmd-line param

mbolivar:
```
Is there a reason why we can't allow both regular projects and special projects at the same time? For example, west list --west foo bar could list the special west project, the foo project, and the bar project.
```

@tejlmand was uncomfortable with this type of option because it leads to commands like `west list --west zephyr`, which would only list zephyr. This is seen as potentially confusing to users, unlike the `--meta` option which is a wholesale switch to the other 'type' of project.

mbolivar:
@ulfalizer would you be ok if we did `s/--meta/--west` and otherwise kept the same behavior? The meaning would be "operate on the `west` directory, not the "normal" projects", which I think is still clean, not overkill, and future-proof.

carlescufi:
> ```
> Is there a reason why we can't allow both regular projects and special projects at the same time? For example, west list --west foo bar could list the special west project, the foo project, and the bar project.
> ```
> @tejlmand was uncomfortable with this type of option because it leads to commands like `west list --west zephyr`, which would only list zephyr. This is seen as potentially confusing to users, unlike the `--meta` option which is a wholesale switch to the other 'type' of project.

I think there is a misunderstanding here. I believe that `west list --west zephyr` would list **both** `west` and `zephyr`.

tejlmand:
> @tejlmand was uncomfortable with this type of option because it leads to commands like west list --west zephyr, which would only list zephyr. This is seen as potentially confusing to users, unlike the --meta option which is a wholesale switch to the other 'type' of project.

my concern was with the `west list --include-meta zephyr` proposal, which on one hand states, list the meta projects, while at the same time only show zephyr project.

I believe I was also open for the earlier `west list --west -p` which is very similar to the now proposed `west list --west zephyr` with only difference `-p` is gone.
Think the irc discussion took of, when @mbolivar pointed out that `-p` was not specific enough, as he needed to be able to specify `west list [project]`

So i'm fine with the adjusted proposal and think that is clear as well.

Just to verify, this would give:
```
west list -> all projects
west list --west -> west only
west list --west zephyr -> west and zephyr project only
west list --manifest zephyr -> manifest and zephyr project only
```

carlescufi:
> west list -> all projects west list --west -> west only west list --west zephyr -> west and zephyr project only west list --manifest zephyr -> manifest and zephyr project only

Yes, that is my expectation.

carlescufi:
> @ulfalizer would you be ok if we did `s/--meta/--west` and otherwise kept the same behavior? The meaning would be "operate on the `west` directory, not the "normal" projects", which I think is still clean, not overkill, and future-proof.

This is also a very good suggestion. Both are very similar in fact, still no real preference on my side.

ulfalizer:
@carlescufi @mbolivar 
Yeah, my thinking was that `west list --west zephyr` would list the special `west` project and the regular `zephyr` project (and `west list --west --manifest foo bar` would list the special `west` and `manifest` projects and the regular `foo` and `bar` projects).

`--west` and `--manifest` take no arguments. They just say to include that special repository.

> @ulfalizer would you be ok if we did `s/--meta/--west` and otherwise kept the same behavior? The meaning would be "operate on the `west` directory, not the "normal" projects", which I think is still clean, not overkill, and future-proof.

Could you give some example commands, similar to above?

One reason I don't like it is that it seems that it'd both add extra argument checking code and make commands longer, while also being less general.

I think the flags are future proof. If another special repository shows up, we could add another flag. I don't think we will ever have lots of special repositories (and you could always add extra helper flags if that ever happens).

src/west/commands/project.py
```diff
@@ -38,16 +38,64 @@ def __init__(self):
...
+        The following arguments are available:
+
+        - name: the name of the project in the manifest
+        - url: the remote full URL of the project as specified by the manifest
+        - path: the relative path to the project as given in the manifest
+          (or relative to the top level directory, if --meta is given)
+        - abspath: an absolute and normalized path to the project as it
+          is cloned on the local file system
+        - revision: the manifest revision of the project
+        - cloned: the string "(cloned)" if the project has been cloned,
```
tejlmand: just thinking out loud here: could status be a better term ?
and then status can be: `cloned`, `not cloned` and at the same extendable with other statuses in future.
comments / other thoughts ?

carlescufi:
@ulfalizer 

> Could you give some example commands, similar to above?

Let me try:

`west list` -> list all "normal" projects
`west list --west` -> list all "west" projects (manifest and west)
`west list zephyr` -> list zephyr only
`west list zephyr kconfiglib` -> list zephyr and kconfiglib
`west list --west manifest` -> list manifest only
`west list west` -> error out, west is not part of the "normal" projects
`west list --west zephyr` -> error out, zephyr is not part of the "west" projects

ulfalizer:
Here's some docs for my proposal by the way, which might clarify it:

```
--west: include the special west project in the list
--manifest: include the special manifest project in the list
```

If we ever get a ton of special projects (probably isn't gonna happen), we could add something like this:

```
--special: include all special projects in the list (this is an alias for --west --manifest ...)
```

ulfalizer:
@mbolivar 
Hmm... yeah, I hadn't thought of that flag collision. We could always just rename either the old or the new flag. Feels like we're throwing the baby out with the bathwater.

I don't understand how the `-a` flag works in the example. What's the difference between `west list west` and `west list -a west`? Was that a typo? Why doesn't `west list -a zephyr` include special projects?

If we're gonna forbid the user from naming projects the same as special projects, then why have `-a` at all? `west list west` is unambiguous on its own.

Not a fan of adding code to check that you don't name a normal project the same as a special project. It's adding code to restrict generality to work around an interface issue. If we go that route, we should document that the `west` and `manifest` names are special at least, and do a magic when passed to `west list`.

The only interface issue with the flags is that you can't give the order for special projects. Wonder if that's really that much of a problem, and it could be fixed.

ulfalizer:
To get things moving along, I'd be fine with just making `west` and `manifest` magic names (though I prefer the flags). At that point, I don't see the point of `-a`.

On the plus side, you can order special projects w.r.t. to other projects. On the minus side, you get magic names and arbitrary restrictions on project names.

mbolivar:
@ulfalizer it works just like `ls -a`, if west and the manifest were hidden directories. Does that help? West list -a west was not a typo.

mbolivar:
The point of -a is that `west list -a` lists everything, not just regular projects.

I don't understand the "sacrifices generality" argument. The `ls` command seems pretty general purpose to me.

carlescufi:
Tested last revision, works as intended:

```
carles@zephyr-nrf ~/src/wncs
$ west list
WARNING: ZEPHYR_BASE=/home/carles/src/zephyr in the calling environment, but has been set to /home/carles/src/wncs/zephyr instead by the manifest file /home/carles/src/wncs/west/manifest/default.yml
zephyr         zephyr             master        https://github.com/zephyrproject-rtos/zephyr (cloned)

carles@zephyr-nrf ~/src/wncs
$ west list -a
WARNING: ZEPHYR_BASE=/home/carles/src/zephyr in the calling environment, but has been set to /home/carles/src/wncs/zephyr instead by the manifest file /home/carles/src/wncs/west/manifest/default.yml
zephyr         zephyr             master        https://github.com/zephyrproject-rtos/zephyr (cloned)
west           west/west          list-enhancements https://github.com/mbolivar/west.git (cloned)
manifest       west/manifest      master        https://github.com/zephyrproject-rtos/manifest (cloned)

carles@zephyr-nrf ~/src/wncs
$ west list zephyr west
WARNING: ZEPHYR_BASE=/home/carles/src/zephyr in the calling environment, but has been set to /home/carles/src/wncs/zephyr instead by the manifest file /home/carles/src/wncs/west/manifest/default.yml
zephyr         zephyr             master        https://github.com/zephyrproject-rtos/zephyr (cloned)
west           west/west          list-enhancements https://github.com/mbolivar/west.git (cloned)

carles@zephyr-nrf ~/src/wncs
$ west list -a manifest
WARNING: ZEPHYR_BASE=/home/carles/src/zephyr in the calling environment, but has been set to /home/carles/src/wncs/zephyr instead by the manifest file /home/carles/src/wncs/west/manifest/default.yml
manifest       west/manifest      master        https://github.com/zephyrproject-rtos/manifest (cloned)

carles@zephyr-nrf ~/src/wncs
$ west list --format={abspath}
WARNING: ZEPHYR_BASE=/home/carles/src/zephyr in the calling environment, but has been set to /home/carles/src/wncs/zephyr instead by the manifest file /home/carles/src/wncs/west/manifest/default.yml
/home/carles/src/wncs/zephyr

carles@zephyr-nrf ~/src/wncs
$ west list -a --format={abspath}
WARNING: ZEPHYR_BASE=/home/carles/src/zephyr in the calling environment, but has been set to /home/carles/src/wncs/zephyr instead by the manifest file /home/carles/src/wncs/west/manifest/default.yml
/home/carles/src/wncs/zephyr
/home/carles/src/wncs/west/west
/home/carles/src/wncs/west/manifest
```

carlescufi:
@ulfalizer @tejlmand please re-review today

carlescufi:
> @mbolivar
> Hmm... yeah, I hadn't thought of that flag collision. We could always just rename either the old or the new flag. Feels like we're throwing the baby out with the bathwater.

No, because your proposal, as discussed at length on IRC, had consistency issues (see IRC log).

> I don't understand how the `-a` flag works in the example. What's the difference between `west list west` and `west list -a west`? Was that a typo? Why doesn't `west list -a zephyr` include special projects?

Because if you specify a project, `-a` does nothing. It only does something if you use it without arguments. See `man ls`.


> If we're gonna forbid the user from naming projects the same as special projects, then why have `-a` at all? `west list west` is unambiguous on its own.

Because we want to be able to do: `west list -a` with no additional arguments, get all repos, special ones included.

> Not a fan of adding code to check that you don't name a normal project the same as a special project. It's adding code to restrict generality to work around an interface issue. If we go that route, we should document that the `west` and `manifest` names are special at least, and do a magic thing when passed to `west list`.

@mbolivar has already documented this in the code.

src/west/commands/project.py
```diff
@@ -32,22 +35,71 @@ def __init__(self):
...
+            # future-proofing if the internal Project representation
+            # ever changes.
+            log.inf(args.format.format(
+                name=project.name,
+                url=project.url,
+                path=project.path,
+                abspath=project.abspath,
+                revision=project.revision,
+                cloned="(cloned)" if _cloned(project) else "(not cloned)",
+                clone_depth=project.clone_depth or "None"))
```
ulfalizer: Missing error checking. Try

```python
"{foo}".format()
```

Should probably catch the `KeyError`.

src/west/commands/project.py
```diff
@@ -32,22 +35,71 @@ def __init__(self):
...
+            By default, lists all project names in the manifest, along with
+            each project's path, revision, URL, and whether it has been cloned.
+
+            The west and manifest repositories in the top-level west directory
+            are not included by default. Use --all or the special project
+            names "west" and "manifest" to include them.'''))
 
     def do_add_parser(self, parser_adder):
-        return _add_parser(parser_adder, self, _project_list_arg)
+        default_fmt = '{name:14} {path:18} {revision:13} {url} {cloned}'
```
ulfalizer: Maybe roll this into the parameter below, like `default='{name:14} {path:18} {revision:13} {url} {cloned}'`. That makes it clear that it's used in a single spot and never updated.

src/west/commands/project.py
```diff
@@ -408,9 +460,9 @@ def _arg(*args, **kwargs):
 
 # Arguments shared between more than one command
 
-_manifest_arg = _arg(
-    '-m', '--manifest',
-    help='path to manifest file (default: west/manifest/default.yml)')
+_MANIFEST_ARG_HELP = 'path to manifest file (default: west/manifest/default.yml)'
```
ulfalizer: Single use. Could be rolled into the assignment below.

src/west/commands/project.py
```diff
@@ -482,17 +537,25 @@ def _projects(args, listed_must_be_cloned=True):
     # listed_must_be_cloned (default: True):
     #   If True, an error is raised if an uncloned project was listed. This
     #   only applies to projects listed explicitly on the command line.
+    #
+    # include_meta (default: False):
+    #   If True, "meta" projects (i.e. west and the manifest) may be given
+    #   in args.projects without raising errors, and are also included in the
+    #   return value if args.projects is empty.
```
ulfalizer: So `west list west` raises an error? That's inconsistent with `ls`, but that's just a note...

`ls .foo` works fine.

src/west/commands/project.py
```diff
@@ -32,22 +35,71 @@ def __init__(self):
...
     def do_add_parser(self, parser_adder):
-        return _add_parser(parser_adder, self, _project_list_arg)
+        default_fmt = '{name:14} {path:18} {revision:13} {url} {cloned}'
+        return _add_parser(
+            parser_adder, self,
+            _arg('-a', '--all', action='store_true',
+                 help='''Include repositories in west/ (i.e. west and the
+                 manifest) in the output. Since these are not part of
+                 the manifest, some of their format values (like "revision")
+                 come from other sources.'''),
```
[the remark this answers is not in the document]
carlescufi: >If this is modeled after ls -a, then it would help to mention it in the documentation.

+1

src/west/commands/project.py
carlescufi: No, `west list west` does not raise an error:

```
C:\Users\Carles\src\wncs>west list west
west           west\west          list-enhancements https://github.com/mbolivar/west.git (cloned)
```

src/west/commands/project.py
mbolivar: I am aware of your antipathy towards single-use variables. The one below is leftovers from rebasing.

This one is used in two places, however:

1. as the default kwarg
2. in the epilog

ulfalizer: Ah, sorry, search fail.

Issue #98, opened 2018-11-06
West crashes if default.yml doesn't exist


ulfalizer:
Looks like there's some new stuff in `set_zephyr_base()` that's causing it:

```
Traceback (most recent call last):
  File "/home/ulf/trywest/west/west/src/west/main.py", line 229, in <module>
    main()
  File "/home/ulf/trywest/west/west/src/west/main.py", line 201, in main
    args, unknown = parse_args(argv)
  File "/home/ulf/trywest/west/west/src/west/main.py", line 184, in parse_args
    set_zephyr_base(args)
  File "/home/ulf/trywest/west/west/src/west/main.py", line 90, in set_zephyr_base
    manifest = Manifest.from_file()
  File "/home/ulf/trywest/west/west/src/west/manifest.py", line 52, in from_file
    return Manifest(source_file=source_file)
  File "/home/ulf/trywest/west/west/src/west/manifest.py", line 81, in __init__
    with open(source_file, 'r') as f:
FileNotFoundError: [Errno 2] No such file or directory: '/home/ulf/trywest/west/manifest/default.yml'
```

ulfalizer:
One bad thing about this is that you can't use `west init` to reset back to a working manifest, because it's using some new `west update` flags.

tejlmand:
@ulfalizer Proposed a fix in #99, which allows you to get out of the situation (but doesn't silence the stack trace)

PR #99, opened 2018-11-06
commands: project: Allow changing configuration options with 'west init' (w. reset)

ulfalizer:
Three commits:

```
Support rerunning 'west init' in an already initialized West
installation to update configuration values in west/config.

Only arguments that are explicitly passed are changed. For example,
'west init --mr foo' will update the manifest revision to 'foo', leaving
all other settings unchanged. Passing no arguments when reinitializing
is a no-op, and prints an error.

The initial west/config creation is treated as an update that happens to
set all configuration values, to reuse some code.

Updating the manifest URL or revision via 'west init' automatically
resets the manifest to the new revision (after fetching it). This is
implemented with a new 'west update --reset-manifest' flag. Updating the
manifest also resets all (cloned) projects, via a new
'west update --reset-projects' flag.

Updating the west URL/revision resets the west repository, via a new
'west update --reset-west' flag.

All resetting is done with 'git reset --keep', which is a bit safer than
'git reset --hard', as it never discards uncommitted changes.

The reset behavior can be suppressed by passing --no-reset to
'west init'.

This commit also adds more comprehensive documentation for the behavior
of 'west init'.

Piggyback some small cleanup:

 - Rename WEST/MANIFEST_DEFAULT to WEST/MANIFEST_URL_DEFAULT. There's
   already WEST/MANIFEST_REV_DEFAULT.

 - Rename find_west_topdir() to west_topdir(). This is what the version
   in the main West code is called (would be nice to reuse things
   directly from there in the future).

   Also make west_topdir() implicitly default to the working directory.

Fixes: #77
```

```
tests: project: Clean up tests a bit

Simplify the projects tests to make them easier to understand and work
with:

 - py.path (https://py.readthedocs.io/en/latest/path.html) functions
   like mkdir() return the resulting path, so operations can be chained

 - py.path paths have a write() method. Use it to write the manifest.
   Get rid of the separate file.

 - Switch to the following directory layout, which is simpler and makes
   the manifest easier to locate:

     <tmpdir>/repos/{kconfiglib,net-tools}
     <tmpdir>/west/.west_marker
     <tmpdir>/manifest.yml

   <tmpdir> is used directly as the working directory.

 - Add some more comments
```

```
tests: project: Add tests for 'west update'

Create some dummy manifest, west, and project upstream repositories, and
a simple configuration file that points to them. Use this setup to test
--reset/update-manifest, --reset/update-west, and --reset-projects.

The --reset-* flags are used by 'west init' when reinitializing an
existing west installation.

Some simple helper functions were added for creating repositories,
adding commits, and getting the subject of a repository's HEAD commit.
```


ulfalizer:
Still a bit WIPish. Maybe some stuff could be factored out to common code.

src/west/_bootstrap/main.py
```diff
@@ -109,74 +120,97 @@ def init(argv):
...
         '-b', '--base-url',
-        help='''Base URL for both 'manifest' and 'zephyr' repositories; cannot
-        be given if either -m or -w are''')
+        help='''Base URL for both 'manifest' and 'zephyr' repositories. Cannot
+             be given if either -m or -w are''')
+
     init_parser.add_argument(
         '-m', '--manifest-url',
-        help='Zephyr manifest fetch URL, default ' + MANIFEST_DEFAULT)
+        help='Manifest repository URL (or remote name) (default: {})'
```
mbolivar: Hm, why not leave this one as-is and let the remote name part be an implementation detail? It's a bit confusing what the remote name is for a first time init, which as you've pointed out is the most common case.

src/west/_bootstrap/main.py
```diff
@@ -109,74 +120,97 @@ def init(argv):
...
+To suppress the reset of the manifest, west, and projects, pass --no-reset to
+'west init'. With --no-reset, only the configuration file will be updated, and
+you will have to handle any resetting yourself.
+'''.format(WEST_MARKER))
+
     init_parser.add_argument(
         '-b', '--base-url',
-        help='''Base URL for both 'manifest' and 'zephyr' repositories; cannot
-        be given if either -m or -w are''')
+        help='''Base URL for both 'manifest' and 'zephyr' repositories. Cannot
```
mbolivar: I'm fine with this change, but FYI, I chose semicolons because none of these help texts end with a period, so I wanted to keep them to a single sentence rather than have them be complete sentences with periods except for the last one.

This is partially to keep consistency with the auto-generated help for the `--help` option.

In general I find it annoying that argparse doesn't capitalize the first words in their help text, but that's a side detail.

src/west/_bootstrap/main.py
```diff
@@ -109,74 +120,97 @@ def init(argv):
...
-        dest='manifest_rev',
-        help='Manifest revision to fetch, default ' + MANIFEST_REV_DEFAULT)
+        '--mr', '--manifest-rev', dest='manifest_rev',
+        help='Manifest revision to fetch (default: {})'
+             .format(MANIFEST_REV_DEFAULT))
+
     init_parser.add_argument(
         '-w', '--west-url',
-        help='West fetch URL, default ' + WEST_DEFAULT)
+        help='West repository URL (or remote name) (default: {})'
```
mbolivar: Same here re:remote

src/west/_bootstrap/main.py
```diff
@@ -196,44 +230,118 @@ def init_bootstrap(directory, args):
...
 
-    config_path = os.path.join(directory, WEST_DIR, 'config')
+    print('=== West initialized ===')
+
+
+def reinit(config_path, args):
+    '''
+    Reinitialize an existing installation.
+
+    Currently, this is just a shorthand for updating the west/config
```
mbolivar: That's not true in this PR, is it?

src/west/_bootstrap/main.py
```diff
@@ -196,44 +230,118 @@ def init_bootstrap(directory, args):
...
+        if args.west_url or args.manifest_url:
+            sys.exit('fatal error: -b is incompatible with -m and -w')
+        west_url = args.base_url.rstrip('/') + '/west'
+        manifest_url = args.base_url.rstrip('/') + '/manifest'
+    else:
+        west_url = args.west_url
+        manifest_url = args.manifest_url
+
+    if not (west_url or args.west_rev or manifest_url or args.manifest_rev):
+        sys.exit('West already initialized. Please pass any settings you '
```
mbolivar: Could running it without any changes possibly be useful?

For example, if the user has accidentally deleted the west directory and tries to run `west init` again to fix it. It seems like it would be more helpful to go ahead and do the reinit anyway than stop here even if nothing "seems" different.

What would the downside of doing that be?

src/west/_bootstrap/main.py
```diff
@@ -109,74 +120,97 @@ def init(argv):
...
+        '--wr', '--west-rev', dest='west_rev',
+        help='West revision to fetch (default: {})'
+             .format(WEST_REV_DEFAULT))
+
     init_parser.add_argument(
-        'directory', nargs='?', default=None,
-        help='Initializes in this directory, creating it if necessary')
+        '--nr', '--no-reset', dest='reset', action='store_false',
+        help='''Suppress automatic reset of the manifest, west, and project
+             repositories when rerunning 'west init' to update the
```
mbolivar: What about "when rerunning 'west init' in an existing installation"?

src/west/commands/project.py
```diff
@@ -336,6 +339,10 @@ def __init__(self):
 
             Pass --update-west or --update-manifest to update just that
             repository. With no arguments, both are updated.
+
+            Except for --reset-manifest/--reset-west/--reset-projects, updates
```
mbolivar: "Unless ... are given, updates" ?

src/west/_bootstrap/main.py
mbolivar: I'm also running into a situation where I have pre-0.3.0 installations around that don't have a west/config. Being able to re-run west init to create one would be a nice feature.

If there's no downside to be avoided that I'm missing, I'm not sure this exit is helpful.

src/west/_bootstrap/main.py
ulfalizer: There was supposed to be a period at the end. Fixed.

src/west/_bootstrap/main.py
ulfalizer: Changed.

Gets a bit weird with the parameter being called `...-url` rather than `...-repo` or the like at least.

src/west/_bootstrap/main.py
ulfalizer: Changed

src/west/_bootstrap/main.py
ulfalizer: Fixed

src/west/_bootstrap/main.py
ulfalizer: Guess `west init` would still re-initialize everything if west/ has been deleted, since there's no west installation anymore.

src/west/_bootstrap/main.py
ulfalizer: Yeah, that sounds better. Changed.

src/west/commands/project.py
ulfalizer: Rephrased it a bit

ulfalizer:
Cleaned up the project tests a bit and added tests for `west update` (which is run directly by `west init` when reinitializing).

src/west/_bootstrap/main.py
ulfalizer: The manifest repository still wouldn't get re-cloned without the check though. What it'd do would be to write `=== Updated configuration written to west/config ===`, but actually do nothing (because all the parameters to `update_conf()` are `None`).

I wouldn't be terribly against that to be honest, but I think the message it prints now is nicer.

At some point, I think the answer to "it hurts when I do this" is "don't do it". Randomly deleting installation files might fall there I think. :)

tests/west/project/test_project.py
```diff
@@ -93,17 +82,28 @@ def clean_west_topdir(tmpdir):
...
+      url: file://{}
+
+  projects:
+    - name: net-tools
+    - name: Kconfiglib
+      revision: zephyr
+      path: sub/Kconfiglib
+'''.format(remote_repos_dir))
+
+    # Switch to the top-level West installation directory
```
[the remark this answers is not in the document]
ulfalizer: Do you think they would ever overlap in practice? This code shorter, simpler to understand, and not any less robust, and that's what counts.

Too much namespacing and directories is a bad thing I think. I think people get carried away with it sometimes, in some quest to look like a serious developer ("bro r u even namespacing properly?"). :)

If there's negligible risk of name collisions, don't namespace. As a bonus, you get tests for whether the code gets confused by non-projects being in there. ;)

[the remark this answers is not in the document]
ulfalizer: Heck, in this case the risk is precisely zero, because we control what projects get put in there for the tests.

src/west/_bootstrap/main.py
ulfalizer: One improvement to the bootstrapper might be to have it do the two-step `git init` + `git fetch` thing that the project commands do though, for the manifest and west repositories. That gives you more flexibility in how you resume failed operations and the like.

Think that falls outside the scope of this PR though, and I'm not sure exactly how we'd set it up.

tests/west/project/test_project.py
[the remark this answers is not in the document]
mbolivar: > Heck, in this case the risk is precisely zero, because we control what projects get put in there for the tests.

... and we have to remember that the tests behave differently than a fresh environment would.

I don't agree that it's simpler to understand. I could see a reader thinking that we were manually placing repositories as if they had been cloned by west.

tests/west/project/test_project.py
```diff
@@ -263,3 +279,73 @@ def test_forall(clean_west_topdir):
...
+
+    # Fetch the net-tools repository
+    cmd('fetch --no-update net-tools')
+
+    # Add commits to the local repos
+    for repo_path in 'west/manifest', 'west/west', NET_TOOLS_PATH:
+        add_commit(repo_path, 'local')
+
+    # Check that resetting the manifest repository removes the local commit
+    cmd('update --reset-manifest')
```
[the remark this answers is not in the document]
mbolivar: I still think we need tests for the feature.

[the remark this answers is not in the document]
ulfalizer: There are tests for the backend commands used by the feature. There are no tests for the bootstrapper though, as far as I can see.

Maybe you could test it while you add some tests for the bootstrapper?

tests/west/project/test_project.py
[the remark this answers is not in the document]
ulfalizer: They behave the same.

Here's the old layout:

```
repositories/kconfiglib/
repositories/net-tools/
repositories/manifest.yml
zephyrproject/west/.west_topdir
zephyrproject/west/west/
zephyrproject/west/manifest/
```

Here's the new layout:

```
repos/kconfiglib
repos/net-tools
manifest.yml
west/.west_topdir
west/west/
west/manifest/
```

The only way they could act different would be if we accidentally added a test that reuses the `repos` name for something else, which is (1) very unlikely, and (2) trivial to fix.

Namespacing when you don't need it is bad. Don't namespace unless you have a huge project where you could realistically get naming collisions.

Heck, I could rename `repos` to `upstream-repos`. That way it'll be obvious even if you don't read the comment in `clean_west_topdir()`, and it drops the risk of collisions somewhere into the national lottery range:

```
upstream-repos/kconfiglib
upstream-repos/net-tools
manifest.yml
west/.west_topdir
west/west/
west/manifest/
```

One thing that confuses me for real is that West/Zephyr mixup by the way. `zephyrproject` made me think it was the Zephyr repository when I first saw it.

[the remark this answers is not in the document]
ulfalizer: There, it's now:

```
remote-repos/kconfiglib
remote-repos/net-tools
remote-repos/west
remote-repos/manifest
manifest.yml
west/.west_topdir
west/west/
west/manifest/
```

ulfalizer:
@tejlmand 
Think this might be related to https://github.com/zephyrproject-rtos/west/issues/98.

ulfalizer:
Added some tests that verify that the bootstrapper generates the expected backend commands when reinitializing (`test_bootstrap_reinit()`).

tejlmand:
> @tejlmand
> Think this might be related to #98.

Not exactly.
Cause looking at the stack trace, it is related to manifest file not obeying the schema, and not a missing manifest, as is the case #98 .

ulfalizer:
@tejlmand 
Both are related to trying to parse the manifest file early at least, and erroring out before commands get a chance to run.

IMO, we should be careful about adding too much "helpful guessing" like that. It can make it difficult to understand where information is coming from, and makes the code more complex too.

src/west/commands/project.py
```diff
@@ -817,12 +855,43 @@ def _update_special(name):
...
+def _update_and_reset_special(name):
+    # Updates one of the special repositories (the manifest and west) by
+    # resetting to the new revision after fetching it (with 'git reset --keep')
+
+    project = _special_project(name)
+    with _error_context(', while updating/resetting special project'):
+        _inf(project,
+             "Fetching and resetting (name-and-path) to '(revision)'")
+        _git(project, 'fetch -- (url) (revision)')
+        if _git(project, 'reset --keep FETCH_HEAD', check=False).returncode:
```
tejlmand: Combined with the way we a repo is cloned in _bootstrap: `git clone -b <rev> -- <url> <dest>`,
resetting this way, will result the repo to still be refer to the original checked out branch, however, now containing the FETCH_HEAD code.
This causes the the local branch to differ from upstream, see below.

I suggest that when bootstrap inits (clones) the remote repo, then we create and switch to local special branch, i.e. similar to what is done for the project, where `manifest-rev` is local only branch, and we go into detached head.

Example, seen from user:
```
$ west init --mr work_branch_1  <-- starting with work_branch_1
=== Updated configuration written to /projects/github/west_test/west/config ===                        
=== Running 'west update --reset-manifest --reset-projects' to update repositories ===                 
=== Fetching and resetting manifest (west/manifest/) to 'work_branch_1'                                
                                                                                                       
$ pushd west/manifest/                                                                                 
$ git status                                                                                           
On branch work_branch_1   <-- We are at local branch work_branch_1, which tracks origin/work_branch_1
Your branch is up-to-date with 'origin/work_branch_1'.                                                 
nothing to commit, working directory clean                                                             
$ popd                                                                                                 
                                                                                                       
$ west init --mr work_branch_2                                                                         
=== Updated configuration written to /projects/github/west_test/west/config ===                        
=== Running 'west update --reset-manifest --reset-projects' to update repositories ===                 
=== Fetching and resetting manifest (west/manifest/) to 'work_branch_2'                                
                                                                                                       
$ pushd west/manifest/                                                                                 
$ git status                                                                                           
On branch work_branch_1 <-- We are still at work_branch_1, which tracks origin/work_branch_1, but now with content of work_branch_2, which could be very different content.                                                                               
Your branch and 'origin/work_branch_1' have diverged,                                                  
and have 1 and 1 different commit each, respectively.                                                  
(use "git pull" to merge the remote branch into yours)                                                 
nothing to commit, working directory clean                                                             
$ popd                                                                                                 
```

ulfalizer: Think it's similar to the things I suggested in https://github.com/zephyrproject-rtos/west/pull/99#discussion_r233570898. Another limitation at the moment is that you can only pass a branch name to `--mr <rev>` when doing the initial clone, because it turns into `git clone -b <rev>`.

What do you mean by this part? Should it both create a `manifest-rev` and check out a detached HEAD?

```
I suggest that when bootstrap inits (clones) the remote repo, then we create and switch to local special branch, i.e. similar to what is done for the project, where manifest-rev is local only branch, and we go into detached head.
```

ulfalizer: FWIW, I'm not a big fan of the project resetting behavior either, because it will mess with whatever branches people have checked out in the same way. I'm just following specs though.

tejlmand:
> @tejlmand
> Both are related to trying to parse the manifest file early at least, and erroring out before commands get a chance to run.
> 
> IMO, we should be careful about adding too much "helpful guessing" like that. It can make it difficult to understand where information is coming from, and makes the code more complex too.

I'm unsure what you mean by guessing.
The point I'm trying to make here, is that a user should not end in an unrecoverable state (seen from an average user's view), where he need to remove and clone everything again, just because what he checked out from somewhere contains invalid/restricted settings.

User should be able to get back to the working manifest `west init --mr work_branch_1`
Currently user is stuck in failing state.

If we don't fix this, it can happen for real in the future, in case we later introduces more restrictions.

ulfalizer:
> I'm unsure what you mean by guessing.

I was thinking of the reason why the manifest is parsed early on. See `main.py`:

```python
# If the user doesn't specify it concretely, use the project
# with path 'zephyr' if that exists, or the ZEPHYR_BASE value
# in the calling environment.
#
# At some point, we need a more flexible way to set environment
# variables based on manifest contents, but this is good enough
# to get started with and to ask for wider testing.
manifest = Manifest.from_file()
for project in manifest.projects:
    if project.path == 'zephyr':
        zb = project.abspath
        zb_origin = 'manifest file {}'.format(manifest.path)
        break
```

src/west/commands/project.py
> What do you mean by this part? Should it both create a `manifest-rev` and check out a detached HEAD?
> 

I was simply referring to how it is done for the projects themselves, e.g. where we end in detached head.
So consistency, and as well to avoid placing changes on a tracking branchm where they do not belong (unless user does so on purpose).

Btw. I have to correct myself. The detached head doesn't happen on the `clone`, but later on the `pull` stage, see below.
Having same behavior for special repos (west/manifest) might make sense, unless there are special reasons not to do so.

Command example
```
$ west clone  <-- Only clones the repo and check out master branch, although it does create the manifest-rev branch
=== Creating repository for zephyr (zephyr/)                                                        
Initialized empty Git repository in /projects/github/west_test/zephyr/.git/                         
=== Fetching changes for zephyr (zephyr/)                                                           
.....                                                                                               
Checking out files: 100% (11967/11967), done.                                                       
Note: checking out 'refs/heads/manifest-rev'.                                                       
                                                                                                    
......                                                                                              
=== Creating branch 'master' in zephyr (zephyr/)                                                    
=== Checking out branch 'master' in zephyr (zephyr/)                                                
Switched to branch 'master'                                                                         
Your branch is up-to-date with 'manifest-rev'.                                                      
$ cd zephyr/                                                                                        
$ git branch                                                                                        
manifest-rev                                                                                        
* master                                                                                            
$ cd ..                                                                                             
$ west pull  <-- This gets us in detached head state                                                                                       
=== Fetching changes for zephyr (zephyr/)                                                           
HEAD is now at d0253d3... samples: reel_board/mesh_badge: Fix various coding style issues           
$ cd zephyr/                                                                                        
$ git branch                                                                                        
* (HEAD detached at refs/heads/manifest-rev) 
manifest-rev                                                                                        
master                     
```

tejlmand: > Think it's similar to the things I suggested in [#99 (comment)](https://github.com/zephyrproject-rtos/west/pull/99#discussion_r233570898). Another limitation at the moment is that you can only pass a branch name to `--mr <rev>` when doing the initial clone, because it turns into `git clone -b <rev>`.
> 
I didn't read your the comment a similar suggestion, but it's nice if were are aligned ;)

I agree that it is a limitation in the existing `west init` behavior, caused by the `git clone -b`

So maybe this should be put as an improvement on the bootstapper ?
but before doing that, we should ensure that we then obtain the desired behavior with this code.
As well as comments from the rest.

tejlmand: @ulfalizer Created: #110 as I don't think this issue should block the PR, as it  is more an edge case for time being, and in order for us to not forget.

tejlmand:
> > I'm unsure what you mean by guessing.
> 
> I was thinking of the reason why the manifest is parsed early on. See `main.py`:
> 
Got it.

Btw. I don't think this issue should block the PR from being merged, but we should take a look at this in the future.

carlescufi:
@mbolivar mind re-reviewing this one?

src/west/_bootstrap/main.py
```diff
@@ -196,44 +230,118 @@ def init_bootstrap(directory, args):
...
+        sys.exit('West already initialized. Please pass any settings you '
+                 'want to change.')
+
+    update_conf(config_path, west_url, args.west_rev, manifest_url,
+                args.manifest_rev)
+
+    print('=== Updated configuration written to {} ==='.format(config_path))
+
+    if args.reset:
+        cmd = ['update']
```
tejlmand: @ulfalizer I suggest to change this into:
```
cmd = ['--zephyr-base', 'none', 'update']
```
This will tell `west` when called for re-init, not to parse the manifest.
As this is a re-init, the Zephyr base is not used, and thus `none` is easy understandable.

It will also allow you to switch back, in case the manifest is not present in the revision, as described in #98.
That way you can always recover with a `west init`

You can add it to this PR, or I can make a PR with the suggestion, once this PR is merged.

Note: we could consider to reserve `none` for zephyr base

PR #100, opened 2018-11-07
commands: project: Allow projects to be specified via their path

ulfalizer:
Make commands that accept project names also accept project paths. For
example, 'west branch fix-foo .' can be run inside the zephyr/
directory, instead of running 'west branch fix-foo zephyr'.

Paths are mapped to projects by checking if the realpath (canonicalized
absolute) of any project is a prefix of the realpath of the supplied
path. That method will map subdirectories of projects to the project as
well.

Switch to using realpath() to calculate project.abspath as well, which
streamlines things and gets rid of gotchas related to symbolic links.

Project names and paths can be mixed in the same command. If both a
project name and a path matches, the name takes precedence (this can
always be worked around with './foo').

### DIFF
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -559,19 +559,48 @@ def _projects(args, listed_must_be_cloned=True, include_meta=False):
         # No projects specified. Return all projects.
         return projects
 
-    # Got a list of projects on the command line. First, check that they exist.
+    # Sort the projects by the length of their absolute paths, with the longest
+    # path first. That way, projects within projects (e.g., for submodules) are
+    # tried before their parent projects, when projects are specified via their
+    # path.
+    projects.sort(key=lambda project: len(project.abspath), reverse=True)
 
-    name_to_project = {project.name: project for project in projects}
-    project_names = name_to_project.keys()
-    nonexistent = set(args.projects) - set(project_names)
-    if nonexistent:
-        log.die('Unknown project{} {} (available projects: {})'
-                .format('s' if len(nonexistent) > 1 else '',
-                        ', '.join(nonexistent),
-                        ', '.join(project_names)))
+    # Listed but missing projects. Used for error reporting.
+    missing_projects = []
 
-    # Return the projects in the order they were listed
-    res = [name_to_project[name] for name in args.projects]
+    def normalize(path):
+        # Returns a case-normalized canonical absolute version of 'path', for
+        # comparisons. The normcase() is a no-op on platforms on case-sensitive
+        # filesystems.
+        return os.path.normcase(os.path.realpath(path))
+
+    res = []
+    for project_arg in args.projects:
+        for project in projects:
+            if project.name == project_arg:
+                # The argument is a project name
+                res.append(project)
+                break
+        else:
+            # The argument is not a project name. See if it is a project
+            # (sub)path.
+            for project in projects:
+                # The startswith() means we also detect subdirectories of
+                # project repositories. Giving a plain file in the repo will
+                # work here too, but that probably doesn't hurt.
+                if normalize(project_arg).startswith(normalize(project.abspath)):
+                    res.append(project)
+                    break
+            else:
+                # Neither a project name nor a project path. We will report an
+                # error below.
+                missing_projects.append(project_arg)
+
+    if missing_projects:
+        log.die('Unknown project name{0}/path{0} {1} (available projects: {2})'
+                .format('s' if len(missing_projects) > 1 else '',
+                        ', '.join(missing_projects),
+                        ', '.join(project.name for project in projects)))
 
     # Check that all listed repositories are cloned, if requested
     if listed_must_be_cloned:
@@ -588,11 +617,9 @@ def _projects(args, listed_must_be_cloned=True, include_meta=False):
         uncloned = [prj.name for prj in res
                     if not _cloned(prj) and prj.name not in META_NAMES]
         if uncloned:
-            log.die('Uncloned project{}: {}.'.
-                    format('s' if len(uncloned) > 1 else '',
-                           ", ".join(uncloned)),
-                    'Please run "west clone {}"'.format(' '.join(uncloned)),
-                    'to clone.')
+            log.die('The following projects are not cloned: {}. Please clone '
+                    "them first with 'west clone'."
+                    .format(", ".join(uncloned)))
 
     return res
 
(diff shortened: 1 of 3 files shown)


ulfalizer:
I want to document this too, but it would currently cause a merge conflict with https://github.com/zephyrproject-rtos/west/pull/96, so I'll wait a bit.

ulfalizer:
CC @mbolivar @tejlmand

src/west/commands/project.py
```diff
@@ -491,31 +491,45 @@ def _projects(args, listed_must_be_cloned=True):
...
                 res.append(project)
                 break
+        else:
+            # The argument is not a project name. See if it is a project
+            # (sub)path.
+            for project in projects:
+                # The startswith() means we also detect subdirectories of
+                # project repositories. Giving a plain file in the repo will
+                # work here too, but that probably doesn't hurt.
+                if os.path.realpath(project_arg).startswith(
```
mbolivar: The project with the "closest" match (i.e. longest common normalized subpath) should win in case there are multiple, rather than halting at the first potential match.

This is important for emulating git submodules with west. You can do that in repo by setting a project path to the correct submodule location and the right revision (it's in fact another use case for having the manifest revision be a plain SHA).

MCUboot is again a real world use case for this, as it has mbedtls as a submodule.

If I have a `project` and a `project/submodule`, and I run `west {whatever} .` from `submodule`, it seems like `project` would match here if it was listed first in the manifest.

src/west/commands/project.py
```diff
@@ -491,31 +491,45 @@ def _projects(args, listed_must_be_cloned=True):
...
+                    os.path.realpath(project.abspath)):
```
mbolivar: Hrm, why take realpath of project.abspath? The value in abspath should be as normalized as possible; if we need to eliminate symbolic links in project abspaths, that should be done at manifest parsing time (and in _special_project). The rest of the code should be able to trust that a Project they get handed has been canonicalized.

src/west/commands/project.py
```diff
@@ -491,31 +491,45 @@ def _projects(args, listed_must_be_cloned=True):
...
+                        ', '.join(missing_projects),
+                        ', '.join(project.name for project in projects)))
 
     # Check that all listed repositories are cloned, if requested
     if listed_must_be_cloned:
         uncloned = [prj.name for prj in res if not _cloned(prj)]
         if uncloned:
             log.die('The following projects are not cloned: {}. Please clone '
-                    "them first (with 'west fetch')."
+                    "them first (with 'west clone' or 'west fetch')."
```
mbolivar: Since you're here, why not just recommend west clone only? It's been discussed a lot that its results match user expectations more since there are no detached heads at first.

carlescufi: I agree, I see `west fetch` as an advanced feature.

ulfalizer: Doesn't hurt to mention it I think. We should trust our users to be able to look up how `west fetch` works, and this is just giving them more information.

src/west/commands/project.py
ulfalizer: Hmm... yeah, I guess it could be normalized earlier.

src/west/commands/project.py
ulfalizer: Fixed by sorting the projects with the longest `abspath` first.

ulfalizer:
The projects-in-projects case has been fixed by sorting the projects with the longest abspath first (so that they get tried first).

src/west/commands/project.py
mbolivar: Forcing people to make choices takes mental effort, so I don't agree with this since we are trying to get out of their way, recommend sane choices, etc. Plain west fetch in this case would lead to results that we know most users do not want.

src/west/manifest.py
```diff
@@ -290,7 +290,7 @@ def __init__(self, name, remote, defaults, path=None, clone_depth=None,
         self.remote = remote
         self.url = remote.url + '/' + name
         self.path = path or name
-        self.abspath = os.path.normpath(os.path.join(util.west_topdir(), self.path))
+        self.abspath = os.path.realpath(os.path.join(util.west_topdir(), self.path))
```
mbolivar: Looking at this, I agree that having normpath alone is not correct: `os.path.normpath('west/src')` on Windows gives me `west\src`, which is not an absolute path, while `os.path.realpath('west/src')` gives me `C:\Users\mbolivar\west\src` (i.e. it's absolutized and backslash-converted).

Care to throw in a call to `normcase` on the result while you're here? Maybe this is a separate issue, but we should probably forbid projects whose names are equal up to case, as that won't work on Windows...

carlescufi: I won't work on Windows and also on macOS, depending on the installation options. That said, I do see this as a separate issue, maybe to be added to the error handling of special names in the manifest.

src/west/commands/project.py
carlescufi: How about we remove the `(with 'west clone' or 'west fetch')` part and let the user figure out how to clone by looking at the doc?

ulfalizer: Tried to be diplomatic and added a longer explanation instead, though I think it's overkill.

src/west/manifest.py
ulfalizer: Added a `normcase()`. Seems to be a no-op on platforms with case-sensitive names.

I'm not sure we should `normcase()` the `abspath` field itself. It might be nice to preserve the name as written there, if it's ever shown, and it could be normalized where it matters.

mbolivar: Normalizing it in this particular situation seems reasonable, since you're using startswith as a proxy for containment.

src/west/commands/project.py
```diff
@@ -588,11 +617,10 @@ def _projects(args, listed_must_be_cloned=True, include_meta=False):
...
                     if not _cloned(prj) and prj.name not in META_NAMES]
         if uncloned:
-            log.die('Uncloned project{}: {}.'.
-                    format('s' if len(uncloned) > 1 else '',
-                           ", ".join(uncloned)),
-                    'Please run "west clone {}"'.format(' '.join(uncloned)),
-                    'to clone.')
+            log.die('The following projects are not cloned: {}. Please clone '
+                    "them first with 'west clone' (or with 'west fetch', "
+                    "if you don't want an initial topic branch created)."
```
mbolivar: Adding additional text does not solve the problem, which is, again:

1. nobody we have ever asked wants a detached head and "west fetch" for an uncloned project would create one
2. making people decide whether they want a detached head or not when we know they don't is wasting their time

ulfalizer: k

Changed to get things moving.

carlescufi:
@mbolivar can you please take another look at this one?

PR #101, opened 2018-11-07
intel_s1000: scripts: Fix the runner script

rgundi:
This reverts commit bc7b1c595ed4090be6d80859a01a0bb372ae030c in
the zephyr main tree to fix the runner script and also does the
below modifications.
1) explicitly terminate processes.
2) implicitly call debugserver when debug is called.

Signed-off-by: Rajavardhan Gundi <rajavardhan.gundi@intel.com>

### DIFF
--- a/src/west/runners/intel_s1000.py
+++ b/src/west/runners/intel_s1000.py
@@ -3,11 +3,11 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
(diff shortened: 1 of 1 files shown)


mbolivar:
Thanks for the fix and sorry I wasn't more help.

If you want to add regression tests please create a file west/tests/west/runners/test_intel_s1000.py and add them there. That could help prevent further breakage if we have to refactor these interfaces.

Issue #108, opened 2018-11-15
Improve help text for west when not yet bootstrapped


tejlmand:
The `west` bootstrapper can do a better job in providing help text to new comers.

Today `west -h` gives the following:
```
$ west -h
Error: not a Zephyr directory (or any parent): /home/<user>
Use "west init" to install Zephyr here
```
which isn't so helpfull, especially not, when the following is actual possible:
```
$ west init -h
usage: west init [-h] [-b BASE_URL] [-m MANIFEST_URL] [--mr MANIFEST_REV]
                 [-w WEST_URL] [--wr WEST_REV]
                 [directory]

Bootstrap initialize a Zephyr installation

positional arguments:
  directory             Initializes in this directory, creating it if
                        necessary

optional arguments:
  -h, --help            show this help message and exit
  -b BASE_URL, --base-url BASE_URL
                        Base URL for both 'manifest' and 'zephyr'
                        repositories; cannot be given if either -m or -w are
  -m MANIFEST_URL, --manifest-url MANIFEST_URL
                        Zephyr manifest fetch URL, default
                        https://github.com/zephyrproject-rtos/manifest
  --mr MANIFEST_REV, --manifest-rev MANIFEST_REV
                        Manifest revision to fetch, default master
  -w WEST_URL, --west-url WEST_URL
                        West fetch URL, default
                        https://github.com/zephyrproject-rtos/west
  --wr WEST_REV, --west-rev WEST_REV
                        West revision to fetch, default master
```

in an uninitialized folder.

Issue #107, opened 2018-11-14
Improve `west --help` formatting


carlescufi:
`west --help` currently renders in a format that is relatively unpleasant to the eye:

```
$ west --help
usage: west [-h] [-z ZEPHYR_BASE] [-v] [-V]
            {build,flash,debug,debugserver,attach,list,clone,fetch,pull,rebase,branch,checkout,diff,status,update,forall}
            ...

The Zephyr RTOS meta-tool.

optional arguments:
  -h, --help            show this help message and exit
  -z ZEPHYR_BASE, --zephyr-base ZEPHYR_BASE
                        Path to the Zephyr base directory. If not given,
                        ZEPHYR_BASE must be defined in the environment, and
                        will be used instead.
  -v, --verbose         Display verbose output. May be given multiple times to
                        increase verbosity.
  -V, --version

commands:
  {build,flash,debug,debugserver,attach,list,clone,fetch,pull,rebase,branch,checkout,diff,status,update,forall}

Run "west <command> -h" for help on each command.
```

Let's compare this to `git`, which does a much better job:

```
$ git --help
usage: git [--version] [--help] [-C <path>] [-c <name>=<value>]
           [--exec-path[=<path>]] [--html-path] [--man-path] [--info-path]
           [-p | --paginate | --no-pager] [--no-replace-objects] [--bare]
           [--git-dir=<path>] [--work-tree=<path>] [--namespace=<name>]
           <command> [<args>]

These are common Git commands used in various situations:

start a working area (see also: git help tutorial)
   clone      Clone a repository into a new directory
   init       Create an empty Git repository or reinitialize an existing one

work on the current change (see also: git help everyday)
   add        Add file contents to the index
   mv         Move or rename a file, a directory, or a symlink
   reset      Reset current HEAD to the specified state
   rm         Remove files from the working tree and from the index

examine the history and state (see also: git help revisions)
   bisect     Use binary search to find the commit that introduced a bug
   grep       Print lines matching a pattern
   log        Show commit logs
   show       Show various types of objects
   status     Show the working tree status

grow, mark and tweak your common history
   branch     List, create, or delete branches
   checkout   Switch branches or restore working tree files
   commit     Record changes to the repository
   diff       Show changes between commits, commit and working tree, etc
   merge      Join two or more development histories together
   rebase     Reapply commits on top of another base tip
   tag        Create, list, delete or verify a tag object signed with GPG

collaborate (see also: git help workflows)
[...]

PR #109, opened 2018-11-16
Be more helpful to newcomers.

mbolivar:
The error output when the bootstrapper is called without arguments or with -h is lacking. Try to improve it.

Tweak the output when called without arguments when an installation is found as well, in the same spirit.

Fixes: #108

### DIFF
--- a/src/west/_bootstrap/main.py
+++ b/src/west/_bootstrap/main.py
@@ -123,7 +123,8 @@ def init(argv):
         formatter_class=argparse.RawDescriptionHelpFormatter,
         description=
 '''
-Initializes a Zephyr installation.
+Initializes a Zephyr installation. Use "west clone" afterwards to fetch the
+sources.
 
 In more detail, does the following:
 
@@ -182,7 +183,7 @@ to handle any resetting yourself.
              installation to update the manifest or west URL/revision''')
 
     init_parser.add_argument(
-        'directory', nargs='?',
+        'directory', nargs='?', default=None,
         help='''Directory to initialize West in. Missing directories will be
              created automatically. (default: current directory)''')
 
@@ -247,7 +248,8 @@ def bootstrap(args):
     with open(os.path.join(directory, WEST_DIR, WEST_MARKER), 'w') as f:
         hide_file(f.name)
 
-    print('=== West initialized ===')
+    print('=== West initialized. Now run "west clone" in {}. ==='.
+          format(directory))
 
 
 def reinit(config_path, args):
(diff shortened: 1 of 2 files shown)


src/west/_bootstrap/main.py
```diff
@@ -247,22 +250,41 @@ def append_to_pythonpath(directory):
...
         else:
-            sys.exit('Error: not a Zephyr directory (or any parent): {}\n'
-                     'Use "west init" to install Zephyr here'
-                     .format(os.getcwd()))
+            print('Error: "{}" is not a Zephyr installation directory.'.
+                  format(start), file=sys.stderr)
+            print('Things to try:', file=sys.stderr)
+            print(' - Run "west init" to set up an installation here.',
+                  file=sys.stderr)
+            print(' - Run "west init -h" for additional information.',
```
tejlmand: Much better.

One comment here though.
There are not many commands the user can actually do, currently only `west init`, so instead of informing the user that additional help is available, when writing `west init -h`, we could potentially just concatenate `west init -h` to him, as that is most like is what he would need, as he already wrote: `west -h`

Maybe:
```
$ west -h
To set up a Zephyr installation here, run "west init".

usage: west init [-h] [-b BASE_URL] [-m MANIFEST_URL] [--mr MANIFEST_REV]
                 [-w WEST_URL] [--wr WEST_REV]
                 [directory]

Bootstrap initialize a Zephyr installation

positional arguments:
  directory             Initializes in this directory, creating it if
                        necessary

optional arguments:
  -h, --help            show this help message and exit
  -b BASE_URL, --base-url BASE_URL
                        Base URL for both 'manifest' and 'zephyr'
                        repositories; cannot be given if either -m or -w are
  -m MANIFEST_URL, --manifest-url MANIFEST_URL
                        Zephyr manifest fetch URL, default
                        https://github.com/zephyrproject-rtos/manifest
  --mr MANIFEST_REV, --manifest-rev MANIFEST_REV
                        Manifest revision to fetch, default master
  -w WEST_URL, --west-url WEST_URL
                        West fetch URL, default
                        https://github.com/zephyrproject-rtos/west
  --wr WEST_REV, --west-rev WEST_REV
                        West revision to fetch, default master
```

carlescufi: +1

carlescufi:
Looks good, but I think that @tejlmand's suggestion also makes sense. I will let you decide on which one to implement.

carlescufi:
@mbolivar by the way, thanks for the PR, this improves the experience by a large amount. While you are at it, would it also make sense to replace `FATAL ERROR: west installation found (in /tmp/test/west), but no command given` with something like `Please specify a command: <list of commands`?

src/west/_bootstrap/main.py
tejlmand: > Otherwise it breaks the convention that, in order to get help for a particular option, you invoke it with west <option> -h.

In principle I agree with you on this. But I decided to make the comment anyway, as `west init -h` is fact the only help message available here, and thus could be nice to present it.

So i'm fine with both options, but wanted to raise it in order to get others opinion.

tejlmand:
> @mbolivar by the way, thanks for the PR, this improves the experience by a large amount. While you are at it, would it also make sense to replace `FATAL ERROR: west installation found (in /tmp/test/west), but no command given` with something like `Please specify a command: <list of commands`?

maybe simply do as `git` and other tools that always requires an argument.
Simply print the help text.

As example:
`git` == `git help`

src/west/_bootstrap/main.py
mbolivar: I thought about that, but I figured it would be confusing to users who -- reasonably -- assume that `pip install west` installs all of west. I know we have been careful to document that isn't true, but most people won't read that.

So I'm trying to balance helpfulness with not suggesting there's only one option here.

mbolivar:
> would it also make sense to replace `FATAL ERROR: west installation found (in /tmp/test/west), but no command given` with something like `Please specify a command: <list of commands`?

I think that would be redundant in this case. Printing the usage handles it:

```
$ west
FATAL ERROR: west installation found (in /home/mbolivar/src/zephyr-west/west), but no command given
usage: west [-h] [-z ZEPHYR_BASE] [-v] [-V]
            {build,flash,debug,debugserver,attach,list,clone,fetch,pull,rebase,branch,checkout,diff,status,update,forall}
            ...
```

This output will also improve with resolution of #107, unlike printing a list of commands here.

mbolivar:
> maybe simply do as `git` and other tools that always requires an argument.
> Simply print the help text.

I think that would be an improvement to just printing the usage as is done now. I've made this change:

```
$ west
FATAL ERROR: west installation found (in /home/mbolivar/src/zephyr-west/west), but no command given
usage: west [-h] [-z ZEPHYR_BASE] [-v] [-V]
            {build,flash,debug,debugserver,attach,list,clone,fetch,pull,rebase,branch,checkout,diff,status,update,forall}
            ...

The Zephyr RTOS meta-tool.

optional arguments:
  -h, --help            show this help message and exit
  -z ZEPHYR_BASE, --zephyr-base ZEPHYR_BASE
                        Path to the Zephyr base directory. If not given,
                        ZEPHYR_BASE must be defined in the environment, and
                        will be used instead.
  -v, --verbose         Display verbose output. May be given multiple times to
                        increase verbosity.
  -V, --version

commands:
  {build,flash,debug,debugserver,attach,list,clone,fetch,pull,rebase,branch,checkout,diff,status,update,forall}

Run "west <command> -h" for help on each command.
```

This seems more helpful than the previous behavior shown in https://github.com/zephyrproject-rtos/west/pull/109#issuecomment-440284235.

mbolivar:
@carlescufi @tejlmand it's unrelated to bootstrap helpfulness, but I noticed that the `--zephyr-base` help was out of date and updated it while we are here.

carlescufi:
@mbolivar can you rebase?

mbolivar:
> @mbolivar can you rebase?

yep, done

PR #114, opened 2018-11-20
Revert "cmake: west: Take the 'kernel' name out of the .bin/.hex/.elf file"

mbolivar:
cc @SebastianBoe 

This reverts commit c466a5617fe7d043bfdf061cecc87f90b04544e9, which
breaks 'west flash' on the latest zephyr mainline:

```
$ west -vvv flash
ZEPHYR_BASE=/home/mbolivar/src/zephyr-west/zephyr (origin: manifest file /home/mbolivar/src/zephyr-west/west/manifest/default.yml)
/home/mbolivar/src/zephyr-west/build is a zephyr build directory
Running CMake: ['/usr/bin/cmake', '--build', '/home/mbolivar/src/zephyr-west/build']
As command: /usr/bin/cmake --build /home/mbolivar/src/zephyr-west/build
ninja: no work to do.
Using runner: pyocd
  File "/home/mbolivar/src/zephyr-west/west/west/src/west/main.py", line 236, in <module>
    main()
  File "/home/mbolivar/src/zephyr-west/west/west/src/west/main.py", line 216, in main
    args.handler(args, unknown)
  File "/home/mbolivar/src/zephyr-west/west/west/src/west/main.py", line 68, in command_handler
    command.run(known_args, unknown_args)
  File "/home/mbolivar/src/zephyr-west/west/west/src/west/commands/__init__.py", line 47, in run
    self.do_run(args, unknown)
  File "/home/mbolivar/src/zephyr-west/west/west/src/west/commands/flash.py", line 26, in do_run
    'ZEPHYR_BOARD_FLASH_RUNNER')
  File "/home/mbolivar/src/zephyr-west/west/west/src/west/commands/run_common.py", line 215, in do_run_common
    cfg = cached_runner_config(build_dir, cache)
  File "/home/mbolivar/src/zephyr-west/west/west/src/west/commands/run_common.py", line 117, in cached_runner_config
    elf_file  = cache['ZEPHYR_RUNNER_CONFIG_ELF_FILE']
  File "/home/mbolivar/src/zephyr-west/west/west/src/west/cmake.py", line 208, in __getitem__
    return self._entries[name].value
KeyError: 'ZEPHYR_RUNNER_CONFIG_ELF_FILE'
```

@SebastianBoe mind taking a look?

### DIFF
--- a/scripts/west-completion.bash
+++ b/scripts/west-completion.bash
@@ -17,9 +17,9 @@ __comp_west()
 	--runner
 	--skip-rebuild
 	--board-dir
-	--elf-file
-	--hex-file
-	--bin-file
+	--kernel-elf
+	--kernel-hex
+	--kernel-bin
 	--gdb
 	--openocd
 	--openocd-search"
--- a/src/west/commands/run_common.py
+++ b/src/west/commands/run_common.py
@@ -75,17 +75,14 @@ def add_parser_common(parser_adder, command):
     #
     # This is how we detect if the user provided them or not when
     # overriding values from the cached configuration.
-
-    command_verb = "flash" if command == "flash" else "debug"
-
     group.add_argument('--board-dir',
                        help='Zephyr board directory')
-    group.add_argument('--elf-file',
-                       help='Path to elf file to {0}'.format(command_verb))
-    group.add_argument('--hex-file',
-                       help='Path to hex file to {0}'.format(command_verb))
-    group.add_argument('--bin-file',
-                       help='Path to binary file to {0}'.format(command_verb))
+    group.add_argument('--kernel-elf',
+                       help='Path to kernel binary in .elf format')
+    group.add_argument('--kernel-hex',
+                       help='Path to kernel binary in .hex format')
+    group.add_argument('--kernel-bin',
+                       help='Path to kernel binary in .bin format')
     group.add_argument('--gdb',
                        help='Path to GDB, if applicable')
     group.add_argument('--openocd',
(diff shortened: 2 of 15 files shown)


SebastianBoe:
https://github.com/zephyrproject-rtos/west/commit/c466a5617fe7d043bfdf061cecc87f90b04544e9 is indeed not compatible with Zephyr.

I assumed we would resolve the incompatibility on the next zephyr-west sync.

Could we do a Zephyr-west sync instead of reverting it?

mbolivar:
> Could we do a Zephyr-west sync instead of reverting it?

We decided to stop syncing and instead head towards removing the copy of west in zephyr entirely.

Do you think you could do this in a way that preserves compatibility for now? E.g. fall back on the old key if not present?

I can hold off on the revert for now if you think so.

SebastianBoe:
I could.

But please, what process should we have for making trivial changes to the API between Zephyr and west?

When should we add an adapter layer, and for how long should the adapter layer live?

SebastianBoe:
Version compatibility adapter:

https://github.com/zephyrproject-rtos/west/pull/116

mbolivar:
> But please, what process should we have for making trivial changes to the API between Zephyr and west?

How to handle dependent cross-repository questions is still open. We have a workflow in my company's internal CI but it will not work for upstream zephyr. I'm looking for guidance from @nashif on how to proceed with upstream CI.

I think for now we will have to go case by case and try to merge dependent pull requests at the same time, with interpersonal communication, until we have something better.
 
> When should we add an adapter layer, and for how long should the adapter layer live?

I don't understand what you mean by "adapter layer", sorry, could you please say more?

SebastianBoe:
The nice thing about having a copy in Zephyr is that it is easy to manage the version compatibility between Zephyr and west.

IMHO, we should not strive for compatibility with multiple versions. But instead just support the copy in Zephyr until it is removed. And then when it is removed we have a pip package with a specified version.

Then there is no need for any version compatiblity adapter layers (which add complexity and extra work).

mbolivar:
> IMHO, we should not strive for compatibility with multiple versions. But instead just support the copy in Zephyr until it is removed. And then when it is removed we have a pip package with a specified version.
> 
> Then there is no need for any version compatiblity adapter layers (which add complexity and extra work).

The goal is that zephyr mainline always works with west mainline.

SebastianBoe:
> The goal is that zephyr mainline always works with west mainline.

Technically, that is an impossible goal because you can't change two repo's simultaneously.

We could strive for within "a day or so".

SebastianBoe:
> I don't understand what you mean by "adapter layer", sorry, could you please say more?

Some code that takes as input an API version X, and adapts it to API version Y for multi-version compatibility reasons.

mbolivar:
> > The goal is that zephyr mainline always works with west mainline.
> 
> Technically, that is an impossible goal because you can't change two repo's simultaneously.
> We could strive for within "a day or so".

We should strive for "within a minute or so". I know there's an atomicity problem but it's the best we can do with GitHub.

This is an issue with any project that evolves alongside its dependencies in other repositories -- no different than if zephyr-net-tools changed in a way that breaks information in zephyr/doc related to the same. The best we can do without a layer around Git like Gerrit (which ick please no) is try to merge cross-repo changes as close to "at the same time" as we can.

> > I don't understand what you mean by "adapter layer", sorry, could you please say more?
> 
> Some code that takes as input an API version X, and adapts it to API version Y for multi-version compatibility reasons.

Yup, I get what you mean now since your comments crossed in the wire -- does it make sense that we're just keeping mainline working with mainline?

This is basically developing west in lockstep with zephyr. Mixing and matching west v1.20 with zephyr v1.21 shouldn't be expected to work.

SebastianBoe:
> Mixing and matching the west released with Zephyr v1.20 with the code base for zephyr v1.21 shouldn't be expected to work.

When you do "west init", and specify an old Zephyr version. How does the bootstrapper know which version of west to clone?

> Yup, I get what you mean now since your comments crossed in the wire with mine -- does it make sense that we're just keeping mainline working with mainline?

I am concerned this will lead to compatibility problems.

I wonder if it would just-work if west was located in the Zephyr repo. Then the bootstrapper would know where to find a version of west that is compatible with the Zephyr version it cloned.

SebastianBoe:
> Mixing and matching the west released with Zephyr v1.20 with the code base for zephyr v1.21 shouldn't be expected to work.

This is reasonable.

Issue #112, opened 2018-11-20
Run a binary onto a board, d'oh


pfalcon:
I'm using whatever west is being run with "make flash" in today's Zephyr master. So, when I run:

> PYTHONPATH=/home/pfalcon/projects-3rdparty/Embedded/Zephyr/zephyr/scripts/meta /usr/bin/python3 /home/pfalcon/projects-3rdparty/Embedded/Zephyr/zephyr/scripts/meta/west/main.py flash --skip-rebuild --help

I get (among other output):

> Flash and run a binary onto a board.

I don't know what "run a binary onto a board" means. That makes me concerned that I'd better flash and run it myself, I only need west to tell me how (likely will be subject of the other ticket).

mbolivar:
Thanks for noticing the typo. I've sent a fix.

mbolivar:
Hi @pfalcon, I hope you'll forgive the piggybacking, but I've tacked another patch onto

mbolivar:
Sorry for the close/reopen, I pressed the "close and comment" button by mistake.

PR #115, opened 2018-11-20
runners: jlink: debug log commander script contents

mbolivar:
@pfalcon does this fix your issue?

As explained in the log, it is certainly our intent that `west -v flash` prints enough information for users to stop using the wrapper if they desire.

### DIFF
--- a/src/west/commands/flash.py
+++ b/src/west/commands/flash.py
@@ -14,7 +14,7 @@ class Flash(WestCommand):
     def __init__(self):
         super(Flash, self).__init__(
             'flash',
-            'Flash and run a binary onto a board.\n\n' +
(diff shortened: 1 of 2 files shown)


mbolivar:
@pfalcon @carlescufi I've tacked on a patch to fix the typo pointed out in #112, since they seem to be related in the reporter's mind.

pfalcon:
@mbolivar, Yes, that should address it, thanks for quick fix!

mbolivar:
> @mbolivar, Yes, that should address it, thanks for quick fix!

Sure thing! 

Looks like topic-west was merged into the Zephyr tree; we'd love to hear any feedback if you have time.

PR #116, opened 2018-11-20
commands: Be compatible with more Zephyr versions

SebastianBoe:
Check for multiple versions of names to be compatible with more
versions of Zephyr.

Signed-off-by: Sebastian Bøe <sebastian.boe@nordicsemi.no>


mbolivar:
Thanks for the quick response. I've tested locally and it resolves the issue. I will take this instead of the revert.

Issue #119, opened 2018-11-21
West pull might end in detached head (unintended)


tejlmand:
As discussed on IRC, some versions of git doesn't show reference when doing:
`git --show-ref --verify HEAD`
instead an error is returned, cause the local clone to go into detached head state.

git 2.7.4 is confirmed to have this behavior.
git 2.19.1 is confirmed to work as expected

See IRC discussion:
```
[2018-11-20 16:00:17] <tejlmand> yesterday when looking into PR#99 I noticed the behavior of `west clone`, `west fetch` and `west pull`
[2018-11-20 16:01:27] <tejlmand> for `west fetch` the help states that you go into detached head, and for `west clone` it states: 
[2018-11-20 16:01:33] <tejlmand> https://www.irccloud.com/pastebin/oBMGr9f1/
[2018-11-20 16:02:25] <tejlmand> which is also the case when doing `west clone`
[2018-11-20 16:02:28] <tejlmand> https://www.irccloud.com/pastebin/r0yPQcOs/
[2018-11-20 16:03:20] <tejlmand> but if doing `west pull` it goes into detached head;
[2018-11-20 16:07:12] <tejlmand> https://www.irccloud.com/pastebin/kTHQzn95/
[2018-11-20 16:07:32] <tejlmand> which I don't think is our intention, as the help states:
[2018-11-20 16:08:17] <tejlmand> usage: west pull [-h] [-m MANIFEST] 0 [PROJECT [PROJECT ...]]
[2018-11-20 16:08:17] <tejlmand> Clone/fetch and rebase projects.
[2018-11-20 16:08:17] <tejlmand> Fetches upstream changes in each of the specified projects (default:
[2018-11-20 16:08:17] <tejlmand> all projects) and rebases the checked-out branch (or detached HEAD
[2018-11-20 16:08:17] <tejlmand> state) on top of 'manifest-rev', effectively bringing the branch up to
[2018-11-20 16:08:17] <tejlmand> date. Repositories that do not already exist are cloned.
[2018-11-20 16:09:29] <tejlmand> what do the rest of you thinks ?
[2018-11-20 17:26:27] <Ulfalize1> tejlmand: it's a side effect of 'west pull' being 'west fetch' + 'west rebase'
[2018-11-20 20:02:19] <tejlmand> yes, and i'm asking, is that what we want ?
[...]

PR #123, opened 2018-11-22
git: Updating use of show-ref to work with older git versions

tejlmand:
Fixes: #119

This commit changes the way show-ref is executed in order to determine
whether HEAD is pointing to a valid reference or if the repository is
in initialized stated.

As discussed on IRC, and described in #119, `git show-ref --quiet --verify HEAD` unexpectedly returns 1 on git version 2.7.4 and probably others.

This PR changes the command into:
`git show-ref --quiet --head <random-64bytes>`
as that returns 0 if `head` is valid and 1 if `head` is not a valid reference. 
Random 64 bytes is used to avoid the reference to match a valid tag or branch.

### DIFF
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -712,7 +712,7 @@ def _fetch(project):
     _git(project, fetch_cmd + ' --tags -- (url) (revision)')
     _git(project, 'update-ref (qual-manifest-rev-branch) FETCH_HEAD^{commit}')
 
-    if not _ref_ok(project, 'HEAD'):
+    if not _head_ok(project):
         # If nothing it checked out (which would usually only happen just after
         # we initialize the repository), check out 'manifest-rev' in a detached
         # HEAD state.
@@ -819,6 +819,20 @@ def _ref_ok(project, ref):
     return _git(project, 'show-ref --quiet --verify ' + ref, check=False) \
            .returncode == 0
 
+def _head_ok(project):
+    # Returns True if the reference 'HEAD' exists and is not a tag or remote
+    # ref (e.g. refs/remotes/origin/HEAD).
+    # Some versions of git will report 1, when doing
+    # 'git show-ref --verify HEAD' even if HEAD is valid, see #119.
+    # 'git show-ref --head <reference>' will always return 0 if HEAD or
+    # <reference> is valid.
+    # We are only interested in HEAD, thus we must avoid <reference> being
+    # valid. '/' can never point to valid reference, thus 'show-ref --head /'
+    # will return:
+    # - 0 if HEAD is present
+    # - 1 otherwise
+    return _git(project, 'show-ref --quiet --head /', check=False) \
+           .returncode == 0
 
 def _checkout(project, branch):
     _inf(project, "Checking out branch '{}' in (name-and-path)".format(branch))


src/west/commands/project.py
```diff
@@ -819,6 +819,16 @@ def _ref_ok(project, ref):
     return _git(project, 'show-ref --quiet --verify ' + ref, check=False) \
            .returncode == 0
 
+def _head_ok(project):
+    # Returns True if the reference 'HEAD' exists and is not a tag/branch.
```
ulfalizer: Maybe this could say "...and is not a tag or remote ref (e.g. refs/remotes/origin/HEAD)", just to be a bit clearer.

src/west/commands/project.py
```diff
@@ -819,6 +819,16 @@ def _ref_ok(project, ref):
...
+    # Some versions of git will report 1, when doing
+    # 'git show-ref --verify HEAD' even if HEAD is valid, see #119.
+    # 'git show-ref --head <reference>' will always return 0 if HEAD or
+    # <reference> is valid.
+    # We are only interested in HEAD, thus we must avoid <reference> being
+    # valid. Using a long random generated string should ensure this.
+    return _git(project, 'show-ref --quiet --head ' + str(os.urandom(64)), check=False) \
```
ulfalizer: Might get spaces and stuff from `os.urandom()`, which makes this a bit iffy. We could probably just hardcode something unlikely instead.

Wonder if `/` might be a guaranteed non-match.

ulfalizer: Maybe `git rev-parse` results on different Git versions could be mentioned too.

tejlmand: > Might get spaces and stuff from `os.urandom()`, which makes this a bit iffy. We could probably just hardcode something unlikely instead.
> 
> Wonder if `/` might be a guaranteed non-match. It's not allowed to appear in refs.

yes, but when converting into `str`, you get something like 0x20, 0x1B, ...
which definitely should fail to match anything a human would start naming a tag or branch.

Good point with the '/'.
Although not allowed, it could be that it is possible, so will do some testing on that.

src/west/commands/project.py
tejlmand: will do

src/west/commands/project.py
tejlmand: > Maybe `git rev-parse` results on different Git versions could be mentioned too.

you mean `git show-ref` results, right ?

src/west/commands/project.py
tejlmand: updated

src/west/commands/project.py
ulfalizer: > When converting into `str`, you get something like 0x20, 0x1B, ...
> which definitely should fail to match anything a human would start naming a tag or branch.

Oh, right, `os.urandom(64)` returns `bytes`, not strings.



> you mean `git show-ref` results, right ?

Was thinking `rev-parse` could be mentioned too, because that's what people usually use for this sort of stuff I think. Won't block for it though.

tejlmand: @ulfalizer  `/` turned out to be allowed as argument to `show-ref` but it is not allowed as end-character on branches and tags, thus will not be able to get a match when used alone with `show-ref`.

Much nicer and simpler solution.

tejlmand:
> Still feels slightly sketchy, since it could break if additional argument verification is added to show-ref in Git. Will do for now though.

agree, could break if `git` changes behavior, but in that case `something/` might also break, and only safe bet is `no-one-will-use-this-name` but I actually like your `/` proposal, as it is very short and unique :)

PR #125, opened 2018-11-22
Read west URL and revision from  manifest instead of config file

### LINKED ISSUE #106: Determine west-based release strategy
We mostly have a plan for tagging a west-based release as discussed in https://github.com/zephyrproject-rtos/west/issues/39#issuecomment-423629912.

The short story is:

1. Tag every project (zephyr gets v1.14.0, hypothetical-other-thing gets v1.14.0, etc.)
2. Set the revisions for each project in the manifest file to v1.14.0; tag the commit that does that v1.14.0 as well

Then you can do `west init --mr v1.14.0`.

I think the thing we did not realize is that this will not prevent users from continually getting updates to west itself! These may in turn be backwards incompatible with the v1.14.0 versions of things. So if the user ever runs 'west pull', say, that will update west itself, which is not safe.

This issue tracks resolving this; it's a blocker for Zephyr v1.14.

carlescufi:
~Early RFC to get feedback, only partially tested~

Full implementation with additional proposal by @tejlmand:
https://github.com/carlescufi/west/pull/1

Also coupled to this PR to the manifest:
https://github.com/zephyrproject-rtos/manifest/pull/6

Fixes #106 (we will be able to simply tag the manifest with this PR)

### DIFF
--- a/src/west/_bootstrap/main.py
+++ b/src/west/_bootstrap/main.py
@@ -9,8 +9,10 @@ import argparse
 import configparser
 import os
 import platform
+import pykwalify.core
 import subprocess
 import sys
+import yaml
 
 import west._bootstrap.version as version
 
@@ -48,6 +50,7 @@ MANIFEST_URL_DEFAULT = 'https://github.com/zephyrproject-rtos/manifest'
 # Default revision to check out of the manifest repository.
 MANIFEST_REV_DEFAULT = 'master'
 
+_SCHEMA_PATH = os.path.join(os.path.dirname(__file__), "west-schema.yml")
 
 #
 # Helpers shared between init and wrapper mode
@@ -71,6 +74,14 @@ def west_dir(start=None):
     '''
     return os.path.join(west_topdir(start), WEST_DIR)
 
+def manifest_dir(start=None):
+    '''
+    Returns the path to the manifest/ directory, searching ``start`` and its
+    parents.
+
+    Raises WestNotFound if no west directory is found.
+    '''
+    return os.path.join(west_topdir(start), MANIFEST)
 
 def west_topdir(start=None):
     '''
@@ -152,11 +163,6 @@ to handle any resetting yourself.
 '''.format(WEST_MARKER))
 
     init_parser.add_argument(
-        '-b', '--base-url',
-        help='''Base URL for both 'manifest' and 'zephyr' repositories. Cannot
-             be given if either -m or -w are.''')
-
-    init_parser.add_argument(
         '-m', '--manifest-url',
         help='Manifest repository URL (default: {})'
              .format(MANIFEST_URL_DEFAULT))
@@ -165,16 +171,6 @@ to handle any resetting yourself.
         '--mr', '--manifest-rev', dest='manifest_rev',
         help='Manifest revision to fetch (default: {})'
              .format(MANIFEST_REV_DEFAULT))
-
-    init_parser.add_argument(
-        '-w', '--west-url',
-        help='West repository URL (default: {})'
-             .format(WEST_URL_DEFAULT))
-
-    init_parser.add_argument(
-        '--wr', '--west-rev', dest='west_rev',
-        help='West revision to fetch (default: {})'
-             .format(WEST_REV_DEFAULT))
 
     init_parser.add_argument(
         '--nr', '--no-reset', dest='reset', action='store_false',
@@ -198,16 +194,10 @@ to handle any resetting yourself.
 def bootstrap(args):
     '''Bootstrap a new manifest + West installation.'''
 
-    if args.base_url:
-        if args.west_url or args.manifest_url:
-            sys.exit('fatal error: -b is incompatible with -m and -w')
-        west_url = args.base_url.rstrip('/') + '/west'
-        manifest_url = args.base_url.rstrip('/') + '/manifest'
-    else:
-        west_url = args.west_url or WEST_URL_DEFAULT
-        manifest_url = args.manifest_url or MANIFEST_URL_DEFAULT
+    west_url = WEST_URL_DEFAULT
+    manifest_url = args.manifest_url or MANIFEST_URL_DEFAULT
 
-    west_rev = args.west_rev or WEST_REV_DEFAULT
+    west_rev = WEST_REV_DEFAULT
     manifest_rev = args.manifest_rev or MANIFEST_REV_DEFAULT
 
     directory = args.directory or os.getcwd()
@@ -231,16 +221,38 @@ def bootstrap(args):
     # Clone the west source code and the manifest into west/. Git will create
     # the west/ directory if it does not exist.
 
-    clone('west repository', west_url, west_rev,
-          os.path.join(directory, WEST_DIR, WEST))
-
     clone('manifest repository', manifest_url, manifest_rev,
           os.path.join(directory, WEST_DIR, MANIFEST))
+
+    # Parse the manifest and look for a section named "west"
+    manifest_file = os.path.join(directory, WEST_DIR, MANIFEST, 'default.yml')
+    with open(manifest_file, 'r') as f:
+        data = yaml.safe_load(f.read())
+
+    if 'west' in data:
+        wdata = data['west']
+        try:
+            pykwalify.core.Core(
+                source_data=wdata,
+                schema_files=[_SCHEMA_PATH]
+            ).validate()
+        except pykwalify.errors.SchemaError as e:
+            sys.exit("Error: Failed to parse manifest file '{}': {}"
+                      .format(manifest_file, e))
+
+        if 'url' in wdata:
+            west_url = wdata['url']
+        if 'revision' in wdata:
+            west_rev = wdata['revision']
+
+    print("cloning {} at revision {}".format(west_url, west_rev))
+    clone('west repository', west_url, west_rev,
+          os.path.join(directory, WEST_DIR, WEST))
 
     # Create an initial configuration file
 
     config_path = os.path.join(directory, WEST_DIR, 'config')
-    update_conf(config_path, west_url, west_rev, manifest_url, manifest_rev)
+    update_conf(config_path, manifest_url, manifest_rev)
     print('=== Initial configuration written to {} ==='.format(config_path))
 
     # Create a dotfile to mark the installation. Hide it on Windows.
@@ -259,36 +271,24 @@ def reinit(config_path, args):
     This updates the west/config configuration file, and optionally resets the
     manifest, west, and project repositories to the new revision.
     '''
-    if args.base_url:
-        if args.west_url or args.manifest_url:
-            sys.exit('fatal error: -b is incompatible with -m and -w')
-        west_url = args.base_url.rstrip('/') + '/west'
-        manifest_url = args.base_url.rstrip('/') + '/manifest'
-    else:
-        west_url = args.west_url
-        manifest_url = args.manifest_url
+    manifest_url = args.manifest_url
 
-    if not (west_url or args.west_rev or manifest_url or args.manifest_rev):
+    if not (manifest_url or args.manifest_rev):
         sys.exit('West already initialized. Please pass any settings you '
                  'want to change.')
 
-    update_conf(config_path, west_url, args.west_rev, manifest_url,
-                args.manifest_rev)
+    update_conf(config_path, manifest_url, args.manifest_rev)
 
     print('=== Updated configuration written to {} ==='.format(config_path))
 
     if args.reset:
-        cmd = ['update']
-        if manifest_url or args.manifest_rev:
-            cmd += ['--reset-manifest', '--reset-projects']
-        if west_url or args.west_rev:
-            cmd.append('--reset-west')
+        cmd = ['update', '--reset-manifest', '--reset-projects', '--reset-west']
         print("=== Running 'west {}' to update repositories ==="
               .format(' '.join(cmd)))
         wrap(cmd)
 
 
-def update_conf(config_path, west_url, west_rev, manifest_url, manifest_rev):
+def update_conf(config_path, manifest_url, manifest_rev):
     '''
     Creates or updates the configuration file at 'config_path' with the
     specified values. Values that are None/empty are ignored.
@@ -298,8 +298,6 @@ def update_conf(config_path, west_url, west_rev, manifest_url, manifest_rev):
     # This is a no-op if the file doesn't exist, so no need to check
     config.read(config_path)
 
-    update_key(config, 'west', 'remote', west_url)
-    update_key(config, 'west', 'revision', west_rev)
     update_key(config, 'manifest', 'remote', manifest_url)
     update_key(config, 'manifest', 'revision', manifest_rev)
 
--- a/src/west/_bootstrap/west-schema.yml
+++ b/src/west/_bootstrap/west-schema.yml
@@ -1,0 +1,17 @@
+## A pykwalify schema for basic validation of the structure of a
+## west YAML file.  (Full validation would require additional work,
+## e.g. to validate that remote URLs obey the URL format specified in
+## rfc1738.)
+##
+
+# The top-level west yaml is a map. The only top-level element is
+# 'west'. All other elements are contained within it. This allows
+# us a bit of future-proofing.
+type: map
+mapping:
+  url:
+    required: false
+    type: str
+  revision:
+    required: false
+    type: str
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -145,8 +145,8 @@ class Clone(WestCommand):
 
     def do_run(self, args, user_args):
         if args.update:
-            _update_west()
-            _update_manifest()
+            _update_manifest(args)
+            _update_west(args)
 
         for project in _projects(args, listed_must_be_cloned=False):
             if args.branch:
@@ -185,8 +185,8 @@ class Fetch(WestCommand):
 
     def do_run(self, args, user_args):
         if args.update:
-            _update_west()
-            _update_manifest()
+            _update_manifest(args)
+            _update_west(args)
 
         for project in _projects(args, listed_must_be_cloned=False):
             _fetch(project)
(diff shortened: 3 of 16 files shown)


src/west/manifest.py
```diff
@@ -23,8 +23,13 @@
...
 from west import util, log
 
+# Todo: take from _bootstrap?
+# Default west repository URL.
+WEST_URL_DEFAULT = 'https://github.com/zephyrproject-rtos/west'
+# Default revision to check out of the west repository.
+WEST_REV_DEFAULT = 'master'
 
-META_NAMES = ['west', 'manifest']
+META_NAMES = ['manifest', 'west']
```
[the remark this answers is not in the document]
carlescufi: In case we go and do `for name in META_NAMES: clone()` at some point

src/west/manifest.py
```diff
@@ -97,15 +102,17 @@ def __init__(self, source_file=None, source_data=None):
...
             self._malformed('manifest contains no manifest element')
-        try:
-            pykwalify.core.Core(
-                source_data=self._data['manifest']
-                schema_files=[_SCHEMA_PATH]
-            ).validate()
-        except pykwalify.errors.SchemaError as e:
-            self._malformed(e)
+
+        for key in self._data:
```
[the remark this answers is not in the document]
carlescufi: I've addressed this now

src/west/manifest.py
```diff
@@ -124,25 +131,31 @@ def __init__(self, source_file=None, source_data=None):
...
 
-    def _load(self, manifest):
+    def _load(self, data):
         # Initialize this instance's fields from values given in the
         # manifest data, which must be validated according to the schema.
 
         projects = []
         project_abspaths = set()
 
+        west = data.get('west', dict())
```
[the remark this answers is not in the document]
carlescufi: I've addressed this as well

carlescufi:
@tejlmand please take another look

tejlmand:
Hi.

Looking better :)
I took the liberty of changing a little bit.
Take a look:
https://github.com/tejlmand/west/commit/873ac11e5f9bd703e53015247bc73616dbdc26d6

Basically it treats `west` as a project, so that `_load` function can parse the `west` section just as any other project, meaning we don't need special WestMeta class.

And it means what you get when instantiating an object of Manifest class is same, just with either normal or special projects.

tests/west/project/test_project.py
```diff
@@ -360,16 +356,10 @@ def save_wrap_args(args):
...
-        (['--wr', 'foo'], ['update', '--reset-west']),
-        (['-m',   'foo', '-w', 'foo'],
-         ['update', '--reset-manifest', '--reset-projects', '--reset-west']),
-        (['-b',   'foo'],
-         ['update', '--reset-manifest', '--reset-projects', '--reset-west']),
-        (['-b',   'foo', '--no-reset'], [])):
+        (['-m',   'foo'], ['update', '--reset-manifest', '--reset-projects',
+                           '--reset-west']),
+        (['--mr', 'foo'], ['update', '--reset-manifest', '--reset-projects',
+                           '--reset-west'])):
```
carlescufi: This test needs fixing, I just removed the stuff that is not applicable anymore

carlescufi:
Fixed tests, I think the coverage should be equal to the one before this PR

src/west/_bootstrap/main.py
```diff
@@ -231,16 +221,35 @@ def bootstrap(args):
...
+
+    if 'west' in data:
+        wdata = data['west']
+        try:
+            pykwalify.core.Core(
+                source_data=wdata,
+                schema_files=[_SCHEMA_PATH]
+            ).validate()
+        except pykwalify.errors.SchemaError as e:
+            raise e
```
ulfalizer: Could remove the `try` block here, since it's catching and raising the same exception.

(Can put just `raise` to re-raise an exception by the way, but not much point if there's no code between catching the exception and re-raising it.)

src/west/_bootstrap/main.py
```diff
@@ -231,16 +221,35 @@ def bootstrap(args):
...
+
+        west_url = wdata['url']
+        west_rev = wdata['revision']
+
+    print("cloning {} at {}".format(west_url, west_rev))
```
ulfalizer: `"cloning {} at revision {}"` maybe.

src/west/manifest.py
```diff
@@ -124,25 +132,39 @@ def __init__(self, source_file=None, source_data=None):
...
                    else ' data:\n{}\n'.format(self._data))
         raise MalformedManifest('Malformed manifest{}(schema: {}):\n{}'
-                                .format(context, _SCHEMA_PATH, complaint))
+                                .format(context, _SCHEMA_PATH[section], complaint))
+
+    def _load(self, data, sections):
+
+        if 'west' in sections:
+            # Parse the west section first, if available
+            west = data.get('west', dict())
```
ulfalizer: Can create an empty dict with `{}`.

src/west/manifest.py
```diff
@@ -217,6 +239,26 @@ class MalformedManifest(Exception):
 
 # Definitions for Manifest attribute types.
 
+class WestMeta:
+
+    __slots__ = 'url revision'.split()
```
ulfalizer: `__slots__` is a bit overkill here by the way. It makes a difference in Kconfiglib, because parsing in Python is slow, but this code isn't very performance-sensitive. :)

src/west/manifest.py
```diff
@@ -217,6 +239,26 @@ class MalformedManifest(Exception):
...
+
+    def __init__(self, url=None, revision=None):
+        if url is None:
+            url = WEST_URL_DEFAULT
+        if revision is None:
+            revision = WEST_REV_DEFAULT
+
+        self.url = url
+        self.revision = revision
```
ulfalizer: Can do

```python
self.url = url or WEST_URL_DEFAULT
self.revision = revision or WEST_REV_DEFAULT
```

src/west/manifest.py
```diff
@@ -217,6 +239,26 @@ class MalformedManifest(Exception):
...
+
+    def __eq__(self, other):
+        raise NotImplemented
```
ulfalizer: Side note, but suspect there isn't any harm in allowing comparisons. It uses the identity by the default (`a == b` if `a` and `b` refer to the same object).

src/west/manifest.py
carlescufi: I copied this from your code :)

src/west/_bootstrap/main.py
carlescufi: Yep, will change.

src/west/manifest.py
carlescufi: Same, I took this from your code in `manifest.py`. Well, I assume it's your code, it might be @mbolivar's.

src/west/manifest.py
carlescufi: thanks, will do

src/west/manifest.py
carlescufi: I also took this from the other classes in `manifest.py`.

src/west/_bootstrap/main.py
carlescufi: Right. I will rework this to either do something useful or remove the `try` block.

tejlmand:
One thing I think should be discussed is `url` naming.

Although my work also uses `url` in `west` section in the updated manifest file, I beleive it is inconsistent use of `url` tag in manifest yaml, as we now have:
```
west:
  url: https://github.com/zephyrproject-rtos/west <-- This url contains full path
  revision: master


manifest:
.....
  remotes:
    - name: upstream
      url: https://github.com/zephyrproject-rtos  <-- This is only a base url which will be concatenated with project name below

  projects:
    - name: zephyr
```
I believe we should find either another name for url in `west` section, or re-name `url` to `url-base` in `manifest` section

carlescufi:
> I believe we should find either another name for url in `west` section, or re-name `url` to `url-base` in `manifest` section

Yes, we can rename `url` to `base-url` or `url-base` in the main manifest section.

carlescufi:
@ulfalizer addressed your comments, thanks for the review!

src/west/manifest.py
ulfalizer: Most of that code got moved from `project.py`. There weren't any classes when it was in there.

src/west/manifest.py
ulfalizer: Hmm... don't remember adding that at least. :)

src/west/_bootstrap/main.py
```diff
@@ -231,16 +221,37 @@ def bootstrap(args):
...
+    if 'west' in data:
+        wdata = data['west']
+        try:
+            pykwalify.core.Core(
+                source_data=wdata,
+                schema_files=[_SCHEMA_PATH]
+            ).validate()
+        except pykwalify.errors.SchemaError as e:
+            print('Error: Failed to parse manifest file: "{}."'.
+                  format(manifest_file), file=sys.stderr)
```
ulfalizer: If the exception from pykwalify has a useful message in it, then you could do this:

```python
sys.exit("Error: Failed to parse manifest file '{}': {}"
         .format(manifest_file, e))
```

`sys.exit()` in Python has a handy feature where it exits with 1 and prints a message if you pass it a string.

src/west/_bootstrap/main.py
```diff
@@ -259,36 +270,25 @@ def reinit(config_path, args):
...
-            sys.exit('fatal error: -b is incompatible with -m and -w')
-        west_url = args.base_url.rstrip('/') + '/west'
-        manifest_url = args.base_url.rstrip('/') + '/manifest'
-    else:
-        west_url = args.west_url
-        manifest_url = args.manifest_url
+    manifest_url = args.manifest_url
 
-    if not (west_url or args.west_rev or manifest_url or args.manifest_rev):
+    if not (manifest_url or args.manifest_rev):
```
ulfalizer: Could remove the `()`.

src/west/_bootstrap/main.py
```diff
@@ -259,36 +270,25 @@ def reinit(config_path, args):
...
 
     print('=== Updated configuration written to {} ==='.format(config_path))
 
     if args.reset:
         cmd = ['update']
-        if manifest_url or args.manifest_rev:
-            cmd += ['--reset-manifest', '--reset-projects']
-        if west_url or args.west_rev:
-            cmd.append('--reset-west')
+        cmd += ['--reset-manifest', '--reset-projects', '--reset-west']
```
ulfalizer: Could just have

```python
cmd = ['update', '--reset-manifest', '--reset-projects', '--reset-west']
```

since the command is fixed now.

src/west/_bootstrap/main.py
ulfalizer: Or not, reading closer. ;)

src/west/_bootstrap/main.py
carlescufi: That's much better, done

src/west/_bootstrap/main.py
carlescufi: Yep, good point. Fixed.

src/west/_bootstrap/main.py
carlescufi: No, I need `!(a || b)`

src/west/manifest.py
mbolivar: I wrote those. I thought about allowing equality in each of those particular cases and decided against it.

src/west/manifest.py
mbolivar: Yup, that's mine.

src/west/_bootstrap/main.py
```diff
@@ -231,16 +221,36 @@ def bootstrap(args):
...
+        try:
+            pykwalify.core.Core(
+                source_data=wdata,
+                schema_files=[_SCHEMA_PATH]
+            ).validate()
+        except pykwalify.errors.SchemaError as e:
+            sys.exit("Error: Failed to parse manifest file '{}': {}"
+                      .format(manifest_file, e))
+
+        west_url = wdata['url']
```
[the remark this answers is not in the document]
tejlmand: My opinion on this is that it is clearer to users of `west` if those are required fields, and it allows us to loosen in future, ie. making them optional, without breaking anything.

Making them optional now, and later deciding to make them required, is much harder.
But besides that, i'm fine with making the optional, if we believe that is ok.

src/west/_bootstrap/west-schema.yml
```diff
@@ -0,0 +1,17 @@
...
+## rfc1738.)
+##
+
+# The top-level west yaml is a map. The only top-level element is
+# 'west'. All other elements are contained within it. This allows
+# us a bit of future-proofing.
+type: map
+mapping:
+  url:
+    required: true
```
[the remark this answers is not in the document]
tejlmand: agree.
If decided to be optional

src/west/commands/project.py
```diff
@@ -1096,7 +1079,6 @@ def _expand_shorthands(project, s):
                      'refs/heads/' + _MANIFEST_REV_BRANCH) \
             .replace('(clone-depth)', str(project.clone_depth))
 
-
```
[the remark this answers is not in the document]
tejlmand: fixed.
Was removed by mistake.

src/west/manifest.py
```diff
@@ -144,22 +152,27 @@ def _malformed(self, complaint, section='manifest'):
...
+        # Initialize this instance's fields from values given in the
+        # manifest data, which must be validated according to the schema.
         if 'west' in sections:
-            # Parse the west section first, if available
             west = data.get('west', {})
-            westmeta = WestMeta(url=west.get('url'),
-                                revision=west.get('revision'))
-            self.westmeta = westmeta
+
+            url = url=west.get('url') or WEST_URL_DEFAULT
```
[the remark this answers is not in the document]
tejlmand: editor (or coder ?) misbehaving.
Fixed.

src/west/manifest.py
```diff
@@ -130,6 +130,14 @@ def __init__(self, source_file=None, source_data=None, sections=META_NAMES):
         Each element's values are fully initialized; there is no need
         to consult the defaults field to supply missing values.'''
 
+
+        self.west_project = None
+        '''Sequence of west.manifest.Project objects representing west
```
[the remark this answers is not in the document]
tejlmand: Fixed

src/west/manifest.py
```diff
@@ -316,13 +319,17 @@ def __init__(self, name, remote, defaults, path=None, clone_depth=None,
                      the project's ``name`` is used.
         :param revision: Project revision as given in the manifest, if present.
         '''
-        if remote is None:
-            raise ValueError('remote may not be None')
-        _wrn_if_not_remote(remote)
+        if remote is None and url is None:
```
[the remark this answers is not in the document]
tejlmand: Fixed, feel free to add new comments to proposal.

tests/west/manifest/test_manifest.py
```diff
@@ -127,6 +127,45 @@ def test_path():
...
+      remotes:
+        - name: testremote
+          url: https://example.com
+      projects:
+        - name: testproject
+          remote: testremote
+          path: sub/directory
+    '''
+    with patch('west.util.west_topdir', return_value=os.path.realpath('/west_top')):
+        # Parsing manifest only, no exception raised
```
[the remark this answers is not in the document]
tejlmand: fixed

src/west/manifest-schema.yml
```diff
@@ -13,124 +13,120 @@
 # 'manifest'. All other elements are contained within it. This allows
 # us a bit of future-proofing.
```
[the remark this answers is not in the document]
tejlmand: fixed.

I used the word `section`, as that is what we used during discussion, but unsure if yaml has a different wording.

src/west/manifest.py
```diff
@@ -43,7 +48,7 @@ class Manifest:
     from_file and from_data helper methods.'''
 
     @staticmethod
-    def from_file(source_file=None):
+    def from_file(source_file=None, sections=META_NAMES):
         '''Create and return a new Manifest object given a source YAML file.
 
         :param source_file: Path to a YAML file containing the manifest.
```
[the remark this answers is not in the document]
tejlmand: Fixed

src/west/manifest.py
```diff
@@ -97,15 +102,18 @@ def __init__(self, source_file=None, source_data=None):
         if not self._data:
             self._malformed('manifest contains no data')
 
-       if not 'manifest' in data:
+        if 'manifest' not in self._data:
```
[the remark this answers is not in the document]
tejlmand: Should be better now.

src/west/_bootstrap/main.py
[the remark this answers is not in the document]
carlescufi: I align with @mbolivar here after having thought about it. I already made this optional in `manifest.py` and I do think we should allow only setting one of the two options. Since the only way to do that is to make them both optional, my preference is to do so.

src/west/manifest-schema.yml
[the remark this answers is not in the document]
carlescufi: I don't think that yaml has a different wording after looking at it, so `section` it is.

src/west/manifest.py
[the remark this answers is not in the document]
carlescufi: I fully agree

src/west/manifest.py
```diff
@@ -97,9 +97,11 @@ def __init__(self, source_file=None, source_data=None):
         if not self._data:
             self._malformed('manifest contains no data')
 
+       if not 'manifest' in data:
```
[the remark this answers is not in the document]
tejlmand: should be fixed

src/west/manifest-schema.yml
```diff
@@ -13,124 +13,120 @@
...
-        required: false
-        type: map
-        mapping:
-          remote:
-            required: false
-            type: str
-          revision:
-            required: false
-            type: str
+ mapping:
```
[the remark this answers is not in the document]
tejlmand: looks ok now.

tejlmand:
> Yes, we can rename `url` to `base-url` or `url-base` in the main manifest section.

Updated to `url-base` and updated test in additional commit as agreed.

src/west/manifest-schema.yml
```diff
@@ -13,120 +13,120 @@
 # 'manifest'. All other elements are contained within it. This allows
 # us a bit of future-proofing.
 type: map
- mapping:
```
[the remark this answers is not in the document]
tejlmand: @mbolivar  commits squashed. Looking much better.

mbolivar:
Alright, here we go!

Issue #102, opened 2018-11-08
Projects in the west directory are not forbidden


mbolivar:
Originally reported by @tejlmand.

For example, even with forbidding projects named ~`west`~ `.west` or `manifest`, this manifest fragment is legal:

```
  projects:
    - name: not-west
      path: .west
```

This creates a structure where the top-level `west` directory contains a `.git` for the `not-west` repository, which is very strange.

Arguably this is user error but the current behavior seems undesirable. We should consider what to do here:

- warn
- error
- skip the project?
- ...

mbolivar:
> And in future we could consider to add `--force` to `west` to allow user to overrule whatever restriction `west` poses

This would require adding `--force` to basically every invocation of west that touches a project or parses a manifest.

Perhaps a `manifest` configuration option?

tejlmand:
@mbolivar you're right, it would be required for every command, expect maybe `west list` ;)

I think my main point was to suggest the error for now, and then take it up, if someone needs this.

mbolivar:
Erroring out seems sane.

mbolivar:
Migrating to enhancement

marc-hb:
Since this issue was filed, WEST_DIR has been renamed to hidden directory `.west` by commit 045b591e17b1bcab and has changed A LOT in nature.

BUT I tested this again with `path: .west/foo` and west version 1.3 still did not mind. So this issue is still valid.

marc-hb:
Fix submitted by @pdgendt in
- #754

PR #132, opened 2018-11-29
project: Restricting projects from being located in west folder

tejlmand:
Fixes: #102

This commit introduces a list of restricted folders that are compared
against project defined paths to ensure a manifest does not contain
paths ending in the west folder, e.g.:
- west
- west/foo
- foo/../west

while still allowing folders, such as:
- west-ext
as project folder.

### DIFF
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -33,6 +33,9 @@ META_NAMES = ['west', 'manifest']
 '''Names of the special "meta-projects", which are reserved and cannot
 be used to name a project in the manifest file.'''
 
+WEST_RESTRICTED_FOLDERS = ['west']
+'''Restricted folders that are used by "west" itself for "meta-projects"'''
+
 MANIFEST_SECTIONS = ['manifest', 'west']
 '''Sections in the manifest file'''
 
@@ -216,6 +219,18 @@ class Manifest:
                                 format(name) +
                                 'be used to name a manifest project')
 
+            path = mp.get('path')
+            if path is not None:
+                path = os.path.normpath(path)
+                for restricted_path in WEST_RESTRICTED_FOLDERS:
+                    # Appending os.path.sep to ensure that eg. restricted path 'west/' and
+                    # 'west-ext/' common prefix does not match on the restricted path 'west/'
+                    restricted_path += os.path.sep
+                    if os.path.commonprefix([restricted_path, path + os.path.sep]) is restricted_path:
+                        self._malformed('the folder "{}" is restricted and '.
+                                        format(restricted_path) +
+                                        'cannot be used for local projects')
+
             # Validate the project remote.
             remote_name = mp.get('remote', default_remote_name)
             if remote_name is None:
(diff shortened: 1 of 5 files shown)


src/west/manifest.py
```diff
@@ -216,6 +219,15 @@ def _load(self, data, sections):
                                 format(name) +
                                 'be used to name a manifest project')
 
+            path = mp.get('path')
+            if path is not None:
+                path = os.path.normpath(path)
+                for restricted_path in WEST_RESTRICTED_FOLDERS:
```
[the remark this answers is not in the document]
tejlmand: commonpath now changed to commonprefix to support 3.4, as commonpath is new in python 3.5

tests/west/manifest/test_manifest.py
```diff
@@ -127,6 +127,29 @@ def test_path():
     assert manifest.projects[0].path == 'sub/directory'
     assert manifest.projects[0].abspath == os.path.realpath('/west_top/sub/directory')
 
+def test_restricted_path():
+    # Projects are restricted from west/ folder, but folders containing west as
+    # part of path name, e.g. west-ext arte still allowed.
```
mbolivar: typo: s/arte/are/

src/west/manifest.py
```diff
@@ -216,6 +219,18 @@ def _load(self, data, sections):
                                 format(name) +
                                 'be used to name a manifest project')
 
+            path = mp.get('path')
+            if path is not None:
```
mbolivar: If not given, the implied path is the name of the project. What if I have a project named `west/subproject` or so? Do we still fail?

If so, would you mind adding some test coverage (another `invalid_xxx.yaml` should do it) to show that is true?

src/west/manifest.py
```diff
@@ -216,6 +219,18 @@ def _load(self, data, sections):
...
+                path = os.path.normpath(path)
+                for restricted_path in WEST_RESTRICTED_FOLDERS:
+                    # Appending os.path.sep to ensure that eg. restricted path 'west/' and
+                    # 'west-ext/' common prefix does not match on the restricted path 'west/'
+                    restricted_path += os.path.sep
+                    if os.path.commonprefix([restricted_path, path + os.path.sep]) is restricted_path:
```
mbolivar: Strings need to be compared with `==`, not `is`.

carlescufi:
Does this really apply now @tejlmand ?

tejlmand:
> Does this really apply now @tejlmand ?

Good question.
I assume we would like to restrict people from placing projects in the `.west` folder.
What do you think ?

Will push an update, if we still consider this relevant.

carlescufi:
Sure, we could restrict people from using `.west` as a name yes, I don't see why not. Better safe than sorry.

mbolivar:
Should we close this for now until we're ready to revisit? We still have the issue open, so we won't forget.

mbolivar:
> Should we close this for now until we're ready to revisit? We still have the issue open, so we won't forget.

@tejlmand ping?

mbolivar:
I'm going to close this for now so it doesn't keep showing up in the slack channel's pull reminders, as it's the only one that is currently doing that. Please reopen if you'd like.

PR #133, opened 2018-12-03
RFC: bootstrap: manifest: Embed the west schema into bootstrapper

tejlmand:
As discussed here: 
https://github.com/zephyrproject-rtos/zephyr/pull/11715#discussion_r237770078
https://github.com/zephyrproject-rtos/zephyr/pull/11715#discussion_r237435896

This RFC provides changes that will allow to fetch and use the `west` bootstrapper without using `pip`.
This means it can be fetched and run bu the use of, eg. `wget`

Note: We might want to consider if the name `main.py` should changes, but for now this RFC just tries to collect additional input and show example on the discussion referenced.


tejlmand:
To try it out:
```
cd <path-to-bootstrapper>
wget https://raw.githubusercontent.com/zephyrproject-rtos/west/3a8c0d348124b5f54a39765dd4ee86ce56025527/src/west/_bootstrap/main.py -O west.py
```

and then go to wherever you want to run the init.
```
cd <project-folder>/
python3 <path-to-bootstrapper>/west.py --init
```
or make the `west.py` executable and place it in your PATH.

tejlmand:
with the `WEST_DIR` option in https://github.com/zephyrproject-rtos/zephyr/pull/11715 this might not be as highly needed, however, we should still consider if we should support those users who for various reasons would like to be able to run `west init` and other `repo` handling commands by simply fetching a single file, as presented in this RFC.

carlescufi:
@tejlmand can we close this since we're anyway going with pip all the way in the near future?

tejlmand:
Closed for now.

PR #136, opened 2018-12-04
[RFC] projects: Add manifest command

carlescufi:
Add a `manifest` command that can display the current manifest or
"freeze" it to the current revision of all its projects.

Sample output:
```
$ west manifest -f -w
manifest:
  defaults:
    remote: upstream
    revision: master
  projects:
  - name: zephyr
    revision: 6a2853f9f449dde0da1ed4a3733842b8140cc806
  remotes:
  - name: upstream
    url: https://github.com/zephyrproject-rtos
west:
  revision: bf7148f43de189bfe97308bd53bd026cf8d5089f
```

### DIFF
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -16,14 +16,65 @@ from west import log
 from west import util
 from west.commands import WestCommand
 from west.manifest import default_path, Remote, Project, SpecialProject, \
-                          Manifest, MalformedManifest, META_NAMES
-
+                          MalformedManifest, META_NAMES
+from west.manifest import Manifest as WestManifest
 
 # Branch that points to the revision specified in the manifest (which might be
 # an SHA). Local branches created with 'west branch' are set to track this
 # branch.
 _MANIFEST_REV_BRANCH = 'manifest-rev'
 
+class Manifest(WestCommand):
+    def __init__(self):
+        super().__init__(
+            'manifest',
+            _wrap('''
+            Display or save the manifest file.
+
+            By default, displays the current manifest on stdout.
+            With additional options it can generate manifest variants based
+            on the state of the different projects and the options
+            provided.'''))
+
+    def do_add_parser(self, parser_adder):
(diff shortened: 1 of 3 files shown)


src/west/commands/project.py
```diff
@@ -16,14 +16,65 @@
...
+            if args.west:
+                log.die("-f required with -w")
+            with open(mp, 'r') as f:
+                print(f.read())
+                return
+
+        manifest = WestManifest.from_file(mp, ['west', 'manifest'])
+        manifest.projects = _projects(args, include_meta=False)
+        for project in manifest.projects:
+            project.revision = _sha(project, 'HEAD')
```
tejlmand: this will fail with an exception if user only did `west init` and then `west manifest -f`.
Or a re-init where an additional repo was added.

I think we can fail more gracefully with an error message.

src/west/commands/project.py
```diff
@@ -16,14 +16,65 @@
...
+
+        sections = ['manifest']
+        if args.west:
+            west = _special_project(args, 'west')
+            west.revision = _sha(west, 'HEAD')
+            manifest.west_project = west
+            sections.append('west')
```
tejlmand: maybe place the printing of `west` section before `manifest` section, to keep the same order as we generally use in manifest

carlescufi: I am fairly sure that `yaml.dump` sorts by alphabetical order, so this won't change anything.

tejlmand: ahh, so it ends back into same yaml issue

tejlmand:
As discussed on slack,
Let's consider maybe:
`-r` or `--sha` instead of `-f`
`-o` for dumping to file, especially useful on windows

For future, we could use `-f` / `--freeze <somewhere>` to do a real freeze, by creating a manifest file and commit it on a branch

carlescufi:
Closing this since it was mistakenly pushed to a branch in this repo.

PR #137, opened 2018-12-06
west: nrfjprog: Do verify after programming.

hakonfam:
Some times the --program operation fails. This patch
ensures that the nrfjprog command returns an error code
and error message if that happenes.

Without this patch, the user is not informed about the
failed program operation.

Signed-off-by: Håkon Øye Amundsen <haakon.amundsen@nordicsemi.no>

### DIFF
--- a/src/west/runners/nrfjprog.py
+++ b/src/west/runners/nrfjprog.py
@@ -89,7 +89,7 @@ class NrfJprogBinaryRunner(ZephyrBinaryRunner):
         else:
             board_snr = self.snr.lstrip("0")
         program_cmd = ['nrfjprog', '--program', self.hex_, '-f', self.family,
-                       '--snr', board_snr]
+                       '--snr', board_snr, '--verify']
 
         print('Flashing file: {}'.format(self.hex_))
         if self.erase:


SebastianBoe:
> I recommend we instead recommend to run west flash --verify to users

Don't we generally want recommended behaviour to also be default behaviour?

carlescufi:
> > I recommend we instead recommend to run west flash --verify to users
> 
> Don't we generally want recommended behaviour to also be default behaviour?

Sorry I meant let's recommend that **if** users see problems when flashing they use the `--verify` flag. I assume that this flag will slow down flashing?

hakonfam:
> > > I recommend we instead recommend to run west flash --verify to users
> > 
> > 
> > Don't we generally want recommended behaviour to also be default behaviour?
> 
> Sorry I meant let's recommend that **if** users see problems when flashing they use the `--verify` flag. I assume that this flag will slow down flashing?

IMO this does not suffice. As a user, I trust my flashing tool (maybe that is wrong?). I would review my code for many hours before I would come to the conclusion that it was the flashing (which seemed to work fine) which actually failed.

SebastianBoe:
Last I checked writing flash is about a thousand times slower than reading it, so unless nrfjprog is doing something silly I assume the performance impact will be less than 1%. But this would have to be tested ...

hakonfam:
> Last I checked writing flash is about a thousand times slower than reading it, so unless nrfjprog is doing something silly I assume the performance impact will be less than 1%. But this would have to be tested ...

```
ls -lha zephyr | grep zephyr.bin
15K zephyr.bin

time nrfjprog --program zephyr/zephyr.hex -r --sectorerase --verify
nrfjprog --program zephyr/zephyr.hex -r --sectorerase --verify  0.34s user 0.42s system 23% cpu 3.268 total

time nrfjprog --program zephyr/zephyr.hex -r --sectorerase
nrfjprog --program zephyr/zephyr.hex -r --sectorerase  0.30s user 0.46s system 23% cpu 3.173 total
```

SebastianBoe:
about 3% then. This is an acceptable price to pay IMHO. But this is of course subjective :)

hakonfam:
Results with a bigger hex file:
```
ls -lha zephyr | grep zephyr.bin
89K zephyr.bin

nrfjprog --program zephyr/zephyr.hex -r --sectorerase --verify  0.23s user 0.30s system 9% cpu 5.306 total

nrfjprog --program zephyr/zephyr.hex -r --sectorerase  0.19s user 0.33s system 13% cpu 3.808 total

SebastianBoe:
Ai, that's pretty far from optimal performance. That's a shame :(

tejlmand:
and from another view point.

With `west flash --verify`, the user has the option of decide to not user `--verify` by not passing that argument as opposed to `west flash` always using `--verify` internal.
With this PR, the user cannot decide to run without `--verify`.

If `--verify` should always be on, maybe that should be a request to `nrfjprog` instead.

oyvindronningstad:
How will the user invoke `ninja flash` with `--verify`? Or should `ninja flash` use `--verify`?

carlescufi:
>How will the user invoke ninja flash with --verify? Or should ninja flash use --verify?

@oyvindronningstad the user can invoke `west flash --verify`. It is not possible to add additional parameters during `ninja` invocation.

carlescufi:
>How will the user invoke ninja flash with --verify? Or should ninja flash use --verify?

@oyvindronningstad the user can invoke `west flash --verify`. It is not possible to add additional parameters during `ninja` invocation.

mbolivar:
This is a Nordic hardware runner so I'm fine with whatever you three collectively decide.

hakonfam:
> > How will the user invoke ninja flash with --verify? Or should ninja flash use --verify?
> 
> @oyvindronningstad the user can invoke `west flash --verify`. It is not possible to add additional parameters during `ninja` invocation.

What about `ninja safeflash` or some variant like that? Personally I would never use an unsafe flash operation if I knew that a safe one existed, even if it meant waiting some extra time. IMO there should be a direct ninja target so that the user does not have to perform two operations (ninja ... west flash)

mbolivar:
> What about `ninja safeflash` or some variant like that?

Handling all the possibilities to satisfy everybody's use case with build system targets does not scale well IMO. The possibilities explode combinatorially pretty quickly.

hakonfam:
> > What about `ninja safeflash` or some variant like that?
> 
> Handling all the possibilities to satisfy everybody's use case with build system targets does not scale well IMO. The possibilities explode combinatorially pretty quickly.

I would argue that a verified flash operation is a more common use case than a non-verified fast flash operation. All testing is useless when you don't know if the hex file you generated corresponds with what is located on the device.

SebastianBoe:
I agree with @mbolivar that we can't introduce a new target for this flag.

Either we wait for the --verify performance to become acceptable, change to another flasher that has an acceptable --verify performance (JLinkExe), or accept that there is a risk of corrupted with "ninja flash" and have users that want different behaviour use west.

mbolivar:
> Either we wait for the --verify performance to become acceptable, change to another flasher that has an acceptable --verify performance (JLinkExe)

I believe all your DKs with an embedded JLink do support the `jlink` runner, so you can just use `west flash -r jlink`.

PR #138, opened 2018-12-11
projects: Add manifest command

carlescufi:
Add a `manifest` command that can display the current manifest or
"freeze" it to the current revision of all its projects.

### DIFF
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -16,14 +16,65 @@ from west import log
 from west import util
 from west.commands import WestCommand
 from west.manifest import default_path, Remote, Project, SpecialProject, \
-                          Manifest, MalformedManifest, META_NAMES
-
+                          MalformedManifest, META_NAMES
+from west.manifest import Manifest as WestManifest
 
 # Branch that points to the revision specified in the manifest (which might be
 # an SHA). Local branches created with 'west branch' are set to track this
 # branch.
 _MANIFEST_REV_BRANCH = 'manifest-rev'
 
+class Manifest(WestCommand):
+    def __init__(self):
+        super().__init__(
+            'manifest',
+            _wrap('''
+            Display or save the manifest file.
+
+            By default, displays the current manifest on stdout.
+            With additional options it can generate manifest variants based
+            on the state of the different projects and the options
+            provided.'''))
+
+    def do_add_parser(self, parser_adder):
+        return _add_parser(
+            parser_adder, self,
+            _arg('-f', '--freeze', action='store_true',
+                 help='''Freeze the current state of the different projects
+                 by generating and printing a manifest that refers to the
+                 specific SHAs currently being pointed to by each project's
+                 HEAD.'''),
+            _arg('-w', '--west', action='store_true',
+                 help='''Include west itself when freezing a manifest. Requires
+                 \'-f\'.'''),
+                )
+
+    def do_run(self, args, user_args):
+        mp = _manifest_path(args)
+
+        # this command does not allow specifying a project subset
+        args.projects = None
+
+        if not args.freeze:
+            if args.west:
+                log.die("-f required with -w")
+            with open(mp, 'r') as f:
+                print(f.read())
+                return
+
+        manifest = WestManifest.from_file(mp, ['west', 'manifest'])
+        manifest.projects = _projects(args, include_meta=False)
+        for project in manifest.projects:
+            project.revision = _sha(project, 'HEAD')
+
+        sections = ['manifest']
+        if args.west:
+            west = _special_project(args, 'west')
+            west.revision = _sha(west, 'HEAD')
+            manifest.west_project = west
+            sections.append('west')
+
+        print(manifest.dump(sections))
 
 class List(WestCommand):
     def __init__(self):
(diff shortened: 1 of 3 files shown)


carlescufi:
Continuation of https://github.com/zephyrproject-rtos/west/pull/136

src/west/manifest.py
```diff
@@ -248,6 +248,45 @@ def _load(self, data, sections):
         self._remotes_dict = remotes_dict
         self.projects = tuple(projects)
 
+    def dump(self, sections):
+        data = {}
+        if 'west' in sections:
+            w = {'url' : self.west_project.url,
+                 'revision' : self.west_project.revision}
+            data['west'] = w
```
mbolivar: cosmetics: could just do

```
            data['west'] = {'url': self.west_project.url,
                            'revision': self.west_project.revision}
```

src/west/manifest.py
```diff
@@ -248,6 +248,45 @@ def _load(self, data, sections):
...
+            # Add the optional defaults section only if present in the
+            # original manifest. It is not overridable so just use the
+            # original data if required.
+            if 'defaults' in sm:
+                m['defaults'] = sm['defaults']
+
+            r = []
+            for remote in self.remotes:
+                r.append({'name' : remote.name, 'url' : remote.url})
+            m['remotes'] = r
```
mbolivar: cosmetics: alternatively:

```
            m['remotes'] = [{'name': r.name, 'url': r.url}
                            for r in self.remotes]
```

src/west/manifest.py
```diff
@@ -248,6 +248,45 @@ def _load(self, data, sections):
...
+
+            p = []
+            for project in self.projects:
+                lp = {}
+                sp = [i for i in sm['projects'] if i['name'] == project.name][0]
+                lp['name'] = project.name
```
mbolivar: Sorry, I'm confused about why all this is necessary. Can't we just properly format `project` as a dict and keep going?

Is the idea to try to "minify" the dumped representation?

src/west/commands/project.py
```diff
@@ -16,14 +16,65 @@
...
+
+        if not args.freeze:
+            if args.west:
+                log.die("-f required with -w")
+            with open(mp, 'r') as f:
+                print(f.read())
+                return
+
+        manifest = WestManifest.from_file(mp, ['west', 'manifest'])
+        manifest.projects = _projects(args, include_meta=False)
```
mbolivar: Why do we need to do this at all? The manifest already knows its own projects, and `_projects(args, include_meta=False)` here seems to just be returning all projects except west and the manifest -- which should already be what's in `manifest.projects`. If it's not, it smells like a bug.

src/west/commands/project.py
```diff
@@ -16,14 +16,65 @@
...
+            _arg('-w', '--west', action='store_true',
+                 help='''Include west itself when freezing a manifest. Requires
+                 \'-f\'.'''),
+                )
+
+    def do_run(self, args, user_args):
+        mp = _manifest_path(args)
+
+        # this command does not allow specifying a project subset
+        args.projects = None
```
mbolivar: Confused about why this is being done, see below.

src/west/commands/project.py
```diff
@@ -16,14 +16,65 @@
...
+
+        if not args.freeze:
+            if args.west:
+                log.die("-f required with -w")
+            with open(mp, 'r') as f:
```
mbolivar: I think it would be nice to load the file and dump it anyway even when not freezing, because this would fail if the manifest is malformed. So this command acts as a manifest validator regardless of whether we're freezing or not.

src/west/commands/project.py
```diff
@@ -16,14 +16,65 @@
...
+        manifest = WestManifest.from_file(mp, ['west', 'manifest'])
+        manifest.projects = _projects(args, include_meta=False)
+        for project in manifest.projects:
+            project.revision = _sha(project, 'HEAD')
+
+        sections = ['manifest']
+        if args.west:
+            west = _special_project(args, 'west')
+            west.revision = _sha(west, 'HEAD')
+            manifest.west_project = west
```
mbolivar: Same here... `manifest` should already know this; we shouldn't need to get it again from the project module.

src/west/commands/project.py
```diff
@@ -16,14 +16,65 @@
...
+            sections.append('west')
+
+        print(manifest.dump(sections))
```
mbolivar: Are you using print instead of log on purpose? Maybe a comment explaining why?

src/west/commands/project.py
```diff
@@ -16,14 +16,65 @@
...
+            Display or save the manifest file.
+
+            By default, displays the current manifest on stdout.
+            With additional options it can generate manifest variants based
+            on the state of the different projects and the options
+            provided.'''))
+
+    def do_add_parser(self, parser_adder):
+        return _add_parser(
+            parser_adder, self,
```
mbolivar: I think a `-o` or `--output-file` argument would be nice here to avoid forcing the user to use shell redirection etc. to save to a file, but I don't think that should block the merge.

carlescufi:
Closing this for now, it is not required after the substantial rearchitecture.

Issue #139, opened 2018-12-11
git: specifying revision to checkout / fetch fails


tejlmand:
Creating a manifest with a specific SHA revision may fail, with an error that it is not advertized by server:

Example snippet:
```
  projects:
    - name: zephyr
      revision: aa35e2df3693c9f797508314abd405940edf1c09
```

Error experienced with `west fetch` (after `west init -m <repo>`)
```
...
Receiving objects: 100% (243333/243333), 128.35 MiB | 9.72 MiB/s, done.
Resolving deltas: 100% (175926/175926), done.
error: Server does not allow request for unadvertised object aa35e2df3693c9f797508314abd405940edf1c09
FATAL ERROR: Command 'git fetch --tags -- https://github.com/zephyrproject-rtos/zephyr aa35e2df3693c9f797508314abd405940edf1c09' failed for zephyr (zephyr/)
```

but the reference is valid, as can be seen be checking out manually:
```
$ cd zephyr/
$ git remote add origin https://github.com/zephyrproject-rtos/zephyr
$ cd zephyr/
$ git fetch origin
remote: Enumerating objects: 8, done.
remote: Counting objects: 100% (8/8), done.
remote: Compressing objects: 100% (4/4), done.
...
$ git checkout aa35e2df3693c9f797508314abd405940edf1c09
Checking out files: 100% (9693/9693), done.
Note: checking out 'aa35e2df3693c9f797508314abd405940edf1c09'.
```

mbolivar:
Is the object in the history of some branch?

tejlmand:
seems that revision is gone (I could have been on a branch).
This revision is on master `bfad2a0cbfee37708e6641f4f740df292d4fd2b7` but it should happen to any valid revision.

marc-hb:
> error: Server does not allow request for unadvertised object aa35e2df3693c9f797

This is a security feature (and an FAQ). Allowing direct SHA1 access can:
- open access to objects deleted but not garbage collected yet
- circumvent refs based permissions like the ones Gerrit and maybe others use.

`allowAnySHA1InWant`

> seems that revision is gone (I could have been on a branch).

Or not: afraid it's too expensive to check.

PR #140, opened 2018-12-11
project: Fixed issue with fetching revision SHA

tejlmand:
Fixes: #139

Fixes the issue of fetching SHAs when specified in manifest file.
It is checked if a revision is a SHA, in which case the fetch is
performed on the url and afterwards the manifest-rev reference is
updated to the SHA.


mbolivar:
I'm good with merging this. At some point we should get the BW optimization back as discussed in https://github.com/zephyrproject-rtos/west/issues/139#issuecomment-447986471.

PR #143, opened 2018-12-18
Fix gdb ctrl-c regression

mbolivar:
This PR fixes a bug in the bootstrapper which caused 'west debug' to stop working properly.

We'll need to make a similar change to zephyr/scripts/west while we maintain that:

https://github.com/zephyrproject-rtos/zephyr/blob/master/scripts/west#L84

Because this is a bootstrapper change, we need to cut a new west release to make this available to users.

If this gets merged, I'll push a tag and update PyPI as well.

### DIFF
--- a/src/west/_bootstrap/main.py
+++ b/src/west/_bootstrap/main.py
@@ -406,18 +406,17 @@ def wrap(argv):
             print('West repository version: unknown; no tags were found')
         sys.exit(0)
 
-    # Replace the wrapper process with the "real" west
-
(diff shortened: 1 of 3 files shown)


mbolivar:
> We'll need to make a similar change to zephyr/scripts/west while we maintain that:
> 
> https://github.com/zephyrproject-rtos/zephyr/blob/master/scripts/west#L84

Sent a separate PR addressing that, https://github.com/zephyrproject-rtos/zephyr/pull/12143

PR #145, opened 2019-01-03
runners: jlink: Default to correct JLink executable

carlescufi:
On Microsoft Windows, the default name of the JLink executable is
"JLink.exe", while on *NIX it is instead "JLinkExe". Detect and
correctly set the default name based on the platform.


carlescufi:
Note that `JLinkGDBServer` is named identically on both Windows and *NIX, so no need to conditionally change the default name there.

PR #147, opened 2019-01-08
Extension commands prep

mbolivar:
I wrote these patches while working on west extension commands. They're not directly related to that feature, but I'd like to get them in first.

They're mostly just cleanups to files I'm going to be touching in the extension commands PR, but there is one patch with significant changes: I'd like to switch our testing to use tox, and I'm overhauling test_project.py.

Rationale and a link to the tox docs are in the next-to-last commit log:

```
To properly test the project commands, it would be best to have a
fresh west bootstrapper package created and installed on PATH, so it
could be used to run commands exactly as they'd happen if we package
and ship the working tree.

To make that easier, add a dependency on tox and use it for testing:

https://tox.readthedocs.io/en/latest/

From now on, we'll test west by running 'tox' from the repository
root.  This has several advantages over running pytest directly:

- "Just run tox": there are no longer any differences in test invocation
  between POSIX OSes and Windows.

- tox creates an sdist package of the current tree using our setup.py
  and installs it into a new virtual environment, then runs tests
  there.  This removes interference from other packages installed on
  the host (like released bootstrappers that are also installed)

- we get to run multiple shell commands in order, should that ever be needed,
  in our test procedures in a way that won't affect users

With that done, we can re-work the multirepo command testing to invoke
the bootstrapper in the virtual environment, adding various tests and
filling in longstanding testing gaps by adding increased checking of
the results (currently, much of the testing just checks whether
commands do or do not error out, which isn't enough).

These changes were made with a view towards the upcoming changes which
are planned before releasing west "into the wild": the test case code
should be mostly the same before and after the changes, so this serves
as a good baseline against regressions introduced by those upcoming
changes.
```

The main reason I want to get this in first is because I need to remove the `-m` argument to all the project commands to implement extension commands, and I don't want to make a change like that without a higher level of testing for `west.commands.project`.


mbolivar:
v2: switched shippable to use tox as well

mbolivar:
@tejlmand @carlescufi OK, ready for review -- I figured out the shippable issues and added a couple of patches to fix them

src/west/manifest.py
```diff
@@ -312,26 +324,33 @@ class Project:
...
+                         If not given, defaults.revision is used instead.
         '''
         if remote is None:
             raise ValueError('remote or url must be given')
         _wrn_if_not_remote(remote)
 
         self.name = name
         self.remote = remote
-        self.url = remote.url + '/' + name
+        self.url = remote.url_base + '/' + name
```
carlescufi: This was a no-brainer. Nice.

tests/west/project/test_project.py
```diff
@@ -1,168 +1,352 @@
...
-
-# Where the projects are cloned to
-NET_TOOLS_PATH = 'net-tools'
-KCONFIGLIB_PATH = 'sub/Kconfiglib'
+import west._bootstrap.main
 
 GIT = shutil.which('git')
+# Assumes this file is west/tests/west/project/test_project.py, returns
+# path to toplevel 'west'
+THIS_WEST = os.path.abspath(dirname(dirname(dirname(dirname(__file__)))))
```
carlescufi: Not the prettiest but I guess in a test it doesn't matter.

tests/west/project/test_project.py
```diff
@@ -1,168 +1,352 @@
...
+def repos_tmpdir(tmpdir):
+    '''Fixture for tmpdir with "remote" repositories, manifest, and west.
+
+    These can then be used to bootstrap an installation and run
+    project-related commands on it with predictable results.
+
+    Switches directory to, and returns, the top level tmpdir -- NOT
+    the subdirectory containing the repositories themselves.
+
+    Initializes placeholder upstream repositories in <tmpdir>/remote-repos/
```
carlescufi: I wish I had had a comment like this when I modified this file.

mbolivar:
> - we get to run multiple shell commands in order, should that ever be needed, in our test procedures in a way that won't affect users

I went ahead and added a patch which takes advantage of this to make sure flake8 is run on the codebase as well.

Doing this before would have avoided Zephyr's checkpatch complaining about trailing whitespace introduced in west when copying west over: https://github.com/zephyrproject-rtos/zephyr/pull/12380

tests/west/project/test_project.py
mbolivar: Yeah, it's pretty hideous.

mbolivar:
Thanks for the review @carlescufi -- I will wait to merge until @tejlmand has had a chance to look at it.

src/west/manifest.py
```diff
@@ -313,32 +324,38 @@ class Project:
...
-                     installation, if present in the manifest. If None,
+                     installation, if present in the manifest. If not given,
                      the project's ``name`` is used.
+        :param clone_depth: Nonnegative integer clone depth if present in
+                            the manifest.
         :param revision: Project revision as given in the manifest, if present.
+                         If not given, defaults.revision is used instead.
         '''
         if remote is None:
             raise ValueError('remote or url must be given')
```
tejlmand: Maybe now that your add it, revert this comment back to:
`raise ValueError('remote or url must be given')`
as `url` is not valid anymore. (leftover from before we made SpecialProject)

mbolivar: Good catch, thank you!

mbolivar:
Latest version tests OK on windows, so merging. Thanks for the review!

PR #150, opened 2019-01-11
Manifest in project repository

tejlmand:
This PR includes the following:

- Ability to have manifest file inside a project repository
- `self` tag introduced in manifest
- Renaming of `default.yml` to `manifest.yml`
- Folder `west` renamed to `.west`
- `west init` is now continued inside `west` repository after bootstrapper has cloned manifest repository

A branch of Zephyr with `manifest.yml` is available for testing:
https://github.com/tejlmand/zephyr/tree/manifest-branch

To test:
`west init -m https://github.com/tejlmand/zephyr --mr manifest-branch`

### DIFF
--- a/src/west/_bootstrap/main.py
+++ b/src/west/_bootstrap/main.py
@@ -13,6 +13,8 @@ import pykwalify.core
 import subprocess
 import sys
 import yaml
+import tempfile
+import shutil
 
 import west._bootstrap.version as version
 
@@ -96,11 +98,12 @@ def west_topdir(start=None):
         cur_dir = parent_dir
 
 
-def clone(desc, url, rev, dest):
-    if os.path.exists(dest):
+def clone(desc, url, rev, dest, exist_ok=False):
+    if not exist_ok and os.path.exists(dest):
         raise WestError('refusing to clone into existing location ' + dest)
 
-    print('=== Cloning {} from {}, rev. {} ==='.format(desc, url, rev))
+    print('=== Cloning {} from {}, rev. {} into {} ==='.format(desc, url, rev,
+                                                               dest))
     subprocess.check_call(('git', 'clone', '-b', rev, '--', url, dest))
 
 
@@ -126,12 +129,15 @@ sources.
 
 In more detail, does the following:
 
-  1. Clones the manifest repository to west/manifest, and the west repository
-     to west/west
+  1. Clones the manifest repository to a temporary folder, and the west
+     repository to .west/west
 
-  2. Creates a marker file west/{}
+  2. Creates a marker file .west/{}
 
-  3. Creates an initial configuration file west/config
+  3. Creates an initial configuration file .west/config
+
+  4. Continues the initialization process through west.main.main() with 'init'
+     as command and '--use-cache' pointing to temporary folder as parameter
 
 As an alternative to manually editing west/config, 'west init' can be rerun on
 an already initialized West instance to update configuration settings. Only
@@ -207,45 +213,59 @@ def bootstrap(args):
 
     # Clone the west source code and the manifest into west/. Git will create
     # the west/ directory if it does not exist.
+    tempdir = tempfile.mkdtemp(prefix='manifest_')
+    try:
+        clone('manifest repository', manifest_url, manifest_rev, tempdir,
+              exist_ok=True)
+        # Parse the manifest and look for a section named "west"
+        manifest_file = os.path.join(tempdir, 'west.yml')
+        with open(manifest_file, 'r') as f:
+            data = yaml.safe_load(f.read())
 
-    clone('manifest repository', manifest_url, manifest_rev,
-          os.path.join(directory, MANIFEST))
+        if 'west' in data:
+            wdata = data['west']
+            try:
+                pykwalify.core.Core(
+                    source_data=wdata,
+                    schema_files=[_SCHEMA_PATH]
+                ).validate()
+            except pykwalify.errors.SchemaError as e:
+                sys.exit("Error: Failed to parse manifest file '{}': {}"
+                         .format(manifest_file, e))
 
-    # Parse the manifest and look for a section named "west"
-    manifest_file = os.path.join(directory, MANIFEST, 'west.yml')
-    with open(manifest_file, 'r') as f:
-        data = yaml.safe_load(f.read())
+            if 'url' in wdata:
+                west_url = wdata['url']
+            if 'revision' in wdata:
+                west_rev = wdata['revision']
 
-    if 'west' in data:
-        wdata = data['west']
-        try:
-            pykwalify.core.Core(
-                source_data=wdata,
-                schema_files=[_SCHEMA_PATH]
-            ).validate()
-        except pykwalify.errors.SchemaError as e:
-            sys.exit("Error: Failed to parse manifest file '{}': {}"
-                     .format(manifest_file, e))
+        print("cloning {} at revision {}".format(west_url, west_rev))
+        clone('west repository', west_url, west_rev,
+              os.path.join(directory, WEST_DIR, WEST))
 
-        if 'url' in wdata:
-            west_url = wdata['url']
-        if 'revision' in wdata:
-            west_rev = wdata['revision']
+        # Create an initial configuration file
 
-    print("cloning {} at revision {}".format(west_url, west_rev))
-    clone('west repository', west_url, west_rev,
-          os.path.join(directory, WEST_DIR, WEST))
+        config_path = os.path.join(directory, WEST_DIR, 'config')
+        update_conf(config_path, manifest_url, manifest_rev)
+        print('=== Initial configuration written to {} ==='
+              .format(config_path))
 
-    # Create an initial configuration file
+        # Create a dotfile to mark the installation. Hide it on Windows.
 
-    config_path = os.path.join(directory, WEST_DIR, 'config')
-    update_conf(config_path, manifest_url, manifest_rev)
-    print('=== Initial configuration written to {} ==='.format(config_path))
+        with open(os.path.join(directory, WEST_DIR, WEST_MARKER), 'w') as f:
+            hide_file(f.name)
 
-    # Create a dotfile to mark the installation. Hide it on Windows.
-
-    with open(os.path.join(directory, WEST_DIR, WEST_MARKER), 'w') as f:
-        hide_file(f.name)
+        # Wrap west init with --cached:
+        # i.e. `west init <options> --use-cache `
+        # Note: main west will try to discover zephyr base and fail, as it is
+        # not fully initialized at this time.
+        # Thus a dummy zephyr_base is provided.
+        os.chdir(directory)
+        cmd = ['--zephyr-base', directory, 'init', '--manifest-url',
+               manifest_url, '--manifest-rev', manifest_rev, '--use-cache',
+               tempdir]
+        wrap(cmd)
+    finally:
+        shutil.rmtree(tempdir)
 
     print('=== West initialized. Now run "west clone" in {}. ==='.
           format(directory))
--- a/src/west/_bootstrap/main.py
+++ b/src/west/_bootstrap/main.py
@@ -37,13 +37,6 @@ WEST = 'west'
 WEST_URL_DEFAULT = 'https://github.com/zephyrproject-rtos/west'
 # Default revision to check out of the west repository.
 WEST_REV_DEFAULT = 'master'
-# File inside of WEST_DIR which marks it as the top level of the
-# Zephyr project installation.
-#
-# (The WEST_DIR name is not distinct enough to use when searching for
-# the top level; other directories named "west" may exist elsewhere,
-# e.g. zephyr/doc/west.)
-WEST_MARKER = '.west_topdir'
 
 # Manifest repository directory under WEST_DIR.
 MANIFEST = 'manifest'
@@ -87,7 +80,7 @@ def west_topdir(start=None):
     cur_dir = start or os.getcwd()
 
     while True:
-        if os.path.isfile(os.path.join(cur_dir, WEST_DIR, WEST_MARKER)):
+        if os.path.isdir(os.path.join(cur_dir, WEST_DIR)):
             return cur_dir
 
         parent_dir = os.path.dirname(cur_dir)
@@ -132,11 +125,9 @@ In more detail, does the following:
   1. Clones the manifest repository to a temporary folder, and the west
      repository to .west/west
 
-  2. Creates a marker file .west/{}
+  2. Creates an initial configuration file .west/config
 
-  3. Creates an initial configuration file .west/config
-
-  4. Continues the initialization process through west.main.main() with 'init'
+  3. Continues the initialization process through west.main.main() with 'init'
      as command and '--use-cache' pointing to temporary folder as parameter
 
 As an alternative to manually editing west/config, 'west init' can be rerun on
@@ -153,7 +144,7 @@ Updating the west URL or revision also runs 'west update --reset-west'.
 To suppress the reset of the manifest, west, and projects, pass --no-reset.
 With --no-reset, only the configuration file will be updated, and you will have
 to handle any resetting yourself.
-'''.format(WEST_MARKER))
+''')
 
     init_parser.add_argument(
         '-m', '--manifest-url',
@@ -248,11 +239,6 @@ def bootstrap(args):
         update_conf(config_path, manifest_url, manifest_rev)
         print('=== Initial configuration written to {} ==='
               .format(config_path))
-
-        # Create a dotfile to mark the installation. Hide it on Windows.
-
-        with open(os.path.join(directory, WEST_DIR, WEST_MARKER), 'w') as f:
-            hide_file(f.name)
 
         # Wrap west init with --cached:
         # i.e. `west init <options> --use-cache `
--- a/src/west/_bootstrap/main.py
+++ b/src/west/_bootstrap/main.py
@@ -46,7 +46,7 @@ WEST_MARKER = '.west_topdir'
 # Manifest repository directory under WEST_DIR.
 MANIFEST = 'manifest'
 # Default manifest repository URL.
-MANIFEST_URL_DEFAULT = 'https://github.com/zephyrproject-rtos/manifest'
+MANIFEST_URL_DEFAULT = 'https://github.com/zephyrproject-rtos/zephyr'
 # Default revision to check out of the manifest repository.
 MANIFEST_REV_DEFAULT = 'master'
 
@@ -73,16 +73,6 @@ def west_dir(start=None):
     Raises WestNotFound if no west directory is found.
     '''
     return os.path.join(west_topdir(start), WEST_DIR)
-
-
-def manifest_dir(start=None):
-    '''
-    Returns the path to the manifest/ directory, searching ``start`` and its
-    parents.
-
-    Raises WestNotFound if no west directory is found.
-    '''
-    return os.path.join(west_topdir(start), MANIFEST)
 
 
 def west_topdir(start=None):
@@ -149,8 +139,8 @@ explicitly passed configuration values (e.g. --mr MANIFEST_REVISION) are
 updated.
 
 Updating the manifest URL or revision via 'west init' automatically runs 'west
-update --reset-manifest --reset-projects' afterwards to reset the manifest to
-the new revision, and all projects to their new manifest revisions.
+update --reset-projects' afterwards to reset all projects to their new manifest
+revisions.
 
 Updating the west URL or revision also runs 'west update --reset-west'.
 
@@ -219,10 +209,10 @@ def bootstrap(args):
     # the west/ directory if it does not exist.
 
     clone('manifest repository', manifest_url, manifest_rev,
-          os.path.join(directory, WEST_DIR, MANIFEST))
+          os.path.join(directory, MANIFEST))
 
     # Parse the manifest and look for a section named "west"
-    manifest_file = os.path.join(directory, WEST_DIR, MANIFEST, 'west.yml')
+    manifest_file = os.path.join(directory, MANIFEST, 'west.yml')
     with open(manifest_file, 'r') as f:
         data = yaml.safe_load(f.read())
 
@@ -279,7 +269,7 @@ def reinit(config_path, args):
     print('=== Updated configuration written to {} ==='.format(config_path))
 
     if args.reset:
-        cmd = ['update', '--reset-manifest', '--reset-projects',
+        cmd = ['update', '--reset-projects',
                '--reset-west']
         print("=== Running 'west {}' to update repositories ==="
               .format(' '.join(cmd)))
--- a/src/west/_bootstrap/main.py
+++ b/src/west/_bootstrap/main.py
@@ -222,7 +222,7 @@ def bootstrap(args):
           os.path.join(directory, WEST_DIR, MANIFEST))
 
     # Parse the manifest and look for a section named "west"
-    manifest_file = os.path.join(directory, WEST_DIR, MANIFEST, 'default.yml')
+    manifest_file = os.path.join(directory, WEST_DIR, MANIFEST, 'west.yml')
     with open(manifest_file, 'r') as f:
         data = yaml.safe_load(f.read())
 
--- a/src/west/_bootstrap/main.py
+++ b/src/west/_bootstrap/main.py
@@ -28,7 +28,7 @@ if sys.version_info < (3,):
 #
 
 # Top-level west directory, containing west itself and the manifest.
-WEST_DIR = 'west'
+WEST_DIR = '.west'
 # Subdirectory to check out the west source repository into.
 WEST = 'west'
 # Default west repository URL.
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -145,7 +145,6 @@ class Clone(WestCommand):
 
     def do_run(self, args, user_args):
         if args.update:
-            _update_manifest(args)
             _update_west(args)
 
         for project in _projects(args, listed_must_be_cloned=False):
@@ -185,11 +184,11 @@ class Fetch(WestCommand):
 
     def do_run(self, args, user_args):
         if args.update:
-            _update_manifest(args)
             _update_west(args)
 
         for project in _projects(args, listed_must_be_cloned=False):
-            _fetch(project)
+            if project.name != 'manifest':
+                _fetch(project)
 
 
 class Pull(WestCommand):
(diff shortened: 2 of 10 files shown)


src/west/manifest-schema.yml
```diff
@@ -130,3 +130,31 @@ mapping:
...
             type: int
+
+  # The "self" key specifies values for the project containing the manifest
+  # file.
+  #
+  # When using West to initialize a project there must be a repository
+  # containing a manifest files with sub-projects.
+  # In case the Git repository containing the manifest file must be located in
+  # The path field can be used to specify a local path, in which the repository
+  # containing the manifest file must be placed.
```
mbolivar: This sentence is hard to understand; is there a typo here somewhere? Also, there is a "- " before the "path: project-path" in the below example. That would mean self's value was a list with one element, which is a map. But the schema says self's value is a map, not a list.

It seems like the overall meaning is something like this (with fixed example and suggestion to make path optional applied):

```
The optional path key specifies the local directory to clone this repository into,
in the same way projects can specify their paths. For example, the following would
cause the manifest repository to be checked out into local path "project-path":

self:
  path: project-path

If the path key is missing, the basename of the path component in the manifest repository
URL will be used by default. For example, if the URL is https://example.com/project-repo, the
manifest repository would be cloned to the directory "project-repo".
```

src/west/manifest.py
```diff
@@ -183,6 +189,18 @@ def _load(self, data, sections):
 
         manifest = data.get('manifest')
 
+        url = config.get('self', 'remote', fallback='origin')
```
mbolivar: This fallback value doesn't look right; are you sure?

src/west/manifest.py
```diff
@@ -41,7 +45,9 @@ def default_path():
     '''Return the path to the default manifest in the west directory.
 
     Raises WestNotFound if called from outside of a west working directory.'''
-    return os.path.join(util.west_dir(), 'manifest', 'manifest.yml')
+    return os.path.join(util.west_topdir(),
+                        config.get('self', 'path', fallback='manifest'),
```
mbolivar: This doesn't seem right. There is no default or fallback: this value simply must be present, or the config file has gotten corrupted and we've lost information needed to track the installation. We can't guess this value since there may be multiple projects with a manifest file inside.

src/west/manifest.py
```diff
@@ -183,6 +189,18 @@ def _load(self, data, sections):
 
         manifest = data.get('manifest')
 
+        url = config.get('self', 'remote', fallback='origin')
+        name = posixpath.basename(urlparse(url).path)
+        revision = config.get(name, 'revision', fallback='master')
+        path = os.path.join(name)
```
mbolivar: Why call os.path.join on only one value?

src/west/manifest-schema.yml
```diff
@@ -130,3 +130,31 @@ mapping:
...
+  # `west init`, example:
+  # `west init -m https://example.com/project-repo` will clone the remote
+  # repository into the local path `<project>/project-repo`
+  #
+  self:
+    required: false
+    type: map
+    mapping:
+      path:
+        required: true
```
mbolivar: I'm not sure making path required is a good idea. Right now it is the only key in the self map's value, sure, but later on we might want to add a second one. If we do that, users will still have to specify path, even if they want the default path.

src/west/manifest.py
```diff
@@ -183,6 +189,18 @@ def _load(self, data, sections):
...
+
+        self_tag = manifest.get('self')
+        if self_tag:
+            path = self_tag['path']
```
mbolivar: This would need a change if path were made optional, of course.

src/west/manifest.py
```diff
@@ -183,6 +189,18 @@ def _load(self, data, sections):
...
+
+        project = SpecialProject(name, revision=revision, path=path, url=url)
+        projects.append(project)
```
mbolivar: The convention that the manifest project is `projects[0]` is a good one. It should be added to the pydoc so users can rely on it as part of the API.

tests/west/manifest/test_manifest.py
```diff
@@ -9,11 +9,27 @@
...
 
+from west import config
 from west.manifest import Manifest, Defaults, Remote, Project, \
-    MalformedManifest
+    SpecialProject, MalformedManifest
 
 THIS_DIRECTORY = os.path.dirname(__file__)
 
+@pytest.fixture
+def config_file_project_setup(tmpdir):
```
mbolivar: This fixture doesn't provide the correct configuration file in all cases: in `test_self_tag`, the west.config.config object is missing a `path=mainproject`. You might want to parametrize the fixture or just use `patch` to substitute a test-specific mock for the `west.config.config` value.

tests/west/project/test_project.py
```diff
@@ -38,8 +40,7 @@ def repos_tmpdir(tmpdir):
     ├── west (branch: master)
     │   └── (contains this west's worktree contents)
     ├── manifest (branch: master)
-    │   └── manifest.yml
-    ├── Kconfiglib (branch: zephyr)
```
mbolivar: It doesn't look like Kconfiglib project should be deleted here. It's still a project which gets created and put into the manifest, and there is still a separate manifest repository. Were these lines changed in error?

tests/west/project/test_project.py
```diff
@@ -502,13 +508,7 @@ def test_update(west_init_tmpdir):
...
     # Add commits to the upstream special repos
     remotes = west_init_tmpdir.join('..', 'repos')
-    for r in remotes.join('manifest'), remotes.join('west'):
-        add_commit(str(r), 'test-update-upstream')
-
-    # Check that updating the manifest repository gets the upstream commit
-    cmd('update --update-manifest')
-    assert head_subject('.west/manifest') == 'test-update-upstream'
-    assert head_subject('.west/west') == west_prev  # Unaffected
+    add_commit(remotes.join('west'), 'test-update-upstream')
```
mbolivar: Why bother to keep adding the commit here if nothing further is being tested?

src/west/_bootstrap/main.py
```diff
@@ -207,12 +209,17 @@ def bootstrap(args):
 
     # Clone the west source code and the manifest into west/. Git will create
     # the west/ directory if it does not exist.
+    tempdir = tempfile.gettempdir()
 
-    clone('manifest repository', manifest_url, manifest_rev,
-          os.path.join(directory, MANIFEST))
+    i = 0
```
mbolivar: Looks like this can just be a call to tempfile.mkdtemp():

https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp

Is there a reason to have the number?

src/west/_bootstrap/main.py
```diff
@@ -207,12 +209,17 @@ def bootstrap(args):
...
+    while os.path.exists(os.path.join(tempdir, 'manifest_{}'.format(str(i)))):
+        i = i + 1
+    manifest_temp = os.path.join(tempdir, 'manifest_{}'.format(str(i)))
+
+    clone('manifest repository', manifest_url, manifest_rev, manifest_temp)
```
mbolivar: It might be worth deleting the temporary directory if the rest of the steps fail.

src/west/_bootstrap/main.py
```diff
@@ -247,6 +254,15 @@ def bootstrap(args):
     with open(os.path.join(directory, WEST_DIR, WEST_MARKER), 'w') as f:
         hide_file(f.name)
 
+    # Wrap west init with --cached:
```
mbolivar: I think the init command should remain the responsibility of the bootstrapper. This behavior is confusing to me and feels a bit brittle, in my opinion. Each bootstrapper should be in charge of its own init procedure. Not to mention making printing help a bit weird...

mbolivar:
Hi @tejlmand, just some review comments without having tested yet. Thanks for this! I think we are on the same page about the approach since we've been discussing this design for some time; I just have some comments on specific issues.

src/west/_bootstrap/main.py
```diff
@@ -222,7 +222,7 @@ def bootstrap(args):
           os.path.join(directory, WEST_DIR, MANIFEST))
 
     # Parse the manifest and look for a section named "west"
-    manifest_file = os.path.join(directory, WEST_DIR, MANIFEST, 'default.yml')
+    manifest_file = os.path.join(directory, WEST_DIR, MANIFEST, 'manifest.yml')
```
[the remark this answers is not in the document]
tejlmand: `west-manifest.yml` is also fine with me.
Just wanted something different that `default.yml`

src/west/util.py
```diff
@@ -36,12 +36,12 @@ def west_dir(start=None):
...
     Raises WestNotFound if no west top-level directory is found.
     '''
-    return os.path.join(west_topdir(start), 'west')
+    return os.path.join(west_topdir(start), '.west')
 
 
 def west_topdir(start=None):
     '''
-    Like west_dir(), but returns the path to the parent directory of the west/
+    Like west_dir(), but returns the path to the parent directory of the .west/
```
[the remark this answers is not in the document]
tejlmand: yes, but I wanted everything to work at first, and think this change is candidate for additional commit.

src/west/_bootstrap/main.py
```diff
@@ -175,10 +177,10 @@ def init(argv):
     try:
         reinit(os.path.join(west_dir(args.directory), 'config'), args)
     except WestNotFound:
-        bootstrap(args)
+        bootstrap(args, argv)
```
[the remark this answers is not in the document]
tejlmand: Sorry, left over from early work where the `argv` was simply passed on to init in `west.main`.
Removed.

src/west/_bootstrap/main.py
carlescufi: I am confused, didn't we discuss this recently and agreed that we would instead defer to the main installation after cloning the manifest repository in a temporary location?

src/west/_bootstrap/main.py
```diff
@@ -296,8 +286,8 @@ def update_conf(config_path, manifest_url, manifest_rev):
     # This is a no-op if the file doesn't exist, so no need to check
     config.read(config_path)
 
-    update_key(config, 'manifest', 'remote', manifest_url)
-    update_key(config, 'manifest', 'revision', manifest_rev)
+    update_key(config, 'self', 'remote', manifest_url)
```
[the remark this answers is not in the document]
carlescufi: I agree with @mbolivar here. I think `[manifest]` is clearer and it still describes well what the configuration is all about.

src/west/_bootstrap/main.py
carlescufi: Yep, not sure about this one either. The tempfile framework in Python ensures that there is no duplicates

src/west/_bootstrap/main.py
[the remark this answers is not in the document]
tejlmand: I see your point.

However, with "self" it is easier to related the section from the config file with the corresponding section in the manifest file.

Especially is we later add additional settings to the "self" section in the manifest file.
`path` needs to be in config file, but `west` can fetch remaining settings from the manifest file itself, and in such case, having same section name in both files feels natural.

@carlescufi any thoughts on this ?

src/west/manifest-schema.yml
```diff
@@ -130,3 +130,31 @@ mapping:
           clone-depth:
             required: false
             type: int
+
+  # The "self" key specifies values for the project containing the manifest
+  # file.
+  #
+  # When using West to initialize a project there must be a repository
+  # containing a manifest files with sub-projects.
```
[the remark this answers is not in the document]
tejlmand: done

src/west/_bootstrap/main.py
```diff
@@ -149,8 +141,8 @@ def init(argv):
 updated.
```
carlescufi: The whole documentation block above is outdated and needs to be adapted to the new reality

src/west/_bootstrap/main.py
tejlmand: Look at code. No folder is created here.
Reason was that creating a folder here would make `clone` fail, if the folder already exists.

But I considered your comment, and changed clone instead, so that it  takes additional argument. Think that looks nicer.

src/west/commands/project.py
```diff
@@ -15,15 +15,68 @@
...
+    def do_add_parser(self, parser_adder):
+        return _add_parser(
+            parser_adder, self,
+            _arg('--manifest-url',
+                 metavar='URL',
+                 help='Manifest repository URL'),
+            _arg('--manifest-rev',
+                 metavar='REVISION',
+                 help='Manifest revision to fetch'),
+            _arg('--use-cache',
```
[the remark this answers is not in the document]
tejlmand: By making a single arg with `nargs='*'` and then parse the list manually.
but the `init` command will still be visible.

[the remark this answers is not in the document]
tejlmand: The `init` can be hidden if we decide to not go through current `wrap` command, i.e. `west.main.main` but that will be a bigger change and break the design as well, so I dislike that.

tejlmand:
> I seem to recall there was a reason for this, but if there is not then I agree with having path as optional.

The `self` tag is optional, but if you include it, I think the `path` should be required, as it is the only setting available.

src/west/_bootstrap/main.py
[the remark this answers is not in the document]
tejlmand: second thought `west-manifest.yml` could imply describing `west` (which it can in optional `west` section), but it is describing projects.
And someone could decide to use something else / manuel parsing the file for fetching dependencies.

How about : `projects.yml` ?
@carlescufi any thought ?

tests/west/project/test_project.py
tejlmand: Fixed.

src/west/manifest-schema.yml
tejlmand: I also think we discussed this, and agreed it was not optional, but can't remember the details.
Although, I kind of like that `path` (and maybe even `self` )is mandatory, as in zephyr, we require a folder named `zephyr` containing `zephyr`.
And if user's fork `zephyr` to `zephyr-my-fork` and `self` is not present, then zephyr-base will fail.

src/west/manifest.py
tejlmand: yep. but that's a minor thing in that case.

tests/west/manifest/test_manifest.py
tejlmand: That doesn't matter in this test, as the `path` is taken from the manifest when creating a list of projects, and not the conf file.

The `path` in the conf file is used to locate the manifest file, although this is not used in the test, as the content of the manifest is fed directly `Manifest.from_data(yaml.safe_load(content))`

Reason for taking path form manifest file when creating the list of projects is that in future it will allow us to detect if `path` inside manifest changes, and thus `config_file.path != manifest_file.self_path`, as example after a `git reset` or local modifications.

tests/west/project/test_project.py
tejlmand: Deleted.

src/west/manifest.py
tejlmand: yep. This was actually something I wanted to raise when I was doing it.
(and then forgot about it again :( )
So thanks for raising this.

`fallback='manifest'` was to have similar behavior as before and it is was required to ensure the tests are passing in this commit (deferred `init` is added in following commit)
(Personally, I prefer passing tests in each commit)

So question is, do we want to default to `zephyr` which will be the default project with manifest in future, or abort with an error ?

src/west/manifest.py
tejlmand: `os.path.join` removed

src/west/manifest.py
tejlmand: No. and it points back to same problem as in: https://github.com/zephyrproject-rtos/west/pull/150/commits/792e7ea1fe4e527158d821470ddd1d712a4c5687#r247477779

What should we do if values are missing ?
In certain cases, we could read a remote from the git repo, if we know the path to it (also part of confiog file).
But then again, if there are multiple remotes defined (add by user), then which to use.

src/west/manifest-schema.yml
tejlmand: Thanks, much better.
Included it (although without `optional`until we've settled on that)

src/west/util.py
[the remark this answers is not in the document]
carlescufi: So it will go away in this PR?

src/west/_bootstrap/main.py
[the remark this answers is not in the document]
carlescufi: I do not like `west-manifest.yml` for the reasons outlined by @tejlmand and because it's too long, and I dislike `projects.yml` because it's as generic as `manifest.yml`, so we are not solving the collision issue. To be honest I would maybe call it `west.yml`. That avoids collisions and, if we ever decide to stop using the word `manifest` then the files will not need to be renamed.

src/west/_bootstrap/main.py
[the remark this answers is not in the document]
carlescufi: But `self` really means "manifest project", not "west" in the manifest itself, so I still think it makes sense to keep using `[manifest]` to indicate that this is where the manifest file lives, not where west lives.

src/west/_bootstrap/main.py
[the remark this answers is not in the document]
tejlmand: @carlescufi a little puzzled on your comment, maybe you can clarify.
First you wrote:
> I do not like west-manifest.yml for the reasons outlined by @tejlmand 

but I was addressing the `west` part in the name `west-manifest.yml` when I wrote:
> west-manifest.yml could imply describing west (which it can in optional west section), but the file is describing projects.
> And someone could decide to use something else / manuel parsing the file for fetching dependencies.

so how does the suggestion `west.yml` address those comments ?

src/west/util.py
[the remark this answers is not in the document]
tejlmand: yep.
It's removed in latest commit (the changes needed was smaller than what I had anticipated at first, nice surprise)

src/west/_bootstrap/main.py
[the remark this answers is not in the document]
tejlmand: Guess my main point is that:
`projects.yml` just lists all projects in use / required for the `self`-project to work.

Then user can use:
- `west` to fetch and parse `projects.yml` to stay up-to-date with latest Zephyr
- Read `projects.yml`manually, and use those to fork / copy /  whatever into his own system, to base his own development on Zephyr. But in this example, he may have no intention of keeping up-to-date with Zephyr development. Kind of what was asked by @aescolar on: How to use Zephyr without `west`
- make his own script / tool to fetch sub-projects

So `projects.yml` has no expectations on how / which tool you use, it just describes projects.
Anything having `west` in its name, starts to imply that `west` is required.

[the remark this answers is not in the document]
carlescufi: >@carlescufi a little puzzled on your comment, maybe you can clarify.

`west-manifest.yml` might imply "this is the manifest for west itself", but `west.yml` for me is like `CMakeLists.txt` or `Makefile`: a file that is processed by west.

The issue with `projects.yml` is that it doesn't solve the problem we set out to solve: collisions. If we are going to ignore the fact that a name can collide with existing files, then I say let's use `manifest.yml`, because we are using the term "manifest repository" extensively.

src/west/_bootstrap/main.py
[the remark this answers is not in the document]
tejlmand: ok. will change back.

src/west/_bootstrap/main.py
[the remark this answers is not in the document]
carlescufi: To summarize our brief conversation with @tejlmand:

- If we dismiss the possibility of ~collusion~ collision with additional files, then `projects.yml` might be a very descriptive name
- If we instead admit that there is a definite risk of ~collusion~ collision, then my personal prefence is `west.yml` because it is consistent with how similar tools operate and with the rest of the files that west creates or uses (`.west/`, `west`, `west.yml`)
- @tejlmand has proposed `west-projects.yml` but I personally think this is too long, especially if we allow nested manifests further down the line

src/west/_bootstrap/main.py
mbolivar: Dropping this concern based on discussion elsewhere. Never mind :)

src/west/_bootstrap/main.py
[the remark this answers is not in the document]
mbolivar: NO COLLUSION! WITCH HUNT! ;)

More seriously, I like your proposal of `west.yml`, @carlescufi.

src/west/_bootstrap/main.py
mbolivar: Very nice, thanks!

src/west/commands/project.py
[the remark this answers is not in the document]
carlescufi: I think @mbolivar has a suggestion with his upcoming PR to alleviate this issue.

src/west/_bootstrap/main.py
```diff
@@ -136,28 +121,28 @@ def init(argv):
 
 In more detail, does the following:
 
-  1. Clones the manifest repository to west/manifest, and the west repository
-     to west/west
+  1. Clones the manifest repository to a temporary folder, and the west
```
carlescufi: Can you change the text to reflect the order this is being done in?
`Clones the west repository to .west/west and the manifest repository to a temporary folder`.

src/west/_bootstrap/main.py
```diff
@@ -136,28 +121,28 @@ def init(argv):
...
+     repository to .west/west
 
-  2. Creates a marker file west/{}
+  2. Creates an initial configuration file .west/config
 
-  3. Creates an initial configuration file west/config
+  3. Continues the initialization process through west.main.main()
```
carlescufi: Would be better to say "invokes `west init` on the main west installation with an added `--use-cache` parameter"?

src/west/_bootstrap/main.py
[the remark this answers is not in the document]
tejlmand: @carlescufi technically you made the `west-projects.yml` in the chat but probably based upon  some of my other comments, and then I just had a little preference on that name compared to `west.yml`

But let's go with `west.yml`

src/west/_bootstrap/main.py
tejlmand: This is the order.
We clone the manifest repository first, in order to know which revision / from where `west` must be cloned.

src/west/_bootstrap/main.py
tejlmand: Well, `west` doesn't invoke `west init` directly, so this wording could lead to confusion in case the reader then expects bootstrapper to execute subprocess.
Bootstrapper actually imports and calls `west.main.main` with an argument list.

But will add some text on `--use-cache`

src/west/_bootstrap/main.py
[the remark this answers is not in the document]
carlescufi: @tejlmand yes sorry, I did propose it myself, you are right.

src/west/_bootstrap/main.py
carlescufi: Oops you are completely right, sorry.

src/west/_bootstrap/main.py
tejlmand: at later stage maybe.
Keeping the temporary folder in case something goes wrong may help the user / developer to investigate what happened. E.g invalid data in manifest, config file, other ?

Same reason the folder is also not removed when `init` completes.
After all it is placed in the system temp folder.

@carlescufi thoughts ?

src/west/_bootstrap/main.py
tejlmand: Take a look at updated text.

src/west/manifest.py
carlescufi: I suggest we bail out with a clear error message that instructs the user to start anew

src/west/manifest.py
tejlmand: As discussed elsewhere, solution for now is to remove the defaults and no specific error handling (meaning default python stack trace if key is missing)

src/west/manifest.py
tejlmand: As discussed elsewhere, solution for now is to remove the defaults and no specific error handling (meaning default python stack trace if key is missing)

tejlmand:
@carlescufi @mbolivar all comments should be addressed / answered.
Please recheck.

src/west/_bootstrap/main.py
carlescufi: I would delete it anyway because the repo is big enough to start causing trouble with people's temp folders getting big and the debugging of what went wrong can be done by looking directly at the original repo's code

src/west/manifest-schema.yml
carlescufi: Makes no sense to have `self` be optional and `path` required. I would go for `path` being optional at this stage, defaulting to the repo name

carlescufi:
I have still 2 items blocking my approval:

- `path` should be optional if `self` is optional
- Delete the manifest repo in the temporary folder

src/west/_bootstrap/main.py
tejlmand: temporary directory now removed when int completes (regardless of success or failure)

src/west/manifest-schema.yml
tejlmand: `path` is now optional, although if `self:` is specified, then `path` is also needed as there are no other keys specified for the map, schema error if one does:
```
self: 
```
without specifying anything.
So in practice `path` is required if specifying `self`

src/west/manifest.py
tejlmand: updated according to schema change

src/west/manifest-schema.yml
mbolivar: Having self without path is strange, for sure, but should it really be an error?

For analogy, a manifest with `defaults:` but nothing underneath it is still not an error.

src/west/manifest.py
mbolivar: Why not at least catch the error if the information is missing and call `log.die()` with a useful message?

src/west/manifest.py
mbolivar: As discussed below I think a call to `log.die()` somewhere is appropriate.

Though upon reflection I don't think anything in `manifest.py` should call this function. I think calling `log.die` should be done by the routines under `src/west/commands` rather than manifest.py, in case people want to use that module for other purposes.

src/west/manifest-schema.yml
tejlmand: This happens automatically by pykwalify.
I have nothing to do with it, just wanted to inform you ;)

` - Value 'None' is not a dict. Value path: '/self'.: Path: '/'>`

src/west/manifest.py
tejlmand: done

src/west/manifest.py
```diff
@@ -144,7 +147,10 @@ def __init__(self, source_file=None, source_data=None,
         projects.
 
         Each element's values are fully initialized; there is no need
-        to consult the defaults field to supply missing values.'''
+        to consult the defaults field to supply missing values.
+
+        Note: The index MANIFEST_PROJECT_INDEX in sequence will hold the
```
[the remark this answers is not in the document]
tejlmand: Guess that is a matter of preference.
When creating a constant, I personally believe it better that people refers to the constant (otherwise why create it), but if writing 0 in the docs, then people might not notice the constant exists and use the value directly.

PR #154, opened 2019-01-22
west update rework

tejlmand:
This PR contains several commits covering the following:
- `west update` is now `west selfupdate`
- Following west commands have been removed: `clone, fetch, pull, rebase, branch, checkout`
- New `update` command which will fetch and checkout in detached HEAD
- Removed support for reinit
- Removed url and revision from config file
- Ability to `west init` with an existing manifest repository


carlescufi:
Proposal for the help text on `west init`:

```
        help='''Directory to initialize West in, unless the -l option is provided, in which case this is the local path to an existing manifest repository. Missing directories will be created automatically. (default: current directory)''
```

```
        '-l', '--local', action='store_true',
        help='''Use an existing local manifest repository directly. The local repository will not be modified in any way. Cannot be
combined with --manifest-url and --manifest-rev''')
```

src/west/commands/project.py
```diff
@@ -162,234 +162,6 @@ def do_run(self, args, user_args):
...
-                # Use the last component of the revision, in case it is a
-                # qualified ref (refs/heads/foo and the like)
-                branch = project.revision.split('/')[-1]
-
-            _fetch(project)
-            _create_branch(project, branch)
-            _checkout(project, branch)
-
-
-class Fetch(WestCommand):
```
mbolivar: I'm confused. In slack we said we could keep west fetch and west rebase.

```
carlesc [12:37 PM]
@mbolivar that really depends. Would fetch and rebase take an argument? `west fetch {yourfork,upstream}` `west rebase {yourfork/branch,upstream/master}`? without a clear definition of remote/revision like we had before I am not sure if I know the answer to your question

mbolivar [12:37 PM]
no changes to the current signature
`west fetch [-h] [--no-update] [PROJECT [PROJECT ...]]`

carlesc [12:38 PM]
ok, and and where would it fetch from? I think I need to know that to answer

mbolivar [12:38 PM]
`west rebase [-h] [PROJECT [PROJECT ...]]`

carlesc [12:38 PM]
in order to see if I find this intuitive
and what would I rebase onto?

mbolivar [12:38 PM]
in both cases the answer is 'the remote revision'
using the discussion we had earlier for how that's defined for project versus manifest repositories

carlesc [12:40 PM]
can you refresh my memory for the manifest repo remote revision? I am juggling many things apologies :slightly_smiling_face:

mbolivar [12:41 PM]
of course
same here :slightly_smiling_face:
1. if the manifest repo is detached, the remote revision is HEAD
2. if the manifest repo branch is tracking a remote branch, that's the remote revision
3. if the manifest repo branch is not tracking a remote branch, the remote revision is HEAD
basically we just use git to try to find a remote revision and default to HEAD otherwise

carlesc [12:45 PM]
ok understood. Then my answer to your question is yes, we should do so, with the following caveat: either as part of those commands or, I believe even better, through a `projects` meta command or similar, we should be able to do the same but leaving the manifest repo untouched
included in my answer is also the fact that I think that is also on your mind: we should drop the manifest url/rev passed to west init

mbolivar [12:46 PM]
yes, i think that's implied
i don't like a projects meta command tbh
parsing is already complicated enough
we are stretching argparse to its limits as it is; if we go any farther we are getting into "fork it" territory

carlesc [12:47 PM]
ok, then `--projects` ?
or maybe `-p/--projects-only`?
is that reasonable?

mbolivar [12:48 PM]
sure, modulo name bikeshedding, i think an option that says "just operate on the non-manifest projects" makes sense (edited) 

carlesc [12:52 PM]
ok. Just want to point something out: unless you use  `git branch -u` to change your upstream, the master upstream will always point to wherever  you cloned from, so you will always fetch from there.  This means that depending on whether you originally cloned from upstream or your own github fork of it you will fetch from one or the other. Both are quite common.
(we can document that)

mbolivar [1:14 PM]
this is for users converting to west, right?
users who bootstrap with west won't be cloning from their personal forks

carlesc [1:15 PM]
For anyone using west init
Why not? If they fork and then clone?

mbolivar [1:15 PM]
you're saying if they west init my-fork-url

carlesc [1:16 PM]
Exactly

mbolivar [1:16 PM]
i understand now. sure, but then hopefully it's clear from context

```

src/west/commands/project.py
```diff
@@ -208,6 +193,49 @@ def do_run(self, args, user_args):
...
+
+    def do_add_parser(self, parser_adder):
+        return _add_parser(parser_adder, self, _no_update_arg,
+                           _project_list_arg)
+
+    def do_run(self, args, user_args):
+        if args.update:
+            _update_west()
+
+        for project in _projects(args, listed_must_be_cloned=False,
```
mbolivar: Shouldn't this be `listed_must_be_cloned=True`? The first thing we do is take the current branch of the project, so it must be cloned for this to work.

src/west/commands/project.py
```diff
@@ -208,6 +193,49 @@ def do_run(self, args, user_args):
...
+
+            _fetch(project)
+            _checkout_detach(project, _MANIFEST_REV_BRANCH)
+            sha = _sha(project, 'HEAD')
+
+            if branch is not None:
+                _inf(project,
+                     '{name} has been updated and is in detached HEAD.')
+                _inf(project,
+                     'To checkout and rebase \'{}\' with \'{}\', run:'
```
mbolivar: Just a nit since I think you're going to need to change this patch, but how about changing "with" to "onto"?

src/west/commands/project.py
```diff
@@ -379,13 +408,20 @@ def _projects(args, listed_must_be_cloned=True, include_west=False):
...
+    #   list.
 
     projects = _all_projects()
     west_project = _west_project()
 
     if include_west:
         projects.append(west_project)
 
+    if exclude_manifest:
+        projects.pop(0)
```
mbolivar: By the discussion in the earlier series, we technically should be using MANIFEST_PROJECT_INDEX.

Things like this are why I think it's worthwhile to just document that it's at index number zero. But I'm not asking for a change here.

src/west/commands/project.py
```diff
@@ -593,6 +629,20 @@ def _create_branch(project, branch):
              .format(branch))
 
 
+def _current_branch(project):
+    # Determine if project is currently on a branch
+    if not _cloned(project):
```
mbolivar: Per discussion above, I think this should error out in this case.

src/west/main.py
```diff
@@ -26,9 +26,8 @@
 from west.commands.build import Build
 from west.commands.flash import Flash
 from west.commands.debug import Debug, DebugServer, Attach
-from west.commands.project import List, Clone, Fetch, Pull, Rebase, Branch, \
```
mbolivar: I just noticed the previous patch breaks `git bisect` - the removed commands should have been removed from the import list there, not here. But we're in a hurry so let's not fix this if there's not time.

src/west/_bootstrap/main.py
```diff
@@ -170,12 +164,9 @@ def init(argv):
     args = init_parser.parse_args(args=argv)
 
     try:
-        reinit(os.path.join(west_dir(args.directory), 'config'), args)
-        bootstrapping = False
+        west_dir(args.directory)
+        sys.exit('West already initialized. Aborting.')
     except WestNotFound:
-        bootstrapping = True
```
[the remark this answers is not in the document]
mbolivar: I see now that I should have put a comment here explaining why the extra step was put in place. Sorry for the confusing code.

src/west/_bootstrap/main.py
```diff
@@ -167,7 +166,9 @@ def init(argv):
         west_dir(args.directory)
         sys.exit('West already initialized. Aborting.')
     except WestNotFound:
-        bootstrap(args)
+        pass
```
mbolivar: Ah, okay... I see you removed the exception cleanup in the previous patch only to replace it in this patch.

This series has a couple places where it's hard to read in patch order or the patches are not bisect friendly. I won't ask for changes to fix this, again, since we're in a hurry, but I think in general it's easier to review if the patches can be read in order (and of course not breaking bisect is a good thing).

src/west/commands/project.py
tejlmand: sorry, maybe I was too quick in removing.
I was just looking at the list @carlescufi  made earlier.

Will put those two commands back in, but restrict them from working on manifest repo

src/west/manifest.py
```diff
@@ -203,21 +201,14 @@ def _load(self, data, sections):
...
-        except configparser.NoOptionError as e:
-            raise MalformedConfig('missing key: \'{}\' in west config file'
-                                  .format(e.args[0])) from e
-
-        name = posixpath.basename(urlparse(url).path)
+        path = config.get('manifest', 'path', fallback=None)
 
         self_tag = manifest.get('self')
-        path = self_tag.get('path') if self_tag else name
+        if path is None:
```
mbolivar: We really need a `path` here or there is nothing more we can do. What about something like this?

```
if path is None:
    path = self_tag.get('path') if self_tag else None
if path is None:
    log.die('no manifest path and self.path is not set in the manifest file')
```

src/west/commands/project.py
tejlmand: You're absolutely right.
Personally I dislike magic numbers myself, so will change to MANIFEST_PROJECT_INDEX.

src/west/commands/project.py
tejlmand: yep, onto sounds better.
But then I should also change order, so end result becomes:
`To checkout and rebase {branch} onto {sha (detached-head)}`
agree ?

src/west/commands/project.py
mbolivar: OK, sounds good.

src/west/commands/project.py
mbolivar: That above discussion is about how they should apply to the manifest repo too

mbolivar: If we don't have time to support them now I suppose we could remove them and open an issue to add them back with the above discussion as the implementation strategy. Your call

src/west/commands/project.py
tejlmand: Current branch is only used for determine if HEAD of project is tracking a branch, so that we can inform user that he is now in detached head, and how he can rebase his branch onto the detached head.

So I don't see the reason why this should error out.
(mayby another function name ? as it sounds like you had expected another behavior based on the name)

src/west/commands/project.py
mbolivar: Zero is a sufficiently magic number that I think it's okay to use sometimes.
But I agree being consistent is better.

src/west/_bootstrap/main.py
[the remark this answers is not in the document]
tejlmand: sorry for that one.
Will put it back in.

src/west/commands/project.py
mbolivar: Agree

src/west/_bootstrap/main.py
tejlmand: I agree with you.

Guess the first change could have been done more carefully :(
I tried to be careful, especially on the Update -> SelfUpdate and then introduction of the new update command.
But not careful enough.
(would have like this as a couple independent PRs as well, but as you mention, hurry)

src/west/manifest.py
tejlmand: `None` is ok.
In case this is first run in PostInit and a url was used for `west init` there is a chance that `self::path` is not present (which is perfectly legal situation).
In that case, `manifest.py` will create manifest project without a path, and then PostInit updates the path and name in the manifest project afterwards based on the url name.

src/west/commands/project.py
tejlmand: I think it should be possible, I just rebase, and don't delete the existing command.
The adjustment to manifest repo to fetch / rebase default remote (usually origin) should not be complicated.

carlescufi: @mbolivar 
That conversation took place before the actual decision to remove those during a call. In fact the list we settled on was:

```
   Init(), - > same
    List(), -> Operate only on projects, except maybe if “-a”
    Clone(), -> Delete
    Fetch(), -> Delete
    Pull(), -> Delete
    Rebase(), -> Delete
    Branch(), -> Delete
    Checkout(), -> Delete
    Diff(), -> Operate only on projects, except maybe if “-a”
    Status(), -> Operate only on projects, except maybe if “-a”
    Update(), -> New command, old update becomes “west update --self-only”
    ForAll(), -> Operate only on projects, except maybe if “-a”
```

I do not think we have the time now to add those commands, especially since it is very likely that additional questions are raised during the addition.
My strong preference is to remove those commands for now, re-introduce later once west is in the tree.

src/west/commands/project.py
carlescufi: Agree with @mbolivar

src/west/commands/project.py
mbolivar: Sorry, as discussed on slack, I must have forgotten this. I'm ok with dropping these.

tejlmand:
> Proposal for the help text on `west init`:
> 
> ```
>         help='''Directory to initialize West in, unless the -l option is provided, in which case this is the local path to an existing manifest repository. Missing directories will be created automatically. (default: current directory)''
> ```

Done, although I still feel this might be little hard for user to read.
As first main sentence refers to all arguments, except `-l` which is then just briefly mentioned.
And then the lat part of the help text then again refers to all arguments.

> 
> ```
>         '-l', '--local', action='store_true',
>         help='''Use an existing local manifest repository directly. The local repository will not be modified in any way. Cannot be
> combined with --manifest-url and --manifest-rev''')
> ```

Done

src/west/commands/project.py
tejlmand: Yep, that was the list I was using.
For a second I thought the slack discussion came after this list.

Although:
```
“west update --self-only”
```
can be executed as
```
west selfupdate
```
as it will be a cleaner cut to completely remote the command, instead of changing signature of an existing command, when `west` goes fully `pip`

src/west/_bootstrap/main.py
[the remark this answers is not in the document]
tejlmand: Cleaned up.
This makes history more readable as those line are no longer removed here.

src/west/main.py
tejlmand: Removal of import is now moved to correct commit, although it doesn't make it bisectable as tests are still failing, due to `west clone` now missing, and `west update is not available until next commit.`

Reorderring commits however, will require updating test code which will anyway be removed due to commands (`clone/fetch/rebase/etc.`) being removed.

PR #156, opened 2019-01-23
Remove runners and build

mbolivar:
This series removes the runner commands **and build command**.

They are moving to the zephyr repository, since their implementations tend to be tightly coupled with the current state of that repository (boards, build system, etc. details matter). That will hopefully also make it easier to find them.


mbolivar:
@carlescufi @tejlmand let's get https://github.com/zephyrproject-rtos/west/pull/154 in first and i'll rebase this next

carlescufi:
This should be ready to be merged, @tejlmand?

tejlmand:
We should get PR/159 in with this one.

src/west/main.py
```diff
@@ -140,9 +140,19 @@ def append(*strings):
...
+            # Perhaps a commands.external_paranoid that if set, uses
+            # thunks, and otherwise just loads the modules and
+            # provides help for each command.
+            #
+            # This has its own wrinkle: we can't let a failed
+            # import break the built-in commands.
             if self.west_externals:
                 for path, specs in self.west_externals.items():
-                    append('commands from project {} at path "{}":'.
+                    append('commands from project at "{}":'.
```
tejlmand: Why removing project name here, i.e. `{}` but still having both `specs[0].project.name` and `path`
in the formatting string below ?
```format(specs[0].project.name, path))```

mbolivar: Thanks for catching that! Fixed

PR #158, opened 2019-01-24
extension_commands: Added test and allow for projects not cloned

tejlmand:
This pr rebases nicely onto PR/156

Added tests to ensure missing extension files (in case a project in
not cloned) will not cause west built-in commands to fail.

### DIFF
--- a/src/west/commands/command.py
+++ b/src/west/commands/command.py
@@ -170,6 +170,10 @@ def _ext_specs(project):
             'west-commands file {} escapes project path {}'.
             format(project.west_commands, project.path))
 
+    # Project may not be cloned yet.
+    if not os.path.exists(spec_file):
+        return []
+
     # Load the spec file and check the schema.
     with open(spec_file, 'r') as f:
         try:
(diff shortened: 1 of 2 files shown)


src/west/commands/command.py
```diff
@@ -170,6 +170,10 @@ def _ext_specs(project):
             'west-commands file {} escapes project path {}'.
             format(project.west_commands, project.path))
 
+    # Project may not be cloned yet.
+    if not os.path.exists(spec_file):
```
carlescufi: Would this be better handled in the `with open()` below catching the exception?

carlescufi: Also, should we verify that the commands themselves exist, beyond the spec file?

tejlmand: can be done later.
currently working on additional test / commit to ensure we don'tr allow duplicates, e.g proj1 defining `ext_cmd_a` as well as proj2 also defining the same command.

but not as important as this PR

tejlmand: > Would this be better handled in the `with open()` below catching the exception?

personally I prefer testing before trying.
but that's just my preference.

mbolivar: > Also, should we verify that the commands themselves exist, beyond the spec file?

This should raise a BadExternalCommand. I'd prefer to handle that with an extra exception handler in main.

carlescufi:
@mbolivar I can rebase this now

PR #159, opened 2019-01-25
extension command: Fixing issue with extension command duplication

tejlmand:
When playing with extension commands, two things were noticed:
1) Same command can be defined in multiple projects
2) When multiple projects are defining extension commands, they show up inside each others lists.

```
external commands from project Kconfiglib at path "subdir/Kconfiglib":
  kconfigtest
  test

external commands from project Kconfiglib at path "net-tools":
  kconfigtest
  test
```

This commit fixes:
- Issue if multiple projects defines the same command.
  This will print a warning for project/command not being available
- Fixed issue where same command could be listed multiple times

Note: doesn't rebase on top of pr/156, looking into that.

### DIFF
--- a/src/west/main.py
+++ b/src/west/main.py
@@ -152,8 +152,14 @@ class WestArgumentParser(argparse.ArgumentParser):
             # import break the built-in commands.
             if self.west_externals:
                 for path, specs in self.west_externals.items():
+                    # This may occur in case a project defines commands already
+                    # defined, in which case it has been filtered out.
+                    if not specs:
+                        continue
+
                     append('commands from project at "{}":'.
                            format(path))
+
                     for spec in specs:
                         self.format_external_spec(append, spec, width)
                     append('')
@@ -440,17 +446,26 @@ def get_external_commands():
         return {}
 
     externals = external_commands()
+    extension_commands = set()
 
-    filtered = []
     for path, specs in externals.items():
-        # Filter out attempts to shadow built-in commands.
+        # Filter out attempts to shadow built-in commands as well as
+        # commands which have names which are already used.
+        filtered = []
         for spec in specs:
             if spec.name in BUILTIN_COMMAND_NAMES:
                 log.wrn('ignoring project {} external command {};'.
                         format(spec.project.name, spec.name),
                         'this is a built in command')
                 continue
+            if spec.name in extension_commands:
+                log.wrn('ignoring project {} external command "{}";'.
+                        format(spec.project.name, spec.name),
+                        'command "{}" already defined as extension command'.
+                        format(spec.name))
+                continue
             filtered.append(spec)
+            extension_commands.add(spec.name)
         externals[path] = filtered
 
     return externals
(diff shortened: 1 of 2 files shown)


src/west/main.py
```diff
@@ -454,6 +459,8 @@ def get_external_commands():
     externals = external_commands()
 
     filtered = []
+    extension_commands = set()
+
     for path, specs in externals.items():
         # Filter out attempts to shadow built-in commands.
```
mbolivar: Maybe update this comment as well?

src/west/main.py
```diff
@@ -462,8 +469,16 @@ def get_external_commands():
...
+            if spec.name in extension_commands:
+                log.wrn('ignoring project {} external command "{}";'.
+                        format(spec.project.name, spec.name),
+                        'command "{}" already defined as extension command'.
+                        format(spec.name))
+                continue
             filtered.append(spec)
+            extension_commands.add(spec.name)
         externals[path] = filtered
+        filtered = []
```
mbolivar: How about just moving `filtered = []` to the line above `for spec in specs`, and deleting the line above `extension_commands = set()`? That might be a little easier to follow.

tests/west/project/test_project.py
```diff
@@ -427,6 +428,170 @@ def test_extension_command_execution(west_init_tmpdir):
...
+                            revision: zephyr
+                            path: subdir/Kconfiglib
+                            west-commands: scripts/west-commands.yml
+                          - name: net-tools
+                            west-commands: scripts/west-commands.yml
+                        self:
+                          path: zephyr
+                      '''.format(west=remote_west, rr=str(rr)))})
+
+    # Initialize the net-tools repository.
```
mbolivar: Shouldn't this be "initialize the kconfiglib repository"?

tests/west/project/test_project.py
```diff
@@ -427,6 +428,169 @@ def test_extension_command_execution(west_init_tmpdir):
...
+    expected = [
+        'commands from project at "subdir/Kconfiglib":',
+        '  kconfigtest',
+        '',
+        'commands from project at "net-tools":',
+        '  test',
+        '',
+        'Run "west <command> -h" for help on each command.',
+        '']
+    assert len(expected) == len(cmd_help), "expected:\n{}\nbut was:\n{}"\
```
mbolivar: Pytest is really smart about runtime type inspecting the arguments of either side of the assert. It will provide very useful feedback if there is a mismatch. Check this output for example:

```
=============================================== FAILURES ===============================================
_______________________________________________ test_foo _______________________________________________

    def test_foo():
        expected = [1,2,3,4]
        actual = [1,2,'a',4]
>       assert expected == actual
E       AssertionError: assert [1, 2, 3, 4] == [1, 2, 'a', 4]
E         At index 2 diff: 3 != 'a'
E         Use -v to get the full diff
```

So I think you can just write `assert expected == actual`, making the test easier to read but not compromising debuggability.

mbolivar:
@tejlmand i took the liberty  of just doing those changes myself

PR #162, opened 2019-01-28
bootstrap: allow bootstrapping a manifest repository using ref to a pr

tejlmand:
The current west bootstrapping does not support:
```
west init --mr pull/<pr>/head
```
as example:
```
west init --mr pull/12666/head
```
fails with error:
```
fatal: Remote branch pull/12666/head not found in upstream origin
```

this PR allows to do `west init` on a pr.

Changing `west init` into using
- git init
- git fetch
- git checkout
as this allows users to specify other refs, such as pull/<no>/head
which is not possible with git clone on github.


marc-hb:
Just for the record, linking to a partial revert of this being discussed in #498

PR #165, opened 2019-01-29
Misc fixes and `west update` with `--keep-descendants` and `--rebase`

mbolivar:
This series is mostly fixes.

The last patch contains a couple of new options for `west update`:

- `--keep-descendants` (or `-k`): if a project has a local branch checked out that is a descendant of the new `manifest-rev`, don't leave my branch behind
- `--rebase` (or `-r`): if a project has a local branch checked out, rebase it onto the new `manifest-rev`

This is a first step towards generalizing this command to support common use cases.

I can split the fixes up into their own PR, but I'm hoping that these options -- since they are opt in and apply only to projects, not the manifest repository -- will be uncontroversial.

### DIFF
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -462,14 +462,11 @@ def _projects(args, listed_must_be_cloned=True, include_west=False,
                     uncloned.append(project.name)
                 break
         else:
-            # The argument is not a project name. See if it is a project
-            # (sub)path.
+            # The argument is not a project name. See if it specifies
+            # an absolute or relative path to a project.
+            proj_arg_norm = normalize(project_arg)
             for project in projects:
-                # The startswith() means we also detect subdirectories of
-                # project repositories. Giving a plain file in the repo will
-                # work here too, but that probably doesn't hurt.
-                if normalize(project_arg).startswith(
-                        normalize(project.abspath)):
+                if proj_arg_norm == normalize(project.abspath):
                     res.append(project)
                     break
             else:
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -15,7 +15,7 @@ import textwrap
 from west.config import config
 from west import log
 from west import util
-from west.commands import WestCommand
+from west.commands import WestCommand, CommandError
 from west.manifest import Manifest, MalformedManifest, MANIFEST_PROJECT_INDEX
 from urllib.parse import urlparse
 import posixpath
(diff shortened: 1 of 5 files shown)


src/west/commands/project.py
```diff
@@ -212,38 +212,142 @@ def __init__(self):
...
             Updates all projects according to the manifest file, `west.yml`,
             in the manifest repository.
-            All projects will be checked out in detached HEAD and the local
-            manifest-rev branch will be reset hard to same HEAD.
 
-            Any local branches will not be altered by this command.
+            By default:
 
-            This command will not touch the manifest repository.
+            1. West itself is updated to the newest revision in the manifest
```
tejlmand: maybe: `to the revision in the manifest file. If revision is a branch, west will update to the tip of the branch`

cause, the revision in the manifest may not be newest revision in the west repo.

src/west/commands/project.py
```diff
@@ -212,38 +212,142 @@ def __init__(self):
...
+            file.
+
+            2. The local per-project manifest-rev branches tracked by
+            west will be hard reset to the updated commits fetched
+            from each project's remote.
```
tejlmand: Should it be mentioned the the branch `manifest-rev` is a west maintained branch, and that no work should be done on this branch ?

src/west/commands/project.py
```diff
@@ -212,38 +212,142 @@ def __init__(self):
...
+
+            3. All projects will have detached HEADs checked out at
+            the new manifest-rev commits.
+
+            4. Any locally checked out project branches will be left
+            behind, unaltered. When this happens, this command
```
tejlmand: I know `git` talks about leaving commits behind, when they are not on a branch, cause they will be garbage collected at some point, and thus being lost.

But I haven't heard anyone talking about, leaving a branch behind.
The branch will stay there, until deleted.

src/west/commands/project.py
```diff
@@ -212,38 +212,142 @@ def __init__(self):
...
+            # the new HEAD (e.g. if this is a topic branch the
+            # user is working on and the remote hasn't changed),
+            # print a message that makes it easy to get back,
+            # no matter where in the installation os.getcwd() is.
+            _wrn(project,
+                 ('left behind {{name}} branch "{}"; '
+                  'to fast forward back, use: git -C {} checkout {}').
+                 format(branch, relpath, branch))
+            log.dbg('(To do this automatically in the future,',
+                    'use "west update --keep-descendants".)')
```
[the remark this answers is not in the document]
tejlmand: So question is, if we should describe that 'rebase' is also an option in this case ?

src/west/commands/project.py
mbolivar: Great suggestion

src/west/commands/project.py
mbolivar: That's what I am trying to say with "per-project manifest-rev branches tracked by west".

Would it be more clear if this said "per-project manifest-rev branches, which are managed exclusively by west"?

src/west/commands/project.py
mbolivar: How about combinining 3. and 4. into:

```
3. All projects will have detached HEADs checked out at
the new manifest-rev commits. Locally checked out branches
are unaffected but will no longer checked out by default.
West will print information on how to get back to those branches
or update them to the new manifest-rev versions.

You can avoid 1. using --no-update. You can (sometimes)
avoid 3. using --keep-descendants and/or --rebase.
See below for more information.
```

src/west/commands/project.py
```diff
@@ -212,38 +212,142 @@ def __init__(self):
...
                 _inf(project,
-                     '{name} has been updated and is in detached HEAD.')
-                _inf(project,
-                     'To checkout and rebase \'{}\' onto \'{}\', run:'
-                     .format(branch, sha))
-                _inf(project,
-                     '`git rebase {} {}`'.format(sha, branch))
+                     'Left branch "{}", a descendant of {}, checked out'.
+                     format(branch, sha))
+            elif try_rebase:
```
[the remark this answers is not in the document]
mbolivar: Good catch. My view is that we should change `_rebase()` to actually rebase. This "never mind" behavior could be enabled with a new kwarg like `unless_descendant`, which defaults to False:

```
def _rebase(project, **kwargs):
    # Rebases the project against the manifest-rev branch
    #
    # Any kwargs are passed on to the underlying _git() call for the
    # rebase operation. A CompletedProcess instance is returned for
    # the git rebase.
    if kwargs.get('unless_descendant', False) and \
           _up_to_date_with(project, _MANIFEST_REV_BRANCH):
        _inf(project,
             '{name_and_path} is up-to-date with {manifest_rev_branch}')
    else:
        _inf(project, 'Rebasing {name_and_path} to {manifest_rev_branch}')
        return _git(project, 'rebase {qual_manifest_rev_branch}', **kwargs)
```

Your view?

src/west/commands/project.py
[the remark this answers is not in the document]
mbolivar: So in that case, why can't the user just `west update --rebase the-project-i-want-to-rebase; west update --keep-descendants --rebase` to rebase exactly what they want to rebase, and just keep descendants (or rebase) the rest?

src/west/commands/project.py
```diff
@@ -548,16 +652,18 @@ def _fetch(project):
...
+    # Any kwargs are passed on to the underlying _git() call for the
+    # rebase operation. A CompletedProcess instance is returned for
+    # the git rebase.
     if _up_to_date_with(project, _MANIFEST_REV_BRANCH):
         _inf(project,
              '{name_and_path} is up-to-date with {manifest_rev_branch}')
     else:
         _inf(project, 'Rebasing {name_and_path} to {manifest_rev_branch}')
-        _git(project, 'rebase {qual_manifest_rev_branch}')
-
```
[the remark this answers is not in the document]
mbolivar: Hm, I'll put it back, thanks

src/west/commands/project.py
```diff
@@ -548,16 +652,18 @@ def _fetch(project):
...
 
-def _rebase(project):
+def _rebase(project, **kwargs):
     # Rebases the project against the manifest-rev branch
-
+    #
+    # Any kwargs are passed on to the underlying _git() call for the
+    # rebase operation. A CompletedProcess instance is returned for
+    # the git rebase.
     if _up_to_date_with(project, _MANIFEST_REV_BRANCH):
```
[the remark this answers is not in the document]
mbolivar: I agree with you and have provided an alternative implementation in another comment -- how do you like that?

src/west/commands/project.py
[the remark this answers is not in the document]
tejlmand: I believe we can simply remove this part:
```
  if kwargs.get('unless_descendant', False) and \
           _up_to_date_with(project, _MANIFEST_REV_BRANCH):
        _inf(project,
             '{name_and_path} is up-to-date with {manifest_rev_branch}')
    else:
```
cause if user wants this behavior, he can specify `west update -k -r` where `-k` will take precedence over `-r` and then we have the behavior suggested with this code.

src/west/commands/project.py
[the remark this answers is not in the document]
tejlmand: He can.
My point was more a question of whether we should suggest him to do `--keep-descendants` as current log message do, or if we should present him with both options, i.e.:
Option 1: `--keep-descendants`
Option 2: `--rebase`

src/west/commands/project.py
tejlmand: I think that highlights better that it's not to be checked out by user.

src/west/commands/project.py
tejlmand: yep.
Minor comments:
> `at the new manifest-rev commits`

suggestion, remove `new`

> but will no longer checked out by default.

should be `but will no longer be checked out by default.`

src/west/commands/project.py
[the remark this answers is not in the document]
tejlmand: I believe, this line (currently 284):
` if args.keep_descendants and is_ancestor:`

removes the need for testing `_up_to_date_with` in `_rebase` method.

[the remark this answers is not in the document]
mbolivar: > I believe we can simply remove this part:

I was worried about other callers of this routine being broken if they expected the old behavior, but now I see there are none. I'll do as you say; I agree with your idea.

src/west/commands/project.py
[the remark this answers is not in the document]
mbolivar: In general I find that many users are scared of rebasing so this is a "friendliness" thing: "don't worry, you can keep your branches totally unchanged with -k" is the output when that is possible. Only if the user needs to rebase will they be given a suggestion to do so.

If both situations apply to different projects in the same `west update` call and the verbosity permits it, they will see both messages.

src/west/commands/project.py
mbolivar: Done, thanks

src/west/commands/project.py
mbolivar: done

src/west/commands/project.py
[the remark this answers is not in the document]
mbolivar: done

src/west/commands/project.py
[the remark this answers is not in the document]
mbolivar: done

mbolivar:
@tejlmand I've responded to your comments and update the PR. Giving the newest version some testing here; just wanted to get it up as soon as possible.

src/west/commands/project.py
carlescufi: Is this done? I still see: `1. West itself is updated to the newest revision in the manifest file.`

mbolivar: botched force push, now fixed

src/west/commands/project.py
```diff
@@ -212,38 +212,141 @@ def __init__(self):
...
-        return _add_parser(parser_adder, self, _no_update_arg,
+        return _add_parser(parser_adder, self,
+                           _no_update_arg,
+                           _arg('-k', '--keep-descendants',
+                                action='store_true',
+                                help=_wrap('''
+                                If a local branch is checked out and is a
+                                descendant commit of the new manifest-rev,
+                                then keep that descendant branch checked out
+                                instead of detaching HEAD. This takes priority
```
carlescufi: Should we specify here also `and if the manifest revision is a SHA`. Since otherwise it will check out a branch. And what happens if you use `-k` and the manifest rev is a branch?

src/west/commands/project.py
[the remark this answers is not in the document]
tejlmand: perfect :+1:

src/west/commands/project.py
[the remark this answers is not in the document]
tejlmand: point taken.
Keep as is :)

src/west/commands/project.py
tejlmand: :+1:

src/west/commands/project.py
tejlmand: :+1:

src/west/commands/project.py
tejlmand: nope, even if revision is a branch, it will still chackout as detached HEAD

mbolivar: > And what happens if you use -k and the manifest rev is a branch?

It will still compare their relationships in the graph:

```
$ git log --oneline --decorate --graph tmp2~..tmp1
* 3e8fc3c0c1 (tmp1) modem: wncm14a2a: Convert to new DT_<COMPAT>_<INSTANCE> defines
* fe8edc2884 (tmp2) display: ssd1306: Convert to new DT_<COMPAT>_<INSTANCE> defines
$ git merge-base --is-ancestor tmp2 tmp1; echo $?
0
$ git merge-base --is-ancestor tmp1 tmp2; echo $?
1
```

carlescufi: >nope, even if revision is a branch, it will still chackout as detached HEAD

Oh right, I guess I wasn't clear about that one then. Thanks for clarifying

carlescufi: >It will still compare their relationships in the graph:

Right, so it will simply use the 2 `HEAD`s of each branch as commits to run `merge-base` on

mbolivar:
> Discussed offline and I agree with the current behavior, with the provision of additional documentation

OK, sounds good. I'm going to write that documentation as a separate PR now, but will merge this first since we are OK with it.

Thanks for the reviews!

mbolivar:
> Looking good now.
> Great work

Thanks!

PR #170, opened 2019-01-29
bootstrap: Removed second fetch and allow for SHA

tejlmand:
This is a follow up PR to pr/162.

Main update is to remove the second `fetch`
Fetching the normal ref-space in first fetch also allows to checkout tags while still keep a reference to the tag name.
This means that when specifying a tag, users will now see, as example:
```
HEAD detached at v0.3.0
```
instead of:
```
HEAD detached at FETCH_HEAD
```

It also allows users to specify a SHA as revision.

edit: Also added `-C <dest>` as the change dir in pr/162 actually broke the directory arg.

### DIFF
--- a/src/west/_bootstrap/main.py
+++ b/src/west/_bootstrap/main.py
@@ -91,32 +91,49 @@ def west_topdir(start=None):
         cur_dir = parent_dir
 
 
+def _is_sha(s):
+    try:
+        int(s, 16)
+    except ValueError:
+        return False
+
+    return len(s) == 40
+
+
 def clone(desc, url, rev, dest, exist_ok=False):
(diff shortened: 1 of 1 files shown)


tejlmand:
> Looks good to me. Should we cut west 0.5.1?

Maybe wait for the `--zephyr-base` fix, in case that also requires an updated bootstrapper.
Thoughts ?

mbolivar:
> Maybe wait for the `--zephyr-base` fix, in case that also requires an updated bootstrapper.
> Thoughts ?
> 
> Of course this can be merged regardless of when we do `0.5.1`

Agreed on both fronts.

PR #174, opened 2019-01-29
[RFC] commands: update: be quieter about normal things

mbolivar:
Requested by @nashif.

Make the _fetch() git output quieter under ordinary circumstances by
capturing stdout and passing --quiet options to git when running 'west
update', and don't check out or print any information when HEAD is
already up to date with the newly fetched manifest-rev.

The new output looks like this even when run multiple times:

```
    $ west update
    === Fetching changes for net-tools (net-tools/)
    === Checking out revision '30b7efa827b04d2e47840716b0372737fe7d6c92' as detached HEAD in net-tools (net-tools/)
    HEAD is now at 30b7efa net-setup: zeth-vlan.conf: Add IP addresses to zeth
    $ west update
    === Fetching changes for net-tools (net-tools/)
    $
```

Projects are now only checked out when they need to change, and some
information that normally goes to stderr reporting progress is
suppressed.

### DIFF
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -618,12 +618,20 @@ def _all_projects():
         log.die(m.args[0])
 
 
-def _fetch(project):
+def _fetch(project, quiet=False):
     # Fetches upstream changes for 'project' and updates the 'manifest-rev'
     # branch to point to the revision specified in the manifest. If the
     # project's repository does not already exist, it is created first.
+    #
+    # The "quiet" kwarg reduces the verbosity of this fetch so that
+    # normal git output is reduced to log.dbg() levels using the
+    # _git_helper capture_stdout kwarg. Exceptional output (such as
+    # having to clone the repository) is never so silenced regardless
+    # of the quiet kwarg's value.
 
     if not _cloned(project):
+        # Don't use _msg or capture_stdout here -- this is an
+        # exceptional circumstance.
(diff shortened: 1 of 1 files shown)


mbolivar:
Just an RFC for the idea. I think the effect of this should be undone if the user says `west -v update`, which this code doesn't provide.

carlescufi:
Hey @mbolivar this looks like a good idea to me. The output is currently too verbose, which makes it hard to actually figure out what is happening. I tested this on our own repo as well, and the output is much more reasonable than before.
+1 for completing this.

@tejlmand thoughts?

mbolivar:
> Hey @mbolivar this looks like a good idea to me. The output is currently too verbose, which makes it hard to actually figure out what is happening. I tested this on our own repo as well, and the output is much more reasonable than before.

I meant to get you the before / after output on this but didn't finish (I had to rebase, then I noticed some issues with my original proposal and started to modify my approach, then I ran out of time); sorry about that. Thanks for the review and feedback

mbolivar:
Rebased. 

Mainline before this patch:

mainline-before

Mainline after this patch:

mainline-after

The Foundries.io downstream before this patch (better example):

foundries-before

And our downstream after this patch:

foundries-after

mbolivar:
I'm going to close this. I think the latest west looks good here:

westupdate

Issue #173, opened 2019-01-29
First 'west update' fails, second succeeds


mbolivar:
This affects 'west update' and 'west selfupdate' equally, of course.

Steps to reproduce:

```
$ cd $(west list -f="{abspath}" west)
$ git fetch upstream
$ git checkout upstream/master~3
$ west selfupdate
=== Updated west (.west/west/) to master (from https://github.com/zephyrproject-rtos/west).
/home/mbolivar/.local/share/virtualenvs/west/bin/python3: can't open file 'selfupdate': [Errno 2] No such file or directory
$ west selfupdate
```

The second one succeeds.

tejlmand:
Observation: when using `--reset-west` the selfupdate works on first try.
```
$ west selfupdate --reset-west
=== Fetching and resetting west (.west/west/) to 'master'
From https://github.com/zephyrproject-rtos/west
 * branch            master     -> FETCH_HEAD
```

mbolivar:
@tejlmand should we close this, or keep it open until we get one more upstream west commit merged so we can test it starting from c5323934df286aaa2e3d400dba417b838bfa9d19?

tejlmand:
You already closed it :p 
Think it's fine to close, cause you can still test it, by checking out an older revision, and then apply this fix manually without committing, and then run `west selfupdate`

mbolivar:
Sorry, stupid combo buttons ;). OK, let's keep this closed, thanks.

PR #176, opened 2019-01-30
selfupdate: Fix for west selfupdate mechanism

tejlmand:
Fixes: #173

This commit adds sys.argv[0] when restarting west after an update.

### DIFF
--- a/src/west/main.py
+++ b/src/west/main.py
@@ -503,7 +503,7 @@ def main(argv=None):
     except WestUpdated:
         # West has been automatically updated. Restart ourselves to run the
         # latest version, with the same arguments that we were given.
-        os.execv(sys.executable, [sys.executable] + argv)
+        os.execv(sys.executable, [sys.executable] + [sys.argv[0]] + argv)
     except KeyboardInterrupt:
         sys.exit(0)
     except CalledProcessError as cpe:


src/west/main.py
```diff
@@ -503,7 +503,7 @@ def main(argv=None):
     except WestUpdated:
         # West has been automatically updated. Restart ourselves to run the
         # latest version, with the same arguments that we were given.
-        os.execv(sys.executable, [sys.executable] + argv)
+        os.execv(sys.executable, [sys.executable] + [sys.argv[0]] + argv)
```
carlescufi: why not use `sys.argv` directly?

tejlmand: I think it has to do with wrapped commands, where main() could have been called with a set of argv != sys.argv.
Line 494:495 show that argv = sys.argv if when main(argv == None)

An example where argv != sys.argv is in post-init, where the wrapped command will has different parameters than the origonal west command.

And I didn't wanted to change that.
However, we still need sys.argv[0] in order to be able to reinvoke west.

carlescufi: Oh I see  yes, I didn't realize that `argv` could be different from `sys.argv[1:]`

PR #177, opened 2019-01-31
Extension command fixes

mbolivar:
A couple of trivial fixes:

- fix the error handling when we get a BadExternalCommand exception
- remove a stupid and unnecessary restriction on the names of extension commands

### DIFF
--- a/src/west/main.py
+++ b/src/west/main.py
@@ -22,7 +22,8 @@ import textwrap
 
 from west import log
 from west import config
-from west.commands import CommandError, CommandContextError, external_commands
+from west.commands import CommandError, CommandContextError, \
+    external_commands, BadExternalCommand
 from west.commands.project import List, Diff, Status, SelfUpdate, ForAll, \
                              WestUpdated, PostInit, Update
 from west.manifest import Manifest, MalformedConfig
@@ -519,6 +520,14 @@ def main(argv=None):
         sys.exit(cce.returncode)
     except CommandError as ce:
         sys.exit(ce.returncode)
+    except BadExternalCommand:
+        log.err('external command', args.command,
+                'was improperly defined and could not be run')
+        if args.verbose:
+            raise
+        else:
+            log.inf(for_stack_trace)
+            sys.exit(1)
 
 
 if __name__ == "__main__":
--- a/src/west/main.py
+++ b/src/west/main.py
@@ -19,11 +19,12 @@ import shutil
 import sys
 from subprocess import CalledProcessError, check_output, DEVNULL
 import textwrap
+import traceback
 
 from west import log
 from west import config
-from west.commands import CommandError, CommandContextError, \
-    external_commands, BadExternalCommand
+from west.commands import external_commands, \
+    CommandError, CommandContextError, ExtensionCommandError
 from west.commands.project import List, Diff, Status, SelfUpdate, ForAll, \
                              WestUpdated, PostInit, Update
 from west.manifest import Manifest, MalformedConfig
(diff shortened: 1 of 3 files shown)


src/west/main.py
```diff
@@ -22,7 +22,8 @@
 
 from west import log
 from west import config
-from west.commands import CommandError, CommandContextError, external_commands
+from west.commands import CommandError, CommandContextError, \
+    external_commands, BadExternalCommand
```
tejlmand: Just noticed here, as the errors are lined-up side-by-side.
Should we rename, `BadExternalCommand` ?
```
CommandError
CommandContextError
BadExternalCommand !!!
```
maybe this is nicer:
```
CommandError
CommandContextError
ExternalCommandError or maybe ExternalCommandInvalidError
```
Thoughts ?
This can also be made into independent issue top be taken later.

src/west/main.py
```diff
@@ -519,6 +520,14 @@ def main(argv=None):
         sys.exit(cce.returncode)
     except CommandError as ce:
         sys.exit(ce.returncode)
+    except BadExternalCommand:
+        log.err('external command', args.command,
+                'was improperly defined and could not be run')
```
tejlmand: are we able to provide more info to the user, on how the command was improperly defined ?

src/west/main.py
```diff
@@ -519,6 +520,14 @@ def main(argv=None):
...
+        if args.verbose:
+            raise
+        else:
+            log.inf(for_stack_trace)
+            sys.exit(1)
```
tejlmand: This code give different flows, depending on the `verbose` setting.
I don't think it's common for `-v` to change flow of the program, only to give higher level of verbosity.
Can you explain a bit more, why ?

src/west/main.py
mbolivar: For now that is all we have. I think since this patch is an improvement and we are short on time we should discuss further improvements later.

src/west/main.py
mbolivar: Sure, that's a good and easy change to make.

src/west/main.py
mbolivar: I disagree with the assessment that it changes the flow of the program. I think it really does just give a higher level of verbosity.

Since this is the top level main() function, the only difference in behavior is whether the stack trace is printed or the short error message is. Either way, after the diagnostic message (trace or error), the program exits immediately with a nonzero exit code. So I think this is just verbosity.

It exists elsewhere in the exception handlers in this same try block btw.

src/west/main.py
mbolivar: (Developers who are trying to debug can use `-v` to get the stack trace  -- users won't care, they just know "it didn't work")

tejlmand: thanks.

src/west/main.py
tejlmand: I think it changes the flow, as we'll do a `sys.exit()` right here, as in principle we can't know what higher level will do with the exception.

But ok to keep as-is.

> It exists elsewhere in the exception handlers in this same try block btw.

Maybe I just didn't notice before.

src/west/main.py
mbolivar: I've added a patch which does all of this. You convinced me :)

src/west/main.py
mbolivar: I've rewritten this in a different way to make that more clear using the traceback module.

mbolivar:
@tejlmand comments addressed, thanks for review!

src/west/main.py
mbolivar: > I think it changes the flow, as we'll do a `sys.exit()` right here, as in principle we can't know what higher level will do with the exception.

In this case we do, because we are main, and the higher layer is the python interpreter itself and its behavior when an exception is not handled is well known.

Even if we weren't the higher layer, it's also worth noting that `sys.exit()` doesn't actually abort the program -- it raises SystemExit. So even in the (unsupported and not stable API) case where someone imported main and ran it, we're just trading that exception for another.

At any rate, I agree that this point is worth making explicit in the code. Thanks again for the review!

src/west/main.py
tejlmand: I just asked out of curiosity, and was satisfied with your answer.
Appears you are easy to convince :)

PR #183, opened 2019-02-04
main: Execute sys.argv[0] on Windows

carlescufi:
On Windows, sys.argv[0] is the path to an executable (.exe)
instead of a Python script. Execute that file directly instead
of relying on Python to do so.

### DIFF
--- a/src/west/main.py
+++ b/src/west/main.py
@@ -505,7 +505,13 @@ def main(argv=None):
     except WestUpdated:
         # West has been automatically updated. Restart ourselves to run the
         # latest version, with the same arguments that we were given.
-        os.execv(sys.executable, [sys.executable] + [sys.argv[0]] + argv)
+        log.dbg("sys.executable:\"{}\" sys.argv[0]:\"{}\" argv:\"{}\"".format(
+                sys.executable, sys.argv[0], argv))
+        if sys.platform == "win32":
+            # On Windows Python generates .exe executables
+            os.execv(sys.argv[0], argv)
+        else:
+            os.execv(sys.executable, [sys.executable] + [sys.argv[0]] + argv)
     except KeyboardInterrupt:
         sys.exit(0)
     except CalledProcessError as cpe:


src/west/main.py
```diff
@@ -505,7 +505,13 @@ def main(argv=None):
     except WestUpdated:
         # West has been automatically updated. Restart ourselves to run the
         # latest version, with the same arguments that we were given.
-        os.execv(sys.executable, [sys.executable] + [sys.argv[0]] + argv)
+        log.inf("sys.executable:\"{}\" sys.argv[0]:\"{}\" argv:\"{}\"".format(
```
mbolivar: at the very least log.dbg

mbolivar: (Otherwise seems fine to me, KISS)

carlescufi: Fixed

mbolivar:
Would like to hear @tejlmand's thoughts

tejlmand:
Noticed this 
```
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
```
in: `~/.local/bin/west` file
which probably is the code that is compiled into an exe in windows.

Seems to be striping `.exe` but I don't see this code in `west` repo.

@mbolivar is this autogenerated when creating the PyPI ?

Secondly, could we be causing troubles for people who might be invoking `_bootstrap/main.py` directly with python ? 
(Not that I assume anyone does it, but should that be supported ?)


Complete file:
```
#!/usr/bin/python3

# -*- coding: utf-8 -*-
import re
import sys

from west._bootstrap.main import main

if __name__ == '__main__':
    sys.argv[0] = re.sub(r'(-script\.pyw?|\.exe)?$', '', sys.argv[0])
    sys.exit(main())
```

src/west/main.py
```diff
@@ -505,7 +505,13 @@ def main(argv=None):
     except WestUpdated:
         # West has been automatically updated. Restart ourselves to run the
         # latest version, with the same arguments that we were given.
-        os.execv(sys.executable, [sys.executable] + [sys.argv[0]] + argv)
+        log.dbg("sys.executable:\"{}\" sys.argv[0]:\"{}\" argv:\"{}\"".format(
+                sys.executable, sys.argv[0], argv))
+        if sys.platform == "win32":
+            # On Windows Python generates .exe executables
+            os.execv(sys.argv[0], argv)
```
tejlmand: Could we be causing troubles for people who might be invoking `_bootstrap/main.py` directly with python ? 
(Not that I assume anyone does it, but should that be supported ?)

carlescufi:
>@mbolivar is this autogenerated when creating the PyPI ?

Yep, this is autogenerated by `pip` or `setuptools`

mbolivar:
> > @mbolivar is this autogenerated when creating the PyPI ?
> 
> Yep, this is autogenerated by `pip` or `setuptools`

In particular, by our `entry_points` in setup.py: https://github.com/zephyrproject-rtos/west/blob/master/setup.py#L39

Issue #186, opened 2019-02-06
west fails to init repo across disks in windows


tejlmand:
When running `west init` on a disk different that the disk having the temp directory, west fails as:

Error
```
=== Creating repository for nrf (nrf\)
Traceback (most recent call last):
  File "c:\python37\lib\shutil.py", line 563, in move
    os.rename(src, real_dst)
OSError: [WinError 17] The system cannot move the file to a different disk drive: 'C:\\Users\\<user>\\AppData\\Local\\Temp\\manifest_p54lzomp' -> 'E:\\ncs2\\nrf'
```

mbolivar:
Dang.

Maybe we need to use .west/tmp as a staging directory during init. If we pass that to mkdtemp in the dir kwarg, we can guarantee the temporary directory is inside the installation.

PR #187, opened 2019-02-06
init: Change west init to clone manifest repo inside .west

tejlmand:
Fixes: #186

Instead of cloning manifest repo into temp directory, which can be a
disk different from final destination, then the manifest repo is now
cloned into .west, and from there, moved to its final location.

### DIFF
--- a/src/west/_bootstrap/main.py
+++ b/src/west/_bootstrap/main.py
@@ -13,7 +13,6 @@ import pykwalify.core
 import subprocess
 import sys
 import yaml
-import tempfile
 import shutil
 
 import west._bootstrap.version as version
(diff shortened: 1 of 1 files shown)


src/west/_bootstrap/main.py
```diff
@@ -285,7 +284,8 @@ def bootstrap(args):
 
     # Clone the west source code and the manifest into west/. Git will create
     # the west/ directory if it does not exist.
-    tempdir = tempfile.mkdtemp(prefix='manifest_')
+    tempdir = os.path.join(os.path.abspath(directory),
+                           WEST_DIR, 'manifest_repo')
```
carlescufi: As stated in Slack, I think this should be `.west/tmp/`

PR #188, opened 2019-02-06
fetch: With revision as a SHA west will fetch default ref-space

tejlmand:
This fixes an issue where west update could fail to checkout a SHA as
revision if that SHA was located on a branch that was not being
fetched.

This should address the issue discover by @nashif 

> playing with adding a new repo into west.yml

```
From https://github.com/zephyrproject-rtos/tinycbor
 * branch            HEAD       -> FETCH_HEAD
fatal: update_ref failed for ref 'refs/heads/manifest-rev': cannot update ref 'refs/heads/manifest-rev': trying to write ref f
FATAL ERROR: Command 'git update-ref refs/heads/manifest-rev 3ef9fedf6d4872a374a0d58f6d207ccbfee670ff' failed for tinycbor (t)
```

> I just added:

  ```projects:
    - name: net-tools
      revision: 30b7efa827b04d2e47840716b0372737fe7d6c92
    - name: tinycbor
      revision: 3ef9fedf6d4872a374a0d58f6d207ccbfee670ff```

### DIFF
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -628,7 +628,8 @@ def _fetch(project):
     # --tags is required to get tags when the remote is specified as an URL.
     if _is_sha(project.revision):
         # Don't fetch a SHA directly, as server may restrict from doing so.
-        _git(project, fetch_cmd + ' --tags -- {url}')
+        _git(project, fetch_cmd + ' --tags -- {url} '
+             'refs/heads/*:refs/west/*')
         _git(project, 'update-ref {qual_manifest_rev_branch} {revision}')
     else:
         _git(project, fetch_cmd + ' --tags -- {url} {revision}')


src/west/commands/project.py
```diff
@@ -628,7 +628,8 @@ def _fetch(project):
     # --tags is required to get tags when the remote is specified as an URL.
     if _is_sha(project.revision):
         # Don't fetch a SHA directly, as server may restrict from doing so.
-        _git(project, fetch_cmd + ' --tags -- {url}')
+        _git(project, fetch_cmd + ' --tags -- {url} '
+             'refs/heads/*:refs/west_refs/*')
```
mbolivar: As discussed on Slack, I think the risk of collisions when using `west` instead of `west_refs` is low enough that the extra ergonomics of having a shorter name is worthwhile.

tejlmand: Well, user is not supposed to be working on those refs.
Just as users are not supposed to work on `manifest_rev` branch.

Those are solely intended for `west` usage.
If github had allowed fetching the SHA directly, then there would be no need for fetching the whole ref-space locally.
The `refs/west_refs/*` is not linked to any git remote, and content will change with a changing url in manifest.
Thus, if user need this kind of functionality, the safest for them will be to create their own own remote, and fetch into that.

And in such case, they are probably way beyond a simple `west update` of a SHA and instead doing specific work, where they have local branches as well as a need to do `git push` to a fork.

So I don't think we should make short refs names, in order to aid the user.
Also more unique name, such as `west_refs` minimizes the (I agree, potentially small) risk of collision.

@carlescufi @nashif Thoughts ?

carlescufi: I agree with @mbolivar. A remote named `west` is so unlikely  that I would rather simplify the code and also enhance the aesthetics of the refspace  for the users that decide to inspect it than ensure lower collision probability.

tejlmand: ok, will chage.
But for the code, it will only change from:
```
refs/heads/*:refs/west_refs/*
```
into
```
refs/heads/*:refs/west/*
```

PR #190, opened 2019-02-08
list: updated abspath to use PurePath.as_posix()

tejlmand:
west list --format={abspath} is intended to use with Zephyr's CMake
build system. In order for CMake not escape path, such as `C:\<path>`
PurePath.as_posix() is used to ensure forward slashes when printing
abspath.

### DIFF
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -121,6 +121,8 @@ class List(WestCommand):
             - path: the relative path to the project from the top level,
               as specified in the manifest where applicable
             - abspath: absolute and normalized path to the project
+            - posixpath: like abspath, but in posix style, that is, with '/'
+              as the separator character instead of '\\'
             - revision: project's manifest revision
             - cloned: "(cloned)" if the project has been cloned, "(not cloned)"
               otherwise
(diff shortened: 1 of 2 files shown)


tejlmand:
@mbolivar @carlescufi I discovered an issue with https://github.com/zephyrproject-rtos/zephyr/pull/13067 when using `west list --format={abspath}` that on windows the abspath would be as:
`c:\<path>` causing charecters to be escaped in CMake.
Instead of having to modify inside CMake, I think it is nicer that `west` prints abspath in posix style.

src/west/commands/project.py
```diff
@@ -120,7 +120,8 @@ def do_add_parser(self, parser_adder):
             - url: full remote URL as specified by the manifest
             - path: the relative path to the project from the top level,
               as specified in the manifest where applicable
-            - abspath: absolute and normalized path to the project
+            - abspath: absolute and normalized path to the project in posix
```
mbolivar: I think this is likely to mess up other tools.

Can we add a new `{posixpath}` key instead?

mbolivar:
@tejlmand I have force pushed your PR with this change

tejlmand:
Great.
Thanks for the fix.

I fully support the `{posixpath}`, cause originally I wanted to introduce a new field like that, but then I assumed it might be easier to get a change on the `abspath` approved.
So i'm very happy you suggested my original preference.

PR #192, opened 2019-02-08
Manifest command selfupdate

tejlmand:
PR #189 with additional commit that makes `west selfupdate` behave similar to `west update` but for the `west repo`

### DIFF
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -296,28 +296,30 @@ class Update(WestCommand):
             'update',
             'update projects described in west.yml',
             _wrap('''
-            Updates all projects according to the manifest file, `west.yml`,
-            in the manifest repository.
+            Updates west and all project repositories according to the
+            manifest file, `west.yml`, in the manifest repository.
 
             By default:
 
-            1. West itself is updated to the revision in the manifest
-            file. If revision is a branch, west will update to the tip
-            of the branch.
+            1. The revisions in the manifest file are fetched from the
+            remote in the west repository in the installation and each
+            project repository.
 
-            2. The local per-project manifest-rev branches (which are
-            managed exclusively by west) will be hard reset to the
-            updated commits fetched from each project's remote.
+            2. The local manifest-rev branches in each repository
+            (which are managed exclusively by west) will be hard reset
+            to the updated commits fetched in step 1.
 
-            3. All projects will have detached HEADs checked out at
-            the manifest-rev commits. Locally checked out branches are
-            unaffected but will no longer be checked out by default.
-            West will print information on how to get back to those
-            branches or update them to the new manifest-rev versions.
+            3. All repositories will have detached HEADs checked out
+            at the new manifest-rev commits. Locally checked out
+            branches are unaffected but will no longer be checked out
+            by default.  West will print information on how to get
+            back to those branches or update them to the new
+            manifest-rev versions.
 
-            You can avoid 1. using --no-update. You can (sometimes)
-            avoid 3. using --keep-descendants and/or --rebase.
-            See below for more information.
+            You can skip updating west using --no-update.
+
+            You can influence the behavior when local branches are
+            checked out using --keep-descendants and/or --rebase.
 
             This command does not change the contents of the manifest
             repository.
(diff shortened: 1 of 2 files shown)


src/west/commands/project.py
```diff
@@ -398,82 +367,45 @@ def do_run(self, args, user_args):
...
-            'west fetch' and 'west pull' automatically update the West and
-            manifest repositories to the latest version before doing anything
-            else.
-
-            Pass --update-west or --update-manifest to update just that
-            repository. With no arguments, both are updated.
-
-            Updates are skipped (with a warning) if they can't be done via
-            fast-forward, unless --reset-west, or --reset-projects is given.
+            'west update' automatically update the West repository to the
```
mbolivar: "updates"

src/west/commands/project.py
```diff
@@ -398,82 +367,45 @@ def do_run(self, args, user_args):
...
             'selfupdate the west repository',
             _wrap('''
-            Updates the manifest repository and/or the West source code
-            repository. The remote to update from is taken from the
-            manifest.remote and manifest.remote configuration settings, and the
-            revision from manifest.revision and west.revision configuration
-            settings.
+            Updates the West source code repository. The remote to update from
+            is taken from the west.remote configuration setting, and the
+            revision from west.revision configuration setting.
```
mbolivar: These are actually in the manifest file, not a configuration file, right? Use of the word "configuration" makes that a bit confusing in my opinion.

src/west/commands/project.py
```diff
@@ -946,19 +855,73 @@ def _update_west():
...
 
-    project = _west_project()
-    with _error_context(', while updating/resetting west'):
+    branch = _current_branch(project)
+    sha = _sha(project, _MANIFEST_REV_BRANCH)
+    if branch is not None:
+        is_ancestor = _is_ancestor_of(project, sha, branch)
+        try_rebase = rebase
+    else:
+        # If no branch is checked out, -k and -r don't matter.
```
mbolivar: Probably want to update the comments in this function to not use -k or -r, since rebase and keep_descendants are formal parameters rather than command line arguments from this function's perspective.

src/west/commands/project.py
```diff
@@ -345,47 +345,16 @@ def do_add_parser(self, parser_adder):
 
     def do_run(self, args, user_args):
         if args.update:
-            _update_west()
+            _update_west(args.rebase, args.keep_descendants)
```
mbolivar: There's nothing wrong with this line but GitHub won't let me leave a line comment outside of the hunks in the diff.

The Update description needs to be changed, because -k and -r now affect 1. " West itself is updated to the revision in the manifest file" as well.

I would suggest a new description along these lines:

```
            Updates west and all project repositories according to the
            manifest file, `west.yml`, in the manifest repository.

            By default:

            1. The revisions in the manifest file are fetched from the
            remote in the west repository in the installation and each
            project repository.

            2. The local manifest-rev branches in each repository
            (which are managed exclusively by west) will be hard reset
            to the updated commits fetched in step 1.

            3. All repositories will have detached HEADs checked out
            at the new manifest-rev commits. Locally checked out
            branches are unaffected but will no longer be checked out
            by default.  West will print information on how to get
            back to those branches or update them to the new
            manifest-rev versions.

            You can skip updating west using --no-update.

            You can influence the behavior when local branches are
            checked out using --keep-descendants and/or --rebase.

            This command does not change the contents of the manifest
            repository.
```

src/west/commands/project.py
tejlmand: agree.
Guess I should have cleaned up a bit more of the old text.

src/west/commands/project.py
tejlmand: thanks for the updated text.
Fixed

src/west/commands/project.py
tejlmand: fixed

src/west/commands/project.py
tejlmand: fixed

mbolivar: There were two comments to update; the other one is:

```
        # A descendant is currently checked out and -k was
        # given, so there's nothing more to do.
```

mbolivar:
@tejlmand I've resolved review conversations as appropriate; one thread remains open.

src/west/commands/project.py
tejlmand: fixed

carlescufi:
@mbolivar please revisit

PR #193, opened 2019-02-08
West v0.5.3

mbolivar:
Just queueing this up so we can merge when ready. I think we ought to get https://github.com/zephyrproject-rtos/west/pull/192 in first before we cut this point release, though.


mbolivar:
@carlescufi @tejlmand pleased review

PR #200, opened 2019-02-13
West v0.5.4

mbolivar:
I screwed up the 0.5.3 upload to PyPI. Then I tried to delete it and
re-upload. This turns out to be impossible -- file names can't be
reused even if the package has been deleted.

So 0.5.3 is the point release that never was, unfortunately. 0.5.4
will be the same, but with a different number. Fun times / whoops.

### DIFF
--- a/src/west/_bootstrap/version.py
+++ b/src/west/_bootstrap/version.py
@@ -2,4 +2,4 @@
 #
 # This is the Python 3 version of option 3 in:
 # https://packaging.python.org/guides/single-sourcing-package-version/#single-sourcing-the-version
-__version__ = '0.5.3'
+__version__ = '0.5.4'


mbolivar:
@carlescufi @tejlmand  some additional details:

I've switched to arch for daily work, where `python` is python 3, but did the upload from my old laptop (which has the yubikey I use to sign these releases configured), where `python` is python 2. So the 0.5.3 sdist that got uploaded to PyPI was built with Python 2. I can't seem to fix this, so we need to just cut a new point release.

Sorry about that.

tejlmand:
guess, 0.5.3 was a bad number anyway :p

carlescufi:
>So the 0.5.3 sdist that got uploaded to PyPI was built with Python 2.

That's why I didn't get it with `pip3` then... :)

PR #202, opened 2019-02-13
manifest: Serialize remote[url-base] properly

doanac:
Since the python object is named url_base, this field is getting
frozen via as_dict with an underscore rather than the dash required
by the schema.

Signed-off-by: Andy Doan <andy@foundries.io>

### DIFF
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -359,14 +359,14 @@ class Remote:
         return self.name == other.name and self.url_base == other.url_base
 
     def __repr__(self):
-        return 'Remote(name={}, url_base={})'.format(repr(self.name),
+        return 'Remote(name={}, url-base={})'.format(repr(self.name),
                                                      repr(self.url_base))
 
     def as_dict(self):
         '''Return a representation of this object as a dict, as it would be
         parsed from an equivalent YAML manifest.'''
         return collections.OrderedDict(
-            ((s, getattr(self, s)) for s in self.__slots__))
+            ((s.replace('_', '-'), getattr(self, s)) for s in self.__slots__))
 
 
 class Project:
--- a/tests/west/project/test_project.py
+++ b/tests/west/project/test_project.py
@@ -266,7 +266,7 @@ def test_manifest_freeze(west_update_tmpdir):
                     '^    revision: master$',
                     '^  remotes:$',
                     '^  - name: test-local$',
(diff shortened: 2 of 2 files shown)


mbolivar:
Nice catch! Can you update test_project.py too? This is what's failing:

```
west_update_tmpdir = local('/tmp/pytest-of-root/pytest-0/test_manifest_freeze0/west_installation')

    def test_manifest_freeze(west_update_tmpdir):
        # We should be able to freeze manifests.
        actual = cmd('manifest --freeze').splitlines()
        # Match the actual output against the expected line by line,
        # so failing lines can be printed in the test output.
        #
        # Since the actual remote URLs and SHAs are not predictable, we
        # don't match those precisely. However, we do expect the output to
        # match project order as specified in our manifest, that all
        # revisions are full 40-character SHAs, and there isn't any random
        # YAML tag crap.
        expected_res = ['^west:$',
                        '^  url: .*$',
                        '^  revision: [a-f0-9]{40}$',
                        '^manifest:$',
                        '^  defaults:$',
                        '^    remote: test-local$',
                        '^    revision: master$',
                        '^  remotes:$',
                        '^  - name: test-local$',
                        '^    url_base: .*$',
                        '^  projects:$',
                        '^  - name: Kconfiglib$',
                        '^    remote: test-local$',
                        '^    revision: [a-f0-9]{40}$',
                        '^    path: subdir/Kconfiglib$',
                        '^  - name: net-tools$',
                        '^    remote: test-local$',
                        '^    revision: [a-f0-9]{40}$',
                        '^    west-commands: scripts/west-commands.yml$',
                        '^  self:$',
                        '^    path: zephyr$']
    
        for eline_re, aline in zip(expected_res, actual):
>           assert re.match(eline_re, aline) is not None, (aline, eline_re)
E           AssertionError: ('    url-base: file:///tmp/pytest-of-root/pytest-0/test_manifest_freeze0/repos', '^    url_base: .*$')
E           assert None is not None
E            +  where None = <function match at 0x7f798d6a6b70>('^    url_base: .*$', '    url-base: file:///tmp/pytest-of-root/pytest-0/test_manifest_freeze0/repos')
E            +    where <function match at 0x7f798d6a6b70> = re.match

/root/src/github.com/zephyrproject-rtos/west/tests/west/project/test_project.py:283: AssertionError
```

PR #204, opened 2019-02-15
main: Add "help" command

carlescufi:
Adds a `west help <command>` command that simply invokes the corresponding `-h`.


src/west/main.py
```diff
@@ -503,16 +509,17 @@ def main(argv=None):
...
                 sys.executable, sys.argv[0], argv))
-        if sys.platform == "win32":
-            # On Windows Python generates .exe executables
-            os.execv(sys.argv[0], argv)
-        else:
-            os.execv(sys.executable, [sys.executable] + [sys.argv[0]] + argv)
+        # Treat the Python script as an executable
+        proc = Popen([sys.argv[0]] + argv)
+        proc.communicate()
+        sys.exit(proc.returncode or 0)
```
carlescufi: I had `or errno.EIO` here but apparently for whatever reason `proc.returncode` is sometimes `None` and the unit tests failed. Any ideas?

src/west/commands/common.py
```diff
@@ -0,0 +1,69 @@
+# Copyright (c) 2019, Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: Apache-2.0
+
+'''West common commands'''
```
mbolivar: Would you mind just putting this in a `help.py` file for now, to make it easy for newcomers to find?

src/west/commands/common.py
```diff
@@ -0,0 +1,69 @@
...
+
+from west.commands import WestCommand
+from west.util import WestReload
+
+class Help(WestCommand):
+    def __init__(self):
+        super().__init__(
+            'help',
+            'print west command help text',
+            _wrap('''
```
mbolivar: When the help text is only a single line, you should be able to just use a single quoted string. We use triple quotes (and the related helper functions for controlling whitespace) only when there are multiple lines.

src/west/commands/common.py
```diff
@@ -0,0 +1,69 @@
...
+            Print help text on any command.
+            '''))
+
+    def do_add_parser(self, parser_adder):
+        return _add_parser(
+            parser_adder, self,
+            _arg('command', metavar='cmd', nargs='?',
+                 help='''Command to get help on''')
```
mbolivar: Ditto here

src/west/commands/common.py
```diff
@@ -0,0 +1,69 @@
...
+    # later passed in parents= to add_parser()
+
+    parser = argparse.ArgumentParser(add_help=False)
+    parser.add_argument(*args, **kwargs)
+    return parser
+
+
+# Arguments shared between more than one command
+
+def _add_parser(parser_adder, cmd, *extra_args, **kwargs):
```
mbolivar: Is it worth moving _arg, _add_parser, and _wrap to helpers importable from `west.command` instead of keeping separate copies in project.py and this new file?

src/west/main.py
```diff
@@ -48,15 +55,14 @@
 # Built-in commands in this West. For compatibility with monorepo
 # installations of West within the Zephyr tree, we only expose the
 # project commands if this is a multirepo installation.
```
mbolivar: Mind updating this comment while you are here?

src/west/main.py
```diff
@@ -508,11 +508,8 @@ def main(argv=None):
         # latest version, with the same arguments that we were given.
         log.dbg("sys.executable:\"{}\" sys.argv[0]:\"{}\" argv:\"{}\"".format(
```
mbolivar: I guess this debug log needs an update too. Maybe just log `sys.argv[0]` and `argv`.

src/west/main.py
```diff
@@ -26,9 +26,16 @@
 from west.commands import external_commands, \
     CommandError, CommandContextError, ExtensionCommandError
 from west.commands.project import List, ManifestCommand, Diff, Status, \
-    SelfUpdate, ForAll, WestUpdated, PostInit, Update
+    SelfUpdate, ForAll, PostInit, Update
+from west.commands.common import Help
 from west.manifest import Manifest, MalformedConfig
-from west.util import quote_sh_list
+from west.util import quote_sh_list, WestReload
```
mbolivar: I kind of feel like WestReload belongs in `west.commands`, since commands should be the only things that can raise the exception. Not married to the idea, though. What do you think?

src/west/main.py
```diff
@@ -517,7 +517,9 @@ def main(argv=None):
         log.dbg("sys.executable:\"{}\" sys.argv[0]:\"{}\" argv:\"{}\"".format(
                 sys.executable, sys.argv[0], argv))
         # Treat the Python script as an executable
-        os.execv(sys.argv[0], argv)
+        proc = Popen([sys.argv[0]] + argv)
```
mbolivar: Does this break control-C in a gdb started by `west debug`? Attempts to run west as a subprocess generally do, because the SIGINT gets delivered to the parent process, which exits, instead of the subprocess west, which is temporarily ignoring it so GDB can handle it.

src/west/main.py
mbolivar: That makes no sense; communicate should wait until the process is done. A None returncode, per documentation, means the subprocess isn't done executing yet.

But please see above about whether this approach is workable.

src/west/commands/common.py
carlescufi: Agreed

src/west/commands/common.py
carlescufi: ok!

src/west/commands/common.py
carlescufi: Yes, I think this makes sense. I will take a look.

src/west/main.py
carlescufi: Will do

src/west/main.py
carlescufi: It might have,  but `west debug` will never raise the exception. Only  commands that update repos and the help command can raise it. So no interactive commands raise the exception.

src/west/main.py
carlescufi: It's strange yes. I will  debug this further and update.

src/west/main.py
carlescufi: To add to this, we will remove the `selfupdate` when we move everything to pypi and also the current `execv` is actually broken on Windows, there's a race and sometimes you can end up with having to press enter to return to the shell. Finally, I think it's more likely that control-C will work reliably with `Popen` than with `execv`, at least on Windows

src/west/main.py
carlescufi: I am indifferent in this case. I guess it could be raised by main itself in the future but I am fine with either

src/west/main.py
carlescufi: Re: control-C you can see that actually both child and parent get the signal [here](https://stackoverflow.com/a/19807536/337201)

src/west/main.py
mbolivar: Let's keep it here for now until we're ready to refactor to commit to a real API

src/west/main.py
carlescufi: Doh, `0 or x` yields x... :|

src/west/commands/common.py
carlescufi: Done

src/west/commands/common.py
carlescufi: Done

src/west/commands/common.py
carlescufi: Done

src/west/commands/common.py
carlescufi: moved

src/west/main.py
carlescufi: updated

src/west/main.py
carlescufi: updated

tests/west/commands/test_commands.py
```diff
@@ -190,6 +190,22 @@ def west_update_tmpdir(west_init_tmpdir):
...
+    assert(h1out == h2out)
+
+    cmds = ['help', 'list', 'manifest', 'diff', 'status', 'update',
+            'selfupdate', 'forall']
+
+    for c in cmds:
+        h1out = cmd('help {}'.format(c))
+        h2out = cmd('{} -h'.format(c))
+        assert(h1out == h2out)
+
```
tejlmand: I think there should be an additional blank line here.

tests/west/commands/test_commands.py
```diff
@@ -282,7 +298,6 @@ def test_manifest_freeze(west_update_tmpdir):
     for eline_re, aline in zip(expected_res, actual):
         assert re.match(eline_re, aline) is not None, (aline, eline_re)
 
-
```
tejlmand: Why is this line removed ?

src/west/commands/project.py
```diff
@@ -857,7 +823,7 @@ def _update_west(rebase, keep_descendants):
             # Signal self-update, which will cause a restart. This is a bit
             # nicer than doing the restart here, as callers will have a
             # chance to flush file buffers, etc.
-            raise WestUpdated()
+            raise WestReload()
```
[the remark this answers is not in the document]
carlescufi: Yes but the issue here is that with `west help` we have *not* updated west, we just want to reload it with different parameters. In the same way that you signal a `RuntimeError` you can signal a `WestReload`.

src/west/commands/command.py
```diff
@@ -129,8 +184,6 @@ def do_run(self, args, unknown):
                         this object, this parameter is an empty sequence.
                         Otherwise, it is an iterable containing all unknown
                         arguments present on the command line.'''
-
-
 class WestExtCommandSpec:
```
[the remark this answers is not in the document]
carlescufi: Too much whitespace :)

tests/west/commands/test_commands.py
carlescufi: Too much whitespace ;)

src/west/commands/project.py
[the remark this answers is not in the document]
tejlmand: Then I understand the need for a `rename`, but I still think `WestReload` is bad, cause we haven't reloaded (yet).
Not sure what a better name would be though.

src/west/commands/help.py
```diff
@@ -0,0 +1,28 @@
...
+    def do_add_parser(self, parser_adder):
+        return self.add_parser_helper(
+            parser_adder,
+            self.add_arg('command', metavar='cmd', nargs='?',
+                         help='Command to get help on')
+        )
+
+    def do_run(self, args, user_args):
+        a = [args.command, '-h']
+        wr = WestReload(argv=[x for x in a if x is not None])
```
tejlmand: Testing this out, I noticed that if an invalid command name is given, then user gets the default `west` usage.
```
$ west help unknown_cmd
usage: west [-h] [-z ZEPHYR_BASE] [-v] [-V] <command> ...
west: error: argument <command>: invalid choice: 'unknown_cmd' (choose from 'list', 'manifest', 'diff', 'status', 'update', 'selfupdate', 'forall', 'post-init', 'help', 'build', 'flash', 'd
```
But in this case, `help` which is a valid command was given.

In contrast , if too many args a given to `west help` you'll get a `west help` error message.
```
$ west help list list
usage: west help [-h] [cmd]
west help: error: unexpected arguments: ['list']
```

Unfortunately, i'm not sure this is so easy to fix.

src/west/commands/help.py
```diff
@@ -0,0 +1,28 @@
...
+        raise wr
```
tejlmand: Just doing some more thinking here.
Are there any reasons why we explicitly need to relaunch `west` ?
It should be possible to just call into `west.main` as we do with bootstrapper and wrapped commands.

The reason why `west update` has to relaunch, is because the `west` source code have changed when the SHAs differs, and thus we need to run rest of the upgrade process using the new `west` code.

tests/west/commands/test_commands.py
tejlmand: I thought you were using a black background on your screen.

src/west/commands/help.py
tejlmand: Not relaunching `west` could also make it easier to fix this https://github.com/zephyrproject-rtos/west/pull/204#discussion_r258152013

tejlmand: This seems to work the same, and avoids the overhead of an exception and relaunching:
```
        import west.main
        west.main.main(argv=[x for x in a if x is not None])
```
instead of:
```
       wr = WestReload(argv=[x for x in a if x is not None])
       raise wr
```
Disclaimer, not fully tested at this point.

tejlmand: Note: `west help init` will need special handling with suggested approach, as `west init -h` is solely handled in the bootstrapper today.

src/west/commands/project.py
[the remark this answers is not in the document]
tejlmand: As commented here: https://github.com/zephyrproject-rtos/west/pull/204#discussion_r258153115

I actually think we should only relaunch `west` if there really is a need, such as when `west` itself has a changed SHA due to an `selfupdate`

src/west/commands/help.py
carlescufi: I am very reluctant to do that because there's a whole bunch of things that are initialized and running inside the same Python VM instance makes me uncomfortable. Also it is very ugly to call back into what just called you IMHO?
@mbolivar thoughts?

src/west/commands/help.py
carlescufi: Yeah, I noticed that too. We can fix this later but it's not going to be easy because of the way we use ArgParse. It's `parse_known_ags` that deals with that automagically for us and displays the help text.

tejlmand:
Should `west help` be moved to the bootstrapper ?

Doing `west help init` fails when not in project:
```
$ west help init
Error: "/home" is not in a west installation.
Things to try:
 - Set ZEPHYR_BASE to a zephyr repository path in a west installation.
 - Run "west init" to set up an installation here.
 - Run "west init -h" for additional information.
```

`west init -h` of course works.

carlescufi:
>Should `west help` be moved to the bootstrapper ?

That's an interesting idea. What I don't like about it is that it's going to be a temporary solution anyway (we are going to go full pypi soon) and that it would not reuse all of the ArgParse/WestCommand architecture.

src/west/commands/help.py
tejlmand: @carlescufi in general I agree with you, but with the `west help` we know exactly where we are.
And seeing the `west help <cmd>` as a wrapper for `west <cmd> -h` I think that is nicer in this particular case.

I think it might be possible to fetch directly into `<command>.arg_parser` to get the help text we need.
That could make it an even nicer solution, than calling `west.main.main`
@mbolivar You have bit more experience with arg parser, would should approach seem feasible ?

My main point is that we don't need to relaunch.

tejlmand:
> > Should `west help` be moved to the bootstrapper ?
> 
> That's an interesting idea. What I don't like about it is that it's going to be a temporary solution anyway (we are going to go full pypi soon) and that it would not reuse all of the ArgParse/WestCommand architecture.

Yep, just wanted to pint out the fact that `west help init` is not working for bootstrapper.
If that is acceptable, I'm fine with that :)

tests/west/commands/test_commands.py
tejlmand: Could you please put the blank line back ?
I believe we are following pep8 for python code:
https://www.python.org/dev/peps/pep-0008/

> Blank Lines
>
> Surround top-level function and class definitions with two blank lines.

mbolivar:
> take a look at the two missing blank lines.

Let me explain what's going on here.

There are 2 blank lines between function and class definitions in several places in the code because indeed, that is pep8 style and flake8's default configuration complains when they are missing.

However (as shown here and in other PRs where @carlesc has snuck in whitespace changes removing them into other patches 😉) it seemed to me that most people didn't like that rule, so I disabled that check in our flake8 configuration:

https://github.com/zephyrproject-rtos/west/blob/master/tox.ini#L10

So in general I am ok with just one blank line. I do agree that in general whitespace "fixes" like this should be kept in their own patches though since they tend to clutter diffs and distract attention from the behavioral changes, also as shown here.

tejlmand:
>  it seemed to me that most people didn't like that rule, so I disabled that check in our flake8 configuration:

Personally, i'm fine with single line and i'm fine with double lines.
What I dislike is mixing, i.e. some places with one line and other with double lines.
So if going for single blank line, I think all code should be adopted to that.

carlescufi:
@mbolivar addressed all your comments now.

mbolivar:
> So if going for single blank line, I think all code should be adopted to that.

I agree but haven't wanted to speak up about this yet, we have a lot of refactoring to do anyway.

Can we agree to hold off on further whitespace changes until we are ready to convert the entire code base?

src/west/commands/help.py
```diff
@@ -0,0 +1,28 @@
+# Copyright (c) 2019, Nordic Semiconductor ASA
+#
+# SPDX-License-Identifier: Apache-2.0
+
+'''West common commands'''
```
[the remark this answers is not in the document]
carlescufi: This file doesn't exist anymore

src/west/commands/command.py
```diff
@@ -129,8 +184,6 @@ def do_run(self, args, unknown):
                         this object, this parameter is an empty sequence.
                         Otherwise, it is an iterable containing all unknown
                         arguments present on the command line.'''
-
```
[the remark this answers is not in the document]
carlescufi: There's actually 2 right now

src/west/commands/command.py
```diff
@@ -100,6 +102,56 @@ def add_parser(self, parser_adder):
...
+        '''
+        parser = argparse.ArgumentParser(add_help=False)
+        parser.add_argument(*args, **kwargs)
+        return parser
+
+    @staticmethod
+    def wrap_help(s):
+        ''' Wraps help texts for commands.
+
+        Some of them have variable length (due to _MANIFEST_REV_BRANCH), so
```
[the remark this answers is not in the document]
carlescufi: This is no longer applicable, `command.py` has not changed anymore

PR #206, opened 2019-02-20
main: Use Popen() + exit instead of execv()

carlescufi:
On Windows execv() is asynchronous, leading to an additional
shell being open concurrently with the existing one.
Use Popen() instead, which ensure synchronicity across all
operating systems.

https://bugs.python.org/issue9148
https://bugs.python.org/issue19124

### DIFF
--- a/src/west/main.py
+++ b/src/west/main.py
@@ -11,13 +11,14 @@
 
 import argparse
 import colorama
+import errno
 from functools import partial
 from io import StringIO
 import itertools
 import os
 import shutil
 import sys
-from subprocess import CalledProcessError, check_output, DEVNULL
+from subprocess import Popen, CalledProcessError, check_output, DEVNULL
 import textwrap
 import traceback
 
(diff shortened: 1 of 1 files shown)


src/west/main.py
```diff
@@ -506,13 +507,23 @@ def main(argv=None):
...
-            # On Windows Python generates .exe executables
-            os.execv(sys.argv[0], argv)
-        else:
-            os.execv(sys.executable, [sys.executable] + [sys.argv[0]] + argv)
+        # Treat the Python script as an executable. This works because on
+        # Unix the script created by pip has a shebang and on Windows it is
+        # actually a binary executable
+        log.dbg("sys.argv[0]:\"{}\" argv:\"{}\"".format(sys.argv[0], argv))
+        # Use Popen + exit instead of execv due to the asynchronous nature of
+        # execv # on Windows, where it creates a new process with a different
```
mbolivar: Looks like these got badly reflowed, note # after execv, and other lines too.

src/west/main.py
```diff
@@ -506,13 +507,23 @@ def main(argv=None):
...
+        # pid # that executes in parallel to the original one instead of
+        # replacing # it as it does on UNIX
+        # https://bugs.python.org/issue9148
+        # https://bugs.python.org/issue19124
+        try:
+            proc = Popen([sys.argv[0]] + argv)
+            proc.communicate()
+        except KeyboardInterrupt:
+            sys.exit(0)
```
mbolivar: Might want to treat this as an error and exit 1.

tejlmand: In that case, I think the line 4 lines below of similar construct should also be changed.

carlescufi:
I tested this with an infinite loop after the update and then pressing ctrl+c on both Linux and Windows. The parent `KeyboardInterrupt` exception is caught first, but the child one is caught as well.

```
carles@zephyr-nrf ~/src/zephyr/.west/west/tests/west/commands (popen *)
$ west manifest --freeze
handling
raising update
popening
handling
looping in list
^Ckbi parent
kbi child
```

```
C:\Users\Carles\src\zephyr\.west\west>west manifest --freeze
handling
raising update
popening
handling
looping in list
kbi parent
kbi child
```

src/west/main.py
carlescufi: Fixed

src/west/main.py
carlescufi: Fixed

PR #209, opened 2019-02-21
project: Rename --no-update to --exclude-west

carlescufi:
Since excluding west from an update procedure is limited to a single
command now (west update), clarify that in the code and use a more
explicit flag name.


carlescufi:
>Seems OK to me but this does break compatibility. Do you want to keep the old option around but make a custom action that sets the new dest but logs a warning about the deprecation?

Honestly I think it's not worth it. I very much doubt anyone is using that option and the few who do can easily find the new one by using `west update -h`.

carlescufi:
>This looks ok, but please update `scripts/west-completion.bash` also

I did?

PR #210, opened 2019-02-21
main.py: trivial comment fix

mbolivar:
Delete some # characters left over from an improperly reflowed paragraph.


carlescufi:
I fixed this locally but never pushed it apparently. Sorry!

PR #211, opened 2019-02-21
zephyr-base: ZEPHYR_BASE env takes precedence over config settings

tejlmand:
As discussed on slack.

Default order is:
1) `--zephyr-base` on cmd line
2) Environment setting, `ZEPHYR_BASE`
3) Config file setting. Config file is now written with `zephyr-base` during init, and on `self-update` if missing

This commit changes the order of precedence so that environment setting
takes precedence over west config option.
It also introduces a new config file setting, to allow to specify which
setting: `environment` or `configfile` that will take precedence.
To simplify search for ZEPHYR_BASE, this setting will now be written to
the local .west/config during init and selfupdate, if non-existing.

### DIFF
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -63,24 +63,27 @@ class PostInit(WestCommand):
 
         manifest_file = os.path.join(args.local or args.cache, 'west.yml')
 
-        project = Manifest.from_file(manifest_file)\
-            .projects[MANIFEST_PROJECT_INDEX]
+        projects = Manifest.from_file(manifest_file).projects
+        manifest_project = projects[MANIFEST_PROJECT_INDEX]
 
         if args.local is not None:
             rel_manifest = os.path.relpath(args.local, util.west_topdir())
             update_config('manifest', 'path', rel_manifest)
         else:
-            if project.path == '':
+            if manifest_project.path == '':
                 url_path = urlparse(args.manifest_url).path
-                project.path = posixpath.basename(url_path)
-                project.abspath = os.path.realpath(
-                    os.path.join(util.west_topdir(), project.path))
-                project.name = project.path
+                manifest_project.path = posixpath.basename(url_path)
+                manifest_project.abspath = os.path.realpath(
+                    os.path.join(util.west_topdir(), manifest_project.path))
+                manifest_project.name = manifest_project.path
 
-            _inf(project, 'Creating repository for {name_and_path}')
-            shutil.move(args.cache, project.abspath)
+            _inf(manifest_project, 'Creating repository for {name_and_path}')
+            shutil.move(args.cache, manifest_project.abspath)
 
-            update_config('manifest', 'path', project.path)
+            update_config('manifest', 'path', manifest_project.path)
+        for project in projects:
+            if project.path == 'zephyr':
+                update_config('zephyr', 'base', project.path)
 
 
 class List(WestCommand):
@@ -852,6 +855,12 @@ def _update_west(rebase, keep_descendants):
 
         old_sha = _sha(project, 'HEAD')
         _update(project, rebase, keep_descendants)
+
+        if config.get('zephyr', 'base', fallback=None) is None:
+            for proj in _all_projects():
+                if proj.path == 'zephyr':
+                    update_config('zephyr', 'base', proj.path)
+                    break
 
         if old_sha != _sha(project, 'HEAD'):
             _inf(project,
--- a/src/west/main.py
+++ b/src/west/main.py
@@ -29,7 +29,7 @@ from west.commands import external_commands, \
 from west.commands.project import List, ManifestCommand, Diff, Status, \
     SelfUpdate, ForAll, WestUpdated, PostInit, Update
 from west.commands.config import Config
-from west.manifest import Manifest, MalformedConfig
+from west.manifest import MalformedConfig
 from west.util import quote_sh_list
 
 PROJECT_COMMANDS = {
@@ -358,10 +358,18 @@ def help_command_handler(west_parser, help_parser, externals, args, *ignored):
 
 
 def set_zephyr_base(args):
-    '''Ensure ZEPHYR_BASE is set, emitting warnings if that's not
-    possible, or if the user is pointing it somewhere different than
-    what the manifest expects.'''
-    zb_env = os.environ.get('ZEPHYR_BASE')
+    '''Ensure ZEPHYR_BASE is set
+    Order of precedence:
+    1) Value given as command line argument
+    2) Value from environment setting: ZEPHYR_BASE
+    3) Value of zephyr.base setting in west config file
+
+    Order of precedence between 2) and 3) can be changed with the setting
+    zephyr.base-prefer.
+    zephyr.base-prefer takes the values 'env' and 'configfile'
+
+    If 2) and 3) has different values and zephyr.base-prefer is unset a warning
+    is printed to the user.'''
 
     if args.zephyr_base:
         # The command line --zephyr-base takes precedence over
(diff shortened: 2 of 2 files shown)


src/west/commands/project.py
```diff
@@ -74,6 +74,7 @@ def do_run(self, args, user_args):
             shutil.move(args.cache, project.abspath)
 
             _update_key(config, 'manifest', 'path', project.path)
+        _update_key(config, 'zephyr', 'zephyr-base', project.path)
```
carlescufi: I think this should probably be:

```
[zephyr]
base=/path/to/zephyr
```

src/west/main.py
```diff
@@ -374,38 +373,44 @@ def set_zephyr_base(args):
...
-        except MalformedConfig as e:
-            log.die('Parsing of manifest file failed during command',
-                    args.command, ':', *e.args)
-        for project in manifest.projects:
-            if project.path == 'zephyr':
-                zb = project.abspath
-                zb_origin = 'manifest file {}'.format(manifest.path)
-                break
+        zb_env = os.environ.get('ZEPHYR_BASE')
+        zb_precedence = config.config.get('zephyr', 'zephyr-base-config',
```
carlescufi: I think this should be:

```
[zephyr]
base-prefer=[env|config]
```

tejlmand: I think we should do full wording on `configfile`,  cause an environment setting may also be a configuration somewhere.
I'm ok with `env` although I also think it is nice to have full word, as it is not something user changes all the time.

src/west/commands/project.py
tejlmand: fine with me.
Will wait a little before changing, to give other a chance to comment also ;)

src/west/main.py
tejlmand: I actually has been in doubt myself on the best word to choose.
So i'm ok with this as well.

carlescufi: Or maybe just `autodetect=[env|configfile]`

tejlmand: Then I like your previous suggestion better.
Cause not much _auto detection_ when we specify how to choose which value has precedence ;)

src/west/commands/project.py
tejlmand: changed to: `base`
This also makes it much nicer where #215 goes in, as you will then write:
`west config zephyr.base` to fetch the value ;)

src/west/main.py
tejlmand: update to use:
`env` and `base-prefer`.

src/west/main.py
```diff
@@ -374,38 +373,44 @@ def set_zephyr_base(args):
         # At some point, we need a more flexible way to set environment
         # variables based on manifest contents, but this is good enough
         # to get started with and to ask for wider testing.
-        try:
-            manifest = Manifest.from_file()
-        except MalformedConfig as e:
-            log.die('Parsing of manifest file failed during command',
-                    args.command, ':', *e.args)
-        for project in manifest.projects:
```
mbolivar: why was this fallback removed? it's the right default

src/west/commands/project.py
```diff
@@ -74,6 +74,7 @@ def do_run(self, args, user_args):
             shutil.move(args.cache, project.abspath)
 
             _update_key(config, 'manifest', 'path', project.path)
+        _update_key(config, 'zephyr', 'base', project.path)
```
tejlmand: This should be updated to ensure the project refering to a `zephyr` project is used instead.

src/west/main.py
tejlmand: Because now `zephyr.base` is always written to `.west/config` so no need to try to detect it.

This further has the advantage that users can now decide to have a copy or side zephyr project next to the upstream zephyr and retarget `zephyr.base` in `.west/config`

As well as simplifying order of precedence.

For out-of-tree, it will be possible to point out a default `zephyr.base` in `~/.westconfig`

For the future when #215 goes in, it will also be possible to extend `CMake` to use `west config zephyr.base` if `ZEPHYR_BASE` is not set in environment, and thus allow user to invoke cmake in a west handled project without having to set `ZEPHYR_BASE`

src/west/commands/project.py
mbolivar: Indeed; this needs to loop over the projects to find the project with path `zephyr`.

Also, shouldn't this be an abspath?

src/west/commands/project.py
```diff
@@ -850,6 +851,12 @@ def _update_west(rebase, keep_descendants):
         old_sha = _sha(project, 'HEAD')
         _update(project, rebase, keep_descendants)
 
+        if config.get('zephyr', 'zephyr-base', fallback=None) is None:
```
mbolivar: I'm confused about why this is in `_update_west()`; it seems like a strange place to put it. What am I missing?

I would have imagined that this could be done in `set_zephyr_base()` in main.py. That way the configuration option, if missing, would get set before a `west build` needs it, for example.

src/west/main.py
mbolivar: OK, I see now that the check was moved to `_update_west()`; sorry for missing that. But I still think the loop and setting of the config option belongs here; why is `_update_west()` a better place for it?

src/west/main.py
```diff
@@ -374,38 +373,44 @@ def set_zephyr_base(args):
...
-        # point to some random place. In practice, this is probably
-        # happening because zephyr-env.sh/cmd was run in some other
-        # zephyr installation, and the user forgot about that.
-        log.wrn('ZEPHYR_BASE={}'.format(zb_env),
-                'in the calling environment, but has been set to',
-                zb, 'instead by the', zb_origin)
-
-    os.environ['ZEPHYR_BASE'] = zb
+            log.err('no --zephyr-base given, ZEPHYR_BASE is unset,',
+                    'and west config contains no zephyr-base setting')
```
mbolivar: shouldn't this be `zephyr.base`?

src/west/main.py
tejlmand: To avoid `hidden` ways of fetching `zephyr.base` then we always write it during `west init`.
That means, a user can always fetch the value using `west config zephyr.base`, and each west managed project has it's own value.

If a user wants to change `zephyr.base`, as example for some local testing, he can then overwrite `zephyr.base` using environment setting, or do `west config zephyr.base <alternative_zephyr>`

As `zephyr.base` is now added during `west init` and must always be present, then the check you refer to is no longer needed.

Unfortunately, there are `west` managed projects that is not having this setting in their config file.
Thus, having the test in the `selfupdate` part is to ensure, that when user update to this `west` revision (or never) then this setting will be added to the conf file.

src/west/commands/project.py
tejlmand: as long as this is inside `west` managed project, this does not matter.
But if user add `zephyr.base` to ~/.westconfig for out of tree (not supported yet), then he would need to specify abs path.

tejlmand: but abs path would still be correct as well, so if abs path is preferred, i will change.

src/west/commands/project.py
tejlmand: as described here: https://github.com/zephyrproject-rtos/west/pull/211#discussion_r264314187
in future, this should not be needed to test, as the value should always be there, but for time being, there are installations out there, where the value is not present, and for those we can simply set the value during part of `selfupdate` and then in future remove this code.

src/west/main.py
tejlmand: fixed

src/west/commands/project.py
tejlmand: but then again.
`manifest.path`  is not an abs path.

tejlmand: now looping.
Actually, it feels kind of silly to loop in order to find the project that has `path == zephyr` cause the resulting setting in that case will actually become `zephyr.base == zephyr` !!!

On the other  hand, it might be valuable in future if deciding to have other way of specifying the `zephyr` project besides the `path`

src/west/main.py
```diff
@@ -374,38 +373,44 @@ def set_zephyr_base(args):
         # At some point, we need a more flexible way to set environment
```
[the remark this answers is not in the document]
tejlmand: done

src/west/main.py
```diff
@@ -359,7 +359,6 @@ def set_zephyr_base(args):
     '''Ensure ZEPHYR_BASE is set, emitting warnings if that's not
     possible, or if the user is pointing it somewhere different than
```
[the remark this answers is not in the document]
tejlmand: done

src/west/main.py
mbolivar: Still confused. It seems like `set_zephyr_base()` can check if the config option is defined when trying to fall back on it, and loop over the projects to set it if undefined. That will keep commands like `west build` even if the user "manually" upgrades west with `git pull` for example. It won't break anything if the user manually sets ZEPHYR_BASE or runs `west config` to set `zephyr.base` either. And that can still be reverted later.

src/west/commands/project.py
mbolivar: Resolving this thread since we have another one open for it

src/west/main.py
tejlmand: yeah, as a final fallback, we could have it, I take your point, although I think it is only us who does manually updates `west`.
I will add an additional commit, similar to the `ConfigParser` fallback commit, then we can decide if changing it in the future.

src/west/commands/project.py
tejlmand: I took a second look at the code, and instead of requesting abspath here, then I will ensure abspath later.

tejlmand:
@mbolivar can you take an extra look and see if you're satisfied.

Major parts.
Additional commit with fall-back as asked for.
Using abspath, but not in config file, instead abspath is used when setting environment in `west`, as this allows users to simply specify path relative to the projects, e.g. `west config zephyr.base zephyr-alternative` and not having to type full path. (Ot they can update `.west/config` by hand)

mbolivar:
Thanks!

PR #222, opened 2019-03-05
bootstrap: Use normpath() to avoid initializing in wrong folder

carlescufi:
On Windows 10, with Python 3.7.1, the following was not working:

west init -l zephyr/
or
west init -l zephyr\

because of the trailing (back)slash being interpreted as an additional
folder and west initializing itself inside the zephyr folder instead of
doing it in the enclosing one.

By normalizing the path we avoid this issue entirely.
Note that this is actually due to a regression that was fixed:
https://bugs.python.org/issue31047


carlescufi:
CC @junqingzou

Issue #201, opened 2019-02-13
west init leaves a branch master in west checked out


mbolivar:
This means the first `west update` issues a warning:

```
WARNING: left behind west branch "master"; to fast forward back, use: git -C .west/west checkout master
```

Issue #191, opened 2019-02-08
Cleanup responsibilities in manifest.py / project.py


tejlmand:
As discussed in https://github.com/zephyrproject-rtos/west/pull/189#discussion_r254714582

the current code is too mixed with no clear responsibilities between `git`, `project objects`, and `manifest writing`.

Discussions on best architecture should be taken and code cleaned up accordingly.

PR #224, opened 2019-03-06
West version checking (Zephyr issue 13209)

mbolivar:
This is the west side of the story in checking if the current west version is up to date.

We also ought to freeze the west version in the Zephyr repository's manifest file for it to work reliably.

### DIFF
--- a/src/west/commands/command.py
+++ b/src/west/commands/command.py
@@ -13,6 +13,7 @@ import sys
 import pykwalify
 import yaml
 
+from west import log
 from west.manifest import Manifest
 from west.util import escapes_directory
 
@@ -44,7 +45,8 @@ class WestCommand(ABC):
 
     All top-level commands supported by west implement this interface.'''
 
-    def __init__(self, name, help, description, accepts_unknown_args=False):
+    def __init__(self, name, help, description, accepts_unknown_args=False,
+                 check_up_to_date=True):
         '''Create a command instance.
 
         Some of the fields in a WestCommand (such as `name`, `help`, and
@@ -62,11 +64,14 @@ class WestCommand(ABC):
                                      in its run() method. Otherwise, passing
                                      unknown arguments will cause
                                      UnknownArgumentsError to be raised.
+        :param check_up_to_date: if true, run() calls check_up_to_date()
+                                 before calling do_run().
         '''
         self.name = name
         self.help = help
         self.description = description
         self._accept_unknown = accepts_unknown_args
+        self._check_up_to_date = check_up_to_date
 
     def run(self, args, unknown):
         '''Run the command.
@@ -82,6 +87,10 @@ class WestCommand(ABC):
         '''
         if unknown and not self._accept_unknown:
             self.parser.error('unexpected arguments: {}'.format(unknown))
+
+        if self._check_up_to_date:
+            self.check_up_to_date()
+
         self.do_run(args, unknown)
 
     def add_parser(self, parser_adder):
@@ -99,6 +108,22 @@ class WestCommand(ABC):
             raise ValueError('no parser was returned')
 
         return self.parser
+
+    def check_up_to_date(self):
+        '''Checks if west itself is up to date, emitting a warning if not.'''
+        try:
+            manifest = Manifest.from_file(sections=['west'])
+            west_project = manifest.west_project
+            if not west_project.is_up_to_date():
+                log.wrn(west_project.format(
+                    '{name_and_path} is at {wsh}, '
+                    'which is not up to date with manifest revision {msh}; '
+                    'consider a "west selfupdate"',
+                    wsh=west_project.sha('HEAD'),
+                    msh=west_project.revision))
+        except:                 # noqa: E722
+            log.err("couldn't check if west is up to date;",
+                    "this installation may be damaged")
 
     #
     # Mandatory subclass hooks
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -35,7 +35,8 @@ class PostInit(WestCommand):
 
             Continue the initialization of the project containing the manifest
             file. You should never need to call this.
-            '''))
+            '''),
+            check_up_to_date=False)
 
     def do_add_parser(self, parser_adder):
         return _add_parser(
@@ -302,7 +303,8 @@ class Update(WestCommand):
 
             This command does not change the contents of the manifest
             repository.
-            '''))
+            '''),
+            check_up_to_date=False)  # we'll check ourselves if necessary
 
     def do_add_parser(self, parser_adder):
         return _add_parser(parser_adder, self,
@@ -333,6 +335,8 @@ class Update(WestCommand):
     def do_run(self, args, user_args):
         if args.update:
             _update_west(args.rebase, args.keep_descendants)
+        else:
+            self.check_up_to_date()
 
         failed_rebases = []
 
@@ -368,7 +372,8 @@ class SelfUpdate(WestCommand):
             There is normally no need to run this command manually, because
             'west update' automatically updates the West repository to the
             latest version before doing anything else.
-            '''))
+            '''),
+            check_up_to_date=False)  # no need, we're about to update west
 
     def do_add_parser(self, parser_adder):
         return _add_parser(parser_adder, self,
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -19,14 +19,11 @@ from west import log
 from west import util
 from west.commands import WestCommand, CommandError
 from west.manifest import Manifest, MalformedManifest, MANIFEST_PROJECT_INDEX
+from west.manifest import MANIFEST_REV_BRANCH as MANIFEST_REV
+from west.manifest import QUAL_MANIFEST_REV_BRANCH as QUAL_MANIFEST_REV
 from urllib.parse import urlparse
 import posixpath
 
-
-# Branch that points to the revision specified in the manifest (which might be
-# an SHA). Local branches created with 'west branch' are set to track this
-# branch.
-_MANIFEST_REV_BRANCH = 'manifest-rev'
 
 class PostInit(WestCommand):
     def __init__(self):
@@ -77,7 +74,7 @@ class PostInit(WestCommand):
                     os.path.join(util.west_topdir(), project.path))
                 project.name = project.path
 
-            _inf(project, 'Creating repository for {name_and_path}')
+            _banner(project.format('Creating repository for {name_and_path}'))
             shutil.move(args.cache, project.abspath)
 
             _update_key(config, 'manifest', 'path', project.path)
@@ -206,32 +203,7 @@ class ManifestCommand(WestCommand):
     def do_run(self, args, user_args):
         # We assume --freeze here. Future extensions to the group
         # --freeze is part of can move this code around.
-
-        m = Manifest.from_file()
-
-        # Build a 'frozen' representation of all projects, except the
-        # manifest project.
-        projects = list(m.projects)
-        del projects[MANIFEST_PROJECT_INDEX]
-        frozen_projects = []
-        for project in projects:
-            sha = _sha(project, '{qual_manifest_rev_branch}')
-            d = project.as_dict()
-            d['revision'] = sha
-            frozen_projects.append(d)
-
-        # We include the defaults value here even though all projects
-        # are fully specified in order to make the resulting manifest
-        # easy to extend by users who want to reuse the defaults.
-        o = collections.OrderedDict()
-        o['west'] = m.west_project.as_dict()
-        o['west']['revision'] = _sha(m.west_project,
-                                     '{qual_manifest_rev_branch}')
-        o['manifest'] = collections.OrderedDict()
-        o['manifest']['defaults'] = m.defaults.as_dict()
-        o['manifest']['remotes'] = [r.as_dict() for r in m.remotes]
-        o['manifest']['projects'] = frozen_projects
-        o['manifest']['self'] = m.projects[MANIFEST_PROJECT_INDEX].as_dict()
+        frozen = Manifest.from_file().as_frozen_dict()
 
         # This is a destructive operation, so it's done here to avoid
         # impacting code which doesn't expect this representer to be
@@ -240,9 +212,9 @@ class ManifestCommand(WestCommand):
 
         if args.out:
             with open(args.out, 'w') as f:
-                yaml.safe_dump(o, default_flow_style=False, stream=f)
+                yaml.safe_dump(frozen, default_flow_style=False, stream=f)
         else:
-            yaml.safe_dump(o, default_flow_style=False, stream=sys.stdout)
+            yaml.safe_dump(frozen, default_flow_style=False, stream=sys.stdout)
 
     def _rep(self, dumper, value):
         # See https://yaml.org/type/map.html for details on the tag.
@@ -292,7 +264,7 @@ class Status(WestCommand):
 
     def do_run(self, args, user_args):
         for project in _cloned_projects(args):
-            _inf(project, 'status of {name_and_path}')
+            _banner(project.format('status of {name_and_path}'))
             _git(project, 'status', extra_args=user_args)
 
 
@@ -369,7 +341,7 @@ class Update(WestCommand):
             returncode = _update(project, args.rebase, args.keep_descendants)
             if returncode:
                 failed_rebases.append(project)
-                _err(project, '{name_and_path} failed to rebase')
+                log.err(project.format('{name_and_path} failed to rebase'))
 
         if failed_rebases:
             # Avoid printing this message if exactly one project
@@ -378,7 +350,7 @@ class Update(WestCommand):
                 log.err(('The following project{} failed to rebase; '
                         'see above for details: {}').format(
                             's' if len(failed_rebases) > 1 else '',
-                            ', '.join(_expand_shorthands(p, '{name_and_path}')
+                            ', '.join(p.format('{name_and_path}')
                                       for p in failed_rebases)))
             raise CommandError(1)
 
@@ -455,8 +427,8 @@ class ForAll(WestCommand):
 
     def do_run(self, args, user_args):
         for project in _cloned_projects(args):
-            _inf(project, "Running '{}' in {{name_and_path}}"
-                          .format(args.command))
+            _banner(project.format("Running '{cmd}' in {name_and_path}",
+                                   cmd=args.command))
 
             subprocess.Popen(args.command, shell=True, cwd=project.abspath) \
                 .wait()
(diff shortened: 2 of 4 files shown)


src/west/commands/command.py
```diff
@@ -82,6 +83,20 @@ def run(self, args, unknown):
         '''
         if unknown and not self._accept_unknown:
             self.parser.error('unexpected arguments: {}'.format(unknown))
+
+        try:
+            manifest = Manifest.from_file(sections=['west'])
```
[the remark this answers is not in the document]
mbolivar: It's safe for postinit; this is the flow:

1. bootstrapper clones west at the manifest revision
2. bootstrapper calls post-init
3. main calls run()
4. west is up to date by definition (we just cloned it and nothing else has happened)

I think though that it should *not* run for commands like `update` and `selfupdate`. Let me add a constructor kwarg to WestCommand which can be used to disable this check and use it in these commands.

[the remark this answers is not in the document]
mbolivar: Hm, I was wrong about that... let me do a bit more updating of this PR

mbolivar:
@carlescufi @tejlmand OK, I've fixed this up and tacked on another commit that cleans up some of the logging output.

I think we ought to merge this first for this west change to work properly: https://github.com/zephyrproject-rtos/zephyr/pull/14170

src/west/commands/project.py
```diff
@@ -744,16 +739,15 @@ def _west_project():
...
         project = _west_project()
-        log.dbg(project.format('Updating {name_and_path}'))
-
         old_sha = _sha(project, 'HEAD')
         _update(project, rebase, keep_descendants)
 
         if old_sha != _sha(project, 'HEAD'):
-            log.inf(project.format(
-                'Updated {name_and_path} to {revision} (from {url}).'))
+            _msg(project.format(
```
mbolivar: We could change `_msg()` signature to `_msg(project, fmt, *args, **kwargs)` and prepend `'{name}: '` to `fmt` if we wanted.

Then calls would look like `_msg(project, 'updated to {revision} (from {url})')`.

I didn't think it was worth it though, but am open to the change.

carlescufi: I like the  proposed change in terms of calls to it would look like, but we can do this in a follow-up PR.

tejlmand:
Just did a quick test, and specified `west` version as:
```
west:
  revision: pull/224/head
```
then did a `west selfupdate` to fetch this PR.

Now, if I run a `west` command, then it tells not a valid object as weel as `west` is out of date:
```
$ west list
fatal: Not a valid object name pull/224/head
WARNING: west (.west/west): git failed with exit code 128; treating as if "pull/224/head" is not an ancestor of "HEAD"
WARNING: west (.west/west) is at 9436a2397d5e1e7a31e8d02c59fed6b422bd91ed, which is not up to date with manifest revision pull/224/head; consider a "west selfupdate"
....
```

src/west/manifest.py
```diff
@@ -436,6 +482,140 @@ def as_dict(self):
...
+
+    def _format_kwargs(self, kwargs):
+        ret = {s: getattr(self, s) for s in self.__slots__}
+        ret['name_and_path'] = '{} ({})'.format(self.name, self.path)
+        ret['remote_name'] = ('None' if self.remote is None
+                              else self.remote.name)
+        ret.update(kwargs)
+        return ret
+
+    def git(self, cmd, extra_args=(), capture_stdout=False, check=True,
```
tejlmand: Just picked a line here.

I thought we wanted to avoid having `git` knowledge in manifest.py.
https://github.com/zephyrproject-rtos/west/pull/189#issuecomment-461498125

If starting to refactor, then maybe then maybe we should take up #191 in order to place `git` handling correctly ;)

tejlmand:
Same issue, if pointing to a branch:
```
west:
  revision: master
```
gives:
```
$ west list
fatal: Not a valid object name master
WARNING: west (.west/west): git failed with exit code 128; treating as if "master" is not an ancestor of "HEAD"
WARNING: west (.west/west) is at 9436a2397d5e1e7a31e8d02c59fed6b422bd91ed, which is not up to date with manifest revision master; consider a "west selfupdate"
```

tejlmand:
btw. thanks for the PR.
Gonna make people happy :)

src/west/commands/project.py
mbolivar: Sure, that's fine.

src/west/manifest.py
mbolivar: > I thought we wanted to avoid having `git` knowledge in manifest.py.
> [#189 (comment)](https://github.com/zephyrproject-rtos/west/pull/189#issuecomment-461498125)

I wasn't against doing it, just against doing an unnecessary refactor during the stabilization period. It is necessary now in order to implement this bug fix, which has been marked high priority for the next release.

> If starting to refactor, then maybe then maybe we should take up #191 in order to place `git` handling correctly ;)

Please take a look at the commit message for the patch:

```
    More clean up for manifest.py / project.py responsibilities
    
    This commit is further progress on #191, but doesn't fix it entirely.
    
    West's design has always been to split code into reusable libraries
    and commands which wrap them. The extra effort is justified since West
    is pluggable by extension commands, which often benefit from being
    able to use the library pieces.
    
    Although the code in project.py started out self-contained, it's
    gradually being refactored to fit the design.
    
    Continue that by factoring two more features into manifest.py: helpers
    for running git commands in a project and formatting strings using
    project attributes.
    
    This lets us remove almost all of the wrappers around west.log in a
    way that makes the code easier to read. The remaining _inf() is
    renamed _banner() to make its purpose clearer.
    
    Beyond being a useful refactoring which makes extensions easier to
    write, this is needed to fix a bug where west doesn't check if the
    currently running version is not a descendant of the manifest
    revision, indicating it is out of date and prompting a warning. Fixing
    that is a high priority issue for Zephyr v1.14.
    
    Delete a fair bit of dead code left over from the older multirepo
    commands which were deleted from v0.4.0 while we are here.
```

I more cleanup is eventually warranted (I only factored out a few nondestructive git based operations for now) but I didn't want to do that without good reason outside of a merge window.

mbolivar:
Thanks for the review!

> Need to fix the issue with getting a warning if specifying branch or a ref, such as pull/<no>/head

As you discovered, I intend this feature to work with revisions, which is why I wrote we should merge https://github.com/zephyrproject-rtos/zephyr/pull/14170 first.

I think these are two separate cases:

1. As far as I know, the `pull/<no>/head` namespace is "magic" in that it isn't mirrored into the local repository by a `git fetch`. I don't think we can test refs like these without actively fetching the given west revision from the remote every time west is invoked, which we cannot do (beyond the performance hit, offline operation for commands like build, list, etc. is required). Furthermore, these will never appear as the revision for west in normal use. I suggest we live with the warnings here as they are fairly clear about what is going on.

2. For branches, this does indeed appear to be a problem -- if the branch name is different than the one that was originally specified in the manifest. This is related to https://github.com/zephyrproject-rtos/west/issues/201. In my testing, with some branch specified for west in the manifest, west init + west update leaves a local branch with the same name in the west checkout, and commands like `west list` showed no warning. With a different branch name `foo` the issue happens, even though `origin/foo` is in the local ref space for `origin`.

To fix the second issue I think we need to keep a special remote name in the `west` project, like `manifest-remote`, that we can use to create a `manifest-remote/<branch-name>` qualified ref when checking up to dateness. Otherwise we cannot trust that `origin` for example has not been pointed somewhere else by the user, e.g. to their private fork.

OK with you?

> Furthermore, I think user should be informed if west revision doesn't match, not only if not descendant.
>
> If using having a west version check then it should ensure correct west revision for a given manifest.
> Even the west revision will be checked out means going back to older version.

I can see advantages and disadvantages to either approach. Having an exact match is nice if older versions of west were exactly required for older versions of Zephyr. But any descendant should work as well as long as we are not breaking backwards compatibility, which we definitely should not do during the stabilization period. And after the merge window opens we are going to eliminate `.west/west` at all, so the west "version" will likely be specified as a comparison against a semantic version like `>=0.6.0` or so.

So I am not sure if an exact match is really *necessary*, and since later versions of west will be both 1. backwards compatible and 2. have nice features like the cleaned up `_msg()` output in the last patch in this series, I think we may be warning users unnecessarily in ways that give them inconsistent versions of west that don't work as well as later versions.

@carlescufi, any opinions about this question (warn if west version is a mismatch at all vs. warn if current version is not a descendant)?

tejlmand:
>     1. As far as I know, the `pull/<no>/head` namespace is "magic" in that it isn't mirrored into the local repository by a `git fetch`.

I'm not saying we should try to detect whether `west` is up-to-date when a special ref is given, such as `pull/<no>/head`.
What I suggest is simply not to test in this case.
If revision is something different from a SHA, then not try to test.
And thus avoid any kind of warnings or errors

> 2. For branches, this does indeed appear to be a problem -- if the branch name is different than the one that was originally specified in the manifest.

A branch is never stable in its revision, thus we can not determine if we look at the right one.
In this case user must determine by himself, when he needs / wants to do `west selfupdate`
All i'm asking, is that we do not print warnings to him.

Simply disaable the `west` revision test in those two cases.

tejlmand:
> But any descendant should work as well as long as we are not breaking backwards compatibility, which we definitely should not do during the stabilization period. 

Exactly, as long as we are not breaking backwards compatibility. But as we are not really testing this, then there is a risk that suddenly things can break. As example, cmake uses `west list` and as we are moving forward, we could suddenly end in a situation where such changes goes unnoticed.
I know we are careful, but things happens.

An alternative could be, that if we are behind, we issue current warning (as this could be critical).
but if we are having a too new  west version, we might just print a small info message, so that people are aware. (i'm not sure how often people are going back in time anyway.)

carlescufi:
@mbolivar @tejlmand 
I think I am aligned with @tejlmand 's proposal of a small info message for the case of the west version being more recent than the currently specified one. But subject to the following: if the manifest specifies a SHA and the current west revision is another SHA (i.e. not a branch) that is actually more recent than the one specified by the manifest, we check out the exact SHA in the manifest. That way regular users not modifying west in any way will actually never see this info message.
Thoughts?

src/west/manifest.py
tejlmand: > Please take a look at the commit message for the patch:

whoops. I see you do refer to #191, my bad :(

I was just under the impression you prefer to have `git` knowledge outside of `manifest.py` module, which is something I would agree with.

From: https://github.com/zephyrproject-rtos/west/pull/189#issuecomment-461498125
> If I understand what you're saying, then a problem with that approach is that that would require the manifest class to know how to run Git to freeze the revisions. That felt like an abstraction violation to me, which is why I chose to do it this way. 

So follow up question, do you think we should factor out the `git` operations into independent module ?
So that they can be imported by manifest.py and project.py ?

mbolivar:
> Simply disaable the west revision test in those two cases.

I'm starting to think we should just use `west --version` output in CMake and call it a day to keep this future-proof. Doing case analysis on whether the manifest revision is a SHA or not would just be something we would delete after v1.14 when we remove the copy of west in the installation.

> I think I am aligned with @tejlmand 's proposal of a small info message for the case of the west version being more recent than the currently specified one. But subject to the following: if the manifest specifies a SHA and the current west revision is another SHA (i.e. not a branch) that is actually more recent than the one specified by the manifest, we check out the exact SHA in the manifest. That way regular users not modifying west in any way will actually never see this info message.

@tejlmand please correct me if I am wrong but a "small info message for the case of the west version being more recent than the currently specified one" is not what you are suggesting; is it?

src/west/manifest.py
mbolivar: Looking at the code, I think it reads pretty naturally when you can do `project.git_related_operation()`, so I'm actually thinking it's nice to have it in `manifest.py`.

carlescufi:
> @tejlmand please correct me if I am wrong but a "small info message for the case of the west version being more recent than the currently specified one" is not what you are suggesting; is it?

Well, @tejlmand wrote "I think user should be informed if west revision doesn't match, not only if not descendant". As I see this means that you output a message whenever the revision is a descendant but doesn't match the exact one in the manifest, which is what I meant. 

>I'm starting to think we should just use west --version output in CMake and call it a day to keep this future-proof. Doing case analysis on whether the manifest revision is a SHA or not would just be something we would delete after v1.14 when we remove the copy of west in the installation.

I'm starting to think the same, we are going down a road that I am beginning to dislike, with all sort of special casing here and there to decide what a valid version means for the user when in fact this is all temporary until we move to standard semantic versioning in PyPi for the whole package. But we need a way to fix the Zephyr issue in the meantime too, maybe we can have a quick call today.

src/west/commands/command.py
```diff
@@ -100,6 +109,22 @@ def add_parser(self, parser_adder):
 
         return self.parser
 
+    def check_up_to_date(self):
+        '''Checks if west itself is up to date, emitting a warning if not.'''
+        try:
+            manifest = Manifest.from_file(sections=['west'])
+            west_project = manifest.west_project
```
tejlmand: Suggested change to handle the only support SHA issue.

Goes together with code suggestion below.
```suggestion
            west_project = manifest.west_project
            int(west_project.revision, 16)
```

src/west/commands/command.py
```diff
@@ -100,6 +109,22 @@ def add_parser(self, parser_adder):
...
+            if not west_project.is_up_to_date():
+                log.wrn(west_project.format(
+                    '{name_and_path} is at {wsh}, '
+                    'which is not up to date with manifest revision {msh}; '
+                    'consider a "west selfupdate"',
+                    wsh=west_project.sha('HEAD'),
+                    msh=west_project.revision))
+        except:                 # noqa: E722
```
tejlmand: Part two of SHA fix.
```suggestion
        except ValueError:
            # Revision is not a SHA, ignore.
            pass
        except:                 # noqa: E722
```

tejlmand:
> Well, @tejlmand wrote "I think user should be informed if west revision doesn't match, not only if not descendant". As I see this means that you output a message whenever the revision is a descendant but doesn't match the exact one in the manifest, which is what I meant.

yep, my thought was something like:
```
            if not west_project.is_up_to_date():
                log.wrn(west_project.format(<same message as today>))
            elif west_project.revision != west_project.sha('HEAD'):
                log.inf(west_project.format('{name_and_path} is newer than manifest revision.'))
```
and I tried this out, and after doing so I would like to revoke my suggestion.
The reason for that is because, `log.wrn` prints to `stderr` which means it does not interfere with `stdout` as used by `CMake` and `west list`.
Whereas, doing this:
```
                log.inf(west_project.format('{name_and_path} is newer than manifest revision.'))
```
will print to `stdout` and then mess up the output for scripts relying on `west list`
(and I never intended this message to be printed as a warning).

But I still think we should only check for SHA, and I added code snippets for that here:
https://github.com/zephyrproject-rtos/west/pull/224#discussion_r264255934
https://github.com/zephyrproject-rtos/west/pull/224#discussion_r264257340

mbolivar:
> But I still think we should only check for SHA, and I added code snippets for that here:

What about the future-proofing issue? Sorry, I am confused.

tejlmand:
> > But I still think we should only check for SHA, and I added code snippets for that here:
> 
> What about the future-proofing issue? Sorry, I am confused.

Sorry for the confusion.
I was just referring to 
a) suggestion of having a message printed if having a `west` that is an ancestor of current revision. Which is no longer suggest, see https://github.com/zephyrproject-rtos/west/pull/224#issuecomment-471574766
b) Only check versions if a SHA is specified in `west.yml`

so to answer your suggestion of using `west --version`, then what would CMake be using, and how would it know what version.
I assume manifest is still specifying a SHA revision.
But a `west --version` might return:
```
$ west --version
West bootstrapper version: v0.5.4 (/projects/<path>/_bootstrap)
West repository version: v0.5.4-8-g1b7bfb2 (/projects/<path>/.west/west)
```

tejlmand:
@mbolivar  If `west --version` can achieve what we want already, that sounds like perfect solution, I just don't see how, so if you can provide some sample code, that would be great.

mbolivar:
> so to answer your suggestion of using west --version, then what would CMake be using, and how would it know what version ?

We would write the minimum bootstrapper and west versions into CMake the same way we do the SDK version, and parse the `west --version` output to get and compare them.

tejlmand:
> > so to answer your suggestion of using west --version, then what would CMake be using, and how would it know what version ?
> 
> We would write the minimum bootstrapper and west versions into CMake the same way we do the SDK version, and parse the `west --version` output to get and compare them.

But if then, what if minimum version is simply specifying `v0.5.4` as now, but user's needs at least the 4th commit after `v0.5.4`, then that revision would look like:
```
$ west --version
West bootstrapper version: v0.5.4 (/projects/<path>/_bootstrap)
West repository version: v0.5.4-4-<short-sha> (/projects/<path>/.west/west)
```
how will that be specified in `west.yml` ?
And will it only be during `CMake` invocation people will be notified about `west selfupdate` ?

Just trying to imagine your proposal :)
So far it sounds like a good approach :+1:

src/west/manifest.py
tejlmand: ok.
not so worreid here.

But maybe at later time we should consider a `git` module, so that we avoid duplicating same `git` commands in multiple modules in future.

tejlmand:
@mbolivar or do you actually intent to keep the version specified to cmake completely independent of what is written in `west.yml` ?
So that `west.yml` is still specifying SHA, branch, etc.

But CMake simply knows just `v0.5.4-n` and checks if that is up-to-date.
If so, that sounds like a good approach.

tejlmand:
something like:
```
set(REQUIRED_WEST_VER 0.5.4-n)
...
if(${REQUIRED_WEST_VER} VERSION_GREATER ${WEST_VERSION})
   <fail>
endif()
```

mbolivar:
@tejlmand @carlescufi please see https://github.com/zephyrproject-rtos/zephyr/pull/14276 for a `west --version` based fix

For now, if the user needs the fourth commit or whatever, we need to cut a new release on PyPI to make it happen. This is slightly inconvenient until v1.14 is released, but the final release manifest will have west version tagged at whatever goes with Zephyr v1.14. And it  will be future-proof once west is obtained entirely from PyPI.

The question remains what to do about this pull request. I think the first and third patches are worth keeping, maybe even for v1.14 now that the work is done. Thoughts?

tejlmand:
> @tejlmand @carlescufi please see [zephyrproject-rtos/zephyr#14276](https://github.com/zephyrproject-rtos/zephyr/pull/14276) for a `west --version` based fix
> 

Great work on https://github.com/zephyrproject-rtos/zephyr/pull/14276 .
And I see it is already in.

> For now, if the user needs the fourth commit or whatever, we need to cut a new release on PyPI to make it happen. This is slightly inconvenient until v1.14 is released, but the final release manifest will have west version tagged at whatever goes with Zephyr v1.14. And it will be future-proof once west is obtained entirely from PyPI.
> 

Agree, and I also think we can live with that.

> The question remains what to do about this pull request. I think the first and third patches are worth keeping, maybe even for v1.14 now that the work is done. Thoughts?

Agree, let's get the other changes in.

mbolivar:
> Agree, let's get the other changes in.

Will do. I would actually argue these changes fix a bug which I have filed as #229. Will mark that.

mbolivar:
Reopening as https://github.com/zephyrproject-rtos/west/pull/230

PR #226, opened 2019-03-10
main.py: fix extra newline in west --help

mbolivar:
Remove trailing whitespace from the format_help() return value. This prevents an extra blank line from appearing in the west --help output.


mbolivar:
@carlescufi this fixes the issue you pointed out on my machine

PR #232, opened 2019-03-14
zephyr base: fixed west config zephyr.base when relative

### LINKED ISSUE #231: `west` incorrectly sets `ZEPHYR_BASE` based on local config file
As a result of https://github.com/zephyrproject-rtos/west/pull/211/, `west` incorrectly sets `ZEPHYR_BASE` based on the local config file in most cases (this is specifically caused by commit e17884f02f44b564919e5055b02f24ac32145137).

`zephyr.base` is set to the `path` in the manifest file for `zephyr` by `project.py` `PostInit` `do_run()`.

https://github.com/zephyrproject-rtos/west/blob/fdb634fd8f2bc017b9120d04e92876da305091e1/src/west/commands/project.py#L84-L86

 If you don't have `ZEPHYR_BASE` set and don't pass it in via the command line, `west` now checks the config file and tries to set `ZEPHYR_BASE` based on that config.

https://github.com/zephyrproject-rtos/west/blob/fdb634fd8f2bc017b9120d04e92876da305091e1/src/west/main.py#L416-L418

The value in the config (`zephyr` for the base `zephyr` project) is then passed to `os.path.abspath`, which gets the incorrect path for `ZEPHYR_BASE` depending on where you run it. 

https://github.com/zephyrproject-rtos/west/blob/fdb634fd8f2bc017b9120d04e92876da305091e1/src/west/main.py#L441-L442

For example, running it in `zephyr/samples/hello_world` results in `ZEPHYR_BASE` being set to `zephyr/samples/hello_world/zephyr`:

```
kasemq@kaer-morhen ‹ master › : ~work/z2/zephyr/samples/hello_world
[0] % west build -b nrf52840_pca10056 -s .
source directory: /home/kasemq/work/z2/zephyr/samples/hello_world
build directory: /home/kasemq/work/z2/zephyr/samples/hello_world/build (created)
BOARD: nrf52840_pca10056
CMake Error at CMakeLists.txt:3 (include):
  include could not find load file:

    /home/kasemq/work/z2/zephyr/samples/hello_world/zephyr/cmake/app/boilerplate.cmake
```

This can be reproduced with the following steps:

```
mkdir z2
cd z2
west init
west update
cd zephyr/samples/hello_world
west build -b nrf52840_pca10056 -s .
```

It does work correctly if you run the build from the `west` top directory (that contains `.west`).

Not sure if the correct fix is to store the absolute path in the `west` config file or assume the `west` config file path is a relative from `west_topdir` within `set_zephyr_base()`

tejlmand:
This commit ensures that zephyr.base will be joined with west_topdir()
when a relative path is used in west config.

The previous solution was using os.path.abspath which would give
different results depending on current location when invoking west
command.

The advantage of allowing relative path in .west/config over requiring
an absolute path, is that it allows a user to rename a project's parent
folders without having to update zephyr.base.

Fixes #231 

### DIFF
--- a/src/west/main.py
+++ b/src/west/main.py
@@ -30,7 +30,7 @@ from west.commands.project import List, ManifestCommand, Diff, Status, \
     SelfUpdate, ForAll, WestUpdated, PostInit, Update
 from west.commands.config import Config
 from west.manifest import Manifest, MalformedConfig
-from west.util import quote_sh_list
+from west.util import quote_sh_list, west_topdir
 
 PROJECT_COMMANDS = {
     'commands for managing multiple git repositories': [
@@ -390,6 +390,8 @@ def set_zephyr_base(args):
         zb_prefer = config.config.get('zephyr', 'base-prefer',
                                       fallback=None)
         zb_config = config.config.get('zephyr', 'base', fallback=None)
+        if zb_config is not None:
+            zb_config = os.path.join(west_topdir(), zb_config)
 
         if zb_prefer == 'env' and zb_env is not None:
             zb = zb_env
@@ -439,7 +441,7 @@ def set_zephyr_base(args):
                         'manifest project has path "zephyr"')
 
     if zb is not None:
-        os.environ['ZEPHYR_BASE'] = os.path.abspath(zb)
+        os.environ['ZEPHYR_BASE'] = zb
 
     log.dbg('ZEPHYR_BASE={} (origin: {})'.format(zb, zb_origin))
 


src/west/main.py
```diff
@@ -390,6 +390,8 @@ def set_zephyr_base(args):
         zb_prefer = config.config.get('zephyr', 'base-prefer',
                                       fallback=None)
         zb_config = config.config.get('zephyr', 'base', fallback=None)
+        if not os.path.isabs(zb_config):
```
carlescufi: What is `os.path.isabs(None)` in case there's no config value for it?

carlescufi:
@QKaseman FYI

src/west/main.py
tejlmand: good catch, fixed.

src/west/main.py
```diff
@@ -390,6 +390,8 @@ def set_zephyr_base(args):
         zb_prefer = config.config.get('zephyr', 'base-prefer',
                                       fallback=None)
         zb_config = config.config.get('zephyr', 'base', fallback=None)
+        if zb_config is not None and not os.path.isabs(zb_config):
+            zb_config = os.path.join(west_topdir(), zb_config)
```
mbolivar: os.path.join(absolute_path1, absolute_path2) is just absolute_path2, so I think the check is unnecessary, no?

tejlmand: but it only joining if `zb_config` is NOT abs path:
```
not os.path.isabs(zb_config)
```

tejlmand: ahh, misread your comment
You mean `not os.path.isabs(zb_config)` can be removed

Will remove when I have verified, that it indeed works that way.

tejlmand: Fixed

PR #235, opened 2019-03-20
git: git fetch now uses -f

tejlmand:
It has been discovered that in the event that a different remote is
added to manifest file which contains same branches but cannot be
fast forwarded into the local refs space, then a west update will fail
with the error:
 ! [rejected] <branch> -> refs/west/<branch> (non-fast-forward)

This commit fixes this issue by introducing '-f' when fetching.

This commit has additional test for verifying the fix.

Fixes: #236 


mbolivar:
Can we get a bug for this? We are pretty far along in https://github.com/zephyrproject-rtos/zephyr/pull/14170, i.e. we have https://github.com/zephyrproject-rtos/ci-dockerfiles/pull/41 queued

PR #241, opened 2019-03-27
Docs ext fixes

mbolivar:


src/west/commands/west-commands-schema.yml
```diff
@@ -4,51 +4,12 @@
...
-# west-commands key's value is a sequence of "command descriptors".
-# Each command descriptor gives the location of a file containing west
-# commands, along with the names of the commands provided by the file
-# (and optionally the names of the classes which define them).
+# --------------------------------------------------------------
 #
-# Here is an example west-commands.yml file:
+#  DO NOT CHANGE THIS FILE WITHOUT UPDATING THE DOCUMENTATION!
 #
-# west-commands:
```
tejlmand: I still think the example and description in the schema file is useful.
If this has been moved to Zephyr docs, it's fine with me, but if not, I think this is still valid and useful to people who are gonna create extension commands.

I see no description of this kind in:
https://docs.zephyrproject.org/latest/guides/west/index.html
https://docs.zephyrproject.org/latest/guides/west/repo-tool.html
https://docs.zephyrproject.org/latest/guides/west/build-flash-debug.html
https://docs.zephyrproject.org/latest/guides/west/west-not-found.html#west-apis

Instead of removing, could we have this information placed elsewhere ?
(Sorry if I have missed a Zephyr PR, tried to look for something related to west / docs but didn't find any)

src/west/commands/west-commands-schema.yml
```diff
@@ -4,51 +4,12 @@
...
 
-# The top-level of the file is a map with a 'west-commands:' key.  The
-# west-commands key's value is a sequence of "command descriptors".
-# Each command descriptor gives the location of a file containing west
-# commands, along with the names of the commands provided by the file
-# (and optionally the names of the classes which define them).
+# --------------------------------------------------------------
 #
-# Here is an example west-commands.yml file:
+#  DO NOT CHANGE THIS FILE WITHOUT UPDATING THE DOCUMENTATION!
```
tejlmand: If someone comes in after us, and needs to maintain `west`, then I think a reference to the documentation would be better.
So instead of 
> `#  DO NOT CHANGE THIS FILE WITHOUT UPDATING THE DOCUMENTATION!`

to something like

> `#  DO NOT CHANGE THIS FILE WITHOUT UPDATING THE WEST DOCUMENTATION FOUND IN ZEPHYR !`
> `# <zephyr-repo>/doc/guides/west/repo-tool.rst`

As it might not be obviously exactly what docs to update.
Or do you actually mean some other docs ?

Although, I haven't found the place where we actually document the yml schema.
See also other comment.

src/west/manifest-schema.yml
```diff
@@ -9,54 +9,32 @@
 ##
 ## However, the features don't map 1:1.
 
-# The top-level manifest is a map. There may be multiple sections in the
-# manifest file. Each section can be validated by their own schema.
-# This schema validates the 'manifest' section.
+# --------------------------------------------------------------
+#
+#  DO NOT CHANGE THIS FILE WITHOUT UPDATING THE DOCUMENTATION!
```
tejlmand: The same comments given to: 
`src/west/commands/west-commands-schema.yml`
also applies to this schema file.

src/west/commands/west-commands-schema.yml
tejlmand: > We'll move this information into
the zephyr documentation.

I noticed this in the commit message, so I assume you're already working on that :)
So guess this answer my question, and you can ignore the comment.

mbolivar: You didn't miss anything -- these docs aren't upstream yet. As you guessed from the commit message, I am almost ready to submit this docs PR:

https://github.com/mbolivar/zephyr/tree/west-docs-v1.14

The only thing left is to document `west sign`, which I am working on now. I need this PR to go in first, though, or the API autodoc directives will break the build.

src/west/commands/west-commands-schema.yml
mbolivar: I did this once before, but then upstream Zephyr moved all the documentation around on me :(. So now I am hesitant to do this again, but I will give it another try.

There is a chicken and egg problem since the documentation is being added to Zephyr in files that do not exist yet in the branch that I linked in my other comment. Can we perhaps add the links after we have merged the documentation? That will be too late for west v0.5.7, but that's OK, because master is where changes to this file would happen.

tejlmand: > I did this once before, but then upstream Zephyr moved all the documentation around on me :(. So now I am hesitant to do this again, but I will give it another try.

I understand, but then may just refer to Zephyr docs in broad terms, so someone updating  this knows that the docs to update is not in this repo.
But let's add that, when making a new commit, as this is low prio.

src/west/commands/west-commands-schema.yml
tejlmand: Great, just as guessed :)

PR #248, opened 2019-04-09
West v0.5.8

mbolivar:
Two bootstrapper fixes:

- check for git with shutil and print a sensible message if it's not found and is needed
- print a better error message if west.main can't be imported


mbolivar:
Switched ModuleNotFoundError for ImportError to keep Python 3.4 and 3.5 working.

Issue #225, opened 2019-03-10
west self-update leads to duplicate output messages


carlescufi:
When west updates itself, it re-runs the same command with the newly updated west copy. But we should probably avoid  updating west itself on the second run, to avoid showing the user west being updated twice.

```
$ west update
=== Fetching changes for west (.west/west/)
From https://github.com/zephyrproject-rtos/west
 * branch            master     -> FETCH_HEAD
=== Checking out revision 'cc5a05fec57a6c98c06056a2665254d968930168' as detached HEAD in west (.west/west/)
HEAD is now at cc5a05f bootstrap: Use normpath() to avoid initializing in wrong folder
WARNING: left behind west branch "tox-macos"; to rebase onto the new HEAD: git -C .west/west rebase cc5a05fec57a6c98c06056a2665254d968930168 tox-macos
=== Updated west (.west/west/) to master (from https://github.com/zephyrproject-rtos/west).
=== Fetching changes for west (.west/west/)
From https://github.com/zephyrproject-rtos/west
 * branch            master     -> FETCH_HEAD
=== Checking out revision 'cc5a05fec57a6c98c06056a2665254d968930168' as detached HEAD in west (.west/west/)
HEAD is now at cc5a05f bootstrap: Use normpath() to avoid initializing in wrong folder
=== Fetching changes for net-tools (net-tools/)
=== Checking out revision '30b7efa827b04d2e47840716b0372737fe7d6c92' as detached HEAD in net-tools (net-tools/)
HEAD is now at 30b7efa net-setup: zeth-vlan.conf: Add IP addresses to zeth
```

Issue #207, opened 2019-02-20
Add support for "west help init"


carlescufi:
Since `west help` is processed by the main installation and `init` is a bootstrapper command, `west help init` does not work today. Once we merge bootstrapper and main installation this should be addressed so that we cover all commands with `west help`.

Issue #179, opened 2019-02-01
west selfupdate does not respect new west remote


mbolivar:
To reproduce, change west remote URL in the manifest repository (e.g. change from upstream to a personal fork) and run `west selfupdate`.

I expect this to update my west installation's remote URL (e.g. change to my personal fork), but it does not. For example, `git remote -v` is the same before and after.

Note that running `west -vv selfupdate` shows the correct remote (e.g. my fork) is being fetched from. So this is not a correctness issue for the command, just a minor issue with the remotes.

tejlmand:
This is not only applicable to `west` remote, but goes for all remotes in the manifest file.
So if changing current behavior, I believe this issue should then be extended to cover manifest projects as well.

There were some discussion on this on slack (unfortunately I can't go so far back).
I remember there were mixed opinions about handling of remotes.
Especially whether `west` should update remotes or not.

Some were strongly against `west` modifying their remotes.
As example, if new remotes are added / deleted, then `west` cannot know, if user had added a git remote manually. In which case, the user probably don't want that remote to be modified by `west`

Personally, I believe we should keep as-is, or decide to remove `west` configured remotes.
The remotes are not used by `west` anyway, and managing them using `west` has a risk of colliding with user controlled remotes.

PR #250, opened 2019-04-24
Unification

mbolivar:
This PR eliminates the bootstrapper. It is prep work for further API cleanup and then implementing the multiple manifest feature under discussion. After merging this, west will no longer self update.

Fixes: #225
Fixes: #179
Fixes: #207


src/west/commands/command.py
```diff
@@ -44,12 +44,20 @@ def __init__(self, **kwargs):
...
+'''
+
 class WestCommand(ABC):
     '''Abstract superclass for a west command.
 
     All top-level commands supported by west implement this interface.'''
 
-    def __init__(self, name, help, description, accepts_unknown_args=False):
+    def __init__(self, name, help, description, accepts_unknown_args=False,
+                 requires_installation=True):
```
carlescufi: Are we breaking the `WestCommand` API here?

src/west/commands/command.py
```diff
@@ -67,13 +75,19 @@ def __init__(self, name, help, description, accepts_unknown_args=False):
...
+                                      with an error.
         '''
         self.name = name
         self.help = help
         self.description = description
         self._accept_unknown = accepts_unknown_args
+        self.requires_installation = requires_installation
 
-    def run(self, args, unknown):
+    def run(self, args, unknown, topdir):
```
carlescufi: Same here, what happens to existing commands that don't implement this? Will they break? If so, how do we handle users that update to the latest west but then want to bisect master?

src/west/commands/command.py
```diff
@@ -84,9 +98,15 @@ def run(self, args, unknown):
...
                         this must be empty if the constructor
                         was passed accepts_unknown_args=False.
+        :param topdir: top level directory of west installation, or None;
+                       this is stored as an attribute before do_run() is
+                       called.
         '''
         if unknown and not self._accept_unknown:
             self.parser.error('unexpected arguments: {}'.format(unknown))
+        if not topdir and self.requires_installation:
+            sys.exit(_NO_TOPDIR_MSG_FMT.format(os.getcwd()))
```
carlescufi: Why not `log.die()` here?

src/west/commands/project.py
```diff
@@ -19,13 +20,211 @@
...
+            local_rev = True
+        except subprocess.CalledProcessError:
+            local_rev = False
+
+        if local_rev or is_sha:
+            subprocess.check_call(('git', 'checkout', rev), cwd=dest)
+        else:
+            subprocess.check_call(('git', 'checkout', 'FETCH_HEAD'), cwd=dest)
+
 class PostInit(WestCommand):
```
carlescufi: Why do we need this at all? Can't this now be moved to the Init() command?

src/west/commands/command.py
mbolivar: It's a backwards compatible change.

src/west/commands/command.py
mbolivar: Nothing breaks because the do_run() signature is not changed incompatibly, and that's what extensions implement.

src/west/commands/command.py
mbolivar: I'm preserving the existing behavior, which doesn't use colorama etc. Happy to change it to use log.die if you want.

src/west/commands/command.py
mbolivar: Well, extensions and built-in commands - but built-ins don't matter, because they are built in and can be kept up to date. There just aren't any changes to make.

src/west/commands/project.py
mbolivar: Bad rebase. i meant to remove it.

src/west/commands/command.py
carlescufi: Yes please, let's be consistent across the board

src/west/commands/project.py
carlescufi: Understood

mbolivar:
@carlescufi comments addressed.

The shippable error was a python 3.4 issue in `util: add require_program()`, which I've added a shim for.

mbolivar:
> The shippable error was a python 3.4 issue in util: add require_program(), which I've added a shim for.

Actually, on second thought, calling `log.die` was never the right thing since this could be used from library code. I'm going to remove that patch entirely and just inline the code where needed.

mbolivar:
>  I'm going to remove that patch entirely and just inline the code where needed.

@carlescufi done, please re-review when convenient.

src/west/commands/project.py
```diff
@@ -18,69 +20,213 @@
...
+from west.manifest import Manifest, MalformedManifest, MalformedConfig, \
+    MANIFEST_PROJECT_INDEX
 from west.manifest import MANIFEST_REV_BRANCH as MANIFEST_REV
 from west.manifest import QUAL_MANIFEST_REV_BRANCH as QUAL_MANIFEST_REV
 from urllib.parse import urlparse
 import posixpath
 
 
-class PostInit(WestCommand):
+class Init(WestCommand):
```
carlescufi: I wonder if this really belongs here. Maybe we should have a `commands/core.py` but let's leave this  for after

mbolivar: IMO it's okay here. This is the file where we keep the multi-repo commands.

PR #261, opened 2019-05-09
Non zephyr projects

### LINKED ISSUE #246: Support use cases other than Zephyr RTOS
West can be already used for other purposes than Zephyr development, even without sacrificing the parts of it that make it convenient for Zephyr.

This issue tracks first-class support for that. The main issue I'm aware of is this error:

```
ERROR: no --zephyr-base given, ZEPHYR_BASE is unset, west config contains no zephyr.base setting, and no manifest project has path "zephyr"
```

It would be nice to generalize the existing mechanism for setting environment variables based on the contents of the manifest so you could do something like this:

```
manifest:
  self:
    path: zephyr
    config-vars:
      - zephyr.base: {abspath}
```

Where the `config-vars` key in the manifest's `self` subsection (or any `project` element) would be a list of `config_variable: value` pairs, where each `value` is a format string like that passable to `west list --format`. These values would be set at `west update` time, as if with `west config --local config_variable $(west list --format="value" path_to_project)`, provided that they do not already have values.

M1cha:
With west you got quite a genius tool and zephyr shouldn't be the only use-case.
this adds support for setting config default values in the manifest and adds the `zephyr.disabled` config to make this possible.

fixes #246

### DIFF
--- a/src/west/main.py
+++ b/src/west/main.py
@@ -389,7 +389,8 @@ class WestApp:
         #   (controversial)
         # - make zephyr extensions that need ZEPHYR_BASE just set it
         #   themselves (easy if above is OK, unnecessary if it isn't)
-        set_zephyr_base(args, self.manifest, self.topdir)
+        if not config.config.getboolean("zephyr", "disabled", fallback=False):
+            set_zephyr_base(args, self.manifest, self.topdir)
 
         command.run(args, unknown, self.topdir, manifest=self.manifest)
 
--- a/src/west/main.py
+++ b/src/west/main.py
@@ -376,6 +376,8 @@ class WestApp:
         # Parse arguments again.
         args, unknown = west_parser.parse_known_args(argv)
 
+        load_manifest_configs(args)
+
         # HACK: try to set ZEPHYR_BASE.
         #
         # Currently required by zephyr extensions like "west build".
@@ -643,6 +645,27 @@ def mve_msg(mve, suggest_upgrade=True):
          'West version: v{}'.format(__version__)] +
         (['Manifest file: {}'.format(mve.file)] if mve.file else []) +
         (['Please upgrade west and retry.'] if suggest_upgrade else []))
+
+def load_manifest_configs(args):
+    try:
+        manifest = Manifest.from_file()
+
+        for section in manifest.configs:
+            if not config.config.has_section(section):
+                config.config.add_section(section)
+
+            for key in manifest.configs[section]:
+                value = manifest.configs[section][key]
+
+                if not config.config.has_option(section, key):
+                    config.config.set(section, key, str(value))
+    except MalformedConfig as e:
+        log.wrn("Can't load manifest configs:",
+                'parsing of manifest file failed during command',
+                args.command, ':', *e.args)
+    except WestNotFound:
+        log.wrn("Can't load manifest configs:",
+                'not currently in a west installation')
 
 def set_zephyr_base(args, manifest, topdir):
     '''Ensure ZEPHYR_BASE is set
--- a/src/west/manifest-schema.yml
+++ b/src/west/manifest-schema.yml
@@ -127,3 +127,8 @@ mapping:
       import:
         required: false
         type: any
+
+  configs:
+    required: false
+    type: map
+    allowempty: true
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -375,6 +375,9 @@ class Manifest:
 
         self.has_imports = False
 
+        self.configs = None
+        '''Map with default configuration values.'''
+
         # Set up the public attributes documented above, as well as
         # any internal attributes needed to implement the public API.
         self._importer = importer or _default_importer
@@ -473,6 +476,7 @@ class Manifest:
         r['manifest'] = collections.OrderedDict()
         r['manifest']['projects'] = frozen_projects
         r['manifest']['self'] = self.projects[MANIFEST_PROJECT_INDEX].as_dict()
+        r['manifest']['configs'] = self.configs
 
         return r
 
@@ -532,6 +536,8 @@ class Manifest:
                     # can have a topdir but no path, and thus no abspath.
                     continue
                 self._projects_by_cpath[util.canon_path(p.abspath)] = p
+
+        self.configs = manifest.get('configs', dict())
 
     def _load_self(self, manifest, path_hint, projects):
         # Handle the "self:" section in the manifest data.


mbolivar:
Ooh, thanks for this! It's quite short. I'll try to use it to see if we can avoid adding zephyr.disabled and just configure the mainline zephyr west manifest to do the right thing to support itself

carlescufi:
@mbolivar this was never followed-up and it's a shame because it does provide functionality beyond Zephyr.
@M1cha could you please rebase this PR?

M1cha:
@carlescufi done :)

bpiepiora:
Hi all,

since I am also using West without Zephyr, I would like to push this issue up again.

M1cha:
@bpiepiora I've rebased th ebranch, that's all I can do.
The shippable failure looks like something that was not caused by my change.

tejlmand:
@M1cha could you rebase.
The issue seens with CI should be fixed, but in meanwhile, manifest import has been introduced, so it could be good to have this PR updated.

M1cha:
@tejlmand done :)

JBPlinn:
We are waiting for this functionality for nearly a year now !
Could you please take a look at it and merge this one ?

M1cha:
yea. @tejlmand can you take another look and report if there are any (conceptional) problems? If there are I'll fix that as well.

tejlmand:
@M1cha Thanks for pinging.
I think @mbolivar-nordic  were the only one with some concerns, see https://github.com/zephyrproject-rtos/west/pull/261#issuecomment-490931808

@mbolivar-nordic with `find_package(Zephyr)` in place, and next Zephyr LTS coming soon, could we completely get rid of setting `ZEPHYR_BASE` in `west` ?

Then all that remains in this PR is the ability to provide a default conf set on initialization.

mbolivar-nordic:
Hi. Long-winded answer ahead.

I've thought about this, and I don't think the best way to handle making west a properly general git wrapper that doesn't have anything specific to do with Zephyr is to add another configuration option specific to Zephyr.

As @tejlmand hints, the only reason we have a `zephyr.base` option is because we relied on it so Zephyr specific extension commands like `build` and `flash` would have ZEPHYR_BASE defined for the sake of CMake, and we now have a better way to do this.

So yes, I think it would be better if west didn't set up a ZEPHYR_BASE environment variable at all.

The only reason we can't do that right now is because west still has to work with Zephyr v1.14 LTS, and that code's `build` and `flash` commands rely on west setting ZEPHYR_BASE.

Once v1.14 is no longer supported, we can eliminate the zephyr-specific bits in west completely.

If you look in the current code, we are already almost there. Besides the code in main.py that tries to set ZEPHYR_BASE, we only mention zephyr in the following places:
```
src/west/build.py
30:def is_zephyr_build(path):
43:        log.dbg(f'{path} is a zephyr build directory',
47:        log.dbg(f'{path} is NOT a valid zephyr build directory',
```

Deprecated module, exists for v1.14 LTS compatibility only. Safe to remove after that's unsupported.

```
src/west/commands.py
65:  - Set ZEPHYR_BASE to a zephyr repository path in a west workspace.
```

West looks for the workspace using the ZEPHYR_BASE environment variable as a fallback if it can't find one from the file system. I never liked this and think we can deprecate it. We only put it in because the application development guide recommended putting your application in `~/app` instead of `~/workspace/app` because... well, there was no workspace before west.

```
src/west/app/project.py
1373:MANIFEST_URL_DEFAULT = 'https://github.com/zephyrproject-rtos/zephyr'
```

We could similarly deprecate this slowly by emitting a warning in `west init` if `-m` is not given, and eventually make it a required argument over a period of a few years.

> We are waiting for this functionality for nearly a year now !
> Could you please take a look at it and merge this one ?

@JBPlinn I'm sorry to have kept you waiting, but I don't think you need this pull request merged to use west without zephyr. I have been doing it for a long time to manage my Arch Linux AUR projects:

https://github.com/mbolivar/aur-packages

All you should have to do is run `west config zephyr.base disabled` (and make sure `zephyr.base-prefer` is set to `configfile` if it's not already). You don't need a separate option like `zephyr.disabled`. This will make the ZEPHYR_BASE environment variable set to some garbage value in any workspace extension commands, but they aren't using zephyr anyway, so that seems fine.

Am I missing something?

M1cha:
well the beauty of this PR was that no additional local commands were necessary for setting up a workspace and to not have any `ZEPHYR_BASE` variable set.
But I agree the way this PR implements that is just a workaround for west doing zephyr-specific things when it shouldn't - and soon won't.

What I don't understand though is why the latest version of west has to support some ancient LTS of zephyr. They should just use the latest west version from the time when that version of zephyr was released. Holding back progress in a tool like this for years because of a project that's using it seems really damaging to me.

In case that's not an option: Couldn't we only set ZEPHYR_BASE if zephyr was actually found? IIRC the only thing that this commit fixes is a warning which we could just remove anyway.

tejlmand:
> What I don't understand though is why the latest version of west has to support some ancient LTS of zephyr. 

There are users relying on LTS, because such a release will have bugfixes backported:
https://github.com/zephyrproject-rtos/zephyr/commits/v1.14-branch

They could have product `A` based on 1.14 LTS, while at the same time be working on new product `B` following zephyr/master.
You can't request such users to install two `west` versions and swap between them.

`west` is a great tool, and should be made as generic as possible, but keep in mind that `west` was initiated in order to facilitate Zephyr development.

>  Couldn't we only set ZEPHYR_BASE if zephyr was actually found? 

how would you know that Zephyr was found ?
It was not until Zephyr CMake package was introduced, we could discover Zephyr automatically.
Until then, users had to specify `ZEPHYR_BASE` in environment or rely on `west` knowing where it was.
One benefit of `west` was that from the introduction of `west` then users no longer had to specify `ZEPHYR_BASE` by themselves. This was a major improvement.

M1cha:
> how would you know that Zephyr was found ?
well west does that already. It looks for `ZEPHYR_BASE` and falls back to looking for a project with name `zephyr`. if neither is found it prints a warning.

tejlmand:
> > how would you know that Zephyr was found ?
> 
> well west does that already. It looks for `ZEPHYR_BASE` and falls back to looking for a project with name `zephyr`. if neither is found it prints a warning.

Got it, so you just want it not to set the environment variable when Zephyr is not found.
That sounds reasonable.

tejlmand:
> So I think dropping the warning 

One thing to remember
Users of Zephyr 1.14 LTS does not have Zephyr CMake package.
And they might be using `west` in an out-of-workspace application, so they still require `ZEPHYR_BASE` or `zephyr.base` in `.west/config`, but later `west` versions DOES NOT set `zephyr.base` in the config file.

So we should consider to keep the warning until a new LTS is in place.

mbolivar-nordic:
> So we should consider to keep the warning until a new LTS is in place.

We could, but honestly, I think users of the LTS got set up a long time ago and already have a nice workflow established where the warning would not be printed anyway.

If there are any new users of Zephyr v1.14 LTS who write their own manifests, somehow forget to include a zephyr project, and get confused about why it doesn't work, just send them to me and I will help them. I expect exactly zero support cases from this offer :wink: .

tejlmand:
> > So we should consider to keep the warning until a new LTS is in place.
> 
> We could, but honestly, I think users of the LTS got set up a long time ago and already have a nice workflow established where the warning would not be printed anyway.
> 

I'm more thinking of `NewGuy` joining `Awesome` company to work on existing `Great` product that is based on Zephyr 1.14 LTS.
He gets a new PC and install latest `west` and fetches company product code, but not realizing that `ZEPHYR_BASE` must be defined because they decided to use `zephyr-awesome` in order to distinguish their Zephyr from vanilla Zephyr.

He will see this warning:
```
WARNING: can't find the zephyr repository
 - no --zephyr-base given
 - ZEPHYR_BASE is unset
 - west config contains no zephyr.base setting
 - no manifest project has name or path "zephyr"
```
if we remove the warning, such users might waste time in understanding what is wrong.
Don't expect everyone to be experienced Zephyr developers.
`NewGuy` doesn't care if he is 1 in 10 or 1 in 1.000.000, he just know he wasted a lot of unnecessary time.
Time that could be save by the existing warning.

All non-zephyr users can already today disable the warning, and it even informs so:
```
If this isn't a Zephyr workspace, you can   silence this warning with something like this:
   west config zephyr.base not-using-zephyr
```
Keeping the warning will not risk wasting anyone's time.
 
> If there are any new users of Zephyr v1.14 LTS who write their own manifests, somehow forget to include a zephyr project, and get confused about why it doesn't work, just send them to me and I will help them. I expect exactly zero support cases from this offer wink .

I'm sure people will figure it out, but could we agree that not everyone are equally experienced, and that we should try to be helpful to everyone, including those that will join projects based on older code and who might never have worked with Zephyr before. We might all have tried, `Ohh, that has been improved a long time ago, please update`, but hey, I'm forced with this version for reason x.y.z, and I just wasted x time on this.

Just look at slack, someone still uses `2.0.99`:

M1cha:
I also have to admit that we recently started a new project based on 1.14.x because bluetooth SIG certification was a requirement and that was only available for 2.2.x since this month.
So yes, @tejlmand s concerns are very valid.

tejlmand:
> OK, so... what I'm hearing is that we're going to just leave everything as-is until we can drop support for zephyr v1.14 and have a new LTS.

I think this:
> Couldn't we only set ZEPHYR_BASE if zephyr was actually found?

would mean that `ZEPHYR_BASE` should not be defined.
Today, if user does:
```
west config zephyr.base not-using-zephyr
```
then `ZEPHYR_BASE` will be set in environment. (it would be set to `not-using-zephyr`)
But suggestion is to not set it at all.

and in this PR there is one more topic that haven't been discussed.
Loading of default settings from the manifest file.
https://github.com/zephyrproject-rtos/west/pull/261/commits/533724a1e656636c9b03e83423e4d4e9b339a61a

mbolivar-nordic:
@tejlmand I don't get your meaning clearly, so I am going to assume you agree with the statement "we're going to just leave everything as-is until we can drop support for zephyr v1.14 and have a new LTS" where this PR (#261) is concerned.

Please say "no, that's not what I mean" if I misunderstood ;). Thanks.

tejlmand:
> Please say "no, that's not what I mean"

"no, that's not what I mean"

1) The environment variable is still being set by `west` (but to a dummy value). We could avoid setting it completely, if we want. But i'm fine "we're going to just leave everything as-is" for `ZEPHYR_BASE`.

2) This PR also contains:
```
 configs:
    required: false
    type: map
    allowempty: true 
```
and 
```
        self.configs = None
        '''Map with default configuration values.'''
```

in this commit: https://github.com/zephyrproject-rtos/west/pull/261/commits/533724a1e656636c9b03e83423e4d4e9b339a61a in the PR.
```
add support for loading default configs from manifest

Signed-off-by: Michael Zimmermann <michael.zimmermann@grandcentrix.net>
```

Is that something we should support ?

M1cha:
> Is that something we should support ?
In other words: Is there any config currently supported where it would make sense to set it from within the manifest so all users automatically have that value?

mbolivar-nordic:
Sorry I was confusing. By "leave everything as is", I meant "do not merge this PR".

I don't think merging support for `zephyr.disabled` is a good idea -- that is what I meant by " I don't think the best way to handle [this] is to add another configuration option specific to Zephyr" earlier.

If `zephyr.disabled` is the only use case we can come up with for setting default values for configuration options from the manifest file, I think we should not merge support for that new feature. If we can think of a use case other than `zephyr.disabled`, let's hear it for sure.

tejlmand:
> If zephyr.disabled is the only use case we can come up with for setting default values for configuration options from the manifest file, I think we should not merge support for that new feature. If we can think of a use case other than zephyr.disabled, let's hear it for sure.

Agree. Currently I do not a see a strong need, but I just wanted to ensure it was not overlooked.

jsiverskog:
isn't it reasonable to either backport support for this into the zephyr v1.14-branch or have users use `virtualenv` (or similar) if they need two versions of `west`?

mbolivar-nordic:
> isn't it reasonable to either backport support for this into the zephyr v1.14-branch or have users use `virtualenv` (or similar) if they need two versions of `west`?

This is a west PR, not a zephyr PR. I'm not sure what you're asking about.

jsiverskog:
> > isn't it reasonable to either backport support for this into the zephyr v1.14-branch or have users use `virtualenv` (or similar) if they need two versions of `west`?
> 
> This is a west PR, not a zephyr PR. I'm not sure what you're asking about.

from what i can see above the arguments against merging this is that zephyr v1.14 (lts) is a blocker.

M1cha:
Also the fix was mostly visual (it removes a warning) and there's a workaround to remove that warning.
My initial intend was to modify zephyr so I can do that workaround from within the manifest repository and the users of my project don't have to do any of this.

But for all the reasons mentioned above this is just not worth the hassle.

mbolivar-nordic:
> > > isn't it reasonable to either backport support for this into the zephyr v1.14-branch or have users use `virtualenv` (or similar) if they need two versions of `west`?
> > 
> > 
> > This is a west PR, not a zephyr PR. I'm not sure what you're asking about.
> 
> from what i can see above the arguments against merging this is that zephyr v1.14 (lts) is a blocker.

Ah, I see. I was confused by "backport support for this into the zephyr v1.14-branch", because this PR of course cannot be backported into that branch, since that branch is in a different code base.

FWIW, I use virtualenvs for everything, and I tried to get support for them merged into the official Zephyr workflow, but didn't have the bandwidth to address the requested changes:

https://github.com/zephyrproject-rtos/zephyr/pull/23979

So a virtualenv-based solution isn't really workable right now in my opinion, though I am personally a bit disappointed about that.

I'm looking forward to the day v1.14 goes away and we can get rid of a lot of this legacy stuff.

tejlmand:
> I'm looking forward to the day v1.14 goes away and we can get rid of a lot of this legacy stuff.

That day is close to come :smile: 

Only to be replaced with a new LTS that risk blocking `west` for introducing awesome new feature X, because that would break that new LTS :disappointed:

hongshui3000:
Is this PR still going on？

sabhiram:
bump - somewhat interested in being able to build closer to metal programs using west, but for the same arch etc so it will be ideal to reuse the toolchain etc.

mbolivar-nordic:
Hi, just in case you are not aware, you can already use west with non-zephyr projects. West will yell at you if it can't find a zephyr project, but just run `west config zephyr.base foo` to silence the warnings, and everything will work fine.

PR #266, opened 2019-05-15
Multimanifest prep

mbolivar:
Prep work commits for the multimanifest work.

The main changes needed to enable multimanifest are:

- `manifest: support explicit project URLs`
- `manifest: project names must be unique`

The rest are just cleanup and minor things.


mbolivar:
@tejlmand regarding `tests: remove empty_config.py`; any insight into why it was added? Just the result of `git commit -a` or so?

mbolivar:
Added some additional fixes to get all our tests passing on Windows 10 as well.

src/west/main.py
```diff
@@ -557,12 +557,10 @@ def main(argv=None):
     except KeyboardInterrupt:
         sys.exit(0)
     except CalledProcessError as cpe:
-        log.err('command exited with status {}: {}'.format(
-            cpe.args[0], quote_sh_list(cpe.args[1])))
```
carlescufi: I can guarantee this worked, I tested it explicitly. That said, this is the right way to do this.

src/west/main.py
```diff
@@ -563,19 +571,22 @@ def main(argv=None):
...
-                'was improperly defined and could not be run{}'.
-                format(': ' + ece.hint if ece.hint else ''))
+        msg = 'extension command "{}" could not be run{}.'.format(
+            args.command, ': ' + ece.hint if ece.hint else '')
         if args.verbose:
+            log.err(msg)
             traceback.print_exc()
         else:
-            log.inf(for_stack_trace)
+            log.err(msg, 'See {} for a traceback.'.format(dump_traceback()))
```
carlescufi: Nice, I like this.

src/west/manifest.py
```diff
@@ -281,6 +282,10 @@ def _load(self, data):
             except ValueError as ve:
                 self._malformed(ve.args[0])
 
+            # Project names must be unique.
+            if project.name in project_names:
```
carlescufi: Doesn't this fix a GitHub issue? I seem to recall there was one

tests/conftest.py
```diff
@@ -206,11 +205,12 @@ def cmd(cmd, cwd=None, stderr=None):
     # stdout from cmd is captured and returned. The command is run in
     # a python subprocess so that program-level setup and teardown
     # happen fresh.
-
-    cmdlst = shlex.split('west ' + cmd)
-    print('running:', cmdlst)
+    cmd = 'west ' + cmd
+    if platform.system() != 'Windows':
```
carlescufi: We typically use `sys.platform`, maybe use that for consistency

carlescufi: Why is this needed btw? We don't usually make this difference in normal invokations of `check_output` et al.

mbolivar: > We typically use sys.platform, maybe use that for consistency

Really? We use platform.system() in configuration.py ,which is what this is testing. I can make that change, though.

> Why is this needed btw? We don't usually make this difference in normal invokations of check_output et al.

It's breaking the tests because shlex splits by POSIX rules, which isn't right for windows shells and led to bad results when applied to paths. Example:

```
In [2]: shlex.split('west foo bar\\baz')
Out[2]: ['west', 'foo', 'barbaz']
```

src/west/manifest.py
mbolivar: This is essential to the design of the multimanifest feature. Projects are identified by their names in that work, it's how we decide what projects to apply overrides of attributes like `url`, `name`, etc. to. This means it doesn't make sense to continue allowing two projects with the same "name" in a single manifest. Arguably it never did, but it was necessary before because some projects might have had the same final path component in a URL but different URL bases. That case can be handled now with an explicit `url:` for at least one of the two.

src/west/main.py
mbolivar: I'm confused. This change is what we discussed in previous PRs, so that users who initialize CalledProcessError with kwargs instead of positionals don't get faced with an IndexError when we try to access `cpe.args`.

mbolivar: IOW: I know you're not asking for a change here, but you specifically asked for this change earlier :). Even though it "works" if you properly initialize CalledProcessError as the python subprocess module does, users who make their own instances won't always, as we discovered.

carlescufi: The change is required and I am completely happy with it. I am just saying that this old code actually worked, to my amazement :)

src/west/manifest.py
carlescufi: Understood, thanks for the explanation

tests/conftest.py
carlescufi: Right, I see. I don't have a strong preference either way, just wanted to keep it consistent. I didn't know we used `platform.system()` in `configuration.py`, but I've always seen `sys.platform` elsewhere, like the `tox.ini` or the `requirements.txt`.

marc-hb: 5 years later, dropping `shlex.split()` in #730

PR #268, opened 2019-05-24
manifest: error if self.path doesn't match directory name

### LINKED ISSUE #267: west init -l is brittle
Reported by @marc-hb:

For instance this doesn't work:
`ZEPHYR_BASE=~/tmp/zephyrsomething west --zephyr-base=zephyrsomething init -l zephyrsomething/`

marc-hb:
As an example, the "Clone the Zephyr Repositories" section of Zephyr's
Getting Started Guide provides the following command to wrap a west
installation around an already cloned, standalone zephyr repository:

  west init -l zephyr/

It feels natural to use the same command with some alternative
zephyr-blue/ or zephyr-red/ directory name, however that's not
compatible with the zephyr-blue/west.wml manifest which has its
self.path hardcoded to "zephyr". Hardcoding self.path is not a issue,
the problem is that the inconsistency is not caught and west pretends
the init command was successful. It's then much more difficult to
understand later why the build fails to find some file(s).

Implement a sanity check which make west init "fail fast" like this:

  FATAL ERROR: Manifest self.path is 'zephyr' but located in \
               directory '/home/marc/work/zephyr-blue'

Fixes #267

### DIFF
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -220,6 +220,15 @@ class Manifest:
         self_tag = manifest.get('self')
         if path is None:
             path = self_tag.get('path') if self_tag else ''
+
+        # Both are named "self.path" depending on where you look at.
+        if path and self.path:
+            realpath = os.path.dirname(self.path)
+            realdir = os.path.basename(realpath)
+            if realdir not in [os.path.basename(path), 'manifest-tmp']:
+                raise MalformedManifest("""Manifest self.path is '{}' \
+but located in directory '{}'""".format(path, realpath))
+
         west_commands = self_tag.get('west-commands') if self_tag else None
 
         project = ManifestProject(path=path, west_commands=west_commands)
--- a/tests/test_manifest.py
+++ b/tests/test_manifest.py
@@ -21,7 +21,7 @@ def config_file_project_setup(tmpdir):
     tmpdir.join('.west/.west_topdir').ensure()
     tmpdir.join('.west/config').write('''
 [manifest]
-path = manifestproject
+path = manifests
 ''')
 
     # Switch to the top-level West installation directory
@@ -128,7 +128,7 @@ def test_no_defaults(config_file_project_setup):
     with patch('west.util.west_topdir', return_value='/west_top'):
         manifest = Manifest.from_data(yaml.safe_load(content))
 
-        expected = [ManifestProject(path='manifestproject'),
+        expected = [ManifestProject(path='manifests'),
                     Project('testproject1', None, path='testproject1',
                             clone_depth=None, revision='rev1', remote=r1),
                     Project('testproject2', None, path='testproject2',
@@ -161,7 +161,7 @@ def test_self_tag(project_setup):
           remote: testremote2
 
       self:
-        path: mainproject
+        path: manifests
     '''
     r1 = Remote('testremote1', 'https://example1.com')
     r2 = Remote('testremote2', 'https://example2.com')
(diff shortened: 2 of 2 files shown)


src/west/manifest.py
```diff
@@ -220,6 +220,15 @@ def _load(self, data):
         self_tag = manifest.get('self')
         if path is None:
             path = self_tag.get('path') if self_tag else ''
+
+        # Both are named "self.path" depending on where you look at.
+        if path and self.path:
+            realpath = os.path.dirname(self.path)
+            realdir = os.path.basename(realpath)
+            if realdir not in [os.path.basename(path), 'manifest-tmp']:
+                raise MalformedManifest("""Manifest self.path is '{}' \
```
mbolivar: Are you sure you don't want to implement this in the init command instead? I think it's legal to let the manifest parser be contained in a directory that has a "wrong" name; the problem comes when used with init -l, no?

mbolivar:
Thanks for the patch! I think this fix might belong in the init implementation in project.py, though.

src/west/manifest.py
marc-hb: The only problem I found is in the init -l command, however:
- Would anyone ever want an incorrect "self.path"?
- How safe would that be? For instance letting west clone in self.path and manually renaming it later seems somewhat reckless.

self.path is optional. If some project makes the effort to configure it then there are some project-specific reason(s) to do so (e.g.: zephyr) and it should be obeyed. Otherwise they can just not use it.

mbolivar: Thinking about this more, I am convinced this problem is specific to `init -l`.

The `west.manifest` module is intended to be used as a general purpose library for parsing and managing manifests, so putting it here is really telling the user what they ought to be doing with their data, rather than letting them choose.

> Would anyone ever want an incorrect "self.path"?

They may be parsing a bunch of manifest files they have dumped in a temporary directory, and only be interested in the project list, for example. In this case, they do not care about self.path at all. Having this exception here prevents them from getting their job done in that case.

Or they might have a bunch of manifests which they are parsing and using to generate some other directory in the right place, by using self.path, even though the directory containing the manifest might not happen to be a valid manifest repository. That's fine, because they're just operating on manifests, not manifest repositories.

Also notice that you can create an instance using `Manifest.from_data()`, so there *is no* directory containing the manifest in general (what if we got the manifest data over the network?).

The `MalformedManifest` exception is for when the contents of the manifest file itself are malformed. It isn't about a context-specific issue in the environment in which the manifest is used.

> How safe would that be? For instance letting west clone in self.path and manually renaming it later seems somewhat reckless.

In my opinion, error handling in that case should also be handled by the commands which operate on an installation, and not the routine whose only job it is to load manifest data.

mbolivar: And yes, I'm aware that there is some code in here which violates the abstraction, e.g. by calling `west_dir()`.

Things are actually cleaner than they used to be. We had arguments early on in the west design if we should be trying to make user-visible APIs or not. "Yes" won out, but code written by the "no" side had already been merged and has slowly been pulled apart into the current `west.manifest` etc. versus `west.commands.xxx`.

We have an issue open to finish the cleanup (https://github.com/zephyrproject-rtos/west/issues/191); let's not make things worse :).

marc-hb: Thanks for all the details, I'll try to find time to change this.

mbolivar: Thanks again for your first west pull request! I'm always glad when we get submissions from other Zephyr developers; I see it as a sign of health.

If you don't find the time, let me know and I'll make the changes.

marc-hb:
> If you don't find the time, let me know and I'll make the changes.

@mbolivar I won't have time to resume work on this particular one, not in the next couple months anyway.

cc @Blacksilver42

mbolivar-nordic:
> @mbolivar I won't have time to resume work on this particular one, not in the next couple months anyway.

OK, I'm closing out stale PRs for now, so please feel free to reopen if you have time again. Thanks for the work so far!

Issue #259, opened 2019-05-07
Support FreeBSD


sdalu:
FreeBSD: 
* system-wide:   `/usr/local/etc/westconfig`
* user-wide: `~/.westconfig` or `$XDG_CONFIG_HOME/west/config`

mbolivar:
What happens now when you run on freebsd? Is there some error?

sdalu:
The following error is due to a missing test in `ConfigFile` (file `configuration.py`), where `paltform.system()` is not tested for FreeBSD

I'm guessing also that `read_config` should be modified as an explicit test is made on `platform.system()`

~~~
Traceback (most recent call last):
  File "/home/sdalu/.local/bin/west", line 11, in <module>
    sys.exit(main())
  File "/home/sdalu/.local/lib/python3.6/site-packages/west/_bootstrap/main.py", line 499, in main
    wrap(wrap_argv)
  File "/home/sdalu/.local/lib/python3.6/site-packages/west/_bootstrap/main.py", line 485, in wrap
    west.main.main(argv)
  File "/home/sdalu/zephyrproject/.west/west/src/west/main.py", line 538, in main
    config.read_config()
  File "/home/sdalu/zephyrproject/.west/west/src/west/configuration.py", line 95, in read_config
    files.append(ConfigFile.SYSTEM.value)
  File "/usr/local/lib/python3.6/enum.py", line 326, in __getattr__
    raise AttributeError(name) from None
AttributeError: SYSTEM
~~~

mbolivar:
Hi @sdalu -- if you have a chance to try out #271 and see if it works for you, I'd appreciate it.

To test, run `pip install -e .` from the west repository with that PR checked out, then try out the use cases that matter to you. If you're happy, I'll mark that PR as fixing this issue.

Thanks again for your report.

sdalu:
Just had time to try zephyr again. And it is now possible to install west on FreeBSD.

PR #271, opened 2019-05-25
Flesh out west config command and west.configuration API

mbolivar:
The goal of this pull request is to add all the features I would like `west config` to have for 1.0.

(I'm not 100% happy with the user friendliness of the `west.configuration` API, though, so I might want to send a follow-up with some deprecations and cleanups.)

The configuration API currently only allows getting and setting individual config options. The test cases are a great starting point for validating this, but they work only at the command level and do not test the underlying APIs, or that the command-line interface and module APIs work together as expected.

To flesh out the interface for both, this:

- adds `west config --list` to print all options and their values
- adds `west config -d [--global | --system | --local] foo.bar` to delete `foo.bar` from an individual file, or the *first* global or local file with a value where it is set
- adds `west config -D foo.bar` to delete `foo.bar` everywhere (including the system file)

It also includes these new features and fixes:

- fixes the behavior  when `XDG_CONFIG_HOME` is set to match the XDG Desktop Specification.
- adds configuration support for BSD operating systems (see #259)
- adds `WEST_CONFIG_SYSTEM`, `WEST_CONFIG_GLOBAL`, and `WEST_CONFIG_LOCAL` environment variables for setting config file locations
- always creates configuration files if they do not exist

Finally, it extends the test coverage quite a bit.

The first patch is a minor behavioral change I needed to make the test updates work nicely.

### DIFF
--- a/src/west/commands/command.py
+++ b/src/west/commands/command.py
@@ -45,7 +45,8 @@ class ExtensionCommandError(CommandError):
         super(ExtensionCommandError, self).__init__(**kwargs)
 
 
-_NO_TOPDIR_MSG_FMT = '''Error: "{}" is not in a west installation.
+_NO_TOPDIR_MSG_FMT = '''\
+no west installation found from "{}"; "west {}" requires one.
 Things to try:
  - Set ZEPHYR_BASE to a zephyr repository path in a west installation.
  - Run "west init" to set up an installation here.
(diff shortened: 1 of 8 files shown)


src/west/commands/command.py
```diff
@@ -45,7 +45,8 @@ def __init__(self, **kwargs):
         super(ExtensionCommandError, self).__init__(**kwargs)
 
 
-_NO_TOPDIR_MSG_FMT = '''Error: "{}" is not in a west installation.
+_NO_TOPDIR_MSG_FMT = '''\
+no west installation found from "{}"; "west {}" requires one.
```
carlescufi: I think we should change the name from `west installation` to something else now that west is not part of the installation

Issue #38, opened 2018-09-21
west library / application / bootstrap structure and API


tejlmand:
from the user point of view `west` is the go-to tool intended for repository handling, building, flashing etc.

Internally, `west` can be seen as:
- bootstrap functionality
- west library (a library which can be reused by build scripts, test, etc)
- application (command line)

With the initial version, v.0.1.0 out and progress is made on v0.2.0, it is important to keep track of the future direction.

This includes, but is not limited to:
- Structure and name spacing
- API
- What features of `west` do we expect other would re-use ?

tejlmand:
IRC discussion found in attachment.
[#zephyr-meta.txt](https://github.com/zephyrproject-rtos/west/files/2404752/zephyr-meta.txt)

ulfalizer:
Thanks for opening this!

Re. the API, I would go for a plain

```python
from west import ESP32BinaryRunner
```

(Much nicer than the ugly separate top-level `runners` thing I suggested earlier.)

I don't think West is large enough to warrant any namespacing in the API. Heavily namespaced code tends to be hard to read, IMO (`west.ESP32BinaryRunner` reads nicer than `west.runners.ESP32BinaryRunner`, and in that case there's no risk for collision).

For the structure, simplest thing would just to put everything at the top level I think. Probably too simple though. At least the tests ought to go in a separate directory.

```
bootstrap.py
westcmd.py
west.py (the library, imported by e.g. westcmd.py)
command_build.py
command_flash.py
...
runner_bossac.py
runner_xtensa.py
...
tests/
```

Adding one layer of subdirectories:

```
bootstrap.py
westcmd.py
west.py
commands/__init__.py
commands/build.py
commands/flash.py
...
runners/__init__.py
runners/bossac.py
runners/xtensa.py
...
tests/
```

`west.py` could populate the API with something like

```
from runners.bossac import BossacBinaryRunner
from runners.xtensa import XtensaBinaryRunner
...
```

I'm sure it could be tweaked (didn't even list a bunch of other files), but personally I prefer fairly shallow and "`cd`-less" hierarchies at least, unless the project is massive.

Any thoughts?

ulfalizer:
I looked into how the Python standard library tends to organize things (can be found in `/usr/lib/python3.6` or the like), and I think it's pretty nice:

  - The majority of the modules are one file each (`os.py`, `argparse.py`, etc.)

[...]

PR #275, opened 2019-06-08
manifest: new Project.west_top to reduce west_topdir() frequency

marc-hb:
On a current and standard Zephyr configuration with 12 projects, a
single 'west list' invocation performs 31, non-recursive invocations of
west_topdir(). Each west_topdir() call invokes os.path.isdir() a few
times (how many varies depending on where west is called from).

Add a new, optional Project.west_top field. Compute it only once and
re-use it when constructing the list of Project objects. In my
configuration this gets the number of west_topdir() invocations down
from 31 to 7.

On local storage the speed improvement is only a couple percents. It
takes barely longer to just load all imports and exit immediately than
actually run the 'west list' command anyway. The goal here is not
performance but the ability to trace and/or debug west_topdir() without
being drowned in noise.

### DIFF
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -290,8 +290,9 @@ class Manifest:
             path = self_tag.get('path') if self_tag else ''
         west_commands = self_tag.get('west-commands') if self_tag else None
 
-        project = ManifestProject(path=path, west_commands=west_commands)
-        projects.insert(MANIFEST_PROJECT_INDEX, project)
+        self_project = ManifestProject(west_top=util.west_topdir(),
+                                       path=path, west_commands=west_commands)
+        projects.insert(MANIFEST_PROJECT_INDEX, self_project)
 
         # Map from each remote's name onto that remote's data in the manifest.
         remotes = tuple(Remote(r['name'], r['url-base']) for r in
@@ -352,7 +353,8 @@ class Manifest:
                                   west_commands=mp.get('west-commands'),
                                   remote=remote,
                                   repo_path=repo_path,
-                                  url=url)
+                                  url=url,
+                                  west_top=self_project.west_top)
             except ValueError as ve:
                 self._malformed(ve.args[0])
 
@@ -479,11 +481,11 @@ class Project:
     Projects are neither comparable nor hashable.'''
 
     __slots__ = ('name remote url path abspath posixpath clone_depth '
-                 'revision west_commands').split()
+                 'revision west_commands west_top').split()
 
     def __init__(self, name, defaults=None, path=None, clone_depth=None,
                  revision=None, west_commands=None, remote=None,
-                 repo_path=None, url=None):
+                 repo_path=None, url=None, west_top=None):
         '''Specify a Project by name, Remote, and optional information.
 
         :param name: Project's user-defined name in the manifest.
@@ -509,6 +511,7 @@ class Project:
                           fetch URL.
         :param url: The project's fetch URL. This cannot be given with *remote*
                     or *repo_path*.
+        :param west_top: Saves a call to util.west_topdir() if given.
         '''
         if remote and url:
             raise ValueError('got both remote={} and url={}'.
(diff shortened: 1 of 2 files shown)


src/west/manifest.py
```diff
@@ -401,11 +403,12 @@ class Project:
 
     Projects are neither comparable nor hashable.'''
 
-    __slots__ = ('name remote url path abspath posixpath clone_depth '
+    __slots__ = ('name remote url path west_top abspath posixpath clone_depth '
```
mbolivar: Just a nit, but `__slots__` is ordered, so would you mind not putting this in between the three xxpath fields? E.g. putting it at the end of the list?

src/west/manifest.py
```diff
@@ -278,7 +278,9 @@ def _load(self, data):
                                   revision=mp.get('revision'),
                                   west_commands=mp.get('west-commands'),
                                   remote=remote,
-                                  url=url)
+                                  url=url,
+                                  west_top=self_project.west_top,
+                                  )
```
mbolivar: Style nit, please write:

```
west_top=self_project.west_top)
```

src/west/manifest.py
```diff
@@ -645,7 +650,9 @@ def __init__(self, path=None, revision='(not set)', url='(not set)',
         self.path = path or self.name
         '''Relative path to the project in the installation.'''
 
-        self.abspath = os.path.realpath(os.path.join(util.west_topdir(),
+        self.west_top = util.west_topdir()
```
mbolivar: I think this is a great idea, but could you do it from the Manifest class instead of ManifestProject?

That will make it easier to make it a kwarg to the manifest constructor and factory methods, which we can then feed in from WestCommand later to make this even better.

src/west/manifest.py
marc-hb: Done but you understand why I did it this way, right?

src/west/manifest.py
marc-hb: I put `west_stop` last here but I had to put it first in the ManifestProject() constructor to make it mandatory. Feedback welcome!

src/west/manifest.py
mbolivar: Trying to minimize diff in later patches?

src/west/manifest.py
mbolivar: > Feedback welcome!

I am not a fan of treating positionals as kwargs (as in `ManifestProject(west_top=foo, ....)` instead of `ManifestProject(foo, ...)`). The formal parameter names of positionals should not be considered API, in my opinion. If https://www.python.org/dev/peps/pep-0570/ were available in Python 3.4 I would use it everywhere.

But this is fine, just a personal maintainer quirk :).

marc-hb:
> ... optional in Project but a positional to ManifestProject. [...]
> Care to explain why they cannot be made consistent?

Optional in Project because the Project() constructor is already used in many places, so making it mandatory would be a relatively big change, not the optional and flexible optimisation this commit is (so far) meant to be.
I think I remember seeing some places where there just can't even be a west_top, for instance when initialising from the network. Sorry it's been a while, I can check that again if you'd like me to.

Mandatory in ManifestProject because the ManifestProject() constructor is used only once and *when possible* and easily done it's better to reduce the number of possible flows. You asked to move the computation of west_top one level up, making the west_top parameter mandatory makes sure of that.

Now I realize adding the above explanation to the commit message wouldn't hurt, just let me know.

marc-hb:
Rebased on top of conflicting bfc67866364c. No change.

marc-hb:
While I thought I was merely implementing some optimization and refactoring, this PR changes a public API and I missed that, thanks @mbolivar  for correcting me on Slack https://docs.zephyrproject.org/latest/guides/west/extensions.html#west-extensions

Other related pointers shared on Slack:
- Issue #38 west library / application / bootstrap structure and API
- https://lwn.net/Articles/795019/ Python and public APIs

mbolivar-nordic:
Closing as stale (no updates for almost a year now). Thanks for the work so far and please feel free to reopen if you have an update.

Issue #283, opened 2019-07-25
Allow to specify repository name


vanwinkeljan:
It would be nice that there is an attribute to specify the repository name (eg. `repo-name`) that together with the `remote` can be combined into a URL to clone from.
This would allow to use the same repo for multiple projects, eg. in the case of multiple API version, without sacrificing the benefits of the remote concept.

See also: https://github.com/zephyrproject-rtos/zephyr/pull/17714#discussion_r307496625

mbolivar:
Thinking about it, that's reasonable and shouldn't be too hard to implement. I'm giving it a go now.

mbolivar:
@vanwinkeljan I've implemented this feature as `repo-path`, to make it clearer that it can contain slashes and is not just a name. Please see #284 -- any testing or feedback is welcome.

PR #284, opened 2019-07-25
West 0.6.0 rc2

mbolivar:
Compared to rc1, this adds a bugfix and a new feature, the project 'repo-path' attribute, to allow continuing use of remotes with non-unique project names.

Fixes: #283

### DIFF
--- a/src/west/manifest-schema.yml
+++ b/src/west/manifest-schema.yml
@@ -56,15 +56,23 @@ mapping:
     sequence:
       - type: map
         mapping:
-          # remote.url-base + / + name is the fetch URL
+          # Project's unique name.
           name:
             required: true
             type: str
-          # Name of the project remote
+          # Name of the project's remote. May not be combined with "url".
           remote:
             required: false
             type: str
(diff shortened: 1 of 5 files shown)


mbolivar:
Once this settles down, I'll update the documentation in https://github.com/zephyrproject-rtos/zephyr/pull/17714. For now, hopefully commit logs and new test cases make things clear.

vanwinkeljan:
Tested it with LVGL PR https://github.com/zephyrproject-rtos/zephyr/pull/17629 and it works as expected.

One observation is that `west --help` and `west --version` do not run if the manifest is malformed (in my case two project with the same name).

carlescufi:
> Tested it with LVGL PR [zephyrproject-rtos/zephyr#17629](https://github.com/zephyrproject-rtos/zephyr/pull/17629) and it works as expected.
> 
> One observation is that `west --help` and `west --version` do not run if the manifest is malformed (in my case two project with the same name).

Sounds like a fix for rc3

Issue #242, opened 2019-03-28
Extra git arguments to `west diff` / `west status` don't work


mbolivar:
To reproduce:

```
west diff -- --no-index
```

Expected behavior: `--no-index` passed to `git diff`.

Actual behavior:

```
$ west diff -- --no-index
FATAL ERROR: Unknown project name/path --no-index (available projects: tinycbor, net-tools, zephyr)
```

Marking as only medium priority since `west forall -c "git <diff|status> --options"` is a workaround. I'll fix the RST documentation to not mention the extra git arguments for now as well.

mbolivar:
Expanding the issue since the same applies to status:

```
$ west status -- --show-stash
FATAL ERROR: Unknown project name/path --show-stash (available projects: tinycbor, net-tools, zephyr)
```

PR #290, opened 2019-08-16
Prep work improvements

mbolivar:
This is a bunch of cleanup and prep work patches I think should be sent on their own (since they're already quite a lot of commits).

It's a basis for two lines of development going on in semi-parallel:

- add a smarter project fetching strategy that avoids fetching projects when the revisions are definitely already available locally
- some downstream-specific extension commands that want to use some of the code that's currently hidden in private project.py methods

I should add:

Fixes: #242 

By removing the unsupported behavior from the test cases and help strings.

### DIFF
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -934,9 +934,6 @@ def _git(project, cmd, extra_args=(), capture_stdout=False, check=True,
          cwd=None):
     # Wrapper for project.git() that by default calls log.die() with a
     # message about the command that failed if CalledProcessError is raised.
-    #
-    # If the global error context value is set, it is appended to the
-    # message.
 
     try:
         res = project.git(cmd, extra_args=extra_args,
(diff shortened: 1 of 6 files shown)


mbolivar:
> some downstream-specific extension commands that want to use some of the code that's currently hidden in private project.py methods

I should note that this one in particular is going to require re-working manifest.py quite a bit to stop assuming that every manifest file being parsed is inside an installation.

The assumption is already unsound within the bounds of our data model since we have Manifest.from_data(), which doesn't assume any file system at all, and is getting in the way of some downstream extensions that want to parse both zephyr/west.yml and downstream-manifest-repository/west.yml and do some comparisons between them.

src/west/commands/project.py
```diff
@@ -29,7 +29,137 @@
...
+        # Listed but missing projects. Used for error reporting.
+        missing_projects = []
+
+        def normalize(path):
+            # Returns a case-normalized canonical absolute version of
+            # 'path', for comparisons. The normcase() is a no-op on
+            # platforms on case-sensitive filesystems.
+            return normcase(realpath(path))
+
+        res = []
```
carlescufi: When does this code actually get run? When the class is defined?

mbolivar: This particular line? It's a method so it gets run when it gets invoked, same as a function. Did you mean some other code?

mbolivar: AH. I think I see the confusion. The normalize() def above is a nested function inside the _projects() method definition.

Note that normalize() gets removed entirely later in the series; this first step is just moving code around.

carlescufi: Ok thanks for the explanation

PR #292, opened 2019-08-19
commands: config: fix error message when providing wrong key formating.

LeBlue:
When requesting config values with the wrong key format (not section.key), west config crashes on producing the error message. Removes the non existent kwarg 'exit_code' and adds the missing format statement.

### DIFF
--- a/src/west/commands/config.py
+++ b/src/west/commands/config.py
@@ -200,8 +200,8 @@ class Config(WestCommand):
     def _sk(self, args):
         name_list = args.name.split(".", 1)
         if len(name_list) != 2:
-            self.parser.error('name {} should be in the form <section>.<key>',
-                              exit_code=3)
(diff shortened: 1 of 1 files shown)


src/west/commands/config.py
```diff
@@ -200,8 +200,8 @@ def write(self, args):
     def _sk(self, args):
         name_list = args.name.split(".", 1)
         if len(name_list) != 2:
-            self.parser.error('name {} should be in the form <section>.<key>',
-                              exit_code=3)
+            self.parser.error("name '{}' should be in the form <section>.<key>".
```
mbolivar: Actually, can you please fix the shippable error and keep this line 80 column clean?

LeBlue: Should be good now.

PR #296, opened 2019-08-30
Path handling improvements

mbolivar:
Sorry this took too long, but there were more edge cases than my original optimism accounted for.

These are the changes we need to finish the downstream NCS extension commands to list out of tree patches, etc. I want to finish the implementations of the downstream commands before we merge this, though, so marking DNM.

### DIFF
--- a/src/west/commands/config.py
+++ b/src/west/commands/config.py
@@ -97,7 +97,8 @@ class Config(WestCommand):
         super().__init__(
             'config',
             'get or set configuration settings in west config files',
-            CONFIG_DESCRIPTION)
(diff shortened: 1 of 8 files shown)


mbolivar:
@marc-hb I think this should supersede https://github.com/zephyrproject-rtos/west/pull/275. It accomplishes the same thing in a way that I think will be more future proof.

mbolivar:
> Quick glance and those all look reasonable to me. In any case we will probably need to iterate further as more requirements arise from downstream usecases.

OK, let's keep iterating then. I'll merge now and we'll hold off on the next point release based on this code until we're more sure.

Issue #310, opened 2019-09-26
RFC: Clean workspace command and/or update option


bjda:
I have on multiple occasions had trouble compiling applications after checking out different versions and `west update`ing our multi-repo downstream (nRF Connect SDK), because of leftover repos and files. It has been fixable by deleting all the repos except the manifest repo, and running `west update`. This is especially prominent when using a single workspace and jumping significantly forward/backward in time.

I propose we introduce a `west clean` command that approximates a fresh workspace, without re-downloading all the repos. The steps could be:
* Remove repos/folders not in the manifest (potentially even moving them to a caching folder)
* Run `git clean -fdx` in each project folder.

Alternatively – or complementarily – adding this functionality as an option in `west update` could be useful. Then, `west update --clean` could be used to bring the rest of the workspace in line with the manifest repo with a single command.

There are several points of discussion, including but probably not limited to:
* Do we even bother?
* Is there some detail about the west/manifest/repo structure that makes implementing this impractical?
* Do we call it `clean`, `reset` or something else

mbolivar:
My first instincts are against this feature request.

- `west forall -c "git clean -fdx"` (or a similar command like `west forall -c "git reset --hard HEAD"`) is enough to clean out every repository west does know about. In general, I want to encourage use of `forall` for special purposes, keeping the number of built-in commands limited to "very common" use cases like `diff` and `status`.
- I would be pretty terrified to introduce a built-in west command that walked the entire installation to find git repositories it didn't know about and just deleted them.
- With a clean build directory, zephyr won't use any repositories that aren't managed by west, as `ZEPHYR_MODULES` is (by default) initialized using `west list`, and `ZEPHYR_EXTRA_MODULES` defaults to the empty list (refer to the zephyr docs for more info on these)

For Linux users (macOS ought to work too, unless its BSD `find` doesn't support some of the flags used here), here is a shell one-liner you can run to compare the git repositories west knows about with all the other ones inside the same workspace:

```
diff -u <(west list -f {abspath} | sort) <(find $PWD -name .git -type d | sed 's/\/\.git//' | sort)
```

You need to run it from the top level directory in the installation.

I can come up with a Windows equivalent if there's interest.

marc-hb:
> I would be pretty terrified to introduce a built-in west command that walked the entire installation to find git repositories it didn't know about and just deleted them.

+1 !

Without going that far, I think one new feature and one change would make cleaning much easier:

[...]

PR #311, opened 2019-09-26
add 'west topdir' command

mbolivar:
This new command will just print the root or top level directory of the west installation (or an error if none is found). We also tweak some of the help text while we're here to keep 'west help' nice and clean.

cc: @marc-hb @bjda

Discussion in https://github.com/zephyrproject-rtos/west/issues/310 brought up a request for a way to print the installation top level directory. We've been discussing whether or not to add this for quite some time now. It's clearly useful to some users, so let's just add it.

### DIFF
--- a/src/west/commands/__init__.py
+++ b/src/west/commands/__init__.py
@@ -50,6 +50,7 @@ class ExtensionCommandError(CommandError):
 _NO_TOPDIR_MSG_FMT = '''\
 no west installation found from "{}"; "west {}" requires one.
 Things to try:
+ - Change directory to a west installation and retry.
  - Set ZEPHYR_BASE to a zephyr repository path in a west installation.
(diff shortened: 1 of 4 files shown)


mbolivar:
> This will fix https://github.com/zephyrproject-rtos/zephyr/pull/16536/files, thanks!

It should be part of version 0.7.0, so it will be usable whenever that's the minimum required version by zephyr.

Note that I don't have a requirement to bump the minimum version higher than 0.6.2 for upstream zephyr any time soon, though.

marc-hb:
Removing my +1 while testing with some symbolic links... too late!

mbolivar:
> Remind me to add a PR to update the autocompletion in the zephyr repo.

@carlescufi this is your reminder

PR #313, opened 2019-09-27
manifest: don't use realpath for abspath; prefer PWD for west_top

marc-hb:
This is enough to make "west list" and "west topdir" not resolve
symbolic links in a number of cases, for instance this has been
successfully tested with symbolic links in
-fmacro-prefix-map=${WEST_TOPDIR}=WEST_TOPDIR as implemented in
https://github.com/zephyrproject-rtos/zephyr/pull/19440

This is enough to avoid too many double dots in module names generated
when ZEPHYR_BASE has symbolic links, e.g. from:

  build/modules/stuff/CMakeFiles/..__..__..__zephyrproject__modules_etc

to:

  build/modules/stuff/CMakeFiles/modules_etc

While successfully tested on very large sanitycheck samples and enough
to make them reproducible, this will likely never support every single
possible use of symbolic links.  See one exception described in the code
comments where "make" is invoked directly. However this makes no
difference when not using symbolic links.

TODO: add a Linux test, skipped on Windows with
https://docs.pytest.org/en/latest/skipping.html#skip-all-test-functions-of-a-class-or-module

### DIFF
--- a/src/west/util.py
+++ b/src/west/util.py
@@ -91,6 +91,30 @@ class WestNotFound(RuntimeError):
     '''Neither the current directory nor any parent has a West installation.'''
 
 
+def get_pwd_or_cwd():
+    '''To align with shells that support symbolic links, prefer the PWD
+    environment variable, otherwise return os.getcwd()
+    '''
+    pwd = os.environ.get('PWD', None)
+    cwd = os.getcwd()
+
+    if pwd is None:
+        return cwd
+
+    # west_topdir is crucial, so let's be paranoid.
+    if not os.path.isdir(pwd):
+        raise os.error("PWD %s is not a directory" % pwd)
+
+    if not os.path.realpath(pwd) == os.path.realpath(cwd):
+        # This can happen if we're not invoked by the shell directly but
+        # by a process which changed cwd without changing PWD. To
+        # reproduce edit some CMakeLists.txt file and run make -C build
+        # which re-runs cmake --check-build-system -> west --version
+        return cwd
+
+    return pwd
+
+
 def west_dir(start=None):
     '''Returns the absolute path of the installation's .west directory.
 
@@ -111,7 +135,7 @@ def west_topdir(start=None, fall_back=True):
     # If you change this function, make sure to update the bootstrap
     # script's find_west_topdir().
 
-    cur_dir = start or os.getcwd()
+    cur_dir = start or get_pwd_or_cwd()
 
     while True:
         if os.path.isdir(os.path.join(cur_dir, '.west')):
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -650,7 +650,7 @@ class Project:
         self.topdir = topdir
         '''Root directory of the west installation this project is inside,
         or None.'''
-        self.abspath = (os.path.realpath(os.path.join(topdir, self.path))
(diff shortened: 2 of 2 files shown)


src/west/util.py
```diff
@@ -111,7 +135,7 @@ def west_topdir(start=None, fall_back=True):
     # If you change this function, make sure to update the bootstrap
     # script's find_west_topdir().
 
-    cur_dir = start or os.getcwd()
+    cur_dir = start or get_pwd_or_cwd()
```
mbolivar: This isn't the only place where we use os.getcwd(), even outside of the test case code.

I know you're still working on this and that it's just a draft, so I assume there will be test cases once you're ready for this to get merged.

src/west/util.py
```diff
@@ -91,6 +91,30 @@ class WestNotFound(RuntimeError):
...
+    if pwd is None:
+        return cwd
+
+    # west_topdir is crucial, so let's be paranoid.
+    if not os.path.isdir(pwd):
+        raise os.error("PWD %s is not a directory" % pwd)
+
+    if not os.path.realpath(pwd) == os.path.realpath(cwd):
+        # This can happen if we're not invoked by the shell directly but
+        # by a process which changed cwd without changing PWD. To
```
mbolivar: And here?

marc-hb:
> I know you're still working on this and that it's just a draft, so I assume there will be test cases once you're ready for this to get merged.

Yes, this is just an early and incomplete (but functional enough) draft. Thanks for providing feedback already anyway!

src/west/util.py
marc-h38: As `/usr/bin/pwd -L` merely returns $PWD I don't see what difference it would make in this `make -C` example.

src/west/util.py
```diff
@@ -91,6 +91,30 @@ class WestNotFound(RuntimeError):
...
+    '''To align with shells that support symbolic links, prefer the PWD
+    environment variable, otherwise return os.getcwd()
+    '''
+    pwd = os.environ.get('PWD', None)
+    cwd = os.getcwd()
+
+    if pwd is None:
+        return cwd
+
+    # west_topdir is crucial, so let's be paranoid.
```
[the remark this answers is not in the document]
marc-h38: Yes, that would be relying on `/usr/bin/pwd -L` to perform this sanity check which seems to be true on the two operating systems I tried `PWD=fubar /usr/bin/pwd -L`[*]. Not sure such a sanity check can be taken for granted but to be honest this first check is totally paranoid; only in case someone evil corrupts PWD?

[*] interestingly, you can successfully fool `-L` like this if PWD is a valid match.

src/west/util.py
marc-h38: Right, as explained in the commit message this place was enough to fix the _limited_ (but useful!) test coverage I tried at the time.
It wouldn't hurt to  fix this problem "incrementally" because it's not like `realpath` or its equivalent is consistently used across the whole zephyr build right now, if you look beyond `west` the situation is already and vastly `abspath/realpath` inconsistent = how I found this issue in the first place.

mbolivar-nordic:
> I have no time to work on this in the short or medium term, it's all yours if anyone wants to start from it.

Tempted to close as stale, but relates to https://github.com/zephyrproject-rtos/west/issues/419 so leaving open for now.

Issue #126, opened 2018-11-27
Manifest Version field


tejlmand:
During work on PR#125 in order to settle on #106 it was identified that in order to be prepared in future for `west` bootstrapper and `west` itself to load older project, i.e. `west` v2.0.0 loading a manifest written when only `west` v1.0.0 was available.

Thus it is important that the manifest contains a manifest version field in order for future `west` versions to be able to load manifest based on old schemas where required fields might have changed.

tejlmand:
There are currently two parts of the manifest that might change:
1) The `west` section
     Changes to the section might require an update to the bootstrapped, and thus it must be possible for old bootstrapper to detect and suggest an update if section cannot be parsed, as well as a new bootstrapper should still be able to at least load an old manifest and fetch the `west` referenced in there.
2) manifest section.
    A manifest file could reference just HEAD of master as the `west` toll to use, so the `west` might need to be able to validate old sections.

Question now is, if there is only a single version field in manifest file, this could potentially result in upgrade of bootstrapper, whenever manifest section changes, which imho is undesired.

But having a version for each section might also proves undesired.
One solution could be to use major number, whenever bootstrapper must be updated, and minor number for changes only related to sections parsed by `west` tool (e.g. manifest section)

mbolivar:
Downgrading to low for now

Issue #221, opened 2019-03-02
West manifest imports


mbolivar:
This issue has evolved over time following a lot of discussion. Its eventual scope settled down to an enhancement to west which allows users to import content from other manifest files into their west.yml.

The final "resolved" manifest contains its own projects, plus projects obtained via imports. This resolved manifest is the logical data which commands like `west list`, `west update`, etc. operate on.

Imports can come from:

- within the manifest repository itself (if I want to split up my manifest into multiple files in one repository)
- a project in my manifest file's `projects:` list (if e.g. I want my own manifest which treats the zephyr repository as a project, and import zephyr/west.yml into my own manifest to get its modules)
- a combination of these

Imports are recursive: I can import manifest A, which itself imports B, and get the projects from B.

The Zephyr [documentation PR 20433](https://github.com/zephyrproject-rtos/zephyr/pull/20433) is the authoritative definition of the scope of this issue (i.e. of west #221).

Definition of done:

- [x] add initial support in west https://github.com/zephyrproject-rtos/west/pull/330/
- [x] documentation for initial support level https://github.com/zephyrproject-rtos/zephyr/pull/20433
- [x] cut a beta release (note: done as 0.6.99.dev4)
- [x] collect user feedback
- [x] implement 'map' imports (planned feature not provided initially)
- [x] update documentation to reflect 'map' import final implementation
- [x] do another beta (note: done as 0.6.99.dev5)
- [x] west.manifest unit tests for all major use cases (see https://github.com/zephyrproject-rtos/west/pull/363)
- [x] integration tests for built-in west commands for all major use cases (see https://github.com/zephyrproject-rtos/west/pull/363)
- [x] collect final user feedback

Specific items to address in follow up work:

- [x] west manifest --resolve implementation (https://github.com/zephyrproject-rtos/west/pull/348)
- [x] whether we can load files from the file system when importing from projects, e.g. https://github.com/zephyrproject-rtos/west/pull/330#issuecomment-566982785 (answer: [no](https://github.com/zephyrproject-rtos/west/issues/221#issuecomment-567401732))
- [x] double-processing some projects in west update https://github.com/zephyrproject-rtos/west/pull/330#pullrequestreview-333329703 (addressed in https://github.com/zephyrproject-rtos/west/pull/364)
[...]

PR #321, opened 2019-10-09
Manifest version check

mbolivar:
Fixes: #126 

The main point of this is to add support for a version field in the manifest:

```
manifest:
     version: 1.0
```

And have west fail unless it's version 1.0 or later in the above example.

I'm really not happy with `commands: handle errors if minimum manifest version is unmet` yet, though.

It gets the job done, but leads to weird output like this:

```
$ west init
WARNING: manifest "/home/mbolivar/zp/zephyr/west.yml" requires west v0.7; you have v0.6.99.
  This will fail if manifest data is required.
  Upgrade west to silence this warning.
FATAL ERROR: already initialized in /home/mbolivar/zp, aborting.

$ west manifest --validate
WARNING: manifest "/home/mbolivar/zp/zephyr/west.yml" requires west v0.7; you have v0.6.99.
  This will fail if manifest data is required.
  Upgrade west to silence this warning.
FATAL ERROR: you must upgrade to west v0.7 or later

$ west list
WARNING: manifest "/home/mbolivar/zp/zephyr/west.yml" requires west v0.7; you have v0.6.99.
  This will fail if manifest data is required.
  Upgrade west to silence this warning.
FATAL ERROR: can't run west list: the manifest was required but could not be read.
  You must upgrade west to run this command.
```

I think this should get cleaned up a bit but thought I'd ask for feedback.

### DIFF
--- a/src/west/manifest-schema.yml
+++ b/src/west/manifest-schema.yml
@@ -17,6 +17,20 @@
 
 type: map
 mapping:
+  # The "version" key is optional, and specifies a minimum version of
+  # west (in semantic versioning, so X.Y <= X.Y.0 < X.Y.1 < (X+1).Y,
+  # etc.) that is required to correctly parse the manifest data.
+  #
+  # This is supported starting with west v0.7 (i.e. it was added after
+  # 0.6 was released). Earlier versions of west will treat manifest
+  # data with a version key as malformed, so it can reliably be used
+  # to prevent parsing on west v0.5 and v0.6 as well, though the error
+  # messages users will receive will be a bit more confusing in that
+  # case.
+  version:
+    required: false
+    type: text
+
   # The "defaults" key specifies some default values used in the
   # rest of the manifest.
   defaults:
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -142,7 +142,8 @@ class Manifest:
     def from_data(source_data, manifest_path=None, topdir=None):
         '''Create and return a new Manifest object given parsed YAML data.
 
-        :param source_data: Parsed YAML data as a Python object.
+        :param source_data: Parsed YAML data as a Python object,
+                            or a string with YAML data inside.
         :param manifest_path: fallback ManifestProject path attribute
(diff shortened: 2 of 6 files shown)


mbolivar:
@carlescufi I am doing this because I think we should fix #126 before tackling #221

src/west/manifest-schema.yml
```diff
@@ -17,6 +17,20 @@
...
+  # west (in semantic versioning, so X.Y <= X.Y.0 < X.Y.1 < (X+1).Y,
+  # etc.) that is required to correctly parse the manifest data.
+  #
+  # This is supported starting with west v0.7 (i.e. it was added after
+  # 0.6 was released). Earlier versions of west will treat manifest
+  # data with a version key as malformed, so it can reliably be used
+  # to prevent parsing on west v0.5 and v0.6 as well, though the error
+  # messages users will receive will be a bit more confusing in that
+  # case.
+  version:
```
carlescufi: Should this be `manifest-version:` ?

src/west/manifest.py
```diff
@@ -228,6 +233,24 @@ def __init__(self, source_file=None, source_data=None,
...
+            # This avoids forcing them to write:
+            #
+            #  version: "1.0"
+            #
+            # by explicitly allowing:
+            #
+            #  version: 1.0
+            if isinstance(min_version, float):
+                min_version = str(min_version)
+            if parse_version(min_version) > _WEST_VERSION:
```
carlescufi: What happens with pre-release versions that have additional metadata after the semantic version? i.e. `0.6.3-rc1` for example.

src/west/manifest-schema.yml
mbolivar: If anything it should be `west-version`; should I change it?

src/west/manifest.py
mbolivar: They just work -- see the test cases.

mbolivar:
Closing this for a bit to self-NACK. I'm really not happy with this and have a better idea.

PR #329, opened 2019-11-04
Project rework

mbolivar:
@carlescufi these are the first patches in my manifest imports queue.

I'd like to get them in now so you can work on git submodules support using the updated Project API, so we don't have to rebase on top of each other.

### DIFF
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -787,9 +787,11 @@ def _update(project, fetch, rebase, keep_descendants):
 def _clone(project):
     log.small_banner(project.format('{name}: cloning and initializing'))
     project.git('init {abspath}', cwd=util.west_topdir())
-    # This remote is only added for the user's convenience. We always fetch
-    # directly from the URL specified in the manifest.
-    project.git('remote add -- {remote_name} {url}')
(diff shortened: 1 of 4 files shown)


src/west/commands/project.py
```diff
@@ -787,9 +787,11 @@ def _update(project, fetch, rebase, keep_descendants):
...
     log.small_banner(project.format('{name}: cloning and initializing'))
     project.git('init {abspath}', cwd=util.west_topdir())
-    # This remote is only added for the user's convenience. We always fetch
-    # directly from the URL specified in the manifest.
-    project.git('remote add -- {remote_name} {url}')
+    # The "origin" remote is only added for the user's convenience. We
+    # always fetch directly from the URL specified in the manifest.
+    # Later changes to the manifest won't change this remote,
+    # unfortunately -- the user will have to fix that themselves.
+    project.git('remote add -- origin {url}')
```
carlescufi: I commented on that in #330. Let's solve this before we accept this change.

carlescufi: Resolved offline. This is a key part of manifest imports, and logically the simplest and cleanest way to solve the problem.

PR #324, opened 2019-10-10
Manifest version check

mbolivar:
A new attempt. GitHub won't let me reopen the previous pull request https://github.com/zephyrproject-rtos/west/pull/321, sorry.

Fixes: #126 

There are 3 different error paths on incompatible versions:

- help path: warn that extension command help will be missing and most commands won't run is printed
- error path: commands which are known to require the manifest (update, list, ...) print a fatal error with similar info for the help path
- watch-out path: commands which don't need the manifest (config, topdir) print a warning that they ought to work, but an upgrade is needed

Examples:

```
$ west [help,--help,-h]
WARNING: west v0.7 or later is required by the manifest
  West version: v0.6.99
  Manifest file: /home/mbolivar/zp/zephyr/west.yml
  Cannot get extension command help, and most commands won't run.
  To silence this warning, upgrade west.
usage: west [-h] [-z ZEPHYR_BASE] [-v] [-V] <command> ...
[... rest of help output ...]

$ west {update,list,manifest,diff,status,forall,build,flash,debug}
FATAL ERROR: west v0.7 or later is required by the manifest
  West version: v0.6.99
  Manifest file: /home/mbolivar/zp/zephyr/west.yml
  Please upgrade west and retry.

$ west config manifest.path
WARNING: west v0.7 or later is required by the manifest
  West version: v0.6.99
  Manifest file: /home/mbolivar/zp/zephyr/west.yml
  This should work, but most commands won't.
  To silence this warning, upgrade west.
zephyr

$ west topdir
WARNING: west v0.7 or later is required by the manifest
  West version: v0.6.99
  Manifest file: /home/mbolivar/zp/zephyr/west.yml
  This should work, but most commands won't.
  To silence this warning, upgrade west.
/home/mbolivar/zp
```

PR #330, opened 2019-11-05
Initial support for manifest imports

mbolivar:
This is incomplete, because support for `import` keys with map values is not supported. However, you can use import keys which are:

- booleans
- files
- directories
- sequences of files and directories

Basic test cases at the west.manifest API level are supported, but I want to write much more detailed test cases before we release a version of west with these features.

However, I think this is ready for review and initial merge to get the feature into the hands of users and start collecting real feedback.

### DIFF
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -194,21 +194,13 @@ class Init(_ProjectCommand):
         if shutil.which('git') is None:
             log.die("can't find git; install it or ensure it's on your PATH")
 
-        # west.manifest will try to read manifest.path and use it when
-        # parsing the manifest. Clear it out for now so we can parse
-        # the manifest without it; local() or bootstrap() will set it
-        # properly.
-        if config.get('manifest', 'path', fallback=None) is not None:
-            config.remove_option('manifest', 'path')
         if args.local:
-            projects, manifest_dir = self.local(args)
+            topdir = self.local(args)
         else:
-            projects, manifest_dir = self.bootstrap(args)
-
-        self.fixup_zephyr_base(projects)
+            topdir = self.bootstrap(args)
 
         log.banner('Initialized. Now run "west update" inside {}.'.
-                   format(self.topdir))
+                   format(topdir))
 
     def local(self, args):
         if args.manifest_rev is not None:
@@ -220,24 +212,19 @@ class Init(_ProjectCommand):
         rel_manifest = basename(manifest_dir)
         west_dir = os.path.join(topdir, WEST_DIR)
 
+        if not exists(manifest_file):
+            log.die('can\'t init: no "west.yml" found in {}'.
+                    format(manifest_dir))
+
         log.banner('Initializing from existing manifest repository',
                    rel_manifest)
-        if not exists(manifest_file):
-            log.die('No "west.yml" found in {}'.format(manifest_dir))
-
+        log.small_banner('Creating {} and local configuration file'.
+                         format(west_dir))
         self.create(west_dir)
         os.chdir(topdir)
-        # This validates the manifest. Note we cannot use
-        # self.manifest from west init, as we are in the middle of
-        # creating the installation right now.
-        projects = self.projects(manifest_file)
-        log.small_banner('Creating {} and local configuration'.
-                         format(west_dir))
         update_config('manifest', 'path', rel_manifest)
 
-        self.topdir = topdir
-
-        return projects, manifest_dir
+        return topdir
 
     def bootstrap(self, args):
         manifest_url = args.manifest_url or MANIFEST_URL_DEFAULT
@@ -248,52 +235,44 @@ class Init(_ProjectCommand):
         log.banner('Initializing in', topdir)
         if not isdir(topdir):
             self.create(topdir, exist_ok=False)
-        os.chdir(topdir)
 
-        # Clone the manifest repository into a temporary directory. It's
-        # important that west_dir exists and we're under topdir, or we
-        # won't be able to call self.projects() without error later.
+        # Clone the manifest repository into a temporary directory.
         tempdir = join(west_dir, 'manifest-tmp')
         if exists(tempdir):
             log.dbg('removing existing temporary manifest directory', tempdir)
             shutil.rmtree(tempdir)
         try:
             self.clone_manifest(manifest_url, manifest_rev, tempdir)
-            temp_manifest_file = join(tempdir, 'west.yml')
-            if not exists(temp_manifest_file):
-                log.die('No "west.yml" in manifest repository ({})'.
-                        format(tempdir))
-
-            projects = self.projects(temp_manifest_file)
-            manifest_project = projects[MANIFEST_PROJECT_INDEX]
-            if manifest_project.path:
-                manifest_path = manifest_project.path
-                manifest_abspath = join(topdir, manifest_path)
-            else:
-                url_path = urlparse(manifest_url).path
-                manifest_path = posixpath.basename(url_path)
-                manifest_abspath = join(topdir, manifest_path)
-
-            log.dbg('moving', tempdir, 'to', manifest_abspath,
-                    level=log.VERBOSE_EXTREME)
-            shutil.move(tempdir, manifest_abspath)
-            log.dbg('setting manifest.path to', manifest_abspath,
-                    level=log.VERBOSE_EXTREME)
-            update_config('manifest', 'path', manifest_path)
-        finally:
+        except subprocess.CalledProcessError:
             shutil.rmtree(tempdir, ignore_errors=True)
+            raise
 
-        self.topdir = topdir
+        # Verify the manifest file exists.
+        temp_manifest = join(tempdir, 'west.yml')
+        if not exists(temp_manifest):
+            log.die('can\'t init: no "west.yml" found in {}\n'
+                    '  Hint: check --manifest-url={} and --manifest-rev={}\n'
+                    '  You may need to remove {} before retrying.'.
+                    format(tempdir, manifest_url, manifest_rev, west_dir))
 
-        return projects, manifest_project.abspath
+        # Parse the manifest to get the manifest path, if it declares one.
+        # Otherwise, use the URL.
+        projects = Manifest.from_file(temp_manifest, topdir=topdir).projects
+        manifest_project = projects[MANIFEST_PROJECT_INDEX]
+        if manifest_project.path:
+            manifest_path = manifest_project.path
+        else:
+            manifest_path = posixpath.basename(urlparse(manifest_url).path)
+        manifest_abspath = join(topdir, manifest_path)
 
-    def projects(self, manifest_file):
-        return Manifest.from_file(manifest_file).projects
+        log.dbg('moving', tempdir, 'to', manifest_abspath,
+                level=log.VERBOSE_EXTREME)
+        shutil.move(tempdir, manifest_abspath)
+        log.dbg('setting manifest.path to', manifest_path,
+                level=log.VERBOSE_EXTREME)
+        update_config('manifest', 'path', manifest_path, topdir=topdir)
 
-    def fixup_zephyr_base(self, projects):
-        for project in projects:
-            if project.path == 'zephyr':
-                update_config('zephyr', 'base', project.path)
+        return topdir
 
     def create(self, directory, exist_ok=True):
         try:
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -268,8 +268,7 @@ class Init(_ProjectCommand):
         log.dbg('moving', tempdir, 'to', manifest_abspath,
                 level=log.VERBOSE_EXTREME)
         shutil.move(tempdir, manifest_abspath)
-        log.dbg('setting manifest.path to', manifest_path,
-                level=log.VERBOSE_EXTREME)
+        log.small_banner('setting manifest.path to', manifest_path)
         update_config('manifest', 'path', manifest_path, topdir=topdir)
 
         return topdir
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -20,7 +20,8 @@ from west.configuration import config, update_config
 from west import log
 from west import util
 from west.commands import WestCommand, CommandError
-from west.manifest import Manifest, MANIFEST_PROJECT_INDEX, ManifestProject
+from west.manifest import ImportFlag, Manifest, MANIFEST_PROJECT_INDEX, \
+    ManifestProject, _manifest_content_at
 from west.manifest import MANIFEST_REV_BRANCH as MANIFEST_REV
 from west.manifest import QUAL_MANIFEST_REV_BRANCH as QUAL_MANIFEST_REV
 from urllib.parse import urlparse
@@ -256,8 +257,11 @@ class Init(_ProjectCommand):
                     format(tempdir, manifest_url, manifest_rev, west_dir))
 
         # Parse the manifest to get the manifest path, if it declares one.
-        # Otherwise, use the URL.
-        projects = Manifest.from_file(temp_manifest, topdir=topdir).projects
+        # Otherwise, use the URL. Ignore imports -- all we really
+        # want to know is if there's a "self: path:" or not.
+        projects = Manifest.from_file(temp_manifest,
+                                      import_flags=ImportFlag.IGNORE,
+                                      topdir=topdir).projects
         manifest_project = projects[MANIFEST_PROJECT_INDEX]
         if manifest_project.path:
             manifest_path = manifest_project.path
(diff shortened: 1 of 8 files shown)


mbolivar:
@carlescufi these are the patches I'd like design feedback on:

- `[wip] manifest: add infrastructure for "simple" imports`
- `[wip] project: add import support to 'west update'`

I started filling in the missing pieces in `west update` but don't have anything working well enough to share yet.

mbolivar:
@carlescufi one last thing.

I wrote these to be mergeable as-is. Using west on manifests without `import` should work the same way as it does now. So if we like the design and want to merge them as a "getting started" PR, that's an option also.

src/west/manifest-schema.yml
```diff
@@ -118,3 +122,8 @@ mapping:
       west-commands:
         required: false
         type: str
+      # Additional manifest data to import from the manifest
+      # repository.
+      import:
```
carlescufi: @tejlmand and I were discussing about this. I think you added the `import` key in the `self` section in order to be able to import additional `.yml` manifests that are present directly in the file system. This is useful for users adding local projects without having to modify the manifest file and also for CI in order to add ci-specific manifests.

I think there are several solutions to this problem:

- Have a hardcoded path `west_topdir/.manifests` that will always be loaded by default, without requiring explicitly setting it into the manifest itself
- The solution you propose here, but I would probably rename `import` in the `self` section to `local-import` or similar
- A new command-line options `--extra-manifests` that allows users to specify the same but via the command line (ideal for CI)

How do you see this proposal you have being used in the context of a downstream? If it's for CI, then the users will be exposed to a folder listed there that has no relevance to them and in the case of CI itself running, there will no indication of whether the extra CI repos have been cloned or not.

src/west/manifest.py
```diff
@@ -537,14 +573,72 @@ def _load_project(self, pdata):
...
+        return project, import_files
+
+    def _imported_files(self, project, import_key):
+        ktype = type(import_key)
+
+        if not import_key or (isinstance(project, ManifestProject) and
+                              ktype == bool):
+            return []
+
+        if ktype == bool:
```
carlescufi: What does `import: true` actually mean in the context of the `self` section?

src/west/manifest.py
```diff
@@ -482,9 +491,24 @@ def _load_self(self, path_hint):
...
             assert self.topdir
 
-        return ManifestProject(path=slf.get('path', path_hint),
-                               topdir=self.topdir,
-                               west_commands=slf.get('west-commands'))
+        # Sanity-check /self/import. We don't do anything else at this
+        # stage: the imports themselves are done last, after
+        # /projects/X/imports and self._data['projects'] are both done.
+        slf_import = slf.get('import')
+        if not (slf_import is None or isinstance(slf_import, bool)):
```
carlescufi: Not clear to us what this check is for

src/west/commands/project.py
```diff
@@ -787,9 +798,11 @@ def _update(project, fetch, rebase, keep_descendants):
...
     log.small_banner(project.format('{name}: cloning and initializing'))
     project.git('init {abspath}', cwd=util.west_topdir())
-    # This remote is only added for the user's convenience. We always fetch
-    # directly from the URL specified in the manifest.
-    project.git('remote add -- {remote_name} {url}')
+    # The "origin" remote is only added for the user's convenience. We
+    # always fetch directly from the URL specified in the manifest.
+    # Later changes to the manifest won't change this remote,
+    # unfortunately -- the user will have to fix that themselves.
+    project.git('remote add -- origin {url}')
```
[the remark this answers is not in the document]
mbolivar: I know it's a big breaking change, and I know that will take some doc rewriting, but I don't have a better alternative.

We need to get rid of this because I believe there's no clean way to allow manifest imports with first-class remotes. If nrf/west.yml defines a remote R with url-base-A and imports zephyr/west.yml, which defines R to have url-base-B, there's no clean way to do this as an "override".

Projects in zephyr/west.yml are completely "inlined" *before* being imported into nrf/west.yml. At that point, the remotes disappear and the URL is what really matters.

If you look at the model for how manifest imports work here, you can see that only a subset of project data is considered for merging when we model them as tuples:

file:///home/mbolivar/zp/zephyr/doc/_build/html/guides/west/manifest.html#west-manifest-formal

The 'remote' is not part of that.

I did it this way because if we have both `remote` and `url` attributes in a west.yml entry for a project, there's no clear way to decide which one should win if a sub-manifest uses one, and a parent manifest uses the other.

So I basically "erase" the remotes after parsing each manifest and just override the URLs.

src/west/manifest.py
```diff
@@ -705,71 +814,86 @@ def __init__(self, name, defaults=None, path=None,
...
         '''Return a representation of this object as a dict, as it
         would be parsed from an equivalent YAML manifest.
         '''
-        ret = collections.OrderedDict(
-            (('name', self.name),
-             ('remote', self.remote.name if self.remote else 'None'),
-             ('revision', self.revision)))
-        if self.path != self.name:
-            ret['path'] = self.path
+        ret = collections.OrderedDict()
```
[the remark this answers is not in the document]
mbolivar: It ensures that the output manifest YAML will order the sections the same way they are added to the dictionary.

Dictionaries are ordered by default in CPython 3.6+, but we still support 3.4 and 3.5.

src/west/manifest.py
```diff
@@ -537,14 +573,72 @@ def _load_project(self, pdata):
...
+                            format(project.name, import_key, ktype))
+
+    def _import(self, project, file):
+        try:
+            data = project.file_contents_at(file)
+        except subprocess.CalledProcessError:
+            # Fall back on the importer.
+            data = self._importer(project, file, project.revision)
+
+        m = Manifest(source_data=data, manifest_path=project.abspath,
```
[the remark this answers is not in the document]
mbolivar: There's no loop detection here yet, but that's indeed where it should go.

(It can't go into the importer for west update, because it needs to be checked every time we import.)

src/west/manifest.py
```diff
@@ -537,14 +573,72 @@ def _load_project(self, pdata):
...
+                     topdir=self.topdir, importer=self._importer)
+
+        assert MANIFEST_PROJECT_INDEX == 0
+        for p in m.projects[1:]:
+            if p.name in self._projects_by_name:
+                self._projects_by_name[p.name]._merge(p)
```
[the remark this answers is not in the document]
mbolivar: This is by design.

Project names are the global unique identifiers that west uses as identities when doing importing. (I introduced the requirement that names be unique in 0.6 precisely to enable this.)

If you are importing a manifest whose names collide with your own, you can use a mapping with a `rename` key to avoid undesired collisions with your projects:

https://builds.zephyrproject.org/zephyrproject-rtos/zephyr/20433/docs/guides/west/manifest.html#west-manifest-import-map

src/west/manifest.py
```diff
@@ -537,14 +573,72 @@ def _load_project(self, pdata):
...
+        else:
+            self._malformed('{}: invalid import key {} (type {})'.
+                            format(project.name, import_key, ktype))
+
+    def _import(self, project, file):
+        try:
+            data = project.file_contents_at(file)
+        except subprocess.CalledProcessError:
+            # Fall back on the importer.
+            data = self._importer(project, file, project.revision)
```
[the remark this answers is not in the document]
mbolivar: > Do we assume that _importer() will actually update the current working tree or just fetch into the local refspace and then git show? 

At this level we make no assumptions about how the data are retrieved. We might not involve the filesystem at all. Try this:

```
$ wget -q -O- https://raw.github.com/zephyrproject-rtos/zephyr/v1.14.1/west.yml
```

This enables doing imports with `Manifest.from_data()` and not just `Manifest.from_file()`. We don't need to assume an installation exists anywhere on disk to parse the manifest data.

src/west/manifest-schema.yml
mbolivar: >  I think you added the import key in the self section in order to be able to import additional .yml manifests that are present directly in the file system. 

It was actually added to enable splitting up the manifest content into multiple files within a single manifest repository:

https://builds.zephyrproject.org/zephyrproject-rtos/zephyr/20433/docs/guides/west/manifest.html#west-manifest-ex2-2

I'm not trying to address additional manifest data in a general place on the file system, i.e. outside of a repository known to west (either the manifest repository or a project).

src/west/manifest.py
mbolivar: Nothing. It's ignored there.

https://builds.zephyrproject.org/zephyrproject-rtos/zephyr/20433/docs/guides/west/manifest.html#option-1-boolean

src/west/manifest.py
mbolivar: This code needs to be reworked. I was originally only going to support this feature if there was a real installation on disk, but then I realized that importers are general enough to not require that.

The purpose of this check is to say that if you're doing an import that has some effect (unlike `import: false` or `self: import: true`, which are ignored / have no effect), then you must have also provided the topdir of the installation.

But we should just call into the importer here instead.

src/west/manifest.py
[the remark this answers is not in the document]
mbolivar: > Also, should this really be called `importer`? This is not really importing anything but actually "fetching" a particular file at a particular revision. Should it be called `fetcher`?

It's the object that is used as a fallback from the `_import()` method when the default `git show` fails. It isn't necessarily fetching anything over the network -- the data could be coming from somewhere else on the file system, if the user of this API has some special knowledge of where to fetch the data from. (Remember that this API is callable from extensions too.)

The in-tree user of this API is `west update`. That code might be running before the repository is even cloned. If that's the case, I think the right thing to do is to clone the new repository into a temporary subdirectory of .west, because we don't know where this project should go yet -- the "parent" manifest might override its `path:`. Only when all the data are available and the manifest is completely resolved can `west update` move newly-cloned repositories into their final locations.

mbolivar:
@carlescufi @tejlmand thanks for the review! I've responded to your comments

mbolivar:
>     * [wip] manifest: add infrastructure for "simple" imports

I've taken the [wip] off of this commit as I now believe the basics are there and reasonably well tested.

We still don't support loop detection or import keys with map values, but I think that's not critical for getting a beta out.

I'll work on finishing the west update etc. portions tomorrow and then I think we're ready to go with this PR (and thus a beta).

mbolivar:
This latest push is able to init, update, and list when using `r/m` as the manifest repository containing the following:

```
$ find r/m -name '*yml' | xargs tail -n+0
==> r/m/west.yml <==
manifest:
  self:
    import: west.d
  defaults:
    remote: r
  remotes:
  - name: r
    url-base: /home/mbolivar/r
  projects:
  - name: p1

==> r/m/west.d/0.yml <==
manifest:
  defaults:
    remote: r
  remotes:
  - name: r
    url-base: /home/mbolivar/r
  projects:
  - name: p2

==> r/m/west.d/1.yml <==
manifest:
  defaults:
    remote: r
  remotes:
  - name: r
    url-base: /home/mbolivar/r
  projects:
  - name: p3
```

mbolivar:
> This latest push is able to init, update, and list when using `r/m` as the manifest repository containing [self imports]

This is still true, and the existing manifest import test cases are now updated (and passing) to reflect the new "ignore" semantics.

Now that self-imports are minimally sanity checked, I'm going to move on to project imports with the new semantics.

mbolivar:
> Now that self-imports are minimally sanity checked, I'm going to move on to project imports with the new semantics.

Progress from today: this latest force-push can init, update, list, and then run the build extension using this manifest:

```
manifest:
  defaults:
    remote: zephyrproject-rtos
  remotes:
  - name: zephyrproject-rtos
  projects:
  - name: zephyr
    revision: master
    import: true
```

Tomorrow I will work on another task, then Monday I will start exercising the west update behavior a bit more carefully. Then I think it's beta time.

mbolivar:
@carlescufi  @tejlmand I dropped DNM. This isn't ready for prime time, but I do think it's ready for a real review.

mbolivar:
Good catch, thanks!

tejlmand:
Great work: :+1: 

A small observation:
```
$ west update nrfxlib
FATAL ERROR: refusing to update just some projects because manifest imports are in use
  Project arguments: nrfxlib
  Manifest file with imports: /projects/github/ncs/nrf/west.yml
  Please run "west update" (with no arguments) instead.
```

If I remember correctly, only imported projects should not support individual updates, but projects directly specified in the top-level manifest file should still allow individual update, such as `nrfxlib` when `NCS` is the manifest repo.

Please correct me if both of you agrees no projects should support individual updates anymore.

Note: I don't see this as blocking for the PR, we can always extend functionality later.

tejlmand:
As mentioned in chat, should we allow special handling of `HEAD` or keep current behavior.
Today, if user specifies
```
    - name: zephyr
      repo-path: fw-nrfconnect-zephyr
      revision: HEAD
```
then `west` will do:
```

**Comment by tejlmand** [comment]
Just did some playing around.

> 
>     * Do not fetch remote but use local HEAD
> 
> 
> ```
>     - name: zephyr
>       repo-path: fw-nrfconnect-zephyr
>       west-commands: scripts/west-commands.yml
>       revision: HEAD
>       import: true
> ```

This is already supported if user does:

```
     - name: zephyr
       url: file:///<path-to>/zephyr
       revision: HEAD
       import: true
 ```
then `west` will fetch the local `HEAD` and update local `HEAD` to itself :smile:

src/west/manifest-schema.yml
tejlmand: @mbolivar can you provide a brief update on this.
How is it used in current PR ?

src/west/main.py
```diff
@@ -286,31 +296,35 @@ def run_command(self, argv):
...
                         '\n  Cannot get extension command help, ' +
                         "and most commands won't run." +
                         '\n  To silence this warning, upgrade west.')
             elif args.command in ['config', 'topdir']:
                 # config and topdir are safe to run, but let's
                 # warn the user that most other commands won't be.
-                log.wrn(mve_msg(self.mve, suggest_ugprade=False) +
+                log.wrn(mve_msg(self.mle, suggest_ugprade=False) +
                         "\n  This should work, but most commands won't." +
                         '\n  To silence this warning, upgrade west.')
```
tejlmand: Is this working as expected ?
When I run:
```
$ west topdir
/tmp/import_test
```
there is no warning.

but on other commands, I do get the help text, as example:
```
$ west status
FATAL ERROR: some imported manifest data is missing. To fix, run:
  west update
```

btw. I assume the hint is wrong. Why would updating `west` help silence the warning ?

src/west/commands/project.py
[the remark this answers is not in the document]
tejlmand: Not sure I understand this, or even why this had to be introduced:
https://github.com/zephyrproject-rtos/west/pull/329/files#r344823583

cause we only set the remote during `_clone` which only happens on first invocation of `west update` after init or when a new project is added to the manifest file.
https://github.com/mbolivar/west/blob/manifest-imports/src/west/commands/project.py#L691-L692

After that point we never touches the remote again, and I don't see this PR changes that behavior.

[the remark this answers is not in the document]
mbolivar: This was needed due to the original 'override' semantics. It's probably no longer needed in the new 'ignore' semantics you and Carles have suggested, but the code remained. We can look into adding them back in as follow-up work.

src/west/manifest-schema.yml
mbolivar: > @mbolivar can you provide a brief update on this.
> How is it used in current PR ?
> see the reason in this commit [7b204e9](https://github.com/zephyrproject-rtos/west/commit/7b204e900ce81b2715ca71a4fa12388ea5d0ae35)

Just in case it's not clear, the reason is for `self: import:`, i.e additional manifest files in the manifest repository itself, as shown in the example I linked to in the built documentation. That link has now expired, but I'll refresh it.

src/west/manifest.py
```diff
@@ -430,69 +490,179 @@ def _load(self, manifest, path_hint):
...
+            # we've implemented this. We'd like to avoid too many git
+            # commands, as subprocesses are slow on windows.
+            log.dbg('searching for the manifest repository root',
+                    level=log.VERBOSE_EXTREME)
+            repo_root = Path(mp.git('rev-parse --show-toplevel',
+                                    capture_stdout=True,
+                                    cwd=str(Path(self.path).parent)).
+                             stdout[:-1].      # chop off newline
+                             decode('utf-8'))  # hopefully this is safe
+        p = Path(repo_root) / imp
```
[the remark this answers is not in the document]
mbolivar: Yeah, this is a small but much appreciated bit of syntactic sugar.

src/west/manifest.py
```diff
@@ -430,69 +490,179 @@ def _load(self, manifest, path_hint):
...
+            # this happens too often, something may be wrong with how
+            # we've implemented this. We'd like to avoid too many git
+            # commands, as subprocesses are slow on windows.
+            log.dbg('searching for the manifest repository root',
+                    level=log.VERBOSE_EXTREME)
+            repo_root = Path(mp.git('rev-parse --show-toplevel',
+                                    capture_stdout=True,
+                                    cwd=str(Path(self.path).parent)).
+                             stdout[:-1].      # chop off newline
+                             decode('utf-8'))  # hopefully this is safe
```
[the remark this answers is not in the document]
mbolivar: I think if that path is found for most west commands, we're likely already in the fast path above, not the fallback path. Can you come up with a case where that's not true?

mbolivar:
> Please correct me if both of you agrees no projects should support individual updates anymore.

Yep, I want to do it that way, but would like to defer that to later on. We will need to tweak `Update.update_some` to use `west.manifest.ImportFlag.IGNORE`, and I'm not totally sure how to make the error messages user friendly, so I deferred this for now.

Please let me know if you'd like to do this now.

mbolivar:
> Note, this only happens when manifest file contains the `import: true` setting.

Nice catch, thanks. The error (and the other error message about updating west) were due to me only partially completing some changes following earlier review. They should be fixed; see handle_builtin_manifest_load_err in main.py.

src/west/main.py
mbolivar: Hint should be fixed, see handle_builtin_manifest_load_err. Thanks.

It's fine to run `west topdir` without a manifest, so the success there is expected. For `topdir` to succeed, a `.west` must be found, but that's it. On the other hand, `west status` really needs a manifest to do its job.

Part of these changes are deliberately splitting up the built-in command error handling depending on whether the commands actually need the manifest or not.

Extension commands by definition are excepted from this because if we can't parse the manifest, we don't know anything about extensions, since that's where they come from.

mbolivar:
@carlescufi @tejlmand thanks for review and testing.

I've responded to comments and collated follow up tasks in a new 'Specific items to address in follow up work' section at the bottom of the issue description in https://github.com/zephyrproject-rtos/west/issues/221. Please let me know if I missed anything you would like to see addressed before west 0.7.

mbolivar:
@carlescufi @tejlmand I've sorted out CI and updated the container to one that uses python 3.6 now.

Let me know if you'd rather see that as a separate PR.

tejlmand:
> > Please correct me if both of you agrees no projects should support individual updates anymore.
> 
> Yep, I want to do it that way, but would like to defer that to later on. We will need to tweak `Update.update_some` to use `west.manifest.ImportFlag.IGNORE`, and I'm not totally sure how to make the error messages user friendly, so I deferred this for now.
> 
> Please let me know if you'd like to do this now.

Deferring it for now is very acceptable :+1:

src/west/main.py
tejlmand: sorry, re-reading my comment I see it might be unclear.

What I was referring to was the message in L317:
```
`To silence this warning, upgrade west.'
```
where I'm unsure how upgrading `west` would help ?

src/west/manifest.py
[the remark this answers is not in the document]
tejlmand: > I think if that path is found for most west commands, we're likely already in the fast path above, not the fallback path. Can you come up with a case where that's not true?

No, I was just wondering that because you added a fallback mechanism here, you might have foreseen a risk, and in that case I just wondered if `.west/config` could be a better option instead of resolving to `git`

Especially as one can do a `west init -l <local-only-filesystem>`.
I don't think we mandate the manifest repo to be a git repo, it could be just local files.

src/west/main.py
```diff
@@ -112,21 +114,95 @@ def load_manifest(self):
...
+                    ctxt += '  Hint: for this to work:\n'
+                    ctxt += f'          - {p.name} must be cloned\n'
+                    ctxt += (f'          - its {MANIFEST_REV_BRANCH} ref '
+                             'must point to a commit with the missing file\n')
+                    ctxt += (f'  Hint: to make sure {p.name} '
+                             f'is cloned and fetch {rev} from:\n')
+                    ctxt += f'          {p.url}\n'
+                    ctxt += (f'        then point {MANIFEST_REV_BRANCH} '
+                             'there, run:\n')
+                    ctxt += '          west update'
```
tejlmand: Better but maybe a bit verbose:
```
FATAL ERROR: failed manifest import in zephyr (zephyr)
  Missing file: "west.yml" from revision "2c43bc81ea7b1de5332b806d99a718148375c981"
  Hint: for this to work:
          - zephyr must be cloned
          - its manifest-rev ref must point to a commit with the missing file
  Hint: to make sure zephyr is cloned and fetch 2c43bc81ea7b1de5332b806d99a718148375c981 from:
          https://github.com/NordicPlayground/fw-nrfconnect-zephyr
        then point manifest-rev there, run:
          west update
```

I think this is enough for users to get going and understand what was wrong:
```
FATAL ERROR: failed manifest import in zephyr (zephyr)
  Missing file: "west.yml" from revision "2c43bc81ea7b1de5332b806d99a718148375c981"
  Hint: for this to work:
          - zephyr must be cloned
          - its manifest-rev ref must point to a commit with the missing file
       To fix, run:
          west update
```

src/west/main.py
mbolivar: > where I'm unsure how upgrading `west` would help ?

As far as I can tell, that message is only ever printed when the specific error is an instance of ManifestVersionError.

Am I missing a code path where it can be printed in some other situation? If not, upgrading west to the latest version is definitely a good fix to suggest to the user, since the latest west by definition will support the latest manifest version.

For context, since you weren't around when this work was done, ManifestVersionError was added as part of closing https://github.com/zephyrproject-rtos/west/issues/126 (which you opened) in https://github.com/zephyrproject-rtos/west/pull/324 (which I submitted).

src/west/main.py
mbolivar: In trying that out, I found that won't always fix it, and we may have bigger problems. If manifest-rev points at the commit, but that commit doesn't have that file, then `git ls-tree SHA nonexistent-commit` just prints nothing, causing west to crash. Will fix.

src/west/manifest.py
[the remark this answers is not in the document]
mbolivar: > No, I was just wondering that because you added a fallback mechanism here, you might have foreseen a risk, and in that case I just wondered if `.west/config` could be a better option instead of resolving to `git`

It wasn't foreseen, it was run into in practice in west init, actually. I have no crystal ball ;).

> Especially as one can do a `west init -l <local-only-filesystem>`.
> I don't think we mandate the manifest repo to be a git repo, it could be just local files.

That still assumes it's a manifest repository, i.e. a git repository.

src/west/main.py
mbolivar: > Will fix.

My initial idea for a fix didn't work out. I'll have to spend more time on this tomorrow.

mbolivar: OK, I fixed it well enough for now. In true American cowboy fashion, I'm going to merge this and go on vacation for 2 weeks if CI passes.

tejlmand: you mean commit and run style :laughing:

PR #331, opened 2019-11-07
git: keep the refs space locally when fetching non-SHA revisions

tejlmand:
Example where fetch of non SHA revisions is kept locally.
Test not updated.

Inspired by: https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/1437

To allow later usage of the fetched refs space then git fetch will
always keep a local ref space even fetching non-SHA refs spaces, such
as branches and PRs.

### DIFF
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -24,6 +24,7 @@ from west.manifest import Manifest, MalformedManifest, MalformedConfig, \
     MANIFEST_PROJECT_INDEX, ManifestProject
 from west.manifest import MANIFEST_REV_BRANCH as MANIFEST_REV
 from west.manifest import QUAL_MANIFEST_REV_BRANCH as QUAL_MANIFEST_REV
+from west.manifest import WEST_REF_SPACE as WEST_REF
 from urllib.parse import urlparse
 import posixpath
 
@@ -726,6 +727,21 @@ class SelfUpdate(_ProjectCommand):
 # Private helper routines.
 #
 
+def _clean_west_refspace(project):
+    # Cleanup fetch if sha and everything != current revision.
+    west_refs_cmd = ('for-each-ref --format="%(refname)" -- ' +
+                     WEST_REF + '**')
+    cp = project.git(west_refs_cmd, capture_stdout=True)
+    west_references = cp.stdout.decode('utf-8').strip()
+
+    for ref in west_references.splitlines():
+        if ref != WEST_REF + project.revision:
+            update_ref_cmd = ('update-ref -m "west fetch: entry for ' +
+                              '{}" {} {}'.format(ref, QUAL_MANIFEST_REV, ref))
+            project.git(update_ref_cmd)
+            update_ref_cmd = 'update-ref -d ' + ref
+            project.git(update_ref_cmd)
+
 def _maybe_sha(rev):
     # Return true if and only if the given revision might be a SHA.
 
@@ -746,6 +762,8 @@ def _update(project, fetch, rebase, keep_descendants):
         log.dbg('skipping unnecessary fetch')
         project.git('update-ref ' + QUAL_MANIFEST_REV +
                     ' {revision}^{{commit}}')
+
+    _clean_west_refspace(project)
 
     try:
         sha = project.sha(QUAL_MANIFEST_REV)
@@ -865,12 +883,12 @@ def _fetch(project):
     update_cmd = 'update-ref ' + QUAL_MANIFEST_REV + ' '
     if _maybe_sha(project.revision):
         # Don't fetch a SHA directly, as server may restrict from doing so.
-        fetch_cmd += 'refs/heads/*:refs/west/*'
+        fetch_cmd += 'refs/heads/*:' + WEST_REF + '*'
         update_cmd += '{revision}'
     else:
         # The revision is definitely not a SHA, so it's safe to fetch directly.
         # This avoids fetching unnecessary ref space from the remote.
-        fetch_cmd += '{revision}'
+        fetch_cmd += '{revision}:' + WEST_REF + '{revision}'
         update_cmd += 'FETCH_HEAD^{{commit}}'
 
     log.small_banner(project.format(msg))
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -35,6 +35,11 @@ MANIFEST_REV_BRANCH = 'manifest-rev'
 #: A fully qualified reference to `MANIFEST_REV_BRANCH`.
 QUAL_MANIFEST_REV_BRANCH = 'refs/heads/' + MANIFEST_REV_BRANCH
 
+#: The west refs space used for current manifest revision.
+#: West ref space is temporary space for fetching of SHAs.
+#: The refspace also useable for future west operations, such as west rebase.
+WEST_REF_SPACE = 'refs/west/'
+
 #: The latest manifest schema version supported by this west program.
 #:
 #: This value changes when a new version of west includes new manifest


tejlmand:
Oh, tests passes, guess we should improve the unit test cases ;)

src/west/commands/project.py
```diff
@@ -868,7 +868,7 @@ def _fetch(project):
     else:
         # The revision is definitely not a SHA, so it's safe to fetch directly.
         # This avoids fetching unnecessary ref space from the remote.
-        fetch_cmd += '{revision}'
+        fetch_cmd += '{revision}:refs/west/{revision}'
```
mbolivar: You know, it's funny, I've been wondering why west/xyz refs were being tracked in the above branch but didn't chase it down.

This is technically illegal behavior. We currently promise that we only ever set the `manifest_rev` ref, so the `west/` namespace is not available to us.

We could of course relax that restriction with good reason, but I want to think about what problems this solves a bit.

tejlmand: this was done here :https://github.com/zephyrproject-rtos/west/pull/188
in order to solve the issue of a SHA being specified, but that SHA is not available on the remote HEAD (usually master).
If the SHA is located on a branch, we wouldn't have it locally.

Secondly, github does not allow SHAs to be fetch directly.
https://github.com/zephyrproject-rtos/west/issues/139

tejlmand: so guess we need to relax this restriction a bit, if we want `west` to work ;)

mbolivar: > this was done here :#188

I have a goldfish's memory :). Thanks for the pointer.

> Secondly, github does not allow SHAs to be fetch directly. #139

Right, but I thought this was for refs like pull/x/head, not SHAs, right?

> so guess we need to relax this restriction a bit, if we want west to work ;)

I'm confused about how this helps pull/x/head, since `west/pull/x/head` is a different local ref name for the same commit.

tejlmand: maybe I started the confusing by not quoting what I answered.

> You know, it's funny, I've been wondering why west/xyz refs were being tracked in the above branch but didn't chase it down.

The reason for tracking `west/xyz` was PR#188, especially a git does not allow SHAs to be fetched directly, which is described here #139, and therefore we just fetch as PR#140:
```
  if _is_sha(project.revision):
        # Don't fetch a SHA directly, as server may restrict from doing so.
        _git(project, fetch_cmd + ' --tags -- (url)')
        _git(project, 'update-ref (qual-manifest-rev-branch) (revision)')
```
but this only fetch remote HEAD, which is pointing to `master` in this case.
So if the SHA was on a branch, it wouldn't be fetched, which lead to: PR#188.
Where the remote `refs/heads/*` is kept locally under `refs/west/*` and therefore we can checkout any SHA.
You even suggested, just using `refs/west/` https://github.com/zephyrproject-rtos/west/pull/188#discussion_r254361240
With PR#188, we can checkout any SHA.

> This is technically illegal behavior. We currently promise that we only ever set the manifest_rev ref, 

and `manifest-rev` is still the only head we use under the repos normal refs space: `refs/heads/manifest-rev`, nothing has changed here, and I don't propose to change that.
Only proposal is to keep a local refs space under `refs/west/*`  for anything we fetch, so that we can use that later.

> so the west/ namespace is not available to us.

Technically, you're right. `west` is not using the knowledge that a commit is available under `refs/west/*` directly. `west` just checks out the SHA and points `manifetst-rev` to point to that SHA. It doesn't care whether the SHA is available locally, or remotely.
Actually, you can specify any SHA you have locally in your repo, even one that doesn't exists on the remote side, and `west` will still work, regardsless of whether that SHA is available remote or not.

So if we start using this knowledge in https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/1437 that is correctly a change in how we use this refs-space.
But as you are only using that refs space `git cat-file` we have not changed the behavior of `west` and we are still not checking out from `refs/west/*`, you are only using it to get information out.

So we are keeping a little more of the remote refs space, locally.
And we are not changing how `west` uses this information, only how an extension command is using that info ;)

mbolivar: Right, I understand that git doesn't allow fetching SHAs directly.

However, the point of the `manifest-rev` ref is that it holds a pointer to the latest `revision` in west.yml, even if it is a SHA. So at least for the most recent `west update`, we do not need `west/*` to support SHA revisions.

Now, if the revisions are all SHAs and they dance around as a repository is rebased, we have a problem. Is that what this is solving?

tejlmand: > Right, I understand that git doesn't allow fetching SHAs directly.
> 
> However, the point of the `manifest-rev` ref is that it holds a pointer to the latest `revision` in west.yml, even if it is a SHA. So at least for the most recent `west update`, we do not need `west/*` to support SHA revisions.
> 
> Now, if the revisions are all SHAs and they dance around as a repository is rebased, we have a problem. Is that what this is solving?

No, it is solving the issue that a SHA may not be fetched if it is on a remote branch.
Imagine that `west.yml` points to a specific SHA, let's use: 
https://github.com/NordicPlayground/fw-nrfconnect-nrf/commit/618add172a20deb96c45a1135bfd93c569d1955c
in this example, and do what `west` would do without PR#188.

```
$ mkdir /tmp/nrf
$ cd /tmp/nrf
$ git init
$ git fetch --tags -- https://github.com/NordicPlayground/fw-nrfconnect-nrf.git 
remote: Enumerating objects: 28, done.
remote: Counting objects: 100% (28/28), done.
remote: Compressing objects: 100% (25/25), done.
remote: Total 19531 (delta 5), reused 8 (delta 3), pack-reused 19503
Receiving objects: 100% (19531/19531), 8.07 MiB | 5.59 MiB/s, done.
Resolving deltas: 100% (13301/13301), done.
From https://github.com/NordicPlayground/fw-nrfconnect-nrf
 * branch              HEAD       -> FETCH_HEAD
...
$ git checkout 9e305316b0d8ef95b0b7c77457a811c75cc48a7e
fatal: reference is not a tree: 9e305316b0d8ef95b0b7c77457a811c75cc48a7e
```
cause the SHA is not on remote HEAD, therefore we fetch the whole refs space into the local ref space `refs/west/*` so that we have all SHAs locally.
As PR#188 adds: `refs/heads/*:refs/west/*`
```
$ git fetch --tags -- https://github.com/NordicPlayground/fw-nrfconnect-nrf.git refs/heads/*:refs/west/*
remote: Enumerating objects: 744, done.
remote: Counting objects: 100% (744/744), done.
remote: Compressing objects: 100% (27/27), done.
remote: Total 1226 (delta 722), reused 723 (delta 716), pack-reused 482
Receiving objects: 100% (1226/1226), 355.50 KiB | 1.09 MiB/s, done.
Resolving deltas: 100% (862/862), completed with 274 local objects.
From https://github.com/NordicPlayground/fw-nrfconnect-nrf
...
$ git checkout 9e305316b0d8ef95b0b7c77457a811c75cc48a7e
Note: checking out '9e305316b0d8ef95b0b7c77457a811c75cc48a7e'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b <new-branch-name>

HEAD is now at 9e305316 drivers: lte_lc: Fix invalid eDRX default value for NB-IoT
```

I hope this clarifies the issue with SHAs located on branches.

marc-hb: > If the SHA is located on a branch [...] So if the SHA was on a branch, [...] No, it is solving the issue that a SHA may not be fetched if it is on a remote branch. [...] I hope this clarifies the issue with SHAs located on branches.

If you had written NON-DEFAULT branch just once I would have understood you 2 or 3 times faster :-)

mbolivar: > > Now, if the revisions are all SHAs and they dance around as a repository is rebased, we have a problem. Is that what this is solving?
> 
> No, it is solving the issue that a SHA may not be fetched if it is on a remote branch.
> Imagine that `west.yml` points to a specific SHA, let's use:
> [NordicPlayground/fw-nrfconnect-nrf@618add1](https://github.com/NordicPlayground/fw-nrfconnect-nrf/commit/618add172a20deb96c45a1135bfd93c569d1955c)
> in this example, and do what `west` would do without PR#188.

I'm sorry I was unclear. By "is that what this is solving", I meant "is that what this [pull request #331] is solving"?

I understand the purpose of #188 :).

tejlmand: Nice, finally we get back on track discussing the actual content of this PR :)

> I'm sorry I was unclear. By "is that what this is solving", I meant "is that what this [pull request #331] is solving"?

no, the PR #331 is not solving that.
The PR's purpose is stated here: https://github.com/zephyrproject-rtos/west/pull/331#issue-337899162

> To allow later usage of the fetched refs space then git fetch will
always keep a local ref space even fetching non-SHA refs spaces, such
as branches and PRs.

and as we are already keeping remote stuff under `refs/west/*` we might as well keep whatever `west` fetches in there, because then we have it available for later use, and thus may reduce number of additional `git fetches` and hence improved performance.

Note, this is NOT about mirroring the remote refs space, simply just storing the information we anyway have fetched, so that we might avoid fetching the same again.

We are fetching the exact same amount of data with or without PR #331, the PR just have the potential to reduce later redundant fetches.

mbolivar: OK, gotcha. Can we make a deal? Will you send a PR to the west documentation saying that west "owns" `manifest-rev` *and* `refs/west/*` before we merge this?

mbolivar: >  Will you send a PR to the west documentation saying that west "owns" manifest-rev and refs/west/* before we merge this?

Or at least promise to do so :)

mbolivar: Actually, hold on. Sorry, I admit I didn't understand refspecs much until today when I RTFM a bit.

Given `revision: foo` in a manifest file, west expects that `foo` is a SHA, a branch, or a tag.

- if `foo` is a SHA, we are in the previous if block, this line is not executed
- if `foo` is a branch: assuming the remote `.git/foo` and `.git/refs/foo` does not exist, fetching plain `foo` gives us `refs/heads/foo` as desired, and it seems OK to stash that in our local `.git/refs/west/foo`, since we're doing that for SHAs with the `refs/heads/*:refs/heads/*` refspec above
- **but** if `foo` is a tag, say `v1.14.1`, this would create `refs/west/v1.14.1`, which is actually a tag object, created outside of `.git/refs/tags`

Try it:

```
$ git fetch https://github.com/zephyrproject-rtos/zephyr v1.14.0:refs/west/v1.14.0
 * [new tag]               v1.14.0    -> refs/west/v1.14.0
$ ls .git/refs/west/v1.14.0
.git/refs/west/v1.14.0
$ git cat-file -t refs/west/v1.14.0
tag
$ git cat-file -t west/v1.14.0
tag
```

But `git tag | grep west` is empty, no results.

Is that really the behavior we want? Serious question, I am really not sure.

mbolivar: (And yes, I know we are also fetching with `--tags` -- that's a separate matter.)

carlescufi: > since we're doing that for SHAs with the refs/heads/*:refs/heads/* refspec above

I assume you mean `refs/heads/*:refs/west/*` here

>OK, gotcha. Can we make a deal? Will you send a PR to the west documentation saying that west "owns" manifest-rev and refs/west/* before we merge this?

West already own `refs/west/*`, so this would be documenting existing behavior. It's unfortunate that Git doesn't allow us to fetch all refs without actually storing them anywhere, just getting the blobs and objects.

marc-hb: > but if foo is a tag,...

I think you meant annotated tag.

> say v1.14.1, this would create refs/west/v1.14.1, which is actually a tag object, created outside of .git/refs/tags

What are the concern(s)? I assumed git would never try to interpret anything outside heads/ and tags/, is this too optimistic? If it too optimistic then maybe just treat tags as a special case and fetch them separately with something equivalent to:  `{revision}^0:refs/west/{revision}`

Or maybe just apply the `^0` treatment to everything fetched, can't hurt...

> It's unfortunate that Git doesn't allow us to fetch all refs without actually storing them anywhere, just getting the blobs and objects.

But then they would go away on the next garbage collection, no?

tejlmand: >     * **but** if `foo` is a tag, say `v1.14.1`, this would create `refs/west/v1.14.1`, which is actually a tag object, created outside of `.git/refs/tags`

Partly right, but not fully.
Remember this line in our fetch: https://github.com/zephyrproject-rtos/west/blob/0d1dfa523fc4b06e421f2fc97d4d7c4febb8a063/src/west/commands/project.py#L862
So the `git fetch --tags -- {url} {revision}:refs/west/*` will ensure the tags still goes to `.git/refs/tags` ;)

tejlmand: Try for yourself:
```
$ ls .git/refs/tags/                                                                                
$ ls .git/refs/west/                                                                                   
ls: cannot access '.git/refs/west/': No such file or directory                                         
$ git tag                                                                                              
$ git fetch ncs -- v1.1.0:refs/west/v1.1.0                                                             
From https://github.com/NordicPlayground/fw-nrfconnect-nrf                                             
 * [new tag]           v1.1.0     -> refs/west/v1.1.0                                                  
 * [new tag]           v0.1.0     -> v0.1.0                                                            
 .....                                                                                                 
$ git tag                                                          
v0.1.0                                                                                                 
v1.1.0                                                                                                 
....                                                                                                   
$ ls .git/refs/tags/                                                                                   
v0.1.0  ... v1.1.0
$ ls .git/refs/west/
v1.1.0
```
`....` a lot of other tags, removed to reduce noise here.

tejlmand: and following up:
```
$ git cat-file -t refs/tags/v1.1.0
tag
```
so is it a problem we keep an additional ref in the `west` ref-space ?

tejlmand: > >    Will you send a PR to the west documentation saying that west "owns" manifest-rev and refs/west/* before we merge this?

> Or at least promise to do so :)

I can promise to send a PR describing that `west` also own the `refs/west` space.
But as that PR would be documenting existing behavior, I don't see why it should be coupled to this PR, but that's just my opinion.

mbolivar: > What are the concern(s)? 

>  I guess, user will never need to worry or care about our additional `refs/west` space.

We're creating an unbounded amount of un-garbage-collectable refs here, including putting tags outside of refs/tags.

Beyond not being able to collect the refs when they are no longer needed, it just feels weird. I freely admit I'm just sketched out by not understanding the potential implications, and at least personally I'm bothered by all the `west/xyz` output I'm seeing showing up in `git log --oneline` output, for example. (E.g. all the backport branches if you have a downstream where zephyr isn't the manifest repository.)

Within the Nordic downstream, I have always created a tag at the old point in history before rewriting history in any repository managed by west. This ensures an explicit pointer to the ref is available in a place where a user can expect to find it. This feels like going behind the user's back in a weird way.

marc-hb:
> as we are already keeping remote stuff under refs/west/* we might as well keep whatever west fetches in there, because then we have it available for later use, and thus may reduce number of additional git fetches and hence improved performance. Note, this is NOT about mirroring the remote refs space, simply just storing the information we anyway have fetched, so that we might avoid fetching the same again.

> > It's unfortunate that Git doesn't allow us to fetch all refs without actually storing them anywhere, just getting the blobs and objects.

> But then they would go away on the next garbage collection, no?

> Beyond not being able to collect the refs when they are no longer needed, it just feels weird. [...]
> at least personally I'm bothered by all the west/xyz output I'm seeing showing up in git log --oneline output, for example. 

Connecting all the dots above, I just remembered that `git` already has a built-in feature to cache objects and automatically expire and garbage collect them  after some configurable time: it's the `reflog`. So here's the pseudo code of a maybe not so crazy idea:

```
git fetch  => refs/west/*

# too many uncollectable refs/west/*
# collapse them all in the reflog a single branch

for ref in  refs/west/*
  git update-ref refs/west/west_global_cache $ref
  git update-ref -d $ref
done
```

mbolivar:
> Connecting all the dots above, I just remembered that `git` already has a built-in feature to cache objects and automatically expire and garbage collect them after some configurable time: it's the `reflog`

Good point!

Since `manifest-rev` is reflogged already, do we need this then?

src/west/commands/project.py
tejlmand: > > What are the concern(s)?
> 
> > I guess, user will never need to worry or care about our additional `refs/west` space.
> 
> We're creating an unbounded amount of un-garbage-collectable refs here, including putting tags outside of refs/tags.
> 

Agree. And a very valid point indeed.
We should look into cleaning up this ref-space.
One option could be to simple wipe the `refs/west/` ref-space upon each `west update`.
That way we ensure remotely deleted branches are removed.
Removed tags are also removed, and so on.
The question is, how will this impact performance of fetching ?
And do we care, as the original reason for PR #188 was about being able to fetch the specific SHA, not about optimizing future fetches.

So if we keep performance out-of the loop, as this was never the intention, we can basically wipe whole `refs/west` as part of `west update`.
Or consider to just keep the refs currently pointed to be `manifest`, so that if manifest points to `my_branch` we keep that refs, as it can be useful for, eg. :
https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/1437
https://github.com/zephyrproject-rtos/west/issues/338 (have some rough ideas on using `refs/west` for checkpoints during a `west rebase`, that are wiped when rebase has completed).
Does that sound like an improvement ?

> Beyond not being able to collect the refs when they are no longer needed, it just feels weird.

Guess I can never take that feeling away from you :(

> and at least personally I'm bothered by all the `west/xyz` output I'm seeing showing up in `git log --oneline` output, for example. (E.g. all the backport branches if you have a downstream where zephyr isn't the manifest repository.)

Agree, valid point.
And cleaning out the parts of `refs/west` we don't actually need will for sure improve this.
Only reason I didn't do that in PR #188, was because I considered, `Why wipe it, if it improves performance of later fetches`.
So i'm fine with cleaning up the `refs/west` space, and then let `git gc` decide for how long potential SHAs are available.
It could mean additional fetches when switching between two very different manifest file, as example, switch between NCS and Zephyr as manifest project.
But I assume any degrade in fetch performance in this case is acceptable, as the performance will be similar to not having `refs/west` at all.

> 
> Within the Nordic downstream, I have always created a tag at the old point in history before rewriting history in any repository managed by west. This ensures an explicit pointer to the ref is available in a place where a user can expect to find it. 

And that is still true. We don't touch that tag, and it will be available until user removes it.
The tag is available where the user expects to find it.

> This feels like going behind the user's back in a weird way.

I disagree, we are not modifying anything the user has done.
But agree, current behavior would keep an additional ref to the same revision and that is what shows up in `git log`, and that seems to bother you, and I do acknowledge that, so will look into improving this.
And doing a little cleanup, as proposed above should help  :-)

tejlmand:
> > Connecting all the dots above, I just remembered that `git` already has a built-in feature to cache objects and automatically expire and garbage collect them after some configurable time: it's the `reflog`
> 
> Good point!
> 
> Since `manifest-rev` is reflogged already, do we need this then?

@marc-hb Good point on updating the reflog, and thus cleaning the `refs/west` will also help te ensure the SHAs are available for a certain time, while not have all those `refs/west` references popping up in `git log`.
As I reply in https://github.com/zephyrproject-rtos/west/pull/331#discussion_r351196763 , there needs to be done some cleanup in `refs/west`, and this appears to be a nice way of doing that.

src/west/commands/project.py
tejlmand: @mbolivar Just got to think.
When we do a `revision: pull/<no>/head` I actually got to think that this revision is NOT in the normal git refs-space.
Github does something similar to our `refs/west/*` refs space, and you can see that, as you can use the full refs as `git fetch <url> refs/pull/<no>/head`.
So special ref spaces, such as `refs/west/*` is not so special. (but must be cleaned up)

src/west/commands/project.py
```diff
@@ -726,6 +727,18 @@ def do_run(self, args, user_args):
...
+    # Cleanup fetch if sha and everything != current revision.
+    west_refs_cmd = ('for-each-ref --format="%(refname)" -- ' +
+                     WEST_REF + '**')
+    cp = project.git(west_refs_cmd, capture_stdout=True)
+    west_references = cp.stdout.decode('utf-8').strip()
+
+    for ref in west_references.splitlines():
+        if ref != WEST_REF + project.revision:
+            update_ref_cmd = 'update-ref -d ' + ref
+            project.git(update_ref_cmd)
```
tejlmand: @marc-hb @mbolivar 
This will clean out any rererences from the `refs/west` ref space, unless revision is a non-SHA and identical to current revision in the manifest file.

As all references west are removed, it means that all SHAs not referenced from elsewhere (as example a tag or local branch) will be a loose object.
Loose objects are per default pruned after two weeks.
https://git-scm.com/docs/git-gc 
I believe, that this approach should be safe, as two weeks is most likely long enough time for objects not referenced elsewhere.

But if desired, it could be considered to make an entry in reflog.
The disadvantage of that, is that it starts making a lot of `west` caused entries, which might be annoying for some users (personally I prefer that reflog is populated mainly to `git` operations I have made).

So I propose to wait for user feedback before populating the reflog.

The following code presents how the reflog could be populated.
Reflog default expires after 90 days.
```suggestion
            update_ref_cmd = 'update-ref --create-reflog -m"west entry: Updating HEAD to:"  HEAD ' + ref
            project.git(update_ref_cmd)
            update_ref_cmd = 'update-ref -d ' + ref
            project.git(update_ref_cmd)
```

tejlmand:
@marc-hb @mbolivar 
This PR has now been updated to clear all refs under `refs/west` after a fetch.
This creates loose objects for SHAs not referenced elsewhere, and those objects are garbage collected after 2 weeks (default).

If revision is non-SHA, it is accessible in: `refs/west/`.
As example, a manifest pointing having: `revision: <SHA>` will result in:
```
$ git log
dc9e1fb (HEAD, ncs/master, manifest-rev) mpsl: Enable zero latency IRQs for MPSL
4f70d0a nfc: documentation: removed obsolete statements
814a529 doc: remove wrong modem limitations in bsdlib doc
```

and when using a non-SHA, such as `pull/<no>/head`
```
$ git log
9ec8d9e (HEAD, refs/west/pull/120/head, manifest-rev) crypto: Updated nrf_oberon to version 3.0.3
dc9e1fb (ncs/master) mpsl: Enable zero latency IRQs for MPSL
4f70d0a nfc: documentation: removed obsolete statements
814a529 doc: remove wrong modem limitations in bsdlib doc
```

The benefit of having `refs/west/pull/120/head` available is that other `west` (extension-)commands can use the knowledge of manifest revision and `west` ref space to determine if `manifest-rev` is pointing to the non-SHA revision in manifest file.

Example, manifest:
```
   revision: pull/120/head
```
if `manifest-rev == refs/west/{project.revision}`, the repo has been updated.
if `manifest-rev != refs/west/{project.revision}`, manifest file has changed since last `west update` was executed.

src/west/commands/project.py
marc-hb: > This could be annoying for some users (personally I prefer that reflog is populated mainly to git operations I have made).

There is one reflog per branch. I didn't suggest to use the reflog of some regular user branch and while I haven't looked at it in detail, the `--create-reflog` in your sample code suggests you wouldn't be polluting any regular reflog either.

Using a reflog would offer some more power and flexibility: unlike loose objects it would let users find anything recent (just like in a regular reflog), it would let users override the default expiration time, maybe others.

tejlmand: > > This could be annoying for some users (personally I prefer that reflog is populated mainly to git operations I have made).
> 
> There is one reflog per branch. I didn't suggest to use the reflog of some regular user branch and while I haven't looked at it in detail, the `--create-reflog` in your sample code suggests you wouldn't be polluting any regular reflog either.
> 
> Using a reflog would offer some more power and flexibility: unlike loose objects it would let users find anything recent (just like in a regular reflog), it would let users override the default expiration time, maybe others.

Good point.
Will look a bit more into the reflog potential.
I haven't played much with reflogs in this way, mostly used reflog when I have done something I regret and want to go back to a place I don't remember ;-)

tejlmand: Now creating a reflog entry in the `manifest-rev` as this reflog is actually owned by `west`.
In order to keep the manifest reflog a little easier to examine, then all entries created during the fetch part of `west update`, will have the message: `west fetch: entry for {ref}`, where `{ref}` is of the form: `refs/west/<name>`.

mbolivar: Interesting, thanks! I will try to take another look at this on Monday morning my time.

mbolivar: > Now creating a reflog entry in the manifest-rev as this reflog is actually owned by west.

You mean this is already done, right? I see a reflog for manifest-rev in my existing west-managed git repos. For example, `git -C ncs/zephyr reflog manifest-rev` works for me.

> In order to keep the manifest reflog a little easier to examine, then all entries created during the fetch part of `west update`, will have the message: `west fetch: entry for {ref}`, where `{ref}` is of the form: `refs/west/<name>`.

It looks like that's true only if the user specifically does something like enable core.logAllRefUpdates, right?

```
LOGGING UPDATES
       If config parameter "core.logAllRefUpdates" is true and the ref is one
       under "refs/heads/", "refs/remotes/", "refs/notes/", or the symbolic
       ref HEAD; or the file "$GIT_DIR/logs/<ref>" exists then git update-ref
       will append a line to the log file "$GIT_DIR/logs/<ref>" (dereferencing
       all symbolic refs before creating the log name) describing the change
       in ref value. 
```

Or did you mean to push a change with `--create-reflog` added?

src/west/manifest.py
```diff
@@ -35,6 +35,11 @@
 #: A fully qualified reference to `MANIFEST_REV_BRANCH`.
 QUAL_MANIFEST_REV_BRANCH = 'refs/heads/' + MANIFEST_REV_BRANCH
 
+#: The west refs space used for current manifest revision.
+#: West ref space is temporary space for fetching of SHAs.
+#: The refspace also useable for future west operations, such as west rebase.
```
mbolivar: Can you remove this line since we haven't agreed west rebase is the name of a feature we will for sure implement?

src/west/commands/project.py
```diff
@@ -865,12 +883,12 @@ def _fetch(project):
...
     if _maybe_sha(project.revision):
         # Don't fetch a SHA directly, as server may restrict from doing so.
-        fetch_cmd += 'refs/heads/*:refs/west/*'
+        fetch_cmd += 'refs/heads/*:' + WEST_REF + '*'
         update_cmd += '{revision}'
     else:
         # The revision is definitely not a SHA, so it's safe to fetch directly.
         # This avoids fetching unnecessary ref space from the remote.
-        fetch_cmd += '{revision}'
+        fetch_cmd += '{revision}:' + WEST_REF + '{revision}'
```
mbolivar: Did we agree to use a reflog on refs/heads/manifest-rev to keep track of old versions of the `revision` field? That's what I thought we agreed to.

If so, why can't we just update manifest-rev directly in this case? That would avoid the need to update the west refspace entirely. We could still keep the code for SHAs as-is, but then also clean up after ourselves after we're done:

```
diff --git a/src/west/commands/project.py b/src/west/commands/project.py
index ca4361f..01c5e4d 100644
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -852,6 +852,9 @@ def _fetch(project):
     msg = "{name}: fetching, need revision {revision}"
     if project.clone_depth:
         msg += " with --depth {clone_depth}"
+
+    log.small_banner(project.format(msg))
+
     # -f is needed to avoid errors in case multiple remotes are present,
     # at least one of which contains refs that can't be fast-forwarded to our
     # local ref space.
@@ -860,20 +863,20 @@ def _fetch(project):
     fetch_cmd = ('fetch -f --tags ' +
                  ('--depth={clone_depth} ' if project.clone_depth else ' ') +
                  '-- {url} ')
-    update_cmd = 'update-ref ' + QUAL_MANIFEST_REV + ' '
     if _maybe_sha(project.revision):
         # Don't fetch a SHA directly, as server may restrict from doing so.
         fetch_cmd += 'refs/heads/*:refs/west/*'
-        update_cmd += '{revision}'
+        update_cmd = 'update-ref ' + QUAL_MANIFEST_REV + ' {revision}'
+        project.git(fetch_cmd)
+        project.git(update_cmd)
+        _clean_west_refspace(project)
     else:
-        # The revision is definitely not a SHA, so it's safe to fetch directly.
-        # This avoids fetching unnecessary ref space from the remote.
-        fetch_cmd += '{revision}'
-        update_cmd += 'FETCH_HEAD^{{commit}}'
-
-    log.small_banner(project.format(msg))
-    project.git(fetch_cmd)
-    project.git(update_cmd)
+        # The revision is definitely not a SHA, so it's safe to fetch
+        # directly into manifest-rev. This avoids fetching unnecessary
+        # refs from the remote, which is faster and saves bandwidth.
+        fetch_cmd += '{revision}:' + QUAL_MANIFEST_REV
+        update_cmd = None
+        project.git(fetch_cmd)
 
     if not _head_ok(project):
         # If nothing it checked out (which would usually only happen just after
```

Fetching directly into manifest-rev leaves a nice clean entry in my reflog:

```
$ git reflog -n1 manifest-rev
132130d17a (manifest-rev) manifest-rev@{0}: fetch https://github.com/zephyrproject-rtos/zephyr +master:refs/heads/manifest-rev: forced-update
```

Bonus, doing it that way means we can just delete *all* the `refs/west/*` using a bulk git operation after updating a "maybe-SHA" instead of looping over them individually, like discussed in this SO: https://stackoverflow.com/a/46229416

That would be faster on Windows, where process creation is expensive.

mbolivar: (Above patch is untested and contains a couple of bad lines, like the `update_cmd = None`, but hopefully makes my idea clear.)

src/west/commands/project.py
tejlmand: > > Now creating a reflog entry in the manifest-rev as this reflog is actually owned by west.
> 
> You mean this is already done, right? I see a reflog for manifest-rev in my existing west-managed git repos. For example, `git -C ncs/zephyr reflog manifest-rev` works for me.
>

Correct, a reflog for manifest-rev is already there with entries on each `git update-ref`.
What I meant, was that a useable entry is now created.
Currently `west update` will create entries such as:
```
a6a98b7854 (HEAD, refs/west/pull/240/head, manifest-rev) manifest-rev@{0}: 
b687d5d022 manifest-rev@{1}: 
a6a98b7854 manifest-rev@{2}:
b687d5d022 manifest-rev@{3}: 
69a4c73530 manifest-rev@{4}: 
b687d5d022 manifest-rev@{5}: 
....
``` 
when running `west update` and `git update-ref` is executed in background.

> > In order to keep the manifest reflog a little easier to examine, then all entries created during the fetch part of `west update`, will have the message: `west fetch: entry for {ref}`, where `{ref}` is of the form: `refs/west/<name>`.
> 
> It looks like that's true only if the user specifically does something like enable core.logAllRefUpdates, right?
> 

No, by actually specifying the `refs/west/<name>` we allow a user that is examine the reflog to actually see that `<shaA>` in the reflog originated from `pull/<no>/head` and `<shaB>` is from `<branch>`, and `shaC` is from same `pull/<no>/head` as `<shaA>`, which could indicate a forced push. 

So now there could be:
```
a6a98b7854 manifest-rev@{2}: west fetch: entry for refs/west/pull/240/head
eedc25e5b2 manifest-rev@{6}: west fetch: entry for refs/west/v2.0.99-ncs1-branch
7e64ac737d manifest-rev@{7}: west fetch: entry for refs/west/v1.14.99-ncs2-branch
1e34490406 manifest-rev@{8}: west fetch: entry for refs/west/nrf91
c373e5578c manifest-rev@{9}: west fetch: entry for refs/west/master
902e57651b manifest-rev@{15}: west fetch: entry for refs/west/pull/240/head <--- A forced push is made later a manifest-rev@{2}
```

src/west/manifest.py
tejlmand: I'm trying to hint what I would like to do when having a little spare time ;)
Bu agree, this comment can be removed.

src/west/commands/project.py
tejlmand: > Did we agree to use a reflog on refs/heads/manifest-rev to keep track of old versions of the revision field? That's what I thought we agreed to.
> 
> If so, why can't we just update manifest-rev directly in this case? 

not sure we actually discussed the reflog part, but that dates back a long time ;)
But you have a valid point that we can actually fetch directly into `manifest-rev` without having to go through `FETCH_HEAD`. Not sure why no one considered that in the first place.

But this doesn't address my main point with this PR, which is actually to be able to verify what revision is currently in use.
With fetching directly `manifest-rev` the user (and other extension commands) can no longer determine if the current `manifest-rev` revision is identical to current manifest file, cause only the SHA is available.

So if I do `git log`, `west list zephyr`, or `west status zephyr`  I might see:
```
$ git log --oneline -n1
b687d5d022 (HEAD, manifest-rev) [nrf noup] settings: Fix NVS initialization with PM enabled

$ west status zephyr
=== status of zephyr (zephyr):
HEAD detached at b687d5d022

$ west list zephyr
zephyr     zephyr     pull/240/head     https://github.com/NordicPlayground/fw-nrfconnect-zephyr
```

but I have no idea whether `b687d5d022` is actually identical to `pull/240/head` or not.
Now, with the proposal in this PR, of keeping only the last updated refs under `refs/west` when it's not a SHA, it provides this:
```
$ git log --oneline -n1
a6a98b7854 (HEAD, refs/west/pull/240/head, manifest-rev) [nrf noup] partition-manager: Fix s1_variant dependency
```
i'm looking at `pull/240/head`, Good to know :)

```
$ git rev-parse refs/west/pull/240/head
a6a98b7854c8ff4e1af7bf22d991379535284521
```
which can then compare to either `HEAD` or current `manifest-rev`, which also means we could do:
```
$ west list zephyr
zephyr  zephyr   pull/240/head (not checked out)    https://github.com/NordicPlayground/fw-nrfconnect-zephyr
```
to indicate to the user that the manifest file has changed, but that he is NOT looking at revision `pull/240/head`, or we could add this info to `west status` instead.

Just as in: https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/1437
that you now have closed, you would be able to determine if current `manifest revision exists in refs/west/{revision}` and thus avoid the additional fetch.

If we only have `manifest-rev` there is no relation back to the manifest file.
With both manifest-rev, `west.yml` and a single `refs/west/<revision>` when not a SHA, we can the do:
- is `manifest-rev` == `west.yml::revision` and available in `refs/west` space

mbolivar: > not sure we actually discussed the reflog part, but that dates back a long time ;)

OK, let me try to summarize the discussion from what I understood, because I think that we have misunderstood each other again.

I wrote a concern about garbage collection of items in refs/west:

https://github.com/zephyrproject-rtos/west/pull/331#discussion_r350387809

>> What are the concern(s)?
>>
>> I guess, user will never need to worry or care about our additional refs/west space.

> We're creating an unbounded amount of un-garbage-collectable refs here, including putting tags outside of refs/tags.

You responded:

https://github.com/zephyrproject-rtos/west/pull/331#discussion_r351196763

> We should look into cleaning up this ref-space.

Marc responded with a comment about the reflog:

https://github.com/zephyrproject-rtos/west/pull/331#issuecomment-558382858

> Connecting all the dots above, I just remembered that git already has a built-in feature to cache objects and automatically expire and garbage collect them after some configurable time: it's the reflog. 

I responded to him:

https://github.com/zephyrproject-rtos/west/pull/331#issuecomment-558635485

> Good point!
>
> Since manifest-rev is reflogged already, do we need this then?

You responded:

https://github.com/zephyrproject-rtos/west/pull/331#issuecomment-559024596

> @marc-hb Good point on updating the reflog, and thus cleaning the refs/west will also help te ensure the SHAs are available for a certain time, while not have all those refs/west references popping up in git log.
> As I reply in #331 (comment) , there needs to be done some cleanup in refs/west, and this appears to be a nice way of doing that.

I assumed by your "this appears to be a nice way of doing that" in response to my question "do we need this", (and by "this", I mean this PR, #331), that you agreed we did not need to keep storing things in refs/west, and we can just use the reflog on the manifest-rev branch.

OK, so that's it so far from my side.

-------------------------------------------

> which can then compare to either HEAD or current manifest-rev, which also means we could do:

```
$ west list zephyr
zephyr  zephyr   pull/240/head (not checked out)    https://github.com/Nor
```

Now given this last example I think I finally see what you want to do with this, and what you really meant in the original pull description when you wrote:

> To allow later usage of the fetched refs space then git fetch will
> always keep a local ref space even fetching non-SHA refs spaces, such
> as branches and PRs.

I really am not sure this is within the scope of west. We certainly originally tried hard to avoid creating any refs at all other than `manifest-rev`, so I think we need to decide what the "real feature" is before we continue to create refs that will show up in people's git log. I personally do not like the refs/west showing up in my git logs, at least.

As you've stated, we started using `refs/west/` as a workaround to fetch SHAs, not because we wanted to track refnames from the remote there.

So can we talk about this in terms of what you ultimately want to do?

tejlmand: > > not sure we actually discussed the reflog part, but that dates back a long time ;)
>
> OK, let me try to summarize the discussion from what I understood, because I think that we have misunderstood each other again.
> 
> I wrote a concern about garbage collection of items in refs/west:

Ahh, now I get it. I thought you were referring back to when we introduced `manifest-rev`, and not the fact pointed out by @marc-hb.

> Now given this last example I think I finally see what you want to do with this, and what you really meant in the original pull description when you wrote:

Yes, this was my original intention with this PR, and then it got sidetracked into cleaning up `refs/west` for all the revisions that was purely kept under `refs/west` because of SHA fetching.
(And a good sidetracking I believe)

So original intention was to keep manifest revision available under `refs/west/<revision>` when it is not a SHA. (and remove all other revisions under `refs/west`).
I found such functionality useful for minimizing need for additional fetching in NordicPlayground/fw-nrfconnect-nrf#1437  and considered it to be potentially useful in other places.
But i'm of course open to hear all opinions on whether this is a valid use-case.


For the other revision under `refs/west` I fully agree that we should not keep those available under `refs/west` just because we need to fetch SHAs.

So that's why I included this cleanup into this PR so that `refs/west` is always cleaned up when pointing to a SHA :-)
Although it was actually a side-track to the original PR.

Maybe this PR should now be split into 2:
1) Cleaning of `refs/west` ref space when fetching SHAs - We definately want this
2) Keep current `refs/west/<revision>` for non-SHA - As this is controversial

mbolivar: > Maybe this PR should now be split into 2:
> 
>     1. Cleaning of `refs/west` ref space when fetching SHAs - We definately want this

+1 to this

>     2. Keep current `refs/west/<revision>` for non-SHA - As this is controversial

And this. I don't mind keeping `refs/west/<revision>` if it helps use cases that matter to people. I closed https://github.com/NordicPlayground/fw-nrfconnect-nrf/pull/1437 unmerged because I think all I need is `west fetch` as you suggested, not `refs/west/xyz`.

>
>   2. Keep current `refs/west/<revision>` for non-SHA - As this is controversial
>
>And this. I don't mind keeping refs/west/<revision> if it helps use cases that matter to people. I closed NordicPlayground/fw-nrfconnect-nrf#1437 unmerged because I think all I need is west fetch as you suggested, not refs/west/xyz.

Fair enough.
This was suggested so that you could avoid a `west fetch` if `revision: pull/<no>/head` was already available.
Which would allow us to stick to current behavior where only `west update` actually touches the network ;)

mbolivar: Hmm, we could do `west update --fetch-only` or so if that's a concern.

tejlmand: > Hmm, we could do `west update --fetch-only` or so if that's a concern.

It's not really a concern for me, as such.

If I remember correctly, you have previously mentioned that today only `west update` touches the network and make changes to projects.

Also you have in general have preferred to minimize fetching.

And I do believe both principles above are good to have in mind when working on `west`.

That was the reason for my comment.

Sorry if I have misunderstood, and I'm referring your intentions wrongly :flushed:

mbolivar: You are right that I prefer to minimize fetching. It's true that right now, west update is the only thing that touches the network (other than west init, of course), but I think it would be OK to add more if there is a good reason. I do want to minimize the number of built-in commands, but it seems like a `west fetch` would be reasonable in this case.

mbolivar-nordic:
@tejlmand any plans to get back to this?

tejlmand:
Every time I look at my list of open PRs, then I feel so sad for this one that it haven't got the attention it deserves.
It's still on my list of things I would like to do.

mbolivar-nordic:
OK, I will leave it open then.

mbolivar-nordic:
@tejlmand please reopen if you have an update.

Issue #319, opened 2019-10-07
Allow for shallow updates 


KaSroka:
For verification purposes, it would be nice to init and fetch only versions from the manifest file.
So issuing:
```
west init
west update --shallow
```
or
```
west init --shallow
west update
```
Would create a minimal version of repository saving a lot of time for big projects.

mbolivar:
> For verification purposes, it would be nice to init and fetch only versions from the manifest file.

This is possible if you use a branch for the project `revision` and set the `clone-depth` attribute in the project.

However, if you use a SHA, you can't do this as easily, because not all Git servers let you fetch a SHA directly. GitHub is one of these servers.

One potential workaround is to add a new project attribute that says "the SHA in the revision field is in this branch over here", but that's a little tricky because the branch could get ahead of the SHA in the revision field, e.g. (for nrf) after merging a PR in the the zephyr project before updating nrf/west.yml to the new SHA. So I agree this would be nice, but it's not totally trivial.

KaSroka:
Yes but clone-depth is useless if you want to have full repo for development and shallow for verification.

Also I'm getting this when using `clone-depth: 1`:
```
=== updating proj (some/dir):
--- proj: fetching, need revision master with --depth 1
error: unknown option `tags--depth=1'
```

mbolivar:
> Also I'm getting this when using clone-depth: 1:

This was fixed in master last week, sorry about that.

> So what would be nice to have an option to override clone-depth from west init or from west update.

I understand and agree. You get what I'm saying about why that isn't really possible to do 100% effectively with SHA revisions, though, right?

mbolivar:
I think something like a combination of specifying the upstream branch plus a fallback strategy would probably be needed, indeed. That makes the feature complex to implement (and test) and less predictable in its results (and thus less useful IMO), which is why it hasn't been a priority, but I do agree it's a valid use case.

yashi:
hmm..   `clone-depth` doesn't seems to work with newly initialized workspace because `_rev_type()` failed with the following error and return "other"

> "git cat-file -t master" exit code: 128 stdout: b'' stderr: b'fatal: Not a valid object name master\n'

This is with git version 2.17.1

koffes:
I just closed #399 as a dupe of this issue. This should work as long as the git installation is >= 2.5.0 on both the client and server. Agree with @KaSroka that as long as we have a fallback, this would make a great feature.

marc-hb:
[...]

PR #344, opened 2019-12-09
[RFC] manifest: allow projects to say where their SHAs are

mbolivar:
Note: "revision-ref" is kind of a placeholder name. I would like to try to come up with something better.

Add a new "revision-ref" project attribute, and accept it in the
defaults section. If provided, it will be fetched during 'west update'
whenever the revision looks like a SHA, instead of syncing the entire
remote ref space.

The new attribute is ignored if the revision doesn't look like a SHA.

There's no need to update the manifest schema version since we haven't
released a version of west with manifest schema versioning yet. 0.6.99
is still correct.

Here is an example change to zephyr/west.yml making use of it for all
projects:

```
    index 67b8001ed2..2d117bd394 100644
    --- a/west.yml
    +++ b/west.yml
    @@ -17,6 +17,7 @@
     manifest:
       defaults:
         remote: upstream
    +    revision-ref: master

       remotes:
         - name: upstream
    @@ -36,6 +37,7 @@ manifest:
           path: modules/lib/civetweb
         - name: esp-idf
           revision: 6835bfc741bf15e98fb7971293913f770df6081f
    +      revision-ref: zephyr
           path: modules/hal/esp-idf
         - name: fatfs
           revision: df96914dd989907f3a5de4cb95b116c2f305820d
    @@ -69,6 +71,7 @@ manifest:
           path: modules/hal/libmetal
         - name: lvgl
           revision: d4708d0a432e95f51bdc712591ba5295b751140c
    +      revision-ref: zephyr
           path: modules/lib/gui/lvgl
         - name: mbedtls
           revision: 4f1e8f5a78dc08aa42a47cc1ad059cce558c26c3
    @@ -90,6 +93,7 @@ manifest:
           path: modules/lib/open-amp
         - name: openthread
           revision: 882e7074b5986027b85cb4f3ba1dc563a11ca013
    +      revision-ref: zephyr
           path: modules/lib/openthread
         - name: segger
           revision: 6fcf61606d6012d2c44129edc033f59331e268bc
    @@ -97,12 +101,15 @@ manifest:
         - name: tinycbor
           path: modules/lib/tinycbor
           revision: 0fc68fceacd1efc1ce809c5880c380f3d98b7b6e
    +      revision-ref: zephyr
         - name: littlefs
           path: modules/fs/littlefs
           revision: fe9572dd5a9fcf93a249daa4233012692bd2881d
    +      revision-ref: zephyr
         - name: mipi-sys-t
           path: modules/debug/mipi-sys-t
           revision: baf51863f19f009b92e762115ba5572a5b996b92
    +      revision-ref: zephyr

       self:
         path: zephyr
    --
    2.24.0
```

This saves network bandwidth and avoid clogging up refs/west/* with
stuff we don't care about.

### DIFF
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -863,7 +863,11 @@ def _fetch(project):
     update_cmd = 'update-ref ' + QUAL_MANIFEST_REV + ' '
     if _maybe_sha(project.revision):
         # Don't fetch a SHA directly, as server may restrict from doing so.
-        fetch_cmd += 'refs/heads/*:refs/west/*'
+        # Respect the project's revision-ref if there is one.
+        if project.revision_ref:
+            fetch_cmd += '{revision_ref}:refs/west/{revision_ref}'
+        else:
+            fetch_cmd += 'refs/heads/*:refs/west/*'
         update_cmd += '{revision}'
     else:
         # The revision is definitely not a SHA, so it's safe to fetch directly.
--- a/src/west/manifest-schema.yml
+++ b/src/west/manifest-schema.yml
@@ -37,8 +37,12 @@ mapping:
       remote:
         required: false
         type: str
-      # The default revision (master if not given).
+      # The default project revision (master if not given).
       revision:
+        required: false
+        type: str
+      # The default project revision-ref value (none if not given).
+      revision-ref:
         required: false
         type: str
 
@@ -90,6 +94,15 @@ mapping:
           revision:
             required: false
             type: text        # SHAs could be only numbers
+          # If revision is a SHA, then revision-ref is an optional git
+          # ref which will be fetched to get the SHA when 'west
+          # update'-ing the project.
+          #
+          # This lets west avoid fetching the entire remote ref space
+          # just to get a SHA.
+          revision-ref:
+            required: false
+            type: str
           # Path to check out, relative to the west installation root.
           # The default is the name.
           path:
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -477,7 +477,8 @@ class Manifest:
             if mdrem not in url_bases:
                 self._malformed('default remote {} is not defined'.
                                 format(mdrem))
-        return _defaults(mdrem, md.get('revision', _DEFAULT_REV))
+        return _defaults(mdrem, md.get('revision', _DEFAULT_REV),
+                         md.get('revision-ref'))
 
     def _load_project(self, pd, url_bases, defaults):
         # pd = project data (dictionary with values parsed from the
@@ -523,7 +524,9 @@ class Manifest:
         return Project(name, url, pd.get('revision', defaults.revision),
                        pd.get('path', name), clone_depth=pd.get('clone-depth'),
                        west_commands=pd.get('west-commands'),
-                       topdir=self.topdir)
+                       topdir=self.topdir,
+                       revision_ref=pd.get('revision-ref',
+                                           defaults.revision_ref))
 
 class MalformedManifest(Exception):
     '''Manifest parsing failed due to invalid data.
@@ -555,6 +558,8 @@ class Project:
     - ``url``: project fetch URL
     - ``revision``: revision to fetch from ``url`` when the
       project is updated
+    - ``revision_ref``: git ref which can be fetched to get ``revision``,
+      if it's a SHA
     - ``path``: relative path to the project within the installation
       (i.e. from ``topdir`` if that is set)
     - ``abspath``: absolute path to the project in the native path name
@@ -574,17 +579,19 @@ class Project:
         return NotImplemented
 
     def __repr__(self):
-        return ('Project({}, {}, revision={}, path={}, clone_depth={}, '
-                'west_commands={}, topdir={}').format(
-                    self.name, self.url, self.revision, self.path,
-                    self.clone_depth, self.west_commands, self.topdir)
+        return ('Project({}, {}, revision={}, revision_ref={}, '
+                'path={}, clone_depth={}, west_commands={}, topdir={}').format(
+                    self.name, self.url, self.revision, self.revision_rev,
+                    self.path, self.clone_depth, self.west_commands,
+                    self.topdir)
 
     def __str__(self):
         return '<Project {} at {}>'.format(
             repr(self.name), repr(self.abspath or self.path))
 
     def __init__(self, name, url, revision=None, path=None,
-                 clone_depth=None, west_commands=None, topdir=None):
+                 clone_depth=None, west_commands=None, topdir=None,
+                 revision_ref=None):
         '''Project constructor.
 
         If *topdir* is ``None``, then absolute path attributes
@@ -598,11 +605,13 @@ class Project:
         :param west_commands: path to west commands directory in the
             project, relative to its own base directory, topdir / path
         :param topdir: the west installation's top level directory
+        :param revision_ref: a ref which contains the fetch revision
         '''
 
         self.name = name
         self.url = url
         self.revision = revision or _DEFAULT_REV
+        self.revision_ref = revision_ref
         self.path = path or name
         self.clone_depth = clone_depth
         self.west_commands = west_commands
@@ -642,6 +651,8 @@ class Project:
         ret['name'] = self.name
         ret['url'] = self.url
         ret['revision'] = self.revision
+        if self.revision_ref:
+            ret['revision-ref'] = self.revision_ref
         if self.path != self.name:
             ret['path'] = self.path
         if self.clone_depth:
@@ -663,6 +674,7 @@ class Project:
             - ``name``
             - ``url``
             - ``revision``
+            - ``revision_ref``
             - ``path``
             - ``abspath``
             - ``posixpath``
@@ -678,7 +690,7 @@ class Project:
         :param s: string (or other object) to call ``format()`` on
         '''
         kw = {s: getattr(self, s) for s in
-              'name url revision path abspath posixpath '
+              'name url revision revision_ref path abspath posixpath '
               'clone_depth west_commands topdir'.split()}
         kw['name_and_path'] = '{} ({})'.format(self.name, self.path)
         kw.update(kwargs)
@@ -868,6 +880,7 @@ class ManifestProject(Project):
       version-controlled by west itself, even though 'west init'
       can fetch a manifest repository from a Git remote
     - ``revision``: ``"HEAD"``
+    - ``revision_ref``: ``"HEAD"``
     - ``clone_depth``: ``None``, because ``url`` is
     '''
 
@@ -939,6 +952,14 @@ class ManifestProject(Project):
         raise ValueError(revision)
 
     @property
+    def revision_ref(self):
+        return 'HEAD'
+
+    @revision_ref.setter
+    def revision_ref(self, revision_ref):
+        raise ValueError(revision_ref)
+
+    @property
     def clone_depth(self):
         return None
 
@@ -956,7 +977,7 @@ class ManifestProject(Project):
             ret['west-commands'] = self.west_commands
         return ret
 
-_defaults = collections.namedtuple('_defaults', 'remote revision')
+_defaults = collections.namedtuple('_defaults', 'remote revision revision_ref')
 _WEST_YML = 'west.yml'
 _SCHEMA_PATH = os.path.join(os.path.dirname(__file__), "manifest-schema.yml")
 _SCHEMA_VER = parse_version(SCHEMA_VERSION)


mbolivar:
cc: @marc-hb 

This is very tangentially related to https://github.com/zephyrproject-rtos/west/pull/331. I've been meaning to add something like this to west for a long time. It's inspired by the grepo `upstream` project attribute.

https://android.googlesource.com/tools/repo/+/HEAD/docs/manifest-format.md

I think this is useful in addition to #331 -- they are not in competition.

If this idea seems good, I will add test cases and remove RFC.

marc-hb:
> Note: "revision-ref" is kind of a placeholder name. I would like to try to come up with something better.

For a hopefully better name, let me.... "cheat" and first change/simplify the feature:

  => drop the "whenever the revision looks like a SHA" condition.

Fetch the new parameter always. Then simply call it `fetch-ref`! Or `fetch-first`, or `fetch-only`.

So yes, if you have:
```yaml
- name: openthread
   revision: master # not a SHA1
   fetch-ref: master2
```
... then `west` would fetch both `master` and `master2` which is a bit odd - but only because the user explicitly asked for something odd!

mbolivar:
`fetch-ref` is much better. Thank you.

But I don't want west to fetch it always (except for `west update --fetch=always`). That would defeat the fetch optimization introduced in 0.6.

mbolivar:
OK, what would this do?

```
revision: deadbeef
fetch-ref: master2
```

The `deadbeef` ref might be a SHA. I think west should not fetch it, and only fetch `master`, in this case.

In your original example:

```
revision: master
fetch-ref: master2
```

We know we want `manifest-rev` to be `master` after we're done fetching. So `master2` doesn't need to be fetched at all in this case, I think, since `master` is definitely not a SHA.

marc-hb:
> So master2 doesn't need to be fetched at all in this case, I think, since master is definitely not a SHA.

OK but why did the user left it there then? Not even commented out?

I think it's simpler to blindly follow the users' orders and make her feel in total control as opposed to guessing that the extra line was a (harmless BTW) configuration mistake. Also simpler to have less logic and less conditionals in the documentation.

> That would defeat the fetch optimization introduced in 0.6.

From grepo experience, fetch performance issues happen when git repos are swarmed with references constantly created or updated by (CI) robots. Fetching 1 ref or 5 refs makes no difference unless someone sadistic keeps force-pushing them to completely different places.

mbolivar:
> > That would defeat the fetch optimization introduced in 0.6.
> 
> From grepo experience, fetch performance issues happen when git repos are swarmed with references constantly created or updated by (CI) robots. Fetching 1 ref or 5 refs makes no difference unless someone sadistic keeps force-pushing them to completely different places.

It's not just a performance issue. Smart fetching lets you run `west update` on an airplane with no wifi as long as your manifest uses SHAs, say during a bisection.

marc-hb:
> It's not just a performance issue. Smart fetching lets you run west update on an airplane with no wifi as long as your manifest uses SHAs, say during a bisection.

OK, but still no big deal: "Wait, why does west try to use the network? Oh, because I told it to fetch! Stupid me. Config mistake fixed."

I just can't stand software that corrects my configuration mistakes *silently*. Sorry :-)

mbolivar:
> OK, but still no big deal: "Wait, why does west try to use the network? Oh, because I told it to fetch! Stupid me. Config mistake fixed."
> 
> I just can't stand software that corrects my configuration mistakes _silently_. Sorry :-)

Hm, then perhaps I need to take back my comment that `fetch-ref` is a better name if you think that name implies that it will always be fetched ;).

Do you agree that smart fetching is a sensible thing to have enabled by default? If so, I think it makes sense to keep discussing here, but if not, I think it's a separate discussion.

mbolivar:
> If this idea seems good, I will add test cases and remove RFC.

Curiously, shippable has failed because the latest colorama dropped Python 3.4 support. Hopefully we can just get https://github.com/zephyrproject-rtos/zephyr/pull/21212 merged so I can stop thinking about that version entirely ;).

marc-hb:
> Hm, then perhaps I need to take back my comment that fetch-ref is a better name if you think that name implies that it will always be fetched ;).

No I don't think `fetch-ref` implies "always" or anything related, there's only so much you can compress in two words :-) At some point you need the --help or documentation.

> Do you agree that smart fetching is a sensible thing to have enabled by default? 

100% because many git repos are "noisy" and the number of people who need to all refs from all repos is approximately zero. A per-repo option to fetch all refs would be nice though, dunno whether `west` has that already. No big deal.

tejlmand:
Great to get this up, can surely improve `west` :)

> `revision: 6835bfc741bf15e98fb7971293913f770df6081f`
> `revision-ref: zephyr`

but I have a small suggestion.
The `@` is already reserved in `git` so no revisions in the manifest file would contain an `@`, so howabout we simply introduce this as:
`revision: 6835bfc741bf15e98fb7971293913f770df6081f@zephyr`

which would even allow users to do:
`revision: 6835bfc741bf15e98fb7971293913f770df6081f@pull/<no>/head`

and it removes the need for yet another field in the manifest file, which is already growing long.
Also current schema is still valid.

+ it removes @marc-hb concern:
``` - name: openthread
   revision: master # not a SHA1
   fetch-ref: master2
```
as the `@` would only be valid when following a SHA.

carlescufi:
> The `@` is already reserved in `git` so no revisions in the manifest file would contain an `@`, so howabout we simply introduce this as:
> `revision: 6835bfc741bf15e98fb7971293913f770df6081f@zephyr`

I am really not convinced that diverging from Git's ref standards is a good idea. This in particular is confusing with another very common pattern that is used in other places (including GitHub) where the `@` is used to denote a revision after a URL.

mbolivar:
Here is my summary of where we are at on this PR.

Please speak up if I have misunderstood :).

- west supports project revisions which are SHAs, this is good, and we like it
- west cannot fetch SHAs directly from a remote
- to avoid having to sync the entire remote ref space to fetch a SHA revision from a remote, therefore, we would need users to be able to specify a git ref -- which I've temporarily called `revision-ref` here -- where a project's SHA revision can be fetched from
- smart fetching is a sensible default, and therefore we should *not* fetch this `revision-ref` by default if we already have a project's SHA revision fetched locally

So the remaining issue is only whether we should overload the `revision:` attribute or have a separate attribute (and what its name should be).

@tejlmand are you convinced by @carlescufi that we should not use `@`? Do you want to suggest an alternative?

Please note that whatever we do, I think we need to be able to specify a default `revision-ref` in the `defaults:` dictionary, since it will usually be `master` in practice. So therefore I lean towards giving it a name which is also usable as a project attribute for consistency and readability even if it is a little more verbose.

Can we agree on a name here?

I still like `fetch-ref`. I thought about `fetch-from` too, but I want to have `ref` in the name.

marc-hb:
I find @tejlmand 's single line idea very interesting and user friendly. In fact `git fetch` itself should implement something like this, if it did the problem would be solved already! Please suggest it at https://public-inbox.org/git/

However and as @carlescufi pointed out, you really don't want to re-use and conflict with the @ syntax that is already used in `git help revisions` for other things. So how about whitespace between the two parameters? That's how parameters are usually separated. Would the yaml parser be OK with that?
```yaml
revision: 9876abcedef  master2
```
Or maybe even:
```yaml
revision: 9876abcedef  @master2
```
I don't know whether git supports whitespace in refs. Even if it does: 1. Who does that? 2. Maybe Yaml can provide quoting help.

> west cannot fetch SHAs directly from a remote

You mean "most git servers" - for security reasons.

mbolivar:
> I don't know whether git supports whitespace in refs.

`HEAD@{5 minutes ago}` is a legal ref in the `git help revisions` output. I am leery of trying to parse the revision format. It's a beast.

>  Even if it does: 1. Who does that? 2. Maybe Yaml can provide quoting help.

Either way:

> So therefore I lean towards giving it a name which is also usable as a project attribute for consistency and readability even if it is a little more verbose.

Any response to this?

IOW, we need a name, no matter what, or we can't put it in defaults like this:

```
manifest:
  defaults:
    fetch-ref: master
```

If we have a name for `defaults`, then I think the most readable and understandable thing to do is use the same name from `projects`.

No?

mbolivar:
> > west cannot fetch SHAs directly from a remote
> 
> You mean "most git servers" - for security reasons.

West can't do things that won't work for "most git servers", so I stand by this wording.

marc-hb:
> HEAD@{5 minutes ago} is a legal ref in the git help revisions output.

Doh! How did I miss this one? Thx for correcting me.

> I am leery of trying to parse the revision format. It's a beast. [...]
> @tejlmand are you convinced by @carlescufi that we should not use @? Do you want to suggest an alternative?

`git help check-ref-format` has a list of characters forbidden in ref names, so I looked there for an alternative. Unfortunately, most of these (`@`, `^`, `~`,... ) are forbidden in ref names because they are used in `git help revisions`, so still not great. The rest is either control characters or poor looking like `?` or `*`.

I found the colon `:` interesting though: it would look "good enough" in `432abcde:master2` IMHO and it is already used in `git help revisions` BUT for nothing I think anyone would sanely use in a `west.yml` file:
- referencing blobs - clearly non-sensical in a `west.yml`
- searching commit messages with regular expressions with `:/`. I hope no west user would want to point his manifest at "the youngest matching commit"? If one does, maybe he can submit later the west enhancement that finds the `/` in `:/` :-)

> West can't do things that won't work for "most git servers", so I stand by this wording.

What I meant, sorry for any misunderstanding: "GIT (not west) cannot fetch SHAs directly from a  remote"

In the lack of more context, I read "west can't do \<git stuff X>" as: "git can do \<stuff X> but west can't" and I suspect most people read it the same. Now we have (too much) context so as long as this sentence doesn't go to any documentation it doesn't matter.

mbolivar:
>>  So therefore I lean towards giving it a name which is also usable as a project attribute for consistency and readability even if it is a little more verbose.

Any response to this? :)

Sorry to nag.

tejlmand:
> EDIT: Furthermore, @ is already used by Git in specifying Git revisions: https://git-scm.com/docs/gitrevisions

Correct, but my suggestion was that only if provided after a SHA, we would have special handling.
Cause SHAs already have special handling in `west`.
The `@` was just an example, but i'm open to other suggestions as well.



> >  I don't know whether git supports whitespace in refs.

> HEAD@{5 minutes ago} is a legal ref in the git help revisions output. I am leery of trying to parse the revision format. It's a beast.

but we only need to do this for SHAs.
and you cannot do: `<SHA>@{5 minutes ago}`, so even though spaces are allowed in git ref, it is not universally allowed.


>   So therefore I lean towards giving it a name which is also usable as a project attribute for consistency and readability even if it is a little more verbose.

I feel we already has many attributes, and I believe this will only grow, so whenever I see an additional I like to think if we can do this in shorter way.
Especially for attributes where it is only valid when another attribute is of a special type.
In this case, the `fetch-ref` is only valid when revision is a SHA.

So I feel that using `revision: <SHA> <your-favorite-separator><branch>`
it is very visible that this is only valid for a SHA and not for branches, tags, PRs, etc.

>
> IOW, we need a name, no matter what, or we can't put it in defaults like this:
> 
> `manifest:`
> `  defaults:`
> `    fetch-ref: master`

I would be careful allowing this in the defaults section, cause suddenly `west` starts fetching only from `master` and people might just update the `SHA` located at a project.
Everything works locally, even when `--fetch always` cause even in this case they already have the SHA (and same could be for a CI).
To other users who runs `west update` it might be cannot find the SHA, and `west update` fails because the SHA was on another branch.

Thus I prefer that the `fetch-ref` or what we end with is ONLY allowed next to the project that uses a SHA.

tejlmand:
> In fact git fetch itself should implement something like this, if it did the problem would be solved already! 

git fetch does allow fetching a SHA directly, it depends on the remote, and not all remotes allows it.
(Such as github).
And git supports fetching of a branch, from where you then can just checkout the SHA afterwards, so not sure what to ask for at https://public-inbox.org/git/

tejlmand:
> I am leery of trying to parse the revision format. It's a beast.

but we already do:
https://github.com/zephyrproject-rtos/west/blob/master/src/west/commands/project.py#L727-L735

This would be the code that determines if it is a SHA.
So this should simply do a `string.split` on the separator and test first part to see if that is a SHA.
If a SHA, then second part is the branch.

If first part is not a SHA, continue as normal.

mbolivar:
> I would be careful allowing this in the defaults section, cause suddenly `west` starts fetching only from `master` and people might just update the `SHA` located at a project.

If brevity is a concern, we really want to put it in the defaults section. That's the only way to make it possible to be a 1-line change to the manifest that uses SHAs for everything (like mainline zephyr/west.yml) -- otherwise, we have to put it in every single project.

That's even more confusing for people who don't know west, I think. Which is easier for a novice to read and modify, say to add a new project or update a project revision?

```
manifest:
  defaults:
    fetch-ref: master

  projects:
    - name: tinycbor
      revision: sha-abc
    - name: mbedtls:
      revision: sha-def
    ...
```

Or:

```
manifest:
  projects:
    - name: tinycbor
      revision: sha-abc@master
    - name: mbedtls:
      revision: sha-def@master
    ...
```

marc-hb:
> not sure what to ask for at https://public-inbox.org/git/

The sugggestion is  to make:

```
git pull origin abc123def%master2 #  '%' is just an example
```
... a shortcut for:
```
git fetch origin master2 # cause origin doesn't allow fetching SHA1s
git update-ref FETCH_HEAD abc123def
git merge FETCH_HEAD
```

If this had been available in git then we wouldn't have had most of the discussion above. Granted, it's too late anyway now, west can't require a bleeding edge git. Never mind.

marc-hb:
`:` is a better separator than  `@` because  git "revisions" with a colon `:` are actually NOT revisions. While `fu@bar` can be pointing at a specific commit, `fu:bar` cannot. While there's little chance anyone would ever make a `fu@bar@master2` mistake, there's absolutely zero chance anyone would ever make a `fu:bar:master2` mistake.


> So this should simply do a string.split on the separator and test first part to see if that is a SHA.
If a SHA, then second part is the branch. If first part is not a SHA, continue as normal.

The problem with mini "parsers" like this one is that you keep finding corner cases that produce utterly confusing error messages. Example, let's start from `abc123def@master2` that works. Then the user finds something suspect in `abc123def` and changes it to `abc123def~1@master2` for a quick test. Boom! Weird and unexpected error message when it should just work.

- Dumb `string.split` -> yes
- "See if that is a SHA1" -> no. Adds no value and can make error messages more confusing. Keep the "syntax" simpler and always assume that the left of `:` (if any) is a usable git revision. Let git or python fail with a simple and relevant error message if it's not.

mbolivar:
> The problem with mini "parsers" like this one is that you keep finding corner cases that produce utterly confusing error messages.

:+1: 

 >: is a better separator than @ because git "revisions" with a colon : are actually NOT revisions. 

This seems like a fair point. Perhaps we could support this as another option for terse users? I still think we need a name to put it in defaults though, and that if we have a name in defaults, we need to support the same name in projects.

yashi:
For someone who didn't know that github doesn't allow you to fetch directly with SHA1, it was hard to realize what was discussed here.  Anyway, for those like me, this "revision-ref" option is a work around to get a commit pointed by sha1.  IOW: a workaround for the following error
```
$ git fetch github c58ae96fc4bb11916b62a96940bb70bb85ea5992:temp
error: Server does not allow request for unadvertised object c58ae96fc4bb11916b62a96940bb70bb85ea5992
```

I thought `git ls-remote` had `--contains` option but didn't.  I agree that it'd be nice addition to git proper.

For the name, `revision-is-in` might be a good name for it.  I doesn't tell west what to do but where to look for.

For default or not, I believe this attribute is a server-wise attribute.  So, write it like so. I mean, If a remote prohibits it, we must use a published name to fetch, anyway.

```yaml
manifest:
  defaults:
    remote: github
       no-hash-fetch: true
       revision-in: master
```

When `revision-in` is specified, `no-hash-fetch` is implied.  So we can omit `no-hash-fetch`.

And, yes, `project` should be allowed to override it.

```yaml
  projects:
    - name: tinycbor
      revision: sha-abc
      revision-in: topic
    - name: mbedtls:
      revision: sha-def
```

tejlmand:
> For someone who didn't know that github doesn't allow you to fetch directly with SHA1, it was hard to realize what was discussed here. Anyway, for those like me, this "revision-ref" option is a work around to get a commit pointed by sha1. IOW: a workaround for the following error
> 
> ```
> $ git fetch github c58ae96fc4bb11916b62a96940bb70bb85ea5992:temp
> error: Server does not allow request for unadvertised object c58ae96fc4bb11916b62a96940bb70bb85ea5992
> ```

Yeah, a long discussion and with some background knowledge.
But thanks for joining. All opinions and views are very welcome.

> For the name, `revision-is-in` might be a good name for it. I doesn't tell west what to do but where to look for.
> 
> For default or not, I believe this attribute is a server-wise attribute. So, write it like so. I mean, If a remote prohibits it, we must use a published name to fetch, anyway.

I like this line of thought.

> 
> ```yaml
> manifest:
>   defaults:
>     remote: github
>        no-hash-fetch: true
>        revision-in: master
> ```

If this is done while allowing to specify multiple names, such as:
> `        revision-in: master,release-branchA,otherBranch`

it might be possible both to reduce fetching (as many automatically created branches will not be fetched) while still minimize user mistakes of updating from a SHA on master, to point to a SHA on a release branch as example.

Remember, with smart fetching, the fetch will only happen when the SHA is not locally available.

> 
> When `revision-in` is specified, `no-hash-fetch` is implied. So we can omit `no-hash-fetch`.
> 
> And, yes, `project` should be allowed to override it.

Override, or simply append ?
After all the `refs/west` space will be cleared, see: https://github.com/zephyrproject-rtos/west/pull/331#discussion_r356279735.

All-in-all I believe it could be valuable for users to have the SHA both from default `revision-in` as well as the project specific, as I believe he is likely to work on code related to any of them from time to time.

And this will minimize future fetching.

> 
> ```yaml
>   projects:
>     - name: tinycbor
>       revision: sha-abc
>       revision-in: topic
>     - name: mbedtls:
>       revision: sha-def
> ```

tejlmand:
> If this had been available in git then we wouldn't have had most of the discussion above. Granted, it's too late anyway now, west can't require a bleeding edge git. Never mind.

Not agreeing.

If you want a fetch optimization, then use `git >= x.y.z`.
Afterall, this is an optimization and we could always fall back on old behavior, as that is still working.

So still worth a thought.

tejlmand:
> The problem with mini "parsers" like this one is that you keep finding corner cases that produce utterly confusing error messages. Example, let's start from abc123def@master2 that works.

True.
Of course care must be taken, especially on selection of a separator, if going in this direction.
This is also why i'm not advocating specifically for the `@` as separator.
And before choosing a separator, then it must be fully examined if it is useful.

It was partly also a comment to @mbolivar that we already do special parsing for SHAs.

This also lead to me discover this issue: 
https://github.com/zephyrproject-rtos/west/issues/345

mbolivar:
> Yeah, a long discussion and with some background knowledge.
> But thanks for joining. All opinions and views are very welcome.

Agreed. We really want users to chime in here :). Just having a small number of developers do discussion for features like these is not good, in my opinion.

> For the name, revision-is-in might be a good name for it. I doesn't tell west what to do but where to look for.

I like that name, thanks for suggesting it! (Actually, I like the shorter `revision-in` idea you use below even more.)

> For default or not, I believe this attribute is a server-wise attribute. So, write it like so. I mean, If a remote prohibits it, we must use a published name to fetch, anyway.

```yaml>
manifest:
  defaults:
    remote: github
       no-hash-fetch: true
       revision-in: master
```

> When revision-in is specified, no-hash-fetch is implied. So we can omit no-hash-fetch.

So that means the semantics are still the same, right? I.e.:

If `revision-in` is present, and `revision` might be a SHA, then west can fetch `revision-in` to get `revision` instead of syncing `refs/heads*`, right?

This is an interesting idea. The server-wise point is a good one. Note that projects can specify their fetch URLs directly, without using remotes. This is a valid manifest:

```yaml
manifest:
  projects:
  - name: west
```

So I think a manifest-wide default may still be useful. We all agree a project-specific override is necessary, for example because some zephyrproject-rtos projects use branch `master`, and some use branch `zephyr` (like `lvgl`), for their `revision-in` values.

What is the use case for `no-hash-fetch`? I'm not sure what the semantics of having it without `revision-in` are.

Can we agree to `revision-in` instead of trying to parse the revision syntax (more than we already do)? Or do people still want to try the `:`?

yashi:
> What is the use case for `no-hash-fetch`? I'm not sure what the semantics of having it without `revision-in` are.

There is none without `revision-in`.  `no-hash-fetch` is just for a placeholder in user's mind.  I mean, it just tell users _why_ they need `revision-in`.  I agree that at some point users need to read the doc.  But, it help users understand the config if `no-hash-fetch` is there.
Technically, it doesn't do anything except helping west generating a warning "No `revision-in` is given while `no-hash-fetch` is present.  Fetching all refs."  So, it's totally possible without it.  It's so biassed by me not knowing github server behaviour.

marc-hb:
> Can we agree to revision-in 

`revision-in` is a great name, makes the line read like plain English. Very pythonic :-)

> instead of trying to parse the revision syntax (more than we already do)? 

The less (duplicate!) parsing the better.

> Or do people still want to try the :?

Yes... but in git first, before west ;-)

tejlmand:
> >    Can we agree to revision-in
>
> revision-in is a great name, makes the line read like plain English. Very pythonic :-)

I like it too.

> If this is done while allowing to specify multiple names, such as:
>
> > `    revision-in: master,release-branchA,otherBranch`

I would like feedback on this.

I think it would be userfriendly to allow to specify multiple default branches to look in.
This can minimize the risk of user accidentally uses SHA on another branch and forgetting to update `revision-in` and at the same time, users are probably going to fetch SHAs from the server side default branches anyway at some time, such as `master`, `releases-branches` etc.
So I see no harm in having pre-fetched those.

Also it avoids users having to remove default branches when updating for temp work.

Comments are welcome :smile:

mbolivar:
>>    So that means the semantics are still the same, right?

> Yes. If this is an optimization and west being a tool does magic (user friendly).
> No, if west is a developer tool and does exactly what it's told to do.

@yashi I'm sorry, but I don't understand this comment. I was trying to summarize what I thought you meant. Did I get your proposal wrong?

>>    What is the use case for no-hash-fetch? I'm not sure what the semantics of having it without revision-in are.

>> There is none without revision-in. no-hash-fetch is just for a placeholder in user's mind. 

Since `revision-in` implies `no-hash-fetch` and there's no use case for having `no-hash-fetch` alone, it seems OK to me to leave it out at first, then.

> > > `    revision-in: master,release-branchA,otherBranch`
> 
> I would like feedback on this.

@tejlmand If we are going to support that, I think it should be a proper YAML sequence instead:

```
revision-in:
    - master
    - release-branchA
    - otherBranch
```

yashi:
Ah, never mind.  I misread what we were talking about. I thought we were talking about how west behave when `revision-in` is present. I mean, west to always or not fetch an advertised object `revision-in` points to.

But you are asking about the semantics with `no-hash-fetch`, right?  Then, yes, `no-hash-fetch` doesn't change the original semantics.

> it seems OK to me to leave it out at first, then.

:+1:

tejlmand:
> 
> @tejlmand If we are going to support that, I think it should be a proper YAML sequence instead:
> 
> ```
> revision-in:
>     - master
>     - release-branchA
>     - otherBranch
> ```

Of course. :smile:

mbolivar:
> Ah, never mind. I misread what we were talking about. I thought we were talking about how west behave when `revision-in` is present. I mean, west to always or not fetch an advertised object `revision-in` points to.

Sorry I was being confusing. I'm asking about both `no-hash-fetch` and the `revision-in` behavior. I see that we agree we can leave out `no-hash-fetch` for now, which is great.

I just want to check if I understand your `revision-in` proposal. For example, in this manifest:

```
manifest:
  projects:
  - name: foo
    url: https://git.example.com/foo
    revision: SOME_SHA
    revision-in: master
```

Then I would expect:

1. if SOME_SHA is already fetched into the local repository, `west update` doesn't have to fetch from `https://git.example.com/foo`
2. but if SOME_SHA is not a SHA that has been fetched, `west update` should fetch the `master` branch from `https://git.example.com/foo`

This is just the default if west's [update.fetch configuration option](https://docs.zephyrproject.org/latest/guides/west/config.html#built-in-configuration-options) is set to `smart`. Of course if `update.fetch` is `always`, west will always fetch, like in 2.

Is that what you propose?

mbolivar:
> > @tejlmand If we are going to support that, I think it should be a proper YAML sequence instead:
> > ```
> > revision-in:
> >     - master
> >     - release-branchA
> >     - otherBranch
> > ```
> 
> Of course. smile

OK. I will support either a string or a list when I send the next revision. For now I will wait on confirmation of the `revision-in` behavior.

marc-hb:
> revision: SOME_SHA
>     revision-in: master
> 
> - if SOME_SHA is already fetched into the local repository, west update doesn't have to fetch from https://git.example.com/foo
> - but if SOME_SHA is not a SHA that has been fetched, west update should fetch the master branch from https://git.example.com/foo

The beauty in this is: nowhere does it matter whether SOME_SHA is actually a SHA. When the `revision-in` field is present, `west` doesn't even need to try to parse SOME_SHA and guess whether it's actually one.

tejlmand:
> > revision: SOME_SHA
> > revision-in: master
> > 
> > * if SOME_SHA is already fetched into the local repository, west update doesn't have to fetch from https://git.example.com/foo
> > * but if SOME_SHA is not a SHA that has been fetched, west update should fetch the master branch from https://git.example.com/foo
> 
> The beauty in this is: nowhere does it matter whether SOME_SHA is actually a SHA. When the `revision-in` field is present, `west` doesn't even need to try to parse SOME_SHA and guess whether it's actually one.

Not understood. If SOME_SHA is not a SHA, then why fetch `revision-in` ?

Like (clearly a user mistake imho):
```
    revision: myBranch
    revision-in: 
    - master
```
then `west` has to fetch `myBranch`, cause if it fetches `master` and then tries to checkout `myBranch`, then it will fail.
Or am I missing something :thinking: 

Side-note: `west` will currently checkout FETCH_HEAD, when revision != SHA.

marc-hb:
> Not understood. If SOME_SHA is not a SHA, then why fetch revision-in ?

Right, this is not about that. This is about west trying and failing to parse a valid SHA like for instance `abc123def~1`. Not a problem because west doesn't need to parse it at all.

> then west has to fetch myBranch, cause if it fetches master and then tries to checkout myBranch, then it will fail.

... with a clear git (not west) error message pointing exactly at the user mistake without west having to duplicate git's parser or do anything, so everything is OK

marc-hb:
> Side-note: west will currently checkout FETCH_HEAD, when revision != SHA.

This can be changed to checking out FETCH_HEAD when there's no `revision-in`

yashi:
> ```
>     revision: SOME_SHA
>     revision-in: master
> ```
> 
> Then I would expect:
> 
> 1. if SOME_SHA is already fetched into the local repository, `west update` doesn't have to fetch from `https://git.example.com/foo`
> 2. but if SOME_SHA is not a SHA that has been fetched, `west update` should fetch the `master` branch from `https://git.example.com/foo`
> 
> This is just the default if west's [update.fetch configuration option](https://docs.zephyrproject.org/latest/guides/west/config.html#built-in-configuration-options) is set to `smart`. Of course if `update.fetch` is `always`, west will always fetch, like in 2.
> 
> Is that what you propose?

What `revision-in` implies, as `no-hash-fetch` also says, is that "the `revision` might not be fetchable from the server, but the user is sure that it's an ancestor of the commit `revision-in` is pointing to".  This also implies that we don't allow a hash value to be given to `revision-in` because that kills whole point of having `revision-in` in the first place.  (we have a corner use case with a server not allowing to fetch the hash pointed by `revision`, but somehow OK with the hash pointed by `revision-in`. But that's another story).

### **Without `revision-in`**

**When the given `revision` is a hash**

1. Check to see we already have the `revision` in our cloned copy (Optimization)
2. If not, go get `revision`.
3. Once we have it, check it out.

**When the given `revision` is not a hash**

1. We _must_ go out to fetch the `revision` because we never know which object the `revision` points to _now_ at the server.
2. Check it out.

### **With `revision-in`**

We have 4 cases:

- `revision-in` is a hash, `revision` is a hash
- `revision-in` is a hash, `revision` is not a hash
- `revision-in` is not a hash, `revision` is a hash
- `revision-in` is not a hash, `revision` is not a hash

**When `revision-in` _is_ a hash**

We'll leave it for future implementation.

**When `revision-in` is not a hash, `revision` is a hash**

1. Check to see we already have the `revision` in our cloned copy (Optimization)
2. If not, go get _`revision-in`_, because that's what the user told us.
3. Once we have the object pointed by `revision-in`, _try_ to check `revision` out.

**When `revision-in` is not a hash, `revision` is not a hash**

1. We _must_ go out to fetch the `revision` because we never know which object `revision` points to *now* at the server.
2. But wait, we have `revision-in`!
2a. So fetch `revision-in` instead of `revision` because that's what a user told us!  But, make sure `revision` is also up to date. Because, again, we never know which object the `revision` points to _now_ at the server.
2b. Or, go fetch both `revision` _and_ `revision-in`.  This one is, I believe, the same as 2a. To make sure the `revision` is up-to-date, we have to fetch it, as well.
3. Check `revision` out.  We know we have it.

### Conclusion
As you can see, you can optimize it (or not to fetch) when
- We are sure that given `revision` is a raw hash value, and
- We already have the object pointed by *a raw hash value*.

You can also say that that's the only case we can optimize.

If we treat `revision` as-is without parsing it.  That is, without knowing whether it's a raw hash value or not, we lose optimization because we have to got fetching the `revision` to see what it points to.

### Additional Thoughts
Git is already started to support SHA256 in addition to SHA1.  So it's not enough to check the `revision` is hex 40 chars.  I haven't tested but even only with SHA1, I think you can have 40 hex char as a ref name.  So I'm not sure how we can make sure the given `revision` is a raw hash value.

marc-hb:
The much simpler version that does the same without trying to parse SHA1s:
```
Check if we have `revision` locally already. whatever `revision` is.
If it's a remote branch then it will fail.

If we have it already
    exit success

if revision-in
   fetch revision-in
else
   fetch revision
```

mbolivar:
> ### **Without `revision-in`**
>
> **When the given `revision` is a hash**
>
>     1. Check to see we already have the `revision` in our cloned copy (Optimization)
>
>     2. If not, go get `revision`.
>
>     3. Once we have it, check it out.

Agreed; that's the current behavior, so this is backwards compatible since revision-in does not exist yet.

>
>
> **When the given `revision` is not a hash**
>
>     1. We _must_ go out to fetch the `revision` because we never know which object the `revision` points to _now_ at the server.
>
>     2. Check it out.

Agreed if we say "is not a hash *or a tag*". That's the current behavior. Tags are assumed to stay the same.

>
>
> ### **With `revision-in`**
>
> We have 4 cases:
>
>     * `revision-in` is a hash, `revision` is a hash
>
>     * `revision-in` is a hash, `revision` is not a hash
>
>     * `revision-in` is not a hash, `revision` is a hash
>
>     * `revision-in` is not a hash, `revision` is not a hash
>
>
> **When `revision-in` _is_ a hash**
>
> We'll leave it for future implementation.

I think it's safe to assume that's a badly written manifest. The point of this pull request is to allow users to provide non-SHA refs where SHA `revision` values can be fetched, so we can assume the new field is not a hash.

>
> **When `revision-in` is not a hash, `revision` is a hash**
>
>     1. Check to see we already have the `revision` in our cloned copy (Optimization)
>
>     2. If not, go get _`revision-in`_, because that's what the user told us.
>
>     3. Once we have the object pointed by `revision-in`, _try_ to check `revision` out.
>

Agreed.

>
> **When `revision-in` is not a hash, `revision` is not a hash**
>
>     1. We _must_ go out to fetch the `revision` because we never know which object `revision` points to _now_ at the server.

Unless `revision` is a tag and `update.fetch` is `smart`, to keep current behavior, I think.

>
>     2. But wait, we have `revision-in`!
>        2a. So fetch `revision-in` instead of `revision` because that's what a user told us!  But, make sure `revision` is also up to date. Because, again, we never know which object the `revision` points to _now_ at the server.
>        2b. Or, go fetch both `revision` _and_ `revision-in`.  This one is, I believe, the same as 2a. To make sure the `revision` is up-to-date, we have to fetch it, as well.
>
>     3. Check `revision` out.  We know we have it.

Great. So I think we agree. If we have both `revision` and `revision-in`, and `revision` is definitely not a hash, we can just fetch `revision`, and ignore `revision-in`, because we should be safe to fetch `revision` directly.

So other than the "... or a tag" parts, I think we're on the same page and I'll make it so (and support lists).

Thanks!

yashi:
> Unless `revision` is a tag

I think it's just a design decision.  A tag can technically be overwritten.  So there is no guarantee that it'll always points to the same object.  But, a tag is usually used as a permanent pointer.  So, I agree we can optimize for general use case.

mbolivar:
> A tag can technically be overwritten. So there is no guarantee that it'll always points to the same object. [...]
> So, I agree we can optimize for the general use case. We should document the behaviour, though.

Right, I know, but I'm proposing this specifically to be consistent with the existing documented behavior for the [update.fetch config option](https://docs.zephyrproject.org/latest/guides/west/config.html#built-in-configuration-options):


> String, one of "smart" (the default behavior starting in v0.6.1)
> or "always" (the previous behavior). If set to "smart", the
> west update command will skip fetching from project remotes
> when those projects’ revisions in the manifest file are SHAs or
> tags which are already available locally. 

Note "SHAs or tags" in the above documentation.

OK, so it sounds like we're all on the same page here, thanks. I will update this PR.

mbolivar-nordic:
I'm closing this as stale for now. I will reopen if I get some time for this. I want to think about whether or how this relates to shallow clones also.

marc-hb:
> If we have both revision and revision-in, and revision is definitely not a hash, we can just fetch revision, and ignore revision-in, because we should be safe to fetch revision directly.

This makes the behavior of `revision-in` inconsistent/surprising and unnecessarily depending on another value. In that case just fetch both`revision-in` and `revision`. In other words keep the interface, code and documentation all simple and fetch `revision-in` always. Done.

If the user provides more than one non-SHA1 ref then `west` should  either consider this as a configuration mistake and return an error or at the very least a warning, or much better west should consider this configuration as **intentional** and then it should _not silently ignore the user's intention_ cause ignoring the user always provides a very bad experience.

This gets even more obvious when the user provides more than one `revision-in`, example:
```yaml
revision: dev1

revision-in:
    - master
    - release-branchA
    - otherBranch
```
In a case like this, do you _really_ want to ignore the entire `revision-in` section too?! In such a case it's pretty clear that this was not a configuration mistake and that the user has made a careful selection of branches that they want to fetch[*]. Don't forget some git repos managed by west are also used for active development, not just to download and build dependencies!

We all found the name `revision-in` better but funny enough for that secondary, "no-SHA1" use case the previous `fetch-ref` variantions were actually better :-)

[*]  excluding all other refs to save  a lot of time https://public-inbox.org/git/CAJo=hJtTp2eA3z9wW9cHo-nA7kK40vVThqh6inXpbCcqfdMP9g@mail.gmail.com/

mbolivar-nordic:
I agree we should do what the user says. I don't follow the example about multiple revision-in fields, though. West doesn't keep those refs around.

mbolivar-nordic:
(And I would expect revision-in to be a str, not a list of str)

marc-hb:
> (And I would expect revision-in to be a str, not a list of str)

This was a (IMHO good) suggestion by @tejlmand on the 12th and 13th of December 2019 and agreed by the other, much better looking @mbolivar on the 16th and 17th December :-)

Long story very short the idea was "convenience" - for instance when switching to a different SHA1 locally. Now I realize it could also be used to mitigate the volatile branch issue you very recently mentioned in #319

marc-hb:
> For someone who didn't know that github doesn't allow you to fetch directly with SHA1,...

I just discovered by (submodule) chance you can actually fetch a SHA1 directly... but only if the SHA1 is full/not abbreviated. See demonstration script in https://github.com/thesofproject/sof/pull/3609

I never expected SHA1s to act as "passwords"...

This works even if the SHA1 is only in a _forked_ repo! Now we know exactly why forking is so fast on github: it is really just an illusion :-)

tejlmand:
> > For someone who didn't know that github doesn't allow you to fetch directly with SHA1,...
> 
> I just discovered by (submodule) chance you can actually fetch a SHA1 directly... but only if the SHA1 is full/not abbreviated. See demonstration script in [thesofproject/sof#3609](https://github.com/thesofproject/sof/pull/3609)

I don't think you have found a loophole.
I think that github has enabled the possibility to fetch SHAs directly.

Git itself has supported this for as long as I remember, but github used to block this.
I have just tested, and it does seems that Github has enabled this possibility :tada:

marc-hb:
I was also aware that options like `allowReachableSHA1InWant` etc. existed. There are a few things new to me and some not just to me:
- I didn't realize these options don't resolve abbreviated SHA1s. Can you confirm they never resolve? The corresponding sections in `git help config` don't seem to mention abbreviated SHA1s :-( Abbreviated SHA1s are obviously very relevant to west.
- As you wrote, this seems new on github. Curious which one(s) of the 3 `allow*SHA1InWant` github changed recently. Is there a way to ask the server what its configuration is? Or just trial and error :-( Again this is very relevant to west which should provide at least good error messages/suggestions in case of subtle `allow*SHA1InWant` differences across mirrors and subsequent breakages.
- While it makes sense from a performance perspective, I'm still amazed this works from a repo forked on github without even submitting a pull request. Again very interesting implications wrt. mirroring.

> Git itself has supported this for as long as I remember

2015? https://stackoverflow.com/questions/14872486/retrieve-specific-commit-from-a-remote-git-repository/30701724

tejlmand:
> I was also aware that options like `allowReachableSHA1InWant` etc. existed. There are a few things new to me and some not just to me:
> 
>     * I didn't realize these options don't resolve abbreviated SHA1s. Can you confirm they never resolve? The corresponding sections in `git help config` don't seem to mention abbreviated SHA1s :-( Abbreviated SHA1s are obviously very relevant to west.
> 

To my knowledge, only 40-hex shas are supported.

>     * As you wrote, this seems new on github. Curious which one(s) of the 3 `allow*SHA1InWant` github changed recently. Is there a way to ask the server what its configuration is? Or just trial and error :-( 

I guess trial and error is needed. Just like trial and error was the only way to discover github didn't support fetching of shas.

> Again this is very relevant to west which should provide at least good error messages/suggestions in case of subtle `allow*SHA1InWant` differences across mirrors and subsequent breakages.
> 

Although it is very nice that github now supports this, then the question is if `west` should rely on that ?
Cause `west` may also be used to work with git repos hosted else-where (where the feature is disabled), or with bitbucket.

>     * While it makes sense from a performance perspective, I'm still amazed this works from a repo forked on github without even submitting a pull request. Again very interesting implications wrt. mirroring.
> 

Not when it is a github fork, then everything probably end up in the same place, as any branch/fork is just a reference.
Which is also why github can easily display: 
image

But if you manually fork a repo, like is done here: https://github.com/nrfconnect/sdk-zephyr
then github does not relate the two repos.

> 
> > Git itself has supported this for as long as I remember
> 
> 2015? https://stackoverflow.com/questions/14872486/retrieve-specific-commit-from-a-remote-git-repository/30701724

That's also long time back :wink: 
I have been using git a bit longer than that, but not all advanced features of git that far back.

marc-hb:
> Although it is very nice that github now supports this, then the question is if west should rely on that ?

Basic west features should of course not depend on specific `allow*SHA1InWant` server configuration but the real question is more complicated than this. For instance one question is: should `west` prevent some users from taking advantage of lenient `allow*SHA1InWant` setting(s) just to keep the experience and documentation consistent across servers? Etc.

mbolivar-nordic:
>  For instance one question is: should `west` prevent some users from taking advantage of lenient `allow*SHA1InWant` setting(s) just to keep the experience and documentation consistent across servers? Should error messages mention `allow*SHA1InWant`? Etc.

The answers to questions like this are always "let the user decide".

Maybe something like:

```yaml
manifest:
  remotes:
  - name: foo-gh
    url-base: https://github.com/foo
    fetchSHA: <true|false>
  projects:
  - name: bar
    remote: foo-gh
  - name: baz
    url: https://git.example.com/bazorg/baz
    fetchSHA: <true|false>
```

I.e. allow a remote configuration to request fetching a SHA directly (and inheriting that in the project) or allowing a project to specify it directly.

I guess even if fetchSHA is true, we would want to keep the fallback code which just fetches everything.

That would seem to make this issue obsolete.

marc-hb:
> ` fetchSHA: <true|false>`

I think I don't like this. Dealing with server variability seems painful enough, adding even more variability on the client side seems to just double the number of possible cases for no reason obvious to me. The combinatorial explosion feels already large to me: SHA or ref? Abbreviated or not? `revision-in` or not? Single `revision-in` or multiple? `allow*SHA1InWant` or not? Three different of them! Also multiply by the number of remotes because each server can be configured differently. Then someone mirrors some repos and changes the configuration again without realizing it, puzzling users why they can use the reference but fail to use the mirror - or the other way round.

I think the design, implementation, documentation, error handling and error messages should strive to be as generic and to cover as many cases as possible in one go because even when trying very hard to be generic there are still going to be painful / unexpected corner cases. The west layer of indirection should also be kept as thin as possible so users can literally copy/paste any failed `git fetch` command and remove `west[.yml]` from the list of suspects as quickly as possible.

mbolivar-nordic:
> The west layer of indirection should also be kept as thin as possible so users can literally copy/paste any failed `git fetch` command and remove `west[.yml]` from the list of suspects as quickly as possible.

I mean... that's been possible since the beginning and "literally copy/paste [...] failed `git fetch`" commands is exactly what the documentation tells users to do:

https://docs.zephyrproject.org/latest/guides/west/repo-tool.html#troubleshooting

(Though that section should probably move to the troubleshooting page)

I'm not really sure what your objection is; could you clarify?

marc-hb:
My objection is to:
> fetchSHA: <true|false>

... and to any logic that does not fit on a post-it note (see below)

> git fetch" commands is exactly what the documentation tells users to do:

Perfect. So maybe I got lost in this (too) long discussion but I still fail to understand why `west` needs to even look at the  values of `revision-in` and `revision`. I still don't understand why `west` can't just pass these values to git directly; "garbage in, garbage out".

- if `revision-in` then fetch (all) `revision-in`
- fetch `revision`
- checkout `revision`

Done! No?

Granted: I omitted possible optimizations like not fetching SHA1s (and tags?) already there but they are hopefully all completely independent from the dead simple logic "documented" above. I mean if the documentation ends up more complicated than these three items then I feel something must be wrong.

tejlmand:
> 
>  - if revision-in then fetch (all) revision-in
>  - fetch revision
>  - checkout revision
>
> Done! No?

I like the approach, especially as it allows people to not add `revision-in` if the remote server supports it.
But there is one hatch to this, see below.

> UPDATE: OK, step 2 is a bit more complicated. How about this:
>    (isSHA1(revision) and found(revision)) else fetch revision

The risk with this is that a user updating the manifest locally may not notice that the remote server doesn't support fetching of SHAs directly, and thus push a manifest with a SHA that cannot be fetched be everyone.

Depending on the setup, a CI might be able to obtain the SHA, in which case everything appears to be working, until someone else comes along, or someone forks the repo to a git server which doesn't allow fetching of SHAs.

But this may also happen today, but because we only fetch the default refs space (branches / tags) it is not that likely to happen.

Also we must be aware that a SHA not connected with a branch/tag will be garbage collected at some point in time and can lead to all sorts of confusions. _I could do `west update` yesterday, but now it fails......_

So allowing fetching of SHAs without a corresponding `revision-in` may lead to all sorts of confusion when SHAs are being garbage collected.

If we can live with this, then I think the approach is sane, and allows fetching of SHAs not on a branch/tag.

marc-hb:
> But this may also happen today, but because we only fetch the default refs space (branches / tags) it is not that likely to happen.

So how about this: if fetching `revision` fails and `revision-in` exists, then suggest (temporarily) commenting out `revision-in`. If `west update` still fails even after commenting out `revision-in`, then it means the SHA1 that the user added to the manifest was so bad that  `revision-in` feature made absolutely no difference, it did not make the problem worse.

marc-hb:
> (isSHA1(revision) and found(revision)) else fetch revision

Actually, make that:

```
isImmutable(revision) and found(revision) else fetch revision
```

... so `abcdef~3` is also supported. Only when `abcdef~3` is local or indirectly fetched of course, speaking of which:

> Depending on the setup, a CI might be able to obtain the SHA, in which case everything appears to be working, until someone else comes along, or someone forks the repo to a git server which doesn't allow fetching of SHAs.

I just realized the risk of "CI might be able to [accidentally] fetch the SHA" is IMHO close to zero because who is going to add a full, 40 hex characters long SHA in a manifest? Plus who is going to blindly approve such unusually long weirdness at manifest review time without raising questions?

All _abbreviated_ SHAs successfully tested in CI/by others have been _indirectly_ fetched, agreed? 

This also explains why you sound like you never had any "I can't find my SHA" complaints with the current, default refs space (branches + tags) fetch. Confirm?

Now there is still the problem of force-pushing `revision-in` + garbage-collection but again there is nothing new here, the exact same problem can happen with the default refs space - and again you never heard anyone complaining about this, correct?

So maybe teams are smart enough not to submit random SHA in their manifests, or they are at least smart enough to realize if/when they've done such a mistake.

In fact I think this new `revision-in` feature should behave almost exactly like "fetch a non-default refs space" with almost no code change, I mean no logical change visible enough to require a significant documentation change. I'm still waiting for a counter example that demonstrates why `revision-in` has to be logically different from "fetch a non-default refs space", sorry if I missed it.

Merely changing the default fetch: very long discussion, very small feature after all :-D

"As simple as possible, but not simpler"

tejlmand:
> All abbreviated SHAs successfully tested in CI/by others have been indirectly fetched, agreed?
>
> This also explains why you sound like you never had any "I can't find my SHA" complaints with the current, default refs space (branches + tags) fetch. Confirm?

Because the default ref-space is fetched all the time, then you get all the SHAs in this ref-space, so you will never experience this situation unless a branch is removed (which contained SHAs that are becoming orphan and thus garbage collected some point in time).

The situation of a `I can't find my SHA` could actually happen here:
https://github.com/nrfconnect/sdk-zephyr/
where history is sometimes re-written to keep the patch-set as clean as possible compared to Zephyr mainline repo.
To avoid SHAs being garbage collected then tagging is done BEFORE a re-write of the repo.

When a repository is forked, then it's the maintainer of the fork responsibility to decide what to include in that fork.
If some tags or branches are left out of such fork, you can actually end in a situation where specific SHAs are not available in a downstream fork.

That is one reason why keeping upstream SHAs available is a good thing, as described:
https://docs.zephyrproject.org/latest/guides/modules.html#allowed-practices
> > This approach is applicable to modules with an upstream project Git repository. The main advantages of this approach is that the upstream repository history (that is, the original commit SHAs) is preserved in the module repository.

Dangling SHAs are never cloned on a fork, so allowing to fetch those in a manifest could also lead to new issues where an upstream projectA is forked downstream to projectA*. Consider a working manifest available that points to dangling SHA in projectA.
Now, someone adjust the manifest to use the downstream fork projectA*, but the missing SHAs are not cloned into that fork.
So trying to do `west update` on that manifest will now fail, but no one in upstream manifest discovered this, cause everything is working there.
This situation may happen to any project that forks Zephyr and one or more of Zephyr's module repos, if we start allowing direct fetching of SHAs, not available in the default ref-space.

Direct fetching of SHAs, compared to fetching the default ref-space, bears the risk that things appears to be working, until they don't.

Remember, most SHAs in the manifest may be updated on regular basis, so that the SHA has been updated to SHA* before SHA was not garbage collected. But if suddenly doing a `west update` on an old manifest (bi-sect ?) could lead to issues later in time.

> In fact I think this new revision-in feature should behave almost exactly like "fetch a non-default refs space"

Seems you missed the main point of the `revision-in`.
The `revision-in` is a performance optimization to avoid fetching the whole default ref-space, but allow to fetch just enough of the ref-space to find the required SHA:
https://github.com/zephyrproject-rtos/west/pull/344#issue-351118357
> > If provided, it will be fetched during 'west update'
whenever the revision looks like a SHA, instead of syncing the entire
remote ref space.

If you want to fetch a non-default ref-space, you can do so already, such as:
```
revision: pull/344/head
```
which is a non-default ref-space.

marc-hb:
> To avoid SHAs being garbage collected then tagging is done BEFORE a re-write of the repo.

Thanks, this is exactly what I wrote: (current) west users seem git-savvy enough to never use dangling SHAs in their manifests. Maybe they're savvy because west is not required to make others' git life miserable; a simple force push is enough. They already learned about dangling SHAs the hard way, long before they used west!


> If some tags or branches are left out of such fork, you can actually end in a situation where specific SHAs are not available in a downstream fork.

So your main fear is that someone adds a SHA1 that belongs to one branch or tag but makes the mistake of pointing `revision-in` at a different branch and does not realize their mistake, correct? OK, but:
- Any CI or person that clones from scratch will detect this very soon AND it will be relatively obvious to debug.
- Of course it's somewhat subjective but he risk of users merely rewriting history and using an old SHA dangling _across all refs_ seems much higher to me than mixing up the branches a SHA belongs to. Yet complaints about the former did not escalate to you; see above. Users are smart.
- again, no risk for abbreviated SHAs or 1fbf627bcc~3

I hope I understand the various risks you described but I still think you overestimate them. I also feel like you didn't clearly separate the description of each risk. Maybe you could focus on just what you think is the ONE most likely risk? This sort of branch mix-up?

> if we start allowing direct fetching of SHAs, not available in the default ref-space,...

Fetching SHAs and not fetching all the refs are related but they're different decisions. I've been struggling to follow which one you're discussing when.

> > In fact I think this new revision-in feature should behave almost exactly like "fetch a non-default refs space" [ correction: in the cases where west currently fetches the default refs ]

> Seems you missed the main point of the revision-in.
> The revision-in is a performance optimization to avoid fetching the whole default ref-space, but allow to fetch just enough of the ref-space to find the required SHA:

At first I felt like just rephrased what I wrote, so thank you for giving this example, it helped me:

> revision: pull/344/head

BTW I'm afraid we have not yet strictly defined what a "SHA" is and used slightly different definitions depending on the context. Does west documentation defines what it considers a SHA versus not? I see some interesting `git rev-parse` commands in `west -v update`...
```
1fbf627bcc739227cd7bb821dc2f0f51b134aa28 # fetchable sometimes
1fbf627bcc   # not fetchable ever?
1fbf627bcc~3 # not a SHA but all the same properties: immutable and unambiguous
deadbeef     # a branch treated as a SHA by west (and by git!?)
```
... and maybe more variations? Multiplied by the `allow*SHAInWant` matrix and multiplied by the number of servers = massive combinatorial explosion, totally untestable! Hence "Don't Parse It, Just Fetch It" as often as possible :-)

tejlmand:
> > To avoid SHAs being garbage collected then tagging is done BEFORE a re-write of the repo.
>
> Thanks, this is exactly what I wrote: (current) west users seem git-savvy enough to never use dangling SHAs in their manifests. > Maybe they're savvy because west is not required to make others' git life miserable; a simple force push is enough. They already learned about dangling SHAs the hard way, long before they used west!

Users doing history re-write are hopefully above the average git user.
This was only an example of how a dangling SHA could appear with current implementation.

If we allow fetching SHAs directly, there will be a higher risk.

Let me give an example, with current design, a manifest update must follow `pull/<no>/head` while there is a dependency to PRs in other repos, like here:
https://github.com/zephyrproject-rtos/zephyr/pull/30410/files
```
  - name: trusted-firmware-m
    path: modules/tee/tfm
    revision: pull/35/head
```
now, it is clear to any reviewer that this PR SHOULD NOT be merged as long as revision is `pull/35/head`.

Because SHAs are not fetched directly, contributers cannot make a mistake.

If we allow fetching a SHA directly, this PR could look as:
```
  - name: trusted-firmware-m
    path: modules/tee/tfm
    revision: 20bc267b9eee613dca61e8612664fb7df2f9ad3f
```
which is the current SHA of `pull/35/head`.
but as a reviewer I cannot judge if this SHA is in the default ref-space or not.
Only way to tell is by examine it manually, which is much more cumbersome then just looking at `pull/35/head`.

Also CI will be able to fetch `20bc267b9eee613dca61e8612664fb7df2f9ad3f`.
but if the `pull/35/head` PR is later forced pushed then the SHA will be dangling, but work until garbage collected.

Also, when PR `pull/35/head` is rebased onto master, then `20bc267b9eee613dca61e8612664fb7df2f9ad3f` will change to a new value, but both SHAs can be fetched, meaning the manifest is valid. However, `20bc267b9eee613dca61e8612664fb7df2f9ad3f` might be garbage collected some point in time.

So opening for direct SHA fetching means we risk merging manifest PRs that points to a SHA not connected to the default ref space.

Thus the risk of a manifest pointing to a dangling SHA increases significantly.
Today, this may only happen in very special situations, like the history re-write case, and hopefully user doing such things knows what they are doing.

But fetching of SHAs means that users can suddenly make manifest updates where revision is a dangling SHA.

This is my main worry.

mbolivar-nordic:
@tejlmand I think that worry comes down to a user error, though. CI should be smart enough (especially now that we have a manifest bot) not to take DNM off until the SHA is available in a branch somewhere.

marc-hb:
> Users doing history re-write are hopefully above the average git user.
> [...]
> Today, this may only happen in very special situations, like the history re-write case,

I don't consider  interactive `git rebase` to be "advanced" git usage. For [gitHUB users maybe](https://github.com/isaacs/github/issues/959#issuecomment-592964757) but not for git users in general.

When I started using git I felt liberated because I could finally use version control for myself, meaning without all my trial and error shared and recorded forever in the project history. The code review process of very many projects expect you apply review comments by rewriting git commits. In fact [Zephyr itself expects this even when github treats force pushes as second class citizens](https://github.com/zephyrproject-rtos/zephyr/pull/14444#issuecomment-575320941).


> which is the current SHA of pull/35/head.

Pointing at an un-merged pull request: a particular but very clear and real example, thank you! Made even riskier by github's "fetch anything from any fork" that I just discovered and reported :-(

I agree this risk is real. I don't think it's huge because if not all submitters, at least manifest reviewers understand what a dangling commit is. If they don't then you have a "bigger problem" IMHO.

Here's a dead simple way to catch this one in CI: mirroring. Don't let CI fetch from github directly. You may be doing this already for performance or firewall reasons.

If you are still scared, maybe one middle ground to reduce the risk would be to fetch SHAs only in `revision-in` but not SHAs in `revision`. This would effectively double the role of `revision-in` as some "Are you sure?" type of confirmation while minimizing changes to the current behaviour. After all the initial purpose of this issue is/was to fetch SHA _indirectly_ :-)

Also keep in mind that `allowTipSHA1InWant` may make it possible to fetch one SHA only to indirectly download another SHA that is not directly fetchable. Now I can't see a real world use case like this but it seems possible :-)

PS: With hindsight I find Zephyr's habit of using full, 40 hex SHA in west.yml surprising.

marc-hb:
There has been a lot of messy (especially from me) but useful (from others) brainstorming in this issue but it has grown too long to still be useful. I think the best next steps would be to write fairly extensive **pseudo-code** for that new feature in a new github issue, something at a level intermediate between future documentation and future Python code. It should include a formal definition of what west (already) considers to be a SHA1, e.g. `deadbeef`. I'm unfortunately not volunteering, I have neither enough knowledge nor enough time.

By the way west should always try to fetch strings like `deadbeef` that are not exactly 40 hex characters long because: 1. they could be refs, 2. if they're actually a SHA, fetching their abbreviated form is  going to fail anyway. I digress again sorry.

tejlmand:
> I don't consider interactive git rebase to be "advanced" git usage. For gitHUB users maybe but not for git users in general.

on a local branch, for sure this may often be done, but I don't hope you re-write history on a master or release branch.
This is something that is hopefully only done by a user above the average git user. A user who knows what a dangling SHA is, and how to avoid that. Not to mention, a user who is not being surprised by a detached head.

I have seen many git users being worried whenever the see `HEAD detached at <SHA>`.

tejlmand:
> @tejlmand I think that worry comes down to a user error, though. CI should be smart enough (especially now that we have a manifest bot) not to take DNM off until the SHA is available in a branch somewhere.

Users make errors. But with today's solution they will not be able to push any SHA, only SHAs valid in the Zephyr default ref-space.
Which means they must push a `pull/<no>/head` revision, and thereby making things clear to everyone.

Also, fixing this in CI means that everyone who uses `west` downstream for projects must implement the same fix.
Why not have this handled directly by `west` as today ?

And there are other git servers out there besides github which may not allow this kind of fetching.

> not to take DNM off until the SHA is available in a branch somewhere.

This SHA: `d16ab3a62398548ee45e2aaaa82f6eb7d7309bac`
can be fetched using:
`git fetch https://github.com/zephyrproject-rtos/zephyr d16ab3a62398548ee45e2aaaa82f6eb7d7309bac`

and is available at a branch somewhere on a Zephyr-fork (but not in Zephyr default ref-space).
So for CI to be able to check this, then CI would need to fetch the complete Zephyr ref-space to ensure that SHA IS NOT available in the Zephyr ref-space (how else can it tell to fail the PR ?) before marking the PR `DNM`.

But with `pull/<no>/head` only that ref needs to be fetched, and not the whole ref-space for the repo.
So what you are suggesting would actually be a performance penalty for CI.

marc-hb:
> on a local branch, for sure this may often be done, but I don't hope you re-write history on a master or release branch. [...] I have seen many git users being worried whenever the see HEAD detached at <SHA>.

There's no technical difference between a local git branch and a remote git branch, the only difference is that one is shared and the other is private. They're created the sae way. People understand what a dangling commit is in either both cases or in none.

I'm assuming that people who use git rebase interactive understand what a dangling commit is. I mean I'm assuming that they notice that old SHAs after rewrite are not valid any more. If they have never noticed that the SHAs are changing then it means they don't even know what a SHA is; in that case why would they add a SHA in a manifest _and_ submit the change without any idea about what they're doing?

> But with today's solution they will not be able to push any SHA, only SHAs valid in the Zephyr default ref-space.

That's not entirely true. Even today, if someone force-pushes a production branch (bad) AND CI does not clone from scratch for performance reasons then you can still have old and dangling commits in the manifest. What you can't submit is more specific:  local/private SHAs - in any solution - and SHA from forks because github... is buggy.

So CI must clone from scratch for correctness anyway + from a nearby mirror for performance and to avoid github's "fetch any SHA from anywhere" false convenience/bug. Once CI does both of these then cloning from scratch works either for everyone or for no one in any solution. You also need CI to clone from scratch every time because of the new `revision-in` feature.

> is available at a branch somewhere on a Zephyr-fork (but not in Zephyr default ref-space).

The more I think about this, the more I consider this to be a github specific bug. The simple workaround is to have a local mirror indirection for both performance and correctness. Or... to have some "advanced" `FetchSHA true/false` setting in the manifest that you want to set to the safer "false" value when dealing with buggy servers like github. Which proves @mbolivar-nordic  was way ahead of us and totally outsmarted us :-) I'm still very concerned about the combinatorial explosion and lack of test coverage.

marc-hb:
> The revision-in is a performance optimization to avoid fetching the whole default ref-space, but allow to fetch just enough of the ref-space to find the required SHA:

This topic came back in #475 and in "shallow" #319  and I suggested the name "narrow clone" for #475 and this 344 here. git-repo calls it `sync-c`

> `no-hash-fetch`, `fetchSHA`

I find these too low-level to be part of the user interface. Giving west higher level goals is IMHO a better user interface than specific steps towards those goals. Here's a high level goal:

=> To save disk and time, please west attempt to keep the clone `--narrow`  if you can. I understand sometimes `uploadpack.allow*SHA1InWant` will force you to fall back (with a warning?) but please give it a try. Please fetch `revision-in` first, it might help to keep the clone narrow.

BTW there are 3 different `uploadpack.allow*SHA1InWant`, meaning some SHA1s can be fetchable and others not from the same server. So `revision-in: SHA` is a  valid use case.

PR #346, opened 2019-12-17
git: clean west ref space after fetching

tejlmand:
As discussed in PR #331 the cleanup of `refs/west` has been split into separate PR.

Note: Cleanup is done BEFORE updating the `manifest-rev` as this ensures the update is the newest entry in reflog, as example:
```
fec2a3c (HEAD, ncs/master, manifest-rev) manifest-rev@{0}: west update: moving to fec2a3c3b4d79b7eb6d1ebdf24e6813476116769
9da9943 (tag: v1.1.0, ncs/v1.1-branch) manifest-rev@{1}: west fetch: entry for refs/west/v1.1-branch
efcbc25 (tag: v1.0.0, ncs/v1.0-branch) manifest-rev@{2}: west fetch: entry for refs/west/v1.0-branch
....
```

### DIFF
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -24,6 +24,7 @@ from west.manifest import ImportFlag, Manifest, MANIFEST_PROJECT_INDEX, \
     ManifestProject, _manifest_content_at
 from west.manifest import MANIFEST_REV_BRANCH as MANIFEST_REV
 from west.manifest import QUAL_MANIFEST_REV_BRANCH as QUAL_MANIFEST_REV
+from west.manifest import QUAL_REFS_WEST as QUAL_REFS
 from urllib.parse import urlparse
 import posixpath
 
@@ -711,11 +712,21 @@ class Update(_ProjectCommand):
             _clone(project)
 
         if self.fs == 'always' or _rev_type(project) not in ('tag', 'commit'):
-            new_manifest_rev = _fetch(project)
+            _fetch(project)
         else:
             log.dbg('skipping unnecessary fetch')
-            new_manifest_rev = '{revision}^{{commit}}'
-        project.git(['update-ref', QUAL_MANIFEST_REV, new_manifest_rev])
+            _update_manifest_rev(project, '{revision}^{{commit}}')
+
+        # Head of manifest-rev is now pointing to current manifest revision.
+        # Thus it is safe to unconditionally clear out the refs/west space.
+        #
+        # Doing this here instead of in _fetch() ensures that it gets cleaned
+        # up when users upgrade from older versions of west (like 0.6.x) that
+        # didn't handle this properly.
+        # In future, this can be moved into _fetch() after the install base of
+        # older west versions is expected to be smaller.
+        _clean_west_refspace(project)
+
         if not _head_ok(project):
             # If nothing is checked out (which usually only happens if we
             # called _clone(project) above), check out 'manifest-rev' in a
@@ -845,6 +856,25 @@ class SelfUpdate(_ProjectCommand):
 # Private helper routines.
 #
 
+def _clean_west_refspace(project):
+    # Clean the refs/west space to ensure they do not show up in 'git log'.
+
+    # Get all the ref names that start with refs/west/.
+    list_refs_cmd = ('for-each-ref --format="%(refname)" -- ' +
+                     QUAL_REFS + '**')
+    cp = project.git(list_refs_cmd, capture_stdout=True)
+    west_references = cp.stdout.decode('utf-8').strip()
+
+    # Safely delete each one.
+    for ref in west_references.splitlines():
+        delete_ref_cmd = 'update-ref -d ' + ref
+        project.git(delete_ref_cmd)
+
+def _update_manifest_rev(project, new_manifest_rev):
+    project.git(['update-ref',
+                 '-m', f'west update: moving to {new_manifest_rev}',
+                 QUAL_MANIFEST_REV, new_manifest_rev])
+
 def _maybe_sha(rev):
     # Return true if and only if the given revision might be a SHA.
 
@@ -947,25 +977,22 @@ def _fetch(project, rev=None):
     fetch_cmd = ('fetch -f --tags ' +
                  ('--depth={clone_depth} ' if project.clone_depth else ' ') +
                  '-- {url} ')
+
+    log.small_banner(project.format(msg))
     if _maybe_sha(rev):
         # We can't in general fetch a SHA from a remote, as many hosts
         # (GitHub included) forbid it for security reasons. Let's hope
         # it's reachable from some branch.
-        fetch_cmd += 'refs/heads/*:refs/west/*'
-        ret = rev
+        fetch_cmd += 'refs/heads/*:' + QUAL_REFS + '*'
+        project.git(fetch_cmd)
+        _update_manifest_rev(project, '{revision}')
     else:
         # The revision is definitely not a SHA, so it's safe to fetch directly.
         # This avoids fetching unnecessary ref space from the remote.
         # We need {{commit}} instead of {commit} because everything gets run
         # through Project.format.
-        fetch_cmd += rev
-        ret = 'FETCH_HEAD^{{commit}}'
-
-    log.small_banner(project.format(msg))
-    project.git(fetch_cmd)
-
-    return ret
-
+        fetch_cmd += '{revision}:' + QUAL_MANIFEST_REV
+        project.git(fetch_cmd)
 
 def _head_ok(project):
     # Returns True if the reference 'HEAD' exists and is not a tag or remote
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -36,6 +36,9 @@ MANIFEST_REV_BRANCH = 'manifest-rev'
 #: A fully qualified reference to `MANIFEST_REV_BRANCH`.
 QUAL_MANIFEST_REV_BRANCH = 'refs/heads/' + MANIFEST_REV_BRANCH
 
+#: Git ref space used by west for internal purposes.
+QUAL_REFS_WEST = 'refs/west/'
+
 #: The latest manifest schema version supported by this west program.
 #:
 #: This value changes when a new version of west includes new manifest


tejlmand:
@marc-hb it's not possible to add you to this review, but feel free to review anyway :smile:

src/west/commands/project.py
```diff
@@ -726,6 +727,19 @@ def do_run(self, args, user_args):
 # Private helper routines.
 #
 
+def _clean_west_refspace(project):
+    west_refs_cmd = ('for-each-ref --format="%(refname)" -- ' +
```
marc-hb: Short (py)doc explaining the rationale. AFAIK:
- garbage collect and save disk space after (configurable!) reflog expiration
- avoid potentially large and unbounded number of refs

src/west/manifest.py
```diff
@@ -35,6 +35,10 @@
 #: A fully qualified reference to `MANIFEST_REV_BRANCH`.
 QUAL_MANIFEST_REV_BRANCH = 'refs/heads/' + MANIFEST_REV_BRANCH
 
+#: The west refs space used for current manifest revision.
+#: West ref space is temporary space for fetching of SHAs.
+WEST_REF_SPACE = 'refs/west/'
```
mbolivar: Should we call this something like QUAL_REFS_WEST for consistency with QUAL_MANIFEST_REV_BRANCH?

marc-hb:
> Can we get unlucky and have our cleaned up space garbage collected before the next update-ref?

Assuming it's an actual risk, a separate `fetch-history` branch would also avoid this.

tejlmand:
> > Note: Cleanup is done BEFORE updating the `manifest-rev` as this ensures the update is the newest entry in reflog, as example:
> 
> Is that safe? Can we get unlucky and have our cleaned up space garbage collected before the next update-ref?

it would require: `gc.reflogExpire=now` and the `git-gc` is also executed.

Unfortunately I have been unable to find a specific list of commands that triggers automatic gc.
The closest reference I've found is:
https://git-scm.com/docs/git-gc
```
When common porcelain operations that create objects are run, 
they will check whether the repository has grown substantially 
since the last maintenance, and if so run git gc automatically.
```
but we are not creating objects with the `update-ref` commands being used, so I would say the proposed behavior should be safe.

On a side-note, I also believe users could experience a very bad performance of `west` if they set `gc.reflogExpire=now` as it could end up in fetching on each `west update` as references could be gone.

tejlmand:
> > Can we get unlucky and have our cleaned up space garbage collected before the next update-ref?
> 
> Assuming it's an actual risk, a separate `fetch-history` branch would also avoid this.

but that would give another `branch` that could potentially show up in `git log`
which is exactly what we try to avoid with cleaning up `refs/west`.

src/west/manifest.py
tejlmand: Consistency is king :+1:

src/west/commands/project.py
```diff
@@ -726,6 +727,19 @@ def do_run(self, args, user_args):
...
+                     WEST_REF + '**')
+    cp = project.git(west_refs_cmd, capture_stdout=True)
+    west_references = cp.stdout.decode('utf-8').strip()
+
```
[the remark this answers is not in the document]
tejlmand: Almost: https://git-scm.com/docs/git-for-each-ref
```
git-for-each-ref - Output information on each ref
```

src/west/commands/project.py
tejlmand: > Short (py)doc explaining the rationale. AFAIK:
> 
>     * garbage collect and save disk space after (configurable!) reflog expiration
> 
>     * avoid potentially large and unbounded number of refs

I don't think our rationale is to save disk space.
The main purposes is to:
1) Be able to fetch a SHA from a git server (reason for `refs/west`)
2) minimize when fetching is needed by keeping a reference to SHAs (west smart fetching)
  Also achieved `refs/west`
3) Avoid `refs/west` shows up, as example in `git log`

So the cleanup is mostly for 3., but the reflog part is to hopefully keep the refs alive for longer time. (not cleaning `refs/west` would have kept them alive until user did a manual clean)

But will add some description :wink:

src/west/commands/project.py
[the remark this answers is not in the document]
tejlmand: cp = capture, I assume.
Just sticking to current style: 
https://github.com/zephyrproject-rtos/west/blob/49d348ce901f6ef38ae92e80d1b647560f8de5e8/src/west/manifest.py#L768
https://github.com/zephyrproject-rtos/west/blob/49d348ce901f6ef38ae92e80d1b647560f8de5e8/src/west/commands/project.py#L756
https://github.com/zephyrproject-rtos/west/blob/49d348ce901f6ef38ae92e80d1b647560f8de5e8/src/west/commands/project.py#L811
etc.

src/west/commands/project.py
```diff
@@ -726,6 +727,19 @@ def do_run(self, args, user_args):
...
+    for ref in west_references.splitlines():
+        update_ref_cmd = ('update-ref -m "west fetch: entry for ' +
+                          '{}" {} {}'.format(ref, QUAL_MANIFEST_REV, ref))
+        project.git(update_ref_cmd)
+        update_ref_cmd = 'update-ref -d ' + ref
+        project.git(update_ref_cmd)
+
```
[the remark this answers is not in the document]
tejlmand: Done

src/west/commands/project.py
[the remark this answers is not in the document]
tejlmand: `west_refs_cmd` --> `list_refs_cmd`

marc-hb:
> but that would give another branch that could potentially show up in git log
which is exactly what we try to avoid with cleaning up refs/west.

I understood this tries to avoid an unbounded and never `--purge`d list of refs. Ver far from 2 refs.

tejlmand:
> > but that would give another branch that could potentially show up in git log
> > which is exactly what we try to avoid with cleaning up refs/west.
> 
> I understood this tries to avoid an unbounded and never `--purge`d list of refs. Very far from 2 refs.

Even two refs are still being debated here:

From https://github.com/zephyrproject-rtos/west/pull/331#discussion_r356221346:
> > @marc-hb Good point on updating the reflog, and thus cleaning the refs/west will also help te ensure the SHAs are available for a certain time, while not have all those refs/west references popping up in git log.
> > As I reply in #331 (comment) , there needs to be done some cleanup in refs/west, and this appears to be a nice way of doing that.
> 
> I assumed by your "this appears to be a nice way of doing that" in response to my question "do we need this", (and by "this", I mean this PR, #331), that you agreed we did not need to keep storing things in refs/west, and we can just use the reflog on the manifest-rev branch.

From https://github.com/zephyrproject-rtos/west/pull/331#discussion_r356279735
> > ```
> > 2. Keep current `refs/west/<revision>` for non-SHA - As this is controversial
> > ```
> 
> And this. I don't mind keeping `refs/west/<revision>` if it helps use cases that matter to people. 

Although two refs are not directly being rejected I personally prefer we start with the cleanup part first.
Regarding `refs/west/fetch-history` I'm not so fond, for the following reasons:
1. West already have `manifest-rev`, and is described here: 
    https://docs.zephyrproject.org/latest/guides/west/repo-tool.html?highlight=manifest%20rev#the-manifest-rev-branch
    This branch is fully under `west` control and its purpose is to be used by `west`. 
    If users want to examine it, as example the reflog, then I see no issue with that, but we should not start creating additional ref-logs, as `manifest-rev` primary target is not user support.
2. We cannot place fetch-history under `refs/west/fetch-history` cause it will be deleted on each update.
3. `fetch-history` may pop-up randomly, depending on last `update-ref` see example below.

Example for 3.
Manifest points to SHA `17b765fa` which is HEAD on master.
```
$ west update -f always nrfxlib
=== updating nrfxlib (nrfxlib):
--- nrfxlib: fetching, need revision 17b765fac9fa3cb58f0bff22f7b04ff1c3e7bf93
From https://github.com/NordicPlayground/nrfxlib
...
--- nrfxlib: checked out 17b765fac9fa3cb58f0bff22f7b04ff1c3e7bf93 as detached HEAD
$ git log -n1
commit 17b765fac9fa3cb58f0bff22f7b04ff1c3e7bf93 (HEAD, manifest-rev)
```

As seen, `fetch-history` does not show up, cause this SHA is not latest update in `fetch-history`
But if SHA is `9da9943d` which is HEAD on `v1.1-branch` then the following is seen.
```
$ west update -f always nrfxlib
=== updating nrfxlib (nrfxlib):
--- nrfxlib: fetching, need revision 9da9943d870096a5c9cc49dbf5e515783d0f9911
From https://github.com/NordicPlayground/nrfxlib
...
--- nrfxlib: checked out 9da9943d870096a5c9cc49dbf5e515783d0f9911 as detached HEAD
$ git log -n1
commit 9da9943d870096a5c9cc49dbf5e515783d0f9911 (HEAD, tag: v1.1.0, manifest-rev, fetch-history)
```

This behavior is random, as it depends on the order we update `fetch-history`, which again depends on the output from `git for-each-ref`

Note: `-f always` is used to ensure entries are created in reflog for `fetch-history`

tejlmand:
> 
> ```
> When common porcelain operations that create objects are run, 
> they will check whether the repository has grown substantially 
> since the last maintenance, and if so run git gc automatically.
> ```
> 
> but we are not creating objects with the `update-ref` commands being used, so I would say the proposed behavior should be safe.

Just did a check-up, and as stated above automatic run of `git gc` only happens on certain porcelain commands, but the `update-ref` is stated under plumbing commands, so we are safe with this approach.
https://git.github.io/htmldocs/git.html

marc-h38:
Sorry, I assumed that:

1. `git log --decorate` showed only  "standard namespaces" by default, I mean only `refs/heads/*`, `refs/tags/*` and `refs/remotes/*/*, HEAD,...` and that you had to explicitly ask for more like `refs/west/*`. Clearly not.
2. Non-standard namespaces like `refs/west/*` got a reflog "for free" (`git help reflog` amusingly lists branches and undefined "others references")

Both seem wrong (but see below).

So yes, `fetch-history` can randomly pop-up in `git log --decorate`. I think it's a very minor inconvenience and small price to pay for not polluting the reflog of `manifest-rev` and for keeping it usable.

> This [manifest-rev] branch is fully under west control and its purpose is to be used by west.
> [...] as manifest-rev primary target is not user support.

"read-only" and "user support" are not mutually exclusive. I believe the reflog of manifest-rev is extremely useful. Is there any other way to log what happened in the module? Please don't answer `git log -p west.yml`, it's not even close.


Now after consulting with my local git guru, I have a fix for 2:
```
 git update-ref logs/refs/west/fetch-history refs/west/fetch-history
 git revlog west/fetch-history
```
This creates a reflog on a non-standard ref.

And then an obvious workaround for 1 follows:
```
 git log --decorate-refs-exclude=west/*
```

mbolivar:
> but we are not creating objects with the `update-ref` commands being used, so I would say the proposed behavior should be safe.

Thanks for checking that.

I need a bit more time to go through the rest of the discussion. BTW, is @marc-h38 a @marc-hb sockpuppet? Did we just unmask you?

src/west/commands/project.py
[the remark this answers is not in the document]
mbolivar: completed process ;)

tejlmand:
> "read-only" and "user support" are not mutually exclusive. 

Agree, but I still want to keep the main focus. `manifest-rev` is primarily for west itself.

> I believe the reflog of manifest-rev is extremely useful.

I don't believe current implementation is useful, cause all you get with `git reflog manifest-rev` today is empty messages:
```
$  git reflog manifest-rev
04a049bda (HEAD, tag: mbedtls-2.16.3, manifest-rev) manifest-rev@{0}: 
53546ea09 (tag: mbedtls-2.13.1) manifest-rev@{1}: 
04a049bda (HEAD, tag: mbedtls-2.16.3, manifest-rev) manifest-rev@{2}: 
53546ea09 (tag: mbedtls-2.13.1) manifest-rev@{3}: 
```
No messages has ever been provided with `update-ref` before :wink:
But I would like to make the reflog of manifest-rev a bit more useful, while still keep focus on main reason for having `manifest-rev`.

>  I think it's a very minor inconvenience and small price to pay for not polluting the reflog of manifest-rev and for keeping it usable.

We are not keeping reflog for manifest-rev usable, we are making it usable.
I still have a very hard time to see the reasons the `west` implementation should do all the tricks you propose when a user can simply do:
```
$ git reflog manifest-rev|grep -v 'west fetch'
9da9943 manifest-rev@{0}: west update: moving to 9da9943d870096a5c9cc49dbf5e515783d0f9911
17b765f manifest-rev@{1}: west update: moving to 17b765fac9fa3cb58f0bff22f7b04ff1c3e7bf93
0f6217f manifest-rev@{6}: fetch -f --tags -- https://github.com/NordicPlayground/nrfxlib pull/102/head:refs/heads/manifest-rev: forced-update
fec2a3c manifest-rev@{7}: 
```
and get all entries due to `west fetch` filtered out.

Note: the empty entry is current behavior without this PR.
> This creates a reflog on a non-standard ref.
> 
> And then an obvious workaround for 1 follows:
> 
> ```
>  git log --decorate-refs-exclude=west/*
> ```

Still a workaround, when a user can still do: `git reflog manifest-rev|grep -v 'west fetch'`
And then all users doing `git log` does not need to care.

marc-hb:
> I still have a very hard time to see the reasons the west implementation should do all the tricks you propose

Maybe we don't understand each other because I don't what "all these tricks" are. I'm merely asking not to hijack a single ref for two completely purposes, simply recommending to use one ref per purpose instead, that's all. The only drawback of this simpler solution is a random and tiny noise in the git log.

> when a user can simply do git reflog manifest-rev|grep -v 'west fetch':

Sure,`grep -v` is an obvious and simple enough way to clean/filter a mixup. But why mix things up in the first place? This raises the bar for the messages BTW.

> No messages has ever been provided with update-ref before 😉

Not mutually exclusive: you can have better reflog messages AND no mixup. I have absolutely no doubt you're making things better; I'm making suggestions to make them even better (and simpler).

> Still a workaround, when a user can just do: ...

I think you missed that particular point because your answer doesn't seem to match it. Not an important one.

tejlmand:
> > I still have a very hard time to see the reasons the west implementation should do all the tricks you propose
> 
> Maybe we don't understand each other because I don't what "all these tricks" are. I'm merely asking not to hijack a single ref for two completely purposes, simply recommending to use one ref per purpose instead, that's all. The only drawback of this simpler solution is a random and tiny noise in the git log.

Agree. Seems I have misunderstood your meaning.
I thought you only found the additional log messages noisy, but in fact your point of view is that the two purposes, updating `manifest-rev` and keeping SHAs in a reflog, should be separated.

And from that view, I agree, it's not a workaround.

Although, I'm personally not convinced that the two purposes should be split into two reflogs.
@mbolivar  any thoughts on this ?



> >    No messages has ever been provided with update-ref before wink

> Not mutually exclusive: you can have better reflog messages AND no mixup. I have absolutely no doubt you're making things better; I'm making suggestions to make them even better (and simpler).

Agree. And your comments are highly appreciated, as your comments can even improve the solution even more.
Although i'm still not convinced :wink:

src/west/commands/project.py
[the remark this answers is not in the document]
mbolivar: @marc-hb comment was addressed, so resolving

tejlmand:
> Happy new year! 

Happy new year :smile: 
> 
> Summary:
> 
>     * I think cleaning up refs/west is orthogonal from reflog maintenance, and this PR was created specifically to get the refs/west clean-up steps merged. So we should focus on that here.
> 

Partly agree. I just found the improving the reflog messages to be a nice low-hanging fruit.

>     * I don't think it's useful to try to maintain a reflog for the revisions in refs/west in this pull request, as that's just for internal use and we don't want to expose too many implementation details on what's inside right now.

Well, if we don't store the refs for the pruned branches, we might do additional (not-needed) fetching later.
Keeping the ref in the reflog for `manifest-rev` avoids the ref to be garbage collected immediately.

With your proposed changes, refs from other branches will be garbage collected faster.
As example, with the proposed changes:
```
mkdir -p /tmp/west-test/nrf
cd /tmp/west-test
git -C nrf init .
git -C nrf fetch file:///<path-to-manifest-repo>/nrf
git -C nrf checkout FETCH_HEAD
west init -l nrf
west -v update nrfxlib
# ...
# --- nrfxlib: fetching, need revision 1d48321863678362e42a72654b67e355b5c1d27d

# invoke a gc.
git -C nrfxlib gc --prune=all 

# Update nrf/west.yml nrfxlib revision to: 0f6217fc8c6bed2dadc7658a5404478bae1e1
# And update the project
west -v update nrfxlib
# ...
# --- nrfxlib: fetching, need revision 0f6217fc8c6bed2dadc7658a5404478bae1e1
```
as seen above, the already fetched SHA needs to be fetched again cause it was garbage collected.

With this PR in its current state, which stores the refs in the reflog, the SHA is still locally available, as seen:
```
# Same initial steps as above: ^^^
west -v update nrfxlib
# ...
# --- nrfxlib: fetching, need revision 1d48321863678362e42a72654b67e355b5c1d27d

# invoke a gc.
git -C nrfxlib gc --prune=all 

# Update nrf/west.yml nrfxlib revision to: 0f6217fc8c6bed2dadc7658a5404478bae1e1
# And update the project
west -v update nrfxlib
# ...
# skipping unnecessary fetch
# --- nrfxlib: checked out 0f6217fc8c6bed2dadc7658a5404478bae1e1d20 as detached HEAD
```

@mbolivar your suggestion is exactly what I initially proposed here: https://github.com/zephyrproject-rtos/west/pull/331#issuecomment-561646897
> > As all references in refs/west are removed, it means that all SHAs not referenced from elsewhere (as example a tag or local branch) will be a loose object.
> > Loose objects are per default pruned after two weeks.

Personally, i'm fine with that approach, but I feel like going in circles now :stuck_out_tongue: 


>     * I would like to clear out refs/west _after_ updating manifest-rev, to avoid any possibility of race conditions where GC kicks in after cleanup, but before the `update-ref`.

As seen above, once the ref is in the reflog, there is no danger of running gc.
Secondly, no commands we are using will trigger the gc, as the `update-ref` is a plumbing command.
See also this comment: https://github.com/zephyrproject-rtos/west/pull/346#issuecomment-567407288

But agree, if we are not keeping the ref in the reflog, things are a little more dangerous.

mbolivar:
> Well, if we don't store the refs for the pruned branches, we might do additional (not-needed) fetching later.
> Keeping the ref in the reflog for `manifest-rev` avoids the ref to be garbage collected immediately.

I did see that in your written motivation, but I don't think it's true for the revisions we really care about, which are the project `revision` attributes in west.yml files.

First, if the older project `revision`s are ancestors of the new ones (i.e. we're developing in a branch and there are no history rewrites), then we won't throw any of the old commits away, because they never become garbage in that branch, and we never change branches. This is the common case.

Second, west fetches (and never deletes) tags from the remote. So if a user wants to rewrite history and then push the results into the same branch, they can force west to keep the old history by pushing it into a new tag. I would call that basic good engineering practice for this type of situation unless the old history is something they really don't want their users to have around anymore, like if it had some security credential they accidentally leaked (and in that case, west ought to throw it away too).

What am I missing?

> @mbolivar your suggestion is exactly what I initially proposed here: https://github.com/zephyrproject-rtos/west/pull/331#issuecomment-561646897
> Personally, i'm fine with that approach, but I feel like going in circles now :stuck_out_tongue:

In further discussion on the same thread, you suggested splitting this work up into two pieces:

>Maybe this PR should now be split into 2:
>
>    1. Cleaning of refs/west ref space when fetching SHAs - We definately want this
>    2. Keep current refs/west/<revision> for non-SHA - As this is controversial

I agreed that splitting this up would be good:

https://github.com/zephyrproject-rtos/west/pull/331#discussion_r356279735

I'm assuming that this PR (i.e. PR https://github.com/zephyrproject-rtos/west/pull/346, which is titled "git: clean west ref space after fetching", after all) is addressing " 1. Cleaning of refs/west ref space when fetching SHAs - We definately want this".

So the only thing I expect in this PR is cleaning up of refs/west when fetching SHAs. And I've already agreed with you that this is a good idea.

This reflog discussion seems to be outside the scope of this PR, then, no? Did I forget something?

marc-hb:
> What am I missing?

The end-user editing her manifest to west fetch and test various branches/commits?

marc-hb:
> The end-user editing her manifest to west fetch and test various branches/commits?

Or, maybe more likely, simply switching her manifest itself to a different branch and running `west update`.

Once upon a time, I asked on the [Slack west channel](https://app.slack.com/client/T18N0JX19/) whether this was a good idea:
```
git branch --set-upstream-to=manifest-rev mywork
```
The very short answer was: "yes, but only until you use west for real" (= branching the manifest itsef)

tejlmand:
> 
> 
> > Well, if we don't store the refs for the pruned branches, we might do additional (not-needed) fetching later.
> > Keeping the ref in the reflog for `manifest-rev` avoids the ref to be garbage collected immediately.
> 
> I did see that in your written motivation, but I don't think it's true for the revisions we really care about, which are the project `revision` attributes in west.yml files.
> 
> ..........
>
> What am I missing?
> 

A use-case which I personally use, and which I know others who works combined upstream/downstream uses, is to edit `.west/config` when switching and then run `west update`
Which could actually be considered branching of the manifest.

As example, `.west/config`:
```
[manifest]
path = nrf
```
Running
```
$ west update
# Do some downstream work
```
update `.west/config`
```
[manifest]
path = zephyr
```
Running
```
$ west update
# Do some upstream work
```
but as the SHA available in the history of manifest-rev reflog for the SHA that is actually checked-out, I consider this to be a little risk to most people.
Remember, any SHA that is ever referenced by the manifest will live in the reflog until expiration.
So only head of branches not checked out will become a loose object, but still live for 2 weeks per default. 
Meaning the additional reflog entry part is purely an optimization for reduced fetching.

And if no SHA on a given remote branch has not been needed for two weeks, it might be fine to get it pruned.

> > @mbolivar your suggestion is exactly what I initially proposed here: [#331 (comment)](https://github.com/zephyrproject-rtos/west/pull/331#issuecomment-561646897)
> > Personally, i'm fine with that approach, but I feel like going in circles now 😛
> 
> In further discussion on the same thread, you suggested splitting this work up into two pieces:
> 
> > Maybe this PR should now be split into 2:
> > 
> > 1. Cleaning of refs/west ref space when fetching SHAs - We definately want this
> > 2. Keep current refs/west/ for non-SHA - As this is controversial
> 
> I agreed that splitting this up would be good:
> 
> [#331 (comment)](https://github.com/zephyrproject-rtos/west/pull/331#discussion_r356279735)
> 
> I'm assuming that this PR (i.e. PR #346, which is titled "git: clean west ref space after fetching", after all) is addressing " 1. Cleaning of refs/west ref space when fetching SHAs - We definately want this".
> 
> So the only thing I expect in this PR is cleaning up of refs/west when fetching SHAs. And I've already agreed with you that this is a good idea.
> 
> This reflog discussion seems to be outside the scope of this PR, then, no? Did I forget something?

The it should have been splitting in 3, I guess:
1. Cleaning of refs/west ref space when fetching SHAs - We definately want this
2. Keep current refs/west/ for non-SHA - As this is controversial
3. Place a reflog entry for cleaned refs/west revision to increase lifetime of SHAs and allow user control through expiration

with only splitting in two, I considered 3 to part of 1. rather than 2. But I wasn't aware that reflog would also be controversial.

tejlmand:
Rebased, and adjusted so no entry is added to the reflog of manifest-rev when cleaning `refs/west`

Then we can always add reflog entries later, if we find that it improve uesr experience.
For now, I think the 2 week default life time for loose objects are sufficient.
Remember that any SHA ever checked out will already have a reflog entry, and hence not be pruned until reflog expiration

mbolivar:
@marc-hb wrote:
> > The end-user editing her manifest to west fetch and test various branches/commits?
> 
> Or, maybe more likely, simply switching her manifest itself to a different branch and running `west update`.

@tejlmand wrote:
> Remember, any SHA that is ever referenced by the manifest will live in the reflog until expiration.
So only head of branches not checked out will become a loose object, but still live for 2 weeks per default.

Right. Any branches which are actually checked out into manifest-rev stick around in the short term in the reflog. So they aren't lost for a while, even if they become unreachable due to switching branches.

Also, I don't know where the "2 week" default is coming from. It seems to be incorrect to me: https://git-scm.com/docs/git-reflog#Documentation/git-reflog.txt---expire-unreachablelttimegt says --expire-unreachable defaults to 30 days (and --expire defaults to 90 days).

@tejlmand I also don't see why the upstream/downstream example you give is relevant. It seems to me that even in this case, `refs/west/master` is going to get overwritten when you switch your manifest.path, since both upstream and downstream SHAs are likely available in upstream/master and downstream/master.

src/west/commands/project.py
```diff
@@ -711,11 +712,21 @@ def update(self, project):
             _clone(project)
 
         if self.fs == 'always' or _rev_type(project) not in ('tag', 'commit'):
-            new_manifest_rev = _fetch(project)
```
mbolivar: Looks good! Almost there, thanks for all your patience.

Just one small change: now that the return value for _fetch() is being thrown away, I think `_fetch()` itself does not need to return anything. Other than that this looks great, thank you again for your work on this.

mbolivar:
> The it should have been splitting in 3, I guess:
> 
>     1. Cleaning of refs/west ref space when fetching SHAs - We definately want this
> 
>     2. Keep current refs/west/ for non-SHA - As this is controversial
> 
>     3. Place a reflog entry for cleaned refs/west revision to increase lifetime of SHAs and allow user control through expiration
> 
> 
> with only splitting in two, I considered 3 to part of 1. rather than 2. But I wasn't aware that reflog would also be controversial.

It's only controversial for me since I am trying to get code merged incrementally, and it's a separate issue from cleaning up the refs space where I don't think the problem statement and use cases are clearly stated yet. Let's definitely tackle the reflog issue next.

tejlmand:
> Also, I don't know where the "2 week" default is coming from. It seems to be incorrect to me: https://git-scm.com/docs/git-reflog#Documentation/git-reflog.txt---expire-unreachablelttimegt says --expire-unreachable defaults to 30 days (and --expire defaults to 90 days).

I'm referring to loose objects.
Most of the time there are not many branches, but let's just consider upstream Zephyr being used as an project.
Most of the time, the SHA might be located in `zephyr/master`, but if `west update` needs to fetch a SHA, then [topic-gpio](https://github.com/zephyrproject-rtos/zephyr/tree/topic-gpio), [topic-sensors](https://github.com/zephyrproject-rtos/zephyr/tree/topic-sensors), etc. branches are all fetched.
During clean-up of `refs/west`, those branches will become loose objects.
In case the manifest is later updated to point to a SHA on `topic-gpio`, as example : [226a9f25](https://github.com/zephyrproject-rtos/zephyr/commit/226a9f25d2d112b80d1a7580c9462a02ffac29e4), then this SHA , although fetched earlier, might no longer be available.

This is where the 2 weeks (default) comes in:
https://git-scm.com/docs/git-gc#Documentation/git-gc.txt---pruneltdategt

Of course, if user runs `git gc --prune=all` it will be pruned immediately.

This is the situation where a reflog entry for `topic-gpio`, `topic-sensors`, etc. would have kept the revision staying around for longer time.

> 
> @tejlmand I also don't see why the upstream/downstream example you give is relevant. It seems to me that even in this case, `refs/west/master` is going to get overwritten when you switch your manifest.path, since both upstream and downstream SHAs are likely available in upstream/master and downstream/master.

No, it's not directly related.
But the chance of jumping between SHAs located on various branches becomes higher when working both up-/downstream, as that means that it is not only branches but also different repos and this might lead to an increased risk of the situation above to happen.

But if only working on a single branch, e.g. master, both up-/downstream, then there is no difference.

For the described situation above to occur, the manifest needs to start pointing to a SHA on a different branch, which were fetched but never checked out (--> loose object).

> Let's definitely tackle the reflog issue next.

The reflog issue is not important to me, cause I believe the 2 week time for loose object is enough.
If user haven't needed a SHA in 2 weeks from any branch which were fetched into `refs/west`, and afterwards deleted, then I find it ok that it's pruned.
If at any time within those 2 weeks the user actually needs such a SHA, it will go into the reflog anyway.

I'm just trying to describe what the creation of a reflog entry will actually do.
And my only reason to create the reflog entry in first place, was per reviewer request :smile: 

So I would normally put the reflog further down prioritization list.

src/west/commands/project.py
tejlmand: good catch.
Fixed.

mbolivar:
> And my only reason to create the reflog entry in first place, was per reviewer request smile
> 
> So I would normally put the reflog further down prioritization list.

FWIW, I am happy with reflog maintenance on manifest-rev only. That's all I was hoping to see. Thanks for the cleanup!

src/west/commands/project.py
```diff
@@ -845,6 +856,25 @@ def do_run(self, args, user_args):
...
+
+    # Get all the ref names that start with refs/west/.
+    list_refs_cmd = ('for-each-ref --format="%(refname)" -- ' +
+                     QUAL_REFS + '**')
+    cp = project.git(list_refs_cmd, capture_stdout=True)
+    west_references = cp.stdout.decode('utf-8').strip()
+
+    # Safely delete each one.
+    for ref in west_references.splitlines():
+        delete_ref_cmd = 'update-ref -d ' + ref
```
marc-hb, line 870: I'm not sure what "safely" meant here but this very ironically... not safe :-) 4 years later, fix submitted in
- #683

PR #348, opened 2020-01-08
west manifest --resolve

mbolivar:
@carlescufi @tejlmand would you mind taking a look?

This is the last loose end that I realized I needed to implement before we can cut the beta for west manifest imports.

### DIFF
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -444,45 +444,67 @@ class ManifestCommand(_ProjectCommand):
             textwrap.dedent('''\
             Manages the west manifest.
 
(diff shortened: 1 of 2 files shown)


mbolivar:
If we get this merged, I will:

1. tag and publish west pre-release v0.6.99.dev3 to pypi
2. send email to users@ and devel@ with the announcement and asking for feedback and testing. I'll include a link to https://github.com/mbolivar/my-zephyr-app to make it easy to test.

mbolivar:
@tejlmand @carlescufi sorry to nag, but as this is a blocker and a relatively small patch, could you please take a look?

Issue #349, opened 2020-01-13
3ac600ac breaks fetches from tags


mbolivar:
Commit 3ac600acaa112fb5d7edfcfd9f1873ed2edc516c breaks setting a revision to a tag.

For example, with https://github.com/mbolivar/my-zephyr-app/blob/202a5ffba528e95255e8c62de919c7510db349ad/west.yml, I get:

```
$ west update
=== updating zephyr (zephyr):
--- zephyr: cloning and initializing
Initialized empty Git repository in /tmp/zephyr-imports/zephyr/.git/
--- zephyr: fetching, need revision v2.1.0
remote: Enumerating objects: 394472, done.
remote: Total 394472 (delta 0), reused 0 (delta 0), pack-reused 394472
Receiving objects: 100% (394472/394472), 284.55 MiB | 8.41 MiB/s, done.
Resolving deltas: 100% (294858/294858), done.
error: cannot update ref 'refs/heads/manifest-rev': trying to write non-commit object a8a050b7385f5c9c5478412f143f45d8e90045f2 to branch 'refs/heads/manifest-rev'
From https://github.com/zephyrproject-rtos/zephyr
 ! [new tag]         v2.1.0         -> manifest-rev  (unable to update local ref)
[...snip...]
https://github.com/mbolivar/my-zephyr-app/blob/202a5ffba528e95255e8c62de919c7510db349ad/west.yml
```

mbolivar:
@tejlmand I am working on a fix. No action on your side is requested.

tejlmand:
> 
> 
> @tejlmand I am working on a fix. No action on your side is requested.

great.
Probably that's why the fetching was done first, and then do `update-ref`

Just restore old behavior for that part.

tejlmand:
https://github.com/zephyrproject-rtos/west/pull/331#discussion_r355721212
> Fetching directly into manifest-rev leaves a nice clean entry in my reflog:

So now we know why this wasn't done that way originally.

mbolivar:
> So now we know why this wasn't done that way originally.

Indeed. Sorry for all the bother.

PR #350, opened 2020-01-13
Tag fixes

mbolivar:
The main point of this PR is commit `update: fix _fetch() when revisions are tags`, which:

Fixes: #349

Everything else is just scaffolding and regression testing.

### DIFF
--- a/src/west/commands/project.py
+++ b/src/west/commands/project.py
@@ -1000,11 +1000,17 @@ def _fetch(project, rev=None):
         _update_manifest_rev(project, '{revision}')
     else:
         # The revision is definitely not a SHA, so it's safe to fetch directly.
-        # This avoids fetching unnecessary ref space from the remote.
-        # We need {{commit}} instead of {commit} because everything gets run
-        # through Project.format.
-        fetch_cmd += '{revision}:' + QUAL_MANIFEST_REV
+        # This avoids fetching unnecessary refs from the remote.
+        #
+        # We update manifest-rev to FETCH_HEAD instead of using a
+        # refspec in case the revision is a tag, which we can't use
+        # from a refspec.
+        #
+        # We need {{commit}} instead of {commit} because everything
+        # gets run through Project.format().
+        fetch_cmd += '{revision}'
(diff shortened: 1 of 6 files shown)


mbolivar:
@carlescufi @tejlmand can you take a look at this? Unfortunately, #349 broke the manifest imports instructions in https://github.com/mbolivar/my-zephyr-app, so we need this before we can do a beta (which will be 0.6.99.dev4 with this PR).

I have tested this PR with my-zephyr-app on Linux, macOS, and Windows, and verified that tox passes on all three also.

src/west/commands/project.py
```diff
@@ -1000,11 +1000,14 @@ def _fetch(project, rev=None):
...
-        # through Project.format.
-        fetch_cmd += '{revision}:' + QUAL_MANIFEST_REV
+        # This avoids fetching unnecessary refs from the remote.
+        #
+        # We update manifest-rev to FETCH_HEAD instead of using a
+        # refspec in case the revision is a tag, which we can't use
+        # from a refspec.
+        fetch_cmd += '{revision}'
         project.git(fetch_cmd)
+        _update_manifest_rev(project, 'FETCH_HEAD^{{commit}}')
```
carlescufi: You are missing the `We need {{commit}} instead of {commit} because everything gets run through Project.format` comment, which was very informative.

mbolivar: > You are missing the We need {{commit}} instead of {commit} because everything gets run through Project.format comment, which was very informative.

@carlescufi fixed, thanks. In the future, I think I want to disable that by default and make this user-selectable with a `fmt=True` keyword argument. It's a bit too magical for an API.

mbolivar: > I think I want to disable that by default and make this user-selectable with a `fmt=True` keyword argument. It's a bit too magical for an API.

You know what, on second thought, I want to remove it without making it user selectable. Experimenting with this a bit, we can just use f-strings now that we require Python 3.6 or later.  The results are quite readable and not magic.

PR #351, opened 2020-01-15
housekeeping: python 3.6

mbolivar:
Now that manifest imports are out and in users' hands, I'd like to take an opportunity to do some housekeeping PRs while we're waiting for feedback.

This is the first such PR, which makes west a more modern python 3.6 code base, primarily by standardizing on f-strings as a way to format information. This removes the west.manifest.Project.format() method and its uses in Project.git() and friends as well: it was always too magical.


src/west/commands/__init__.py
```diff
@@ -45,8 +45,9 @@ def __init__(self, **kwargs):
         self.hint = kwargs.pop('hint', None)
         super(ExtensionCommandError, self).__init__(**kwargs)
 
-_NO_TOPDIR_MSG_FMT = '''\
-no west installation found from "{}"; "west {}" requires one.
+def _no_topdir_msg(cwd, name):
```
carlescufi: Not sure why you defined a function instead of keeping the global with the `{cwd}` and `{name}` inside them. Isn't the f-string evaluated whenever you actually use it? or does that happen when you define it?

src/west/manifest.py
```diff
@@ -1062,6 +1052,10 @@ def posixpath(self):
             self._posixpath = PurePath(self.abspath).as_posix()
         return self._posixpath
 
+    @property
```
carlescufi: I didn't realize property decorators were also not supported in 3.4

tests/test_project.py
```diff
@@ -533,16 +532,16 @@ def do_run(self, args, ignored):
                               print('Testing kconfig test command')
                       '''),
                       })
-    west_tmpdir = repos_tmpdir.join('west_installation')
-    cmd('init -m "{}" "{}"'.format(str(repos_tmpdir.join('repos', 'zephyr')),
-                                   str(west_tmpdir)))
+    west_tmpdir = repos_tmpdir / 'west_installation'
```
carlescufi: is `repos_tmpdir` created with pathlib? or does this `/` work with standard `os.path` objects or even with strings?

src/west/manifest.py
mbolivar: They are, it's just so we can use {name_and_path} from f-strings.

src/west/commands/__init__.py
mbolivar: To be honest, I don't know if I could have done this with an f-string, but I don't think so because the format arguments aren't in global scope, so I put them on the function stack to make them available.

tests/test_project.py
mbolivar: This repos_tmpdir is an os.PathLike object that ultimately is derived from a pytest 'tmpdir' fixture object, which is a pytest LocalPath instance. I'm not sure, but my guess is that the design of LocalPath predates and influenced the standard library pathlib.Path types, since they work in very similar ways.

mbolivar:
Thanks for review, I will rebase and merge the rest of the open PRs.

PR #357, opened 2020-01-18
manifest: support maps of imports

mbolivar:
This completes the feature set for west manifest imports. We include a
test of name whitelists from the documentation as a basic sanity
check. Further testing is left to future work.

To close out this feature after this, we should only need more error
handling and testing. In particular, beyond general coverage issues,
we need to try to reject any obvious import cycles, and add tests for
the same.

### DIFF
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -636,7 +636,7 @@ class Manifest:
             for subimp in imp:
                 self._import_from_self(mp, subimp, projects)
         elif imptype == dict:
-            self._import_map_from_self(mp, imp, projects)
+            self._import_map(mp, imp, self.import_path_from_self, projects)
         else:
             self._malformed(f'{mp.abspath}: "self: import: {imp}" '
                             f'has invalid type {imptype}')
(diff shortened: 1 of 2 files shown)


carlescufi:
LGTM, no objections

src/west/manifest.py
```diff
@@ -892,8 +890,67 @@ def _import_content_from_project(self, project, path):
...
+
+        imap = self._load_imap(p, imp)
+
+        all_imported = {}
+        base_importer(p, imap.file, all_imported)
+
+        for name, project in all_imported.items():
+            if _is_imap_ok(project, imap):
+                if name in imap.rename:
+                    project.name = imap.rename[name]
```
tejlmand: Seems you're only doing a rename of the project name.
What would happen if `path` had been defined in the project, and this path is colliding with the path of a similar project in downstream `west.yml` ?

tejlmand:
@carlescufi  seems you're a bit fast 😛

carlescufi:
> @carlescufi seems you're a bit fast 😛

You haven't been reviewing a lot lately and @mbolivar needs to make progress. I read through the code, thought it was reasonable and merged it. We can iterate on master.

src/west/manifest.py
mbolivar: I'm not sure I understand your question. Renaming is meant to catch cases where I have something called 'app' both upstream and downstream, and I want to have both without having to track the upstream version in my west.yml. I don't see how a path collision is possible, since project.path is not changed and we already check for project path collisions by calling _check_paths_are_unique() from _load().

mbolivar: Perhaps we should remove the 'rename' feature to get started, since it doesn't make as much sense now that we've moved to your suggested 'ignore duplicate definitions' semantics?

Looking at it again, it has the feature of something that hasn't been thought through enough yet. It is easy to put back if we get users asking for it.

Issue #149, opened 2019-01-10
Global west state makes it harder to unit test 


mbolivar:
West has some global state that makes it harder to write proper unit tests. Examples:

- global configuration state is written into a module-level object in west.config
- log verbosity level is too

This makes it harder to write proper unit tests and causes other annoyances.

mbolivar:
This bugs me all the time when I write tests, and I'm constantly having to hack around it, as shown in #271. Bumping priority to medium as a result (since it's a problem that is slowing down development, despite not being user-visible).

mbolivar-nordic:
We're down to `west.log` at this point. We can close this issue once that deprecated module is removed sometime in the west 2.0 or later timeline.

marc-hb:
Progress from @pdgendt in https://github.com/zephyrproject-rtos/zephyr/pull/79171 and others.

marc-hb:
More in
- https://github.com/zephyrproject-rtos/zephyr/issues/92475
- https://github.com/zephyrproject-rtos/zephyr/issues/97228

PR #358, opened 2020-01-22
west.app reorg

mbolivar:
This PR adds a new west.app package which contains all the non-API code, fixing #38.

I want to do that as a setup step for replacing direct use of west.log from API modules like west.manifest (related to #149 and general good practice) before reworking the west.manifest diagnostics so that `west -v manifest --resolve` prints extra debugging info helpful to users (as discussed offline with @carlescufi).

I tossed in a .gitignore patch as well.


mbolivar:
Thanks! Will follow up with manifest parse diagnostics.

PR #359, opened 2020-01-22
Logging reorg

mbolivar:
This replaces direct use of west.log from west.manifest with the standard python logging module. There are a few miscellaneous fixes I found along the way.

While doing this, I figured out why our logging in `west -v manifest --resolve` is incorrectly printing messages about some projects being added (in particular, it prints them twice). I won't fix this in this PR, though, as we'll need to start passing an additional "don't log" keyword argument when doing some types of recursive manifest loads, and respecting it.

I will need a bit more time to look at the code paths and implement that.


mbolivar:
Never mind; I'm going to resubmit as a series of bugfixes with the logging included as one of them.

PR #361, opened 2020-01-23
github: Use a github action for testing

galak:
Introduce a github action that will run tox on a number of python
versions and host OSes for testing.

Signed-off-by: Kumar Gala <kumar.gala@linaro.org>

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,23 @@
+name: Python Test
+
+on: [push,pull_request]
+
+jobs:
+  build:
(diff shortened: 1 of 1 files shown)


galak:
For west testing I think we should move to GH actions away from shippable.  Plus side is easy to test several python variants and different host OSes (windows, macos x, and linux).

galak:
Can see run here:

https://github.com/galak/west/actions

mbolivar:
I'll merge this, but I might ask you for help if I need to make changes since I don't know GH Actions at all.

PR #363, opened 2020-01-27
More import tests

mbolivar:
Add some more regression tests.

PR #364, opened 2020-01-30
Manifest import tweaks

mbolivar:
DNM, depends on https://github.com/zephyrproject-rtos/west/pull/363

This pull request covers some remaining work in manifest imports:

- git repository remote names match west manifest remote names on clone again 
- we allow west update on projects defined in the manifest repository, even if there are imports
- don't update projects twice when we've already done so via imports

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -641,22 +641,31 @@ class Update(_ProjectCommand):
         # call our importer whenever it encounters an import statement
         # in a project, allowing us to control the recursion so it
         # always uses the latest manifest data.
+        self.updated = set()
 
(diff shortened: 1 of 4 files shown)


mbolivar:
> manifest: factor private context into a namedtuple 

This last commit is just prep work for adding a 'path-prefix' map import key that lets users remap where imported projects should live on the file system. It's not supported yet -- still TODO. We need to decide if that's a release blocker.

mbolivar:
>  imported project is set back in time, prior to the manifest, as example. importing an old Zephyr tag.

Sorry, what do you mean?

You can use `import: true` with `revision: v1.0` with no troubles.

PR #366, opened 2020-02-03
0.7.0rc1

mbolivar:
This is west 0.7.0.rc1, containing

- import loop detection (at least a hacky version)
- a couple of minor fixes
- updates to west.manifest.SCHEMA_VERSION and version.py

### DIFF
--- a/.shippable.yml
+++ b/.shippable.yml
@@ -18,4 +18,5 @@ build:
     - export PATH=$PATH:$HOME/.local/bin
     - pip3 install tox
     - mkdir -p shippable/{testresults,codecoverage}
+    - export WEST_SKIP_SLOW_TESTS=0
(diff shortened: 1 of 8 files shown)


mbolivar:
> Think we should consider some import list, and test if file / source is already imported, and then bail out instead of the (slow) RecursionError to trigger.

I wanted to do that, but it's tricky because of constructors like `Manifest.from_data()` which do not have an associated file. Since the important thing is to get the feature out, I decided to see if this would be good enough for now.

Thanks for review.

marc-hb:
> import loop detection (at least a hacky version)

Would you have guessed in 2020 that this humble and "hacky import loop detection" would end up catching regressions in cpython 3.13 back ports! Please review:
- #919

PR #371, opened 2020-02-05
(Low priority) more west update --stats

mbolivar:
This does a better accounting of update-related work.

It also cleans up the 'cloning and initializing' output line since it's not really what's going on, and people are likely to get confused about why it's so fast to clone repositories, then so slow to fetch from them, when in fact what's going on is that we're just initializing a local repository then fetching the remote.


mbolivar:
@galak wins the nerd sniper of the week award

mbolivar:
I'm going to wait until after 0.7 for this

Issue #374, opened 2020-02-10
west 0.7.0 broke building on windows


tejlmand:
West 0.7.0 broke building with compiler checking in windows.

Path in windows becomes wrong style:

```
c:\..\ncs\nrf\applications\asset_tracker>west build -bnrf9160_pca10090ns
<SNIP>
-- Found assembler: C:/opt/armgcc7_2018q2/bin/arm-none-eabi-gcc.exe
CMake Error at C:/Program Files/CMake/share/cmake-3.14/Modules/CheckCCompilerFlag.cmake:41 (set):
  Syntax error in cmake code at

    C:/Program Files/CMake/share/cmake-3.14/Modules/CheckCCompilerFlag.cmake:41

  when parsing string

    -fmacro-prefix-map=C:\projects\github\ncs=WEST_TOPDIR

  Invalid character escape '\p'.
Call Stack (most recent call first):
  C:/projects/github/ncs/zephyr/cmake/extensions.cmake:1345 (check_c_compiler_flag)
  C:/projects/github/ncs/zephyr/cmake/extensions.cmake:844 (check_compiler_flag)
  C:/projects/github/ncs/zephyr/cmake/extensions.cmake:1384 (zephyr_check_compiler_flag)
  C:/projects/github/ncs/zephyr/cmake/extensions.cmake:1360 (target_cc_option_fallback)
  C:/projects/github/ncs/zephyr/cmake/extensions.cmake:118 (target_cc_option)
  C:/projects/github/ncs/zephyr/CMakeLists.txt:344 (zephyr_cc_option)


-- Configuring incomplete, errors occurred!
See also "C:/projects/github/ncs/nrf/applications/asset_tracker/build/CMakeFiles/CMakeOutput.log".
See also "C:/projects/github/ncs/nrf/applications/asset_tracker/build/CMakeFiles/CMakeError.log".
FATAL ERROR: command exited with status 1: 'C:\Program Files\CMake\bin\cmake.EXE' '-BC:\projects\github\ncs\nrf\applications\asset_tracker\build' '-SC:\projects\github\ncs\nrf\applications\asset_tracker' -GNinja -DBOARD=nrf9160_pca10090ns
```

Issue #273, opened 2019-06-07
Path handling needs overhaul


mbolivar:
The path handling in west was originally written and tested on Linux, and did things like using string comparison operators to test path equality after calling os.path.abspath or realpath.

After user testing on windows, we started throwing some normpath calls around to resolve issues related to case insensitivity and the POSIX vs. Windows path separators (i.e. slash vs. backslash).

The whole thing has evolved organically to keep things "working", but needs to be overhauled, with all path handling using the pathlib module and its platform-specific helpers.

This should result in more readable code, anyway, since pathlib has various conviences. E.g. calls to os.path.join(a,b) can be replaced with a / b.

mbolivar:
cc @marc-hb

marc-hb:
As can be seen in https://github.com/zephyrproject-rtos/zephyr/pull/16537 and a number of links from there, I've been masochistically testing _symbolic links_ in the Zephyr build. Among various Zephyr issues proving that: 1. I'm the first one attempting this; 2. git killed `/mnt/` and network filesystems; I found this more specific west issue:
When ZEPHYR_BASE has symbolic links, then the output of `west list` is **different** depending on whether the current directory is inside the west installation versus not. That's because `west` has this useful and desired logic: "search [real] getcwd() first otherwise fall back on [symbolic] ZEPHYR_BASE". The fun effect on cmake's module_name = RELPATH(...) is left as an exercise for the reader[*]

After experimenting with some west changes I feel like fixing west issues with symbolic links would take neither a lot of west code changes, nor a lot of effort and could be performed gradually. **However**, I found a somewhat "code style" / naming convention issue. `west list -h`, `git grep -i -C 3 -E 'rea?lpath'` and other searches shows a widespread pattern of confusion between absolute paths and real paths, I mean stuff like this:
```
self.abspath = os.path.realpath(...
```
After this introduction that I didn't spend enough time making shorter, two questions for @mbolivar and others:

1. Considering a grander move to pathlib is desired, there's not point even starting to fix the pattern of symbolic confusion in the current `os.path` code style. Correct?

2. If/whenever this west migration to pathlib happens, it should initially be "bug-for-bug" compatible with all existing symbolic link issues. Because you should always be careful to change only one thing at a time and such refactoring should ideally have zero functional change, at least initially. Correct? 

cc @SebastianBoe , @carlescufi, @ulfalizer 

[...]

PR #375, opened 2020-02-10
topdir: west topdir prints posix style path.

tejlmand:
Fixes: #374

This commit ensures that west topdir always prints in posix style and
thus fixes an issue in windows where CMake fails to parse the path when
'\' is used as separator.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -1062,7 +1062,7 @@ class Topdir(_ProjectCommand):
         return self._parser(parser_adder)
 
     def do_run(self, args, user_args):
-        log.inf(self.topdir)
+        log.inf(PurePath(self.topdir).as_posix())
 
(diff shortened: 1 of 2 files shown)


marc-hb:
Should be the other way round: topdir should be `C:/Users/...` and converted for the dinosaurs still using `CMD.EXE` :-)

mbolivar:
> Should be the other way round: topdir should be `C:/Users/...` and converted for the dinosaurs still using `CMD.EXE` :-)

Sorry? This patch makes it C:/Users/ from C:\Users\.

marc-hb:
I half-jokingly meant: why not make the `self.topdir` field itself a POSIX style path.

mbolivar:
Hmm, that's an interesting idea :). Though I think I'd rather revisit this and add a --posix switch when we have a chance to look at #273 .

PR #377, opened 2020-02-10
Add MAINTAINERS.rst

mbolivar:
I've collected lessons learned from past releases in this file. We can use it going forward as a basis for continuing improvements to our release process.

### DIFF
--- a/MAINTAINERS.rst
+++ b/MAINTAINERS.rst
@@ -1,0 +1,83 @@
+Notes for maintainers cutting a release.
+
+Pre-release test plan
+---------------------
+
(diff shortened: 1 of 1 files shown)


mbolivar:
> This is definitely needed.
> Thanks.

Thanks for review!

@carlescufi -- if you have any comments, I'll address them in another PR.

> Now the only thing which remains is to ensure people releasing is following this guide.

Not sure what you mean. Of course that is the idea! :smile:

carlescufi:
@mbolivar no objections, this is fine.

Issue #380, opened 2020-02-13
west update is not verbose enough by default


carlescufi:
I am labeling this as a bug because it has already caused confusion among our users.

The following 2 cases should warn the user by default, with an option to disable the warning:

- `west update` with a project in the manifest's SHA, but with uncommitted changes, This should warn the user by default so that the user **is informed** that west update did not bring the "revision" to the one specified in the manifest. Many users make changes to the projects while testing, and then forget about them but really meant to undo them, and they are simply not aware right now.

- `west update` with a committed SHA (but detached HEAD) that is a descendant of the manifest's SHA. Here a user might have forgotten to create a branch, but has committed a change to the repo. Unless the user knows how to use `git reflog` it will "lose" the commit it created. Incidentally, `--keep-descendants` does not work in this case, and west reverts back to the manifest SHA.

Alternatively we can add a `west reset --hard` command. I think we need this anyway at some point so that stale folders are removed from the workspace.

CC @koffes

mbolivar:
A fix is in #382.

Issue #381, opened 2020-02-13
west update: generalize -k / -r to include commits, not just branches


mbolivar:
This is a behavioral change originally suggested in #380, so I'm opening an issue to solicit feedback.

Currently, `west update --keep-descendants` and `west update --rebase` only work if there is a branch checked out. They ignore detached heads and do not try to keep or rebase them.

This issue tracks generalizing their behavior to also work when HEAD is detached. We could have a configuration option to keep the old behavior.

cc: @carlescufi @tejlmand @marc-hb

marc-hb:
Having lost detached HEADs many times when using grepo, I couldn't agree more. grepo's "solution" to this problem is (was?) "don't forget to create branches". This is silly: we use computers precisely because they don't forget what we do.

The reason users regularly "forget to create branches" is super obvious: it's simply because plain git uses branches by default whereas grepo and west do not use branches by default. Inconsistent usability - for valid reason(s) I'm sure but inconsistent usability still.

I can think of a possibly simpler/cheaper middle ground between 1. #381 automagically rebasing everything even detached HEADs and 3. #382 just warn when something bad happens: 2. Refuse to `west rebase` if any detached HEAD is found. This could be opt-in in `~/.westconfig`

By the way I wonder whether west more generally supports the concept of a **transactional west update**. Last time I tried it did not. A plain git rebase or merge is either successful or it's not, I mean it either commits all the files or none. Plain git also has the incredibly useful `--abort` feature. Does `west` have the ability to "successfully `west rebase` all repos or none"? Part of why we have manifests is to control dependencies across repos. #338 is related.

mbolivar:
> By the way I wonder whether west more generally supports the concept of a **transactional west update**. Last time I tried it did not.

It does not, and I don't know how to do it. Is there some way to ask git "would this rebase work if I tried it?"

Without that, I'm not sure how such a thing could be done.

marc-hb:
> Is there some way to ask git "would this rebase work if I tried it?"

Well, west could try and `git --rebase abort` if it fails! Or it could implement its own git checkpoint system...

Without dreaming and digressing that far, `west` could perform a few sanity checks like looking for detached HEADs before performing anything with a side effect. Maybe it performs some other per-repo checks already?

mbolivar:
> Well, west could try and `git --rebase abort` if it fails! Or it could implement its own git checkpoint system...

[...]

PR #382, opened 2020-02-13
Fix update verbosity

mbolivar:
Fixes: #380 

I've tested the following matrix of west updates:

{descendant branch checked out + dirty tree,
 detached head + dirty tree,
 branch checked out + clean tree,
 detached descendant HEAD + clean tree}

x

{default behavior,
 --keep-descendants,
 --rebase}

And the output always warns when something surprising happens. The win is mostly due to removing the --quiet from the 'git checkout --detach' and letting git do the work for us.

I want to backport these changes to 0.7.2, so I'm intentionally preserving the existing --keep-descendants behavior. We can defer changes to that to 0.8.0 pending resolution of https://github.com/zephyrproject-rtos/west/issues/381.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -1314,8 +1314,6 @@ def die_if_no_git():
 # Top-level west directory, containing west itself and the manifest.
 WEST_DIR = '.west'
 
-# Manifest repository directory under WEST_DIR.
-MANIFEST = 'manifest'
 # Default manifest repository URL.
 MANIFEST_URL_DEFAULT = 'https://github.com/zephyrproject-rtos/zephyr'
 # Default revision to check out of the manifest repository.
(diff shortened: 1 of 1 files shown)


carlescufi:
LGTM, thanks. This is enough for the time being.

carlescufi:
Forgot to say, tested both with local uncommitted changes and with an extra commit on a detached HEAD.

src/west/app/project.py
```diff
@@ -863,11 +868,9 @@ def update(self, project):
             # out the new detached HEAD, then print some helpful context.
             if take_stats:
                 start = perf_counter()
-            project.git('checkout --detach --quiet ' + sha)
+            project.git('checkout --detach ' + sha)
```
tejlmand: unsure if this could become too noisy to some users.
But, only one way to find out.

Alternative could be to do this when `-v` is given.

mbolivar: It's the same number of lines of output when the tree is clean

PR #385, opened 2020-02-22
Fix west command duplicate warnings

mbolivar:
The import feature is printing spurious warnings for duplicated west extensions if the user imports a manifest *and* gives a west-commands that the imported manifest already declares. Fix it.

See commit log for the second patch for a detailed example.

### DIFF
--- a/src/west/commands.py
+++ b/src/west/commands.py
@@ -216,6 +216,12 @@ class WestExtCommandSpec:
         It may do some additional steps (like importing the definition of
         the command) before constructing it, however.'''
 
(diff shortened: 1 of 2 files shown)


mbolivar:
@carlescufi @tejlmand I found this while porting the NCS to use west imports. The warning is harmless but might confuse people, so I'd like to get this backported to 0.7-branch and included in 0.7.2 as well.

mbolivar:
I've tested v0.7.2a1 on pypi with e885843 as described in #386

PR #386, opened 2020-02-23
MAINTAINERS.rst: some changes from going through this once

mbolivar:
A few process changes from doing a point release, mainly that we'll
cut an alpha release instead of an RC. That's being done because
uploading an alpha is something a maintainer can just do, without
having to go through the usual "bump version.py" review process. This
makes it easier to do things like testing over the weekend.

### DIFF
--- a/MAINTAINERS.rst
+++ b/MAINTAINERS.rst
@@ -23,37 +23,38 @@ Pre-release test plan
    Make sure to check if west.manifest.SCHEMA_VERSION also needs an update
    before uploading. ::
 
-   git clean -ffdx
-   WEST_VERSION=X.YaN python3 setup.py sdist bdist_wheel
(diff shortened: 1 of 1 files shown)


mbolivar:
@tejlmand thanks for review. I've fixed a typo and tweaked the formatting but no semantic changes were made since then.

I'll give this a few more days to see if you want to review again, and then probably merge either way.

tejlmand:
> @tejlmand thanks for review. I've fixed a typo and tweaked the formatting but no semantic changes were made since then.
> 

I can only see you added an `or`.
Funny when reading text how the brain automatically can add words that are missing.

mbolivar:
> I can only see you added an `or`.

The formatting changes were done in a separate commit.

PR #391, opened 2020-03-07
Output improvements

mbolivar:
depends on https://github.com/zephyrproject-rtos/west/pull/390

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -556,16 +556,30 @@ class Diff(_ProjectCommand):
         self._setup_logging(args)
 
         failed = []
+        no_diff = 0
+        # We may need to force git to use colors if the user wants them,
+        # which it won't do ordinarily since stdout is not a terminal.
+        color = ['--color=always'] if log.use_color() else []
+
         for project in self._cloned_projects(args):
-            log.banner(f'diff for {project.name_and_path}:')
             # Use paths that are relative to the base directory to make it
             # easier to see where the changes are
-            try:
-                project.git(['diff', f'--src-prefix={project.path}/',
-                             f'--dst-prefix={project.path}/'])
-            except subprocess.CalledProcessError:
+            cp = project.git(['diff', f'--src-prefix={project.path}/',
+                              f'--dst-prefix={project.path}/',
+                              '--exit-code'] + color,
+                             capture_stdout=True, capture_stderr=True,
+                             check=False)
+            if cp.returncode == 0:
+                no_diff += 1
+            if cp.returncode == 1 or log.VERBOSE > log.VERBOSE_NONE:
+                log.banner(f'diff for {project.name_and_path}:')
+                log.inf(cp.stdout.decode('utf-8'))
+            elif cp.returncode:
                 failed.append(project)
-        self._handle_failed(args, failed)
+        if failed:
+            self._handle_failed(args, failed)
+        elif log.VERBOSE <= log.VERBOSE_NONE:
+            log.inf(f"Empty diff in {no_diff} projects.")
 
 class Status(_ProjectCommand):
     def __init__(self):
--- a/src/west/log.py
+++ b/src/west/log.py
@@ -166,20 +166,25 @@ def msg(*args, color=None, stream=sys.stdout):
     if _use_colors():
         _reset_colors(stream)
 
+def use_color():
+    '''Returns True if the configuration requests colored output.'''
+    return _use_colors(warn=False)
+
 _COLOR_UI_WARNED = False
 
-def _use_colors():
+def _use_colors(warn=True):
     # Convenience function for reading the color.ui setting
     try:
         return config.config.getboolean('color', 'ui', fallback=True)
     except ValueError as e:
-        global _COLOR_UI_WARNED
-        if not _COLOR_UI_WARNED:
-            print(f"WARNING: invalid color.ui value: {e}.",
-                  file=sys.stderr)
-            print('         To fix, run: "west config color.ui <true|false>"',
-                  file=sys.stderr)
-            _COLOR_UI_WARNED = True
+        if warn:
+            global _COLOR_UI_WARNED
+            if not _COLOR_UI_WARNED:
+                print(f"WARNING: invalid color.ui value: {e}.",
+                      file=sys.stderr)
+                print('         To fix: "west config color.ui <true|false>"',
+                      file=sys.stderr)
+                _COLOR_UI_WARNED = True
         return False
 
 def _reset_colors(file):


carlescufi:
@mbolivar please rebase

mbolivar:
> @mbolivar please rebase

done

src/west/app/project.py
```diff
@@ -556,14 +556,22 @@ def do_run(self, args, ignored):
...
-            log.banner(f'diff for {project.name_and_path}:')
             # Use paths that are relative to the base directory to make it
             # easier to see where the changes are
-            try:
-                project.git(['diff', f'--src-prefix={project.path}/',
-                             f'--dst-prefix={project.path}/'])
-            except subprocess.CalledProcessError:
+            cp = project.git(['diff', f'--src-prefix={project.path}/',
+                              f'--dst-prefix={project.path}/',
+                              '--exit-code'] + color,
```
mbolivar: BTW, I tried to do the same thing for 'git status', but ran into a potential problem. The `--porcelain=2` option is not available on Ubuntu 16.04 and I believe that was required to get machine-readable output from `git status` which can be colorized. I think we need to either:

- cache the git version in the WestApp and pass it to the WestCommand, which can check for features with something like `if self.git_version >= SOME_MIN_VERSION: # use git status --porcelain=2`, to try to keep supporting older versions of git, OR
- start tracking a minimum version of git that west's core supports and testing against it and only use features that are always available in the minimum version

This gets tricky and is a general problem, IMO.

src/west/log.py
```diff
@@ -166,20 +166,25 @@ def msg(*args, color=None, stream=sys.stdout):
...
-                  file=sys.stderr)
-            print('         To fix, run: "west config color.ui <true|false>"',
-                  file=sys.stderr)
-            _COLOR_UI_WARNED = True
+        if warn:
+            global _COLOR_UI_WARNED
+            if not _COLOR_UI_WARNED:
+                print(f"WARNING: invalid color.ui value: {e}.",
+                      file=sys.stderr)
+                print('         To fix: "west config color.ui <true|false>"',
```
tejlmand: didn't knew `west` had this feature, and the changes themselves looks good to me.

But as we are using: `config.ui`, and git is also having a `config.ui` setting, then maybe we should align allowed values.

That is, `auto`, `always`, `false`

From https://www.git-scm.com/book/en/v2/Customizing-Git-Git-Configuration
> color.ui
>
> Git automatically colors most of its output, but there’s a master switch if you don’t like this behavior. To turn off all Git’s colored terminal output, do this:
>
> `$ git config --global color.ui false`
>
> The default setting is auto, which colors output when it’s going straight to a terminal, but omits the color-control codes when the output is redirected to a pipe or a file.
>
> You can also set it to always to ignore the difference between terminals and pipes. You’ll rarely want this; in most scenarios, if you want color codes in your redirected output, you can instead pass a --color flag to the Git command to force it to use color codes. The default setting is almost always what you’ll want.

src/west/app/project.py
```diff
@@ -556,14 +556,22 @@ def do_run(self, args, ignored):
...
+                             capture_stdout=True, capture_stderr=True,
+                             check=False)
+            if cp.returncode == 1 or log.VERBOSE > log.VERBOSE_NONE:
+                log.banner(f'diff for {project.name_and_path}:')
+                log.inf(cp.stdout.decode('utf-8'))
```
tejlmand: are we sure users want this ?

If I run `west diff` today, I might see:
```
=== diff for zephyr (zephyr):
=== diff for nffs (modules/fs/nffs):
=== diff for segger (modules/debug/segger):
=== diff for mbedtls (modules/crypto/mbedtls):
....
```
but if we remove those lines per default, users might be unsure if the command was actually executed in all repos.
At minimum inform something like:
```
=== no diff for zephyr (zephyr):
=== no diff for nffs (modules/fs/nffs):
=== no diff for segger (modules/debug/segger):
=== no diff for mbedtls (modules/crypto/mbedtls):
....
```

But maybe even better could be to collect repos with no diff, and then print a final message, such as:
```
=== No diff in projects: zephyr, nffs, segger, mbedtls, ....
```
I don't think the path is important for projects without changes.

src/west/log.py
mbolivar: While I don't disagree with the suggestion, that feels like scope creep.

src/west/app/project.py
mbolivar: Good idea. Updated. It now prints something like this:

```
$ west  diff
=== diff for manifest (zephyr):
diff --git zephyr/scripts/dts/gen_defines.py zephyr/scripts/dts/gen_defines.py
index 4270eb7b4f..2d3d33a309 100755
--- zephyr/scripts/dts/gen_defines.py
+++ zephyr/scripts/dts/gen_defines.py
@@ -431,7 +431,7 @@ def phandle_macros(prop, macro):
             out_comment(f"No label macro emitted for {prop.node.name} "
                         f"property {prop.name}:\n"
                         f"{prop.val.path} has no label\n",
-                        blank_before=False)
+                        blank_before=False)  # hello
     else:
         for i, node in enumerate(prop.val):
             ret[f"{macro}_IDX_{i}_VAL_LABEL"] = quote_str(node.label)

Empty diff in 31 projects.
```

src/west/log.py
tejlmand: > While I don't disagree with the suggestion, that feels like scope creep.

That was not my intention.
I was more thinking that it would be just passing on the setting, as well as not reinventing the wheel.

But it is in principle not something which needs discussion in this PR.

Out of curiosity, why is this needed as a west config also, when user can already do:
```
git config --global color.ui auto
```
and we could just honor that setting ?

mbolivar: The west option has the same name as the git one for familiarity, but it controls west's color output, e.g. for log.wrn().

mbolivar: ... and therefore subprocesses which are capable of colored output (like git) should be encouraged to print in color when the west user has requested west colors.

tejlmand: Thanks.

src/west/app/project.py
```diff
@@ -556,16 +556,30 @@ def do_run(self, args, ignored):
...
+            elif cp.returncode:
                 failed.append(project)
-        self._handle_failed(args, failed)
+        if failed:
+            self._handle_failed(args, failed)
+        elif log.VERBOSE <= log.VERBOSE_NONE:
+            log.inf(f"Empty diff in {no_diff} projects.")
```
marc-hb, line 582: I found 3 error handling issues there, see #731 and #732

PR #393, opened 2020-03-27
west manifest: add --path action

mbolivar-nordic:
This prints the absolute path of the manifest file:

west manifest --path

Requested by @SebastianBoe

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -476,6 +476,10 @@ class ManifestCommand(_ProjectCommand):
               If the manifest can be parsed, print nothing and exit
               successfully.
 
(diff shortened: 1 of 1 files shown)


mbolivar-nordic:
@SebastianBoe does your work need this feature backported to 0.7.x? It would come out in 0.8.0 at the earliest if not, which is likely to be some time away.

SebastianBoe:
There is no pressing need yet.

SebastianBoe:
This is useful in a support situation to help users find out where their manifest file is, and it might eventually be used to solve the depedency that exists between changes to the manifest file and running west update.

Great work.

PR #401, opened 2020-05-12
RFC: main: add global --color=when cli argument

mniestroj:
Implement commandline option similar to git, which allows to select when
colors should be printed. Based on that pass strip= keyword argument to
colorama.init().

This is just a scratch of the implementation and I would like to fallback into `color.ui`. Also I think that  things like `if not _use_colors():` could be removed then.

I have found following comment in 1967d3d47a983334bcd6932f4ba0f637ddf3c978:
> I was thinking of toggling colors at a higher level with colorama
at first, but colorama.init() makes it a bit tricky to get right for
both Linux and Windows. Probably not too bad to make it explicit like
this.

So the question is: what were the issues with `colorama.init()` approach? Unfortunately I don't have windows to test. On Linux at least I didn't notice issues with current `--color=when` implementation.

### DIFF
--- a/src/west/app/main.py
+++ b/src/west/app/main.py
@@ -296,6 +296,10 @@ class WestApp:
                             version=f'West version: v{__version__}',
                             help='print the program version and exit')
 
+        parser.add_argument('--color', default='auto',
+                            choices=['always', 'auto', 'never'],
+                            help='select when colored output should be used')
+
         subparser_gen = parser.add_subparsers(metavar='<command>',
                                               dest='command')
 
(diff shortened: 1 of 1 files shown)


mniestroj:
@ulfalizer I would like to ask about `colorama.init()` approach when enabling/disabling colors. Could you comment on what type of issues did you encounter?

mbolivar-nordic:
> This is just a scratch of the implementation and I would like to fallback into `color.ui`

Yes, if you can please implement a fallback onto the config option value, I'd be happy to test and merge this.

I didn't do the colorama implementation so I don't know what the issues were on Windows; @ulfalizer is the right one to ask though he has since moved on from working on west actively.

I can run some windows tests if you want to tell me some commands I can lazily copy/paste and screenshot the output of :smile: 

Thanks for the patch!

PR #405, opened 2020-06-10
More typing

mbolivar-nordic:
Following https://github.com/zephyrproject-rtos/west/commit/26b69f987e2715a69e789c214d40d51eee277a7b, let's add type annotations for manifest.py, which I think is west's most important module.

Fix a few issues caught by mypy before doing so. Also do some clean-ups and refactors before adding type annotations to make the results more pleasant.

### DIFF
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -508,7 +508,7 @@ class Project:
 
     @property
     def posixpath(self):
-        if self._posixpath is None and self.topdir:
+        if self._posixpath is None and self.abspath is not None:
             self._posixpath = PurePath(self.abspath).as_posix()
         return self._posixpath
 
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -784,30 +784,11 @@ class ManifestProject(Project):
         self.west_commands = _west_commands_list(west_commands)
 
     @property
-    def path(self):
-        return self._path
-
-    @path.setter
-    def path(self, path):
-        self._path = path
-
-        # Invalidate the absolute path attributes. They'll get
-        # computed again next time they're accessed.
-        self._abspath = None
-        self._posixpath = None
-
-    @property
     def abspath(self):
         if self._abspath is None and self.topdir and self.path:
             self._abspath = os.path.realpath(os.path.join(self.topdir,
                                                           self.path))
         return self._abspath
-
-    @property
-    def posixpath(self):
-        if self._posixpath is None and self.abspath:
-            self._posixpath = PurePath(self.abspath).as_posix()
-        return self._posixpath
 
     def as_dict(self):
         '''Return a representation of this object as a dict, as it would be
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -387,7 +387,7 @@ class ManifestImportFailed(Exception):
     - ``filename``: the missing file
     '''
 
-    def __init__(self, project, filename):
+    def __init__(self, project: 'Project', filename: str):
         self.project = project
         self.filename = filename
 
@@ -400,7 +400,7 @@ class ManifestVersionError(Exception):
     current version.
     '''
 
-    def __init__(self, version, file=None):
+    def __init__(self, version: str, file: str = None):
         self.version = version
         '''The minimum version of west that was required.'''
 
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -16,7 +16,7 @@ import os
 from pathlib import PurePath, PurePosixPath, Path
 import shlex
 import subprocess
-from typing import Any, Callable, List, Optional, Union
+from typing import Any, Callable, Dict, Iterable, List, Optional, Union
 
 from packaging.version import parse as parse_version
 import pykwalify.core
@@ -487,9 +487,13 @@ class Project:
         path_repr = repr(self.abspath or self.path)
         return f'<Project {self.name} ({path_repr}) at {self.revision}>'
 
-    def __init__(self, name, url, revision=None, path=None,
-                 clone_depth=None, west_commands=None, topdir=None,
-                 remote_name=None):
+    def __init__(self, name: str, url: str,
+                 revision: Optional[str] = None,
+                 path: Optional[str] = None,
+                 clone_depth: Optional[int] = None,
+                 west_commands: Optional[WestCommandsType] = None,
+                 topdir: Optional[str] = None,
+                 remote_name: Optional[str] = None):
         '''Project constructor.
 
         If *topdir* is ``None``, then absolute path attributes
@@ -518,40 +522,40 @@ class Project:
         self.remote_name = remote_name or 'origin'
 
     @property
-    def path(self):
+    def path(self) -> str:
         return self._path
 
     @path.setter
-    def path(self, path):
-        self._path = str(path)
+    def path(self, path: PathType) -> None:
+        self._path: str = str(path)
 
         # Invalidate the absolute path attributes. They'll get
         # computed again next time they're accessed.
-        self._abspath = None
-        self._posixpath = None
+        self._abspath: Optional[str] = None
+        self._posixpath: Optional[str] = None
 
     @property
-    def abspath(self):
+    def abspath(self) -> Optional[str]:
         if self._abspath is None and self.topdir:
             self._abspath = os.path.realpath(os.path.join(self.topdir,
                                                           self.path))
         return self._abspath
 
     @property
-    def posixpath(self):
+    def posixpath(self) -> Optional[str]:
         if self._posixpath is None and self.abspath is not None:
             self._posixpath = PurePath(self.abspath).as_posix()
         return self._posixpath
 
     @property
-    def name_and_path(self):
+    def name_and_path(self) -> str:
         return f'{self.name} ({self.path})'
 
-    def as_dict(self):
+    def as_dict(self) -> Dict:
         '''Return a representation of this object as a dict, as it
         would be parsed from an equivalent YAML manifest.
         '''
-        ret = {}
+        ret: Dict = {}
         ret['name'] = self.name
         ret['url'] = self.url
         ret['revision'] = self.revision
@@ -569,11 +573,13 @@ class Project:
     # Git helpers
     #
 
-    def git(self, cmd, extra_args=(), capture_stdout=False,
-            capture_stderr=False, check=True, cwd=None):
+    def git(self, cmd: Union[str, List[str]],
+            extra_args: Iterable[str] = (),
+            capture_stdout: bool = False,
+            capture_stderr: bool = False,
+            check: bool = True,
+            cwd: Optional[PathType] = None) -> subprocess.CompletedProcess:
         '''Run a git command in the project repository.
-
-        Returns a ``subprocess.CompletedProcess``.
 
         :param cmd: git command as a string (or list of strings)
         :param extra_args: sequence of additional arguments to pass to
@@ -626,7 +632,7 @@ class Project:
             return subprocess.CompletedProcess(popen.args, popen.returncode,
                                                stdout, stderr)
 
-    def sha(self, rev, cwd=None):
+    def sha(self, rev: str, cwd: Optional[PathType] = None) -> str:
         '''Get the SHA for a project revision.
 
         :param rev: git revision (HEAD, v2.0.0, etc.) as a string
@@ -642,7 +648,8 @@ class Project:
         # It'll be fun when we find out that was wrong and how...
         return cp.stdout.decode('ascii').strip()
 
-    def is_ancestor_of(self, rev1, rev2, cwd=None):
+    def is_ancestor_of(self, rev1: str, rev2: str,
+                       cwd: Optional[str] = None) -> bool:
         '''Check if 'rev1' is an ancestor of 'rev2' in this project.
 
         Returns True if rev1 is an ancestor commit of rev2 in the
@@ -666,7 +673,7 @@ class Project:
         else:
             raise RuntimeError(f'unexpected git merge-base result {rc}')
 
-    def is_up_to_date_with(self, rev, cwd=None):
+    def is_up_to_date_with(self, rev: str, cwd: Optional[str] = None) -> bool:
         '''Check if the project is up to date with *rev*, returning
         ``True`` if so.
 
@@ -680,7 +687,7 @@ class Project:
         '''
         return self.is_ancestor_of(rev, 'HEAD', cwd=cwd)
 
-    def is_up_to_date(self, cwd=None):
+    def is_up_to_date(self, cwd: Optional[str] = None) -> bool:
         '''Check if the project HEAD is up to date with the manifest.
 
         This is equivalent to ``is_up_to_date_with(self.revision,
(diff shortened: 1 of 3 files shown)


mbolivar-nordic:
@gmarull could you please take a look also, if you have time?

src/west/manifest.py
```diff
@@ -48,9 +54,229 @@
...
+    try:
+        return yaml.safe_load(data)
+    except yaml.scanner.ScannerError as e:
+        raise MalformedManifest(data) from e
+
+def _west_commands_list(west_commands: Optional[Union[str, List[str]]]) -> \
+        List[str]:
+    # Convert the raw data from a manifest file to a list of
+    # west_commands locations. (If it's already a list, make a
+    # defensive copy.)
```
gmarull, line 92: Would be nice to use same public docstring format for internal functions.

src/west/manifest.py
```diff
@@ -48,9 +54,229 @@
...
+    else:
+        return west_commands
+
+def _west_commands_merge(wc1: List[str], wc2: List[str]) -> List[str]:
+    # Merge two west_commands lists, filtering out duplicates.
+
+    if wc1 and wc2:
+        return wc1 + [wc for wc in wc2 if wc not in wc1]
+    else:
+        return wc1 or wc2
```
gmarull, line 117: This function may be reduced to:

```python
return wc1 + list(set(wc2) - set(wc1)))
```

src/west/manifest.py
```diff
@@ -48,9 +54,229 @@
...
+
+def _mpath(cp: Optional[configparser.ConfigParser] = None,
+           topdir: Optional[str] = None):
```
gmarull, line 120: Personal suggestion: `black` autoformatter would provide nicer format:

```python
def _mpath(
    cp: Optional[configparser.ConfigParser] = None,
    topdir: Optional[str] = None
) -> None:
```

src/west/manifest.py
```diff
@@ -48,9 +54,229 @@
...
+    # Return the value of the manifest.path configuration option
```
gmarull, line 121: Comment not in line with function (does not return)

src/west/manifest.py
```diff
@@ -48,9 +54,229 @@
...
+    except (configparser.NoOptionError, configparser.NoSectionError) as e:
+        raise MalformedConfig('no "manifest.path" config option is set') from e
+
+# Manifest import handling
+
+def _default_importer(project: 'Project', file: str):
+    raise ManifestImportFailed(project, file)
+
+def _manifest_content_at(project: 'Project', path: str,
+                         rev: str = QUAL_MANIFEST_REV_BRANCH):
```
gmarull, line 144: Missing return type

src/west/manifest.py
```diff
@@ -48,9 +54,229 @@
...
+_import_map = collections.namedtuple('_import_map',
+                                     'file '
+                                     'name_whitelist path_whitelist '
+                                     'name_blacklist path_blacklist')
+
+# Type for an import map filter function, which takes a Project and
+# returns a bool. The various whitelists and blacklists are used to
+# create these filter functions. A None value is treated as a function
+# which always returns True.
+_ImapFilterFnType = Optional[Callable[['Project'], bool]]
```
gmarull, line 200: Would probably be nice to group custom types to the top or in a module.

src/west/manifest.py
```diff
@@ -48,9 +54,229 @@
...
+
+def _is_imap_list(value: Any) -> bool:
```
gmarull, line 202: `Any` too generic?

src/west/manifest.py
```diff
@@ -48,9 +54,229 @@
...
+    # Returns either None (if no filter is necessary) or a
+    # filter function for the given import map.
+
+    if any([imap.name_whitelist, imap.path_whitelist,
+            imap.name_blacklist, imap.path_blacklist]):
+        return lambda project: _is_imap_ok(imap, project)
+    else:
+        return None
+
+def _ensure_list(item: Union[str, List[str]]) -> List:
```
gmarull, line 220: `-> List[str]`

src/west/manifest.py
```diff
@@ -71,7 +297,7 @@ def manifest_path():
         raise OSError(errno.ENOENT, os.strerror(errno.ENOENT), ret)
     return ret
 
-def validate(data):
+def validate(data: Any) -> None:
     '''Validate manifest data
 
     Returns if the manifest data is valid and can be loaded by this
```
gmarull, line 303: Does not return, so comment should be in line with function.

src/west/manifest.py
```diff
@@ -111,17 +337,68 @@ def validate(data):
...
+
+    Attributes:
+
+    - ``project``: the Project instance with the missing manifest data
+    - ``filename``: the missing file
+    '''
+
+    def __init__(self, project: 'Project', filename: str):
+        self.project = project
+        self.filename = filename
```
gmarull, line 376: Better use `@property`

src/west/manifest.py
```diff
@@ -150,1504 +427,1302 @@ def _flags_ok(flags):
...
+        return f'<Project {self.name} ({path_repr}) at {self.revision}>'
 
-            - If only *source_file* is given, *topdir* is found
-              starting there. The directory containing *source_file*
-              doesn't have to be ``manifest.path`` in this case.
+    def __init__(self, name: str, url: str,
+                 revision: Optional[str] = None,
+                 path: Optional[str] = None,
+                 clone_depth: Optional[int] = None,
+                 west_commands: Optional[Union[str, List[str]]] = None,
```
gmarull, line 478: As `Optional[Union[str, List[str]]]` is used in quite a few places a custom type may be useful.

src/west/manifest.py
```diff
@@ -150,1504 +427,1302 @@ def _flags_ok(flags):
...
+            ret['path'] = self.path
+        if self.west_commands:
+            ret['west-commands'] = \
+                _west_commands_maybe_delist(self.west_commands)
+        return ret
 
-        ``ValueError`` is raised if:
+ManifestDataType = Union[str, Dict]
+ImportedContent = Optional[Union[str, List[str]]]
+ImporterType = Callable[[Project, str], ImportedContent]
```
gmarull, line 828: Would be nice to group custom types.

src/west/manifest.py
```diff
@@ -150,1504 +427,1302 @@ def _flags_ok(flags):
...
-    def __str__(self):
-        path_repr = repr(self.abspath or self.path)
-        return f'<Project {self.name} ({path_repr}) at {self.revision}>'
+        _logger.debug(f'loaded {loading_what}')
 
-    def __init__(self, name, url, revision=None, path=None,
-                 clone_depth=None, west_commands=None, topdir=None,
-                 remote_name=None):
-        '''Project constructor.
+    def _load_self(self, manifest: Dict[str, Any], path_hint: Optional[str],
```
gmarull, line 1316: Regarding lists or dictionaries you may also explore using https://docs.python.org/3/library/typing.html#generics (i.e. `Iterable, Mapping...`.

src/west/manifest.py
mbolivar-nordic: Hmm, are you sure the set() calls and set difference operation preserve order in the same way the above does?

src/west/manifest.py
mbolivar-nordic: We actually don't do this on purpose, so that these comments don't show up in help() calls in the REPL.

src/west/manifest.py
mbolivar-nordic: Is that a common convention? I was trying to keep the type definitions close to their first point of use. Happy to go either way.

src/west/manifest.py
mbolivar-nordic: I think it's the best we can do. The `value` argument comes from untrusted user input in a parsed YAML file, so I think Any is appropriate. Am I missing something?

src/west/manifest.py
mbolivar-nordic: Sorry, I don't see the problem here. The comment is saying that the function returns if the manifest data is valid, which it does. It only raises an exception if the manifest data are invalid. Is there a bug in the docstring?

src/west/manifest.py
mbolivar-nordic: I tried that, but flake8 doesn't like it. Given that, I hope we can keep black vs flake8 out of scope of this PR if it's OK with you.

src/west/manifest.py
gmarull: If function returns something `-> None` is inconsistent.

src/west/manifest.py
gmarull: Yes, just a suggestion. `flake8` with `--max-line-length=88` will be happy 99% of the times with `black` output.

src/west/manifest.py
gmarull: Was not aware that order is relevant, should be tested.

src/west/manifest.py
gmarull: got it, said that because `isinstance` is used to check types `str` or `List[str]`

src/west/manifest.py
mbolivar-nordic: Could you provide some rationale why that would be better? These are just simple attributes.

gmarull: `properties` have explicit get (r) or get/set (rw). I tend to encapsulate class data as properties, it's ok as it is though :-)

src/west/manifest.py
gmarull: I understand the decision, fair enough. Afaik there's no rule, I've even seen some codebases to have a module with shared types, e.g. https://github.com/marshmallow-code/marshmallow/blob/dev/src/marshmallow/types.py

src/west/manifest.py
mbolivar-nordic: The annotations applied to internal data fields are the concrete types returned by the yaml module's load function. I looked for public APIs involving sequence types, and came up with this:

- Manifest.get_projects() and Manifest.projects() return List[Project], but the "list" here is part of the documented API; these don't return a tuple or any other iterable or sequence, but a real list
- Manifest.as_dict() and Manifest.as_frozen_dict() return Dict, which is similarly part of the contract
- Similar comments apply to Project.listdir_at(): the return value should be a list
- Project.git()'s extra_args kwarg is an Iterable already, as is Manifest.get_projects() "project_ids" kwarg

So I think argument types are already permissive in terms of what they accept, but deliberately conservative in terms of what they return. That matches my intent for their behavior. Hopefully that's what you meant.

mbolivar-nordic:
@gmarull I think I've addressed all your comments in this latest force push. Thanks very much for the detailed review.

src/west/manifest.py
mbolivar-nordic: It either returns control flow back to the caller or raises an exception. If it returns, it returns None. So I think this is correct. No?

src/west/manifest.py
mbolivar-nordic: OK, let's keep this as-is for now and see if @carlescufi or @tejlmand have a preference then. Thanks for the pointer!

src/west/manifest.py
gmarull: The text says that returns if data is valid (suggesting a boolean return). In this case invalid means that exception is raised (so nothing is returned), so maybe `Raises exception if the manifest....` would be more explanatory.

mbolivar-nordic: I see. The docstring is unclear. Here, "returns" is used intransitively. It doesn't return any value, it simply returns (at all). Thanks, will try to fix.

mbolivar-nordic:
@carlescufi @tejlmand please review

mbolivar-nordic:
Thanks @carlescufi. Hopefully you or @tejlmand will be willing to give a +1 so I don't have to use admin privileges to merge it; I'd really rather avoid setting that precedent.

carlescufi:
> Thanks @carlescufi. Hopefully you or @tejlmand will be willing to give a +1 so I don't have to use admin privileges to merge it; I'd really rather avoid setting that precedent.

Of course. If @tejlmand doesn't review in the next couple of day I'll give you the +1 after looking a bit more through it. I just don't feel comfortable giving it myself right now with my limited review.

src/west/manifest.py
```diff
@@ -303,7 +303,7 @@ def validate(data):
         pykwalify.core.Core(source_data=data,
                             schema_files=[_SCHEMA_PATH]).validate()
     except pykwalify.errors.SchemaError as se:
-        raise MalformedManifest(se._msg) from se
+        raise MalformedManifest(se.msg) from se
```
tejlmand, line 353: Did you intent to mean:
```suggestion
        raise MalformedManifest(se.msg()) from se
```

src/west/manifest.py
tejlmand: no strong opinions.
I can live with whatever decision @mbolivar-nordic prefers.

but I do feel the code is easier to read, when custom type definitions are group together, simply because the indentation is more `clean`/uniform in the method definition part this way.

src/west/manifest.py
mbolivar-nordic: At least on my system's pykwalify, msg is a property and therefore should be accessed like an attribute, so this wouldn't work.

Do you have a specific pykwalify version where this is required?

mbolivar-nordic:
Given both reviewers prefer custom type definitions in one place, I've rearranged them that way and cleaned them up a bit.

src/west/manifest.py
tejlmand: The reason I ask, is because I see both a method:
https://github.com/Grokzen/pykwalify/blob/master/pykwalify/errors.py#L94

and a property:
https://github.com/Grokzen/pykwalify/blob/master/pykwalify/errors.py#L55

but in your commit message you write:
> The SchemaError type's error message should be accessed via the msg()
property,

so bacause you used `msg()` I simply got unsure whether you meant the property or the method `msg()`.
that was the only reason for asking.

tejlmand: Thanks for clarifying.
Approved.

mbolivar-nordic: > The reason I ask, is because I see both a method:
> https://github.com/Grokzen/pykwalify/blob/master/pykwalify/errors.py#L94
> 
> and a property:
> https://github.com/Grokzen/pykwalify/blob/master/pykwalify/errors.py#L55

The code that starts on line 94 is written in a tricky style that's IMO a bit hard to read, but it's a property, not a method. I'm going to copy/paste the entire thing here:

```
def msg():
    doc = """ """

    def fget(self):
        return self._msg

    def fset(self, value):
        assert isinstance(value, basestring), "argument is not string"
        self._msg = value

    return locals()
msg = property(**msg())
```

Notice the final `msg = property(**msg())`. The first definition of `msg`, using `def`, is overwritten immediately by this statement.

So the final value of `msg` is the return value of `property()`, whose arguments are taken by passing the return value of `msg()` as kwargs.

In turn, the `msg()` call returns the dictionary `locals()`, which looks like this:

```
>>> msg()
{'doc': ' ',
 'fget': <function __main__.msg.<locals>.fget(self)>,
 'fset': <function __main__.msg.<locals>.fset(self, value)>}
```

So it's as if you defined

```
msg = property(doc=' ', fget=<some function>, fset=<another function>)
```

and its final value is a "real" property with a getter and a setter function. That's why errors.py on line 55 can treat it like a property; it is one.

tejlmand: I know.
The thing I tried to say, was that because your commit message said `msg()` but you used the property `msg` I simply was unsure what your really intented to do.

mbolivar-nordic: Ah, OK, I understand your meaning now, thanks.

PR #409, opened 2020-06-29
.gitignore: add htmlcov/

mbolivar-nordic:
This includes coverage data in human-readable format. It's generated
by tox.


.gitignore
```diff
@@ -8,4 +8,5 @@ shippable/
 .tox/
 .coverage*
 .mypy_cache/
-
+# htmlcov is generated by tox due to coverage gathering in pytest
+htmlcov/
```
tejlmand, line 12: as this is generated as part of the build, could you explain the reason for a dedicated folder in the root of the repo, instead of having this content in the build folder ?

mbolivar-nordic: This is just the default directory created by the coverage test engine.

mbolivar-nordic:
>  we could consider to configure tox to add this folder under the build, as part of the complete west build at a later time.

The build directory is really the responsibility of setuptools etc. to manage; "building" west is different from testing it (building is done with setup.py, testing with tox).

Tox does build west, but it puts its artifacts in a separate virtual environment directory under .tox. Pytest, which is run by tox, also has some files that are ignored, as you can see in the rest of the file.

The htmlcov directory is created by pytests's coverage plugin, so it's not really done as part of a west "build". The location of the output directory is configurable but I am just imitating other projects, like typeshed, by ignoring the default location:

https://github.com/python/typeshed/blob/01ce742298ea6cd65ffe13795a429083c2e526dc/.gitignore#L35

tejlmand:
guess we need some CMake files to build, test, and deploy `west` in a uniform way 😈

mbolivar-nordic:
> guess we need some CMake files to build, test, and deploy `west` in a uniform way smiling_imp

Thanks but no thanks :stuck_out_tongue:

Issue #367, opened 2020-02-03
Add 'path-prefix' key in west import map


mbolivar:
Example syntax:

```
manifest:
  projects:
  - name: zephyr
    url: https://github.com/zephyrproject-rtos/zephyr
    import:
      directory: zephyr_projects
```

Semantics:

- the `zephyr` project, and all imported projects, have `zephyr_projects` as an additional path component. e.g. `zephyr_projects/zephyr` instead of `zephyr` as the path for `zephyr` project
- the `directory` value cannot escape the workspace, e.g. `directory: ../..` is an error
- the `directory` value is cumulative, e.g. if the imported manifest in turn does an import with `directory: foo`, then the combined `directory` is `zephyr_projects/foo`

Alternate suggested spellings for `directory`:

- `path`
- `path-prefix`

mbolivar:
cc @henrikbrixandersen @tejlmand

henrikbrixandersen:
LGTM. Personally I prefer `directory`.

DenvE:
Directory is fine fo me too

marc-cpdesign:
I'd love to see this implemented also !

I'll take whatever keyword I get for 'directory', but also consider some alternatives:
- folder
- destination
- destination-prefix
- base-prefix
- base-destination

(Having 'prefix' or 'base' in there appeals to me because the _actual_ destination is the path+name)

marc-hb:
> (Having 'prefix' or 'base' in there appeals to me because the actual destination is the path+name)

Agreed, "directory" or "path" are super generic terms, way too vague when left alone.

Inspired by "destination" and trying to describe the action rather than the elusive nature, how about `into:`?

```
manifest:
  projects:
  - name: zephyr
    url: https://github.com/zephyrproject-rtos/zephyr
    import:
      into: all_imports/zephyr_clones/
```
By the way will this new prefix support more than one multiple directory level?

mbolivar:
> By the way will this new prefix support more than one multiple directory level?

It should IMO.

marc-cpdesign:
+1 for "into"
+1 for recursive/ multi-level import support

Let me know if I can help somehow.

euripedesrocha:
I would like to have something like this. But I would prefer a syntax that adds a prefix to the defaults section of the manifest:
```yaml
defaults:
    path_prefix: prefix_path
```
My goal at this moment is to have only the projects I´m using by setting import as false.

tejlmand:
[...]

Issue #327, opened 2019-10-21
apis: west.manifest.Manifest.projects should not include the manifest itself


mbolivar:
Trying to treat the manifest repository as an ordinary Project was doomed to failure; they're simply too different. We should instead move the valid state that's in ManifestProject today to Manifest attributes, removing MANIFEST_PROJECT_INDEX and the fact that projects[MANIFEST_PROJECT_INDEX] contains the manifest.

This will require a bit of thought into how `west list` should behave, though. Maybe we need to resurrect `west list --all` to include manifest project data like `{path}` and `{west_commands}`? And perhaps that should abort if you `west list --all -f {url}` or try to access some other attribute that's not defined for the manifest repository.

tejlmand:
> Trying to treat the manifest repository as an ordinary Project was doomed to failure

what kind of issues are observed with the current behavior.
Could you try to list the issues you experiences when the manifest repo is treated as an ordinary project.

Cause I remember there were long discussion back and forth on how to treat the manifest project.

Note, I'm not advocating against changing it, just would like to see a list of issues experienced so that everyone looking at this issue know why to change API.

mbolivar:
Welcome back :).

The manifest repository is a *repository* for sure, and its Python representation should have some Project attributes (like west_commands, etc.), but it's not truly a Project:

- it doesn't have a `revision`
- or a `url`
- or, in some cases (when parsing a manifest from data without `self: path: foo`), a `path`

So by "not truly a Project", I mean this in a Liskov substitution principle sense.

Right now we hack it by setting these fields to None, but I've noticed again and again we've messed up updating the docstrings or code in ManifestProject following changes to Project, leading to errors, which is also a bad sign.

From personal experience, I also find myself writing downstream code (nrf/scripts/ncs_west_commands/ncs_commands.py) that wants to iterate over projects and ending up special-casing the manifest repository as a result of crashes on code that naively looped over them. The manifest repository doesn't really have a 'name' in the project sense, and trying to treat it as if it were a project has led to weird edge cases.

I think it was a worthy attempt and what we did made sense at the time, but experience has shown it was not a good idea in retrospect.

mbolivar:
>  leading to errors, which is also a bad sign.

For example, `west forall -h` (still* says that the projects list defaults to "all projects in the manifest", which is wrong: it's actually all projects in the manifest.... plus the manifest repository, because that's not filtered out.

tejlmand:
> Welcome back :).

[...]

Issue #243, opened 2019-03-28
Non-git "projects" support


mbolivar:
Though west is currently tied to git for projects and the manifest repository, there's no reason it has to stay that way forever. Some ideas for generalization:

- support projects and manifests managed by other version control systems (e.g. hg, p4)
- support projects which are just tarballs (.tar.gz, .tar.xz, .tar.bz2, .zip, .lzma, etc. -- see [dtrx](https://github.com/moonpyk/dtrx) for inspiration)
- support artifacts which are binary objects (e.g. firmware blobs needed by downstream distributions)

mbolivar-nordic:
After getting a couple of years of experience here, I don't think this is worth doing. West is a git wrapper, pretty much by design. Some Zephyr users complained when it was introduced that it was too git specific, but Zephyr uses git, for better or worse. I think it is incumbent upon those users to deal with this as they see fit.

Note that converting between the west.yml format and other formats is possible, I dare say easy, using the west API. I posted an example for this a while back in the list:

https://lists.zephyrproject.org/g/devel/message/6437

PR #411, opened 2020-07-01
Overhaul path handling

mbolivar-nordic:
Fixes: #273 

This makes west's internal representations for paths use pathlib objects whenever relevant or convenient. In particular, this is useful whenever path comparisons are necessary, since pathlib handles the canonicalization properly.

It also makes west accept os.PathLike in its public APIs in addition to str. This is consistent with the general direction of the python standard library and just generally reasonable practice for python in 2020.

I'm leaning pretty heavily on the type checker to catch my mistakes here. It definitely helped during development.

I tried to make this as symlink friendly as I could, but I can't say with 100% honesty that running west commands from `/some/symlink/to/a/workspace` doesn't ever leak the physical path behind the symlink.

Now that we are dealing with paths in a nicer way, we can move on to https://github.com/zephyrproject-rtos/west/issues/367 after this.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -261,7 +261,7 @@ class Init(_ProjectCommand):
         if not exists(temp_manifest):
             log.die(f'can\'t init: no "west.yml" found in {tempdir}\n'
                     f'  Hint: check --manifest-url={manifest_url} and '
-                    '--manifest-rev={manifest_rev}\n'
+                    f'--manifest-rev={manifest_rev}\n'
                     f'  You may need to remove {west_dir} before retrying.')
 
         # Parse the manifest to get the manifest path, if it declares one.
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -9,8 +9,8 @@ import argparse
 from functools import partial, lru_cache
 import logging
 import os
-from os.path import join, relpath, basename, dirname, exists, isdir
-from pathlib import PurePath
+from os.path import abspath, relpath, exists
+from pathlib import PurePath, Path
 import shutil
 import shlex
 import subprocess
@@ -207,17 +207,17 @@ class Init(_ProjectCommand):
     def die_already(self, where, also=None):
         log.die(f'already initialized in {where}, aborting.{also or ""}')
 
-    def local(self, args):
+    def local(self, args) -> Path:
         if args.manifest_rev is not None:
             log.die('--mr cannot be used with -l')
 
-        manifest_dir = util.canon_path(args.directory or os.getcwd())
-        manifest_file = join(manifest_dir, 'west.yml')
-        topdir = dirname(manifest_dir)
-        rel_manifest = basename(manifest_dir)
-        west_dir = os.path.join(topdir, WEST_DIR)
+        manifest_dir = Path(args.directory or os.getcwd())
+        manifest_file = manifest_dir / 'west.yml'
+        topdir = manifest_dir.parent
+        rel_manifest = manifest_dir.name
+        west_dir = topdir / WEST_DIR
 
-        if not exists(manifest_file):
+        if not manifest_file.is_file():
             log.die(f'can\'t init: no "west.yml" found in {manifest_dir}')
 
         log.banner('Initializing from existing manifest repository',
@@ -225,15 +225,15 @@ class Init(_ProjectCommand):
         log.small_banner(f'Creating {west_dir} and local configuration file')
         self.create(west_dir)
         os.chdir(topdir)
-        update_config('manifest', 'path', rel_manifest)
+        update_config('manifest', 'path', os.fspath(rel_manifest))
 
         return topdir
 
-    def bootstrap(self, args):
+    def bootstrap(self, args) -> Path:
         manifest_url = args.manifest_url or MANIFEST_URL_DEFAULT
         manifest_rev = args.manifest_rev or MANIFEST_REV_DEFAULT
-        topdir = util.canon_path(args.directory or os.getcwd())
-        west_dir = join(topdir, WEST_DIR)
+        topdir = Path(abspath(args.directory or os.getcwd()))
+        west_dir = topdir / WEST_DIR
 
         try:
             already = util.west_topdir(topdir, fall_back=False)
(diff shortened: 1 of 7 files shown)


mbolivar-nordic:
@tejlmand I hope you will now agree by the number of places where we have to either jump through hoops or play tricks on the type checker to make ManifestProject work that https://github.com/zephyrproject-rtos/west/issues/327 is a real issue, and should be fixed.

tejlmand:
>  manifest: use a crystal ball to aid the next commit 

Can I have a copy ?

mbolivar-nordic:
> Can I have a copy ?

Sure, I'll give you mine in exchange for a review ;)

tejlmand:
> @tejlmand I hope you will now agree by the number of places where we have to either jump through hoops or play tricks on the type checker to make ManifestProject work that #327 is a real issue, and should be fixed.

Yep, in the context of VCS projects.
Which means we need to reserve projects to VCS-projects, as was also mentioned in that link.
Always remember that someday we might want to extend to handle non-VCS projects.

> >    Take a look at: #243
>
> Good point. In this case I think as long as the non-git projects show up in manifest: projects: in the manifest file we should be able to keep the model consistent.
>
> In fact, making the change here (#327) may lead us to add another section besides projects: for fetching some non-git "things", e.g. large files, if they aren't cleanly "repositories of some kind with a name, URL, revision, path, (maybe) clone depth, and optional extension commands".

src/west/configuration.py
```diff
@@ -37,13 +37,13 @@
 
 import configparser
 import os
-import pathlib
+from pathlib import PureWindowsPath, Path
```
tejlmand, line 40: Comment to whole commit, not this line in particular.
So nice that Python 3.6 is minimum required.

src/west/app/project.py
```diff
@@ -261,7 +261,7 @@ def bootstrap(self, args):
         if not exists(temp_manifest):
             log.die(f'can\'t init: no "west.yml" found in {tempdir}\n'
                     f'  Hint: check --manifest-url={manifest_url} and '
-                    '--manifest-rev={manifest_rev}\n'
+                    f'--manifest-rev={manifest_rev}\n'
```
tejlmand, line 266: No worries, no users ever makes mistakes with `west init` ;)

src/west/app/project.py
```diff
@@ -202,63 +202,65 @@ def do_run(self, args, ignored):
         else:
             topdir = self.bootstrap(args)
 
-        log.banner(f'Initialized. Now run "west update" inside {topdir}.')
+        log.banner('Initialized. Now run "west update" inside',
+                   topdir)
```
tejlmand, line 206: just curios, here you change into two args, but in line 245, you do the opposite:
```
log.banner(f'Initializing in {topdir}')
```
any particular reason ?

src/west/configuration.py
mbolivar-nordic: Amen, f-strings and os.PathLike are both great. Also many thanks to @gmarull for contributing a type checking framework that helps convince me we're not doing anything egregiously wrong in this PR.

src/west/app/project.py
mbolivar-nordic: It was when I started changing this file that I began learning more about the differences between `f'{foo}'`, `str(foo)`, `repr(foo)`, and `os.fspath(foo)` when `foo` is either a general-purpose `os.PathLike` or specifically a `pathlib.Path`. I didn't know `os.fspath` existed when I started this work and so I was relying on behavior that is specific to `pathlib.Path` in some cases, which I later learned was incorrect. Live and learn... and then you read the docs.

Any inconsistencies here are due to me not cleaning up after learning more. I will double check and make sure everything is OK. Thanks for asking.

mbolivar-nordic:
> Which means we need to reserve projects to VCS-projects, as was also mentioned in that link.
> Always remember that someday we might want to extend to handle non-VCS projects.

Yep, I agree with this. We can handle non-VCS "projects" (really non-git repository artifacts) with a separate list in the manifest file once we have a real use case.

mbolivar-nordic:
>> Can I have a copy ?
>
> Sure, I'll give you mine in exchange for a review ;)

@tejlmand as promised: :crystal_ball: use in peace and good health :smile: 

Thanks for review

src/west/app/project.py
mbolivar-nordic: >  I will double check and make sure everything is OK.

I didn't notice any bugs being introduced, but I went back and removed the useless changes to print / log lines to not clutter the diff with unrelated changes.

For reference:

- `f'{foo}'` uses the `__format__` method from `foo`
- pathlib objects do not override `__format__`, so `f'{pathobj}'` ends up calling `object.__format__` on `pathobj` with an empty `format_spec`
- [`object.__format__`](https://docs.python.org/3/reference/datamodel.html#object.__format__) with an empty `format_spec`  ends up calling `str()` on the object being formatted (with some differences between python 3.6 and later versions that don't matter to us)
- for pathlib objects, `os.fspath(pathobj)` and `str(pathobj)` are the same thing, because `__fspath__` delegates to `str()`
- `west.log` ends up calling `print()` on its arguments, which will print them out just like `str()` would format them

So TL;DR `west.log.inf(f'blah blah f{pathobj}')` and `west.log.inf('blah blah ', pathobj)` produce the same results, because `__format__` for these objects works like `str()`, which works like `os.fspath()`, so there's no differences in printing them with f-strings or as separate args, they always produce the right filesystem representation as a string, and we might as well not introduce pointless churn.

Where this *does* matter is in places like west.manifest, which accept `os.PathLike`. In the general case, `str(some_os_PathLike)` and `os.fspath(some_os_PathLike)` are *not* the same, because the first uses `__str__` and the second uses `__fspath__`, which don't need to be the same (though they are for pathlib).

That's why I tried to use os.fspath in places in project.py where we were not obviously just printing things out.

tejlmand: > Any inconsistencies here are due to me not cleaning up after learning more.

I think you missed my point.
When looking at the diff: `index b0ed0d5..5c68170 100644`
then I noticed:
```
-        log.banner(f'Initialized. Now run "west update" inside {topdir}.')
+        log.banner('Initialized. Now run "west update" inside',
+                   topdir)
```
and then later:
```
-        log.banner('Initializing in', topdir)
+        log.banner(f'Initializing in {topdir}')
```
INSIDE the same commit

No bugs, just curios why one place was changed from `(f'text {topdir}')`--> `('text', topdir)`, and at the second location you went from: `('text', topdir)` --> `(f'text {topdir}')`

tejlmand:
Thanks, will hand the 🔮 back when i'm done with it.
Those are always handy to have.

src/west/app/project.py
mbolivar-nordic: > I think you missed my point.

I don't think I did. I went back and made those consistent by dropping the unrelated changes to format strings... did I miss one? Anyway, not important.

PR #412, opened 2020-07-01
remove .shippable.yml

mbolivar-nordic:
Experience has borne out the hope that the CI system using GitHub
actions offers the same amount of testing on Linux platforms as
existing shippable infrastructure. It's also configured to run on
macOS and Windows, which is nicer. So there's no need to continue to
run shippable based testing now.


mbolivar-nordic:
@nashif not sure if there's anything else I need to do to make shippable stop running on the west repository now, but I'm ready when you are. The GH actions that @galak added work very well for this project.

mbolivar-nordic:
> @nashif not sure if there's anything else I need to do to make shippable stop running on the west repository now,

Can you please help disable shippable? It's showing status FAILED with this PR. I was hoping removing the file would remove the integration but that doesn't seem to be the case. I never turned it on in the first place so I don't know how to turn it off now.

Issue #249, opened 2019-04-17
Be more forgiving about the manifest file name


mbolivar:
- support `--mf <file_name>` in `west init` and add `manifest.file` to the config
- also fall back on `west.yaml` if `west.yml` is missing

mbolivar-nordic:
`--mf` is supported now; I'm closing. This means the resolution for 'fall back on west.yaml' is "won't do".

PR #414, opened 2020-07-10
add support for manifest filenames other than west.yml

M1cha:
partially solves #249

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -124,7 +124,7 @@ class Init(_ProjectCommand):
               1. Creates a .west directory and clones a manifest repository
                  from a git URL to a temporary subdirectory of .west,
                  .west/<tmpdir>.
-              2. Parses the manifest file, .west/<tmpdir>/west.yml.
+              2. Parses the manifest file, .west/<tmpdir>/<manifest.file>.
                  This file's contents can specify manifest.path, the location
                  of the manifest repository in the workspace, like so:
 
@@ -161,6 +161,8 @@ class Init(_ProjectCommand):
         parser.add_argument('--mr', '--manifest-rev', dest='manifest_rev',
                             help='''manifest revision to check out and use;
                             cannot be combined with -l''')
+        parser.add_argument('--mf', '--manifest-file', dest='manifest_file',
+                            help='manifest file name to use')
         parser.add_argument('-l', '--local', action='store_true',
                             help='''use an existing local manifest repository
                             instead of cloning one; cannot be combined with
@@ -212,13 +214,15 @@ class Init(_ProjectCommand):
             log.die('--mr cannot be used with -l')
 
         manifest_dir = Path(args.directory or os.getcwd())
-        manifest_file = manifest_dir / 'west.yml'
+        manifest_filename = args.manifest_file or 'west.yml'
+        manifest_file = manifest_dir / manifest_filename
         topdir = manifest_dir.parent
         rel_manifest = manifest_dir.name
         west_dir = topdir / WEST_DIR
 
         if not manifest_file.is_file():
-            log.die(f'can\'t init: no "west.yml" found in {manifest_dir}')
+            log.die(f'can\'t init: no {manifest_filename} found in '
+                    f'{manifest_dir}')
 
         log.banner('Initializing from existing manifest repository',
                    rel_manifest)
@@ -226,6 +230,7 @@ class Init(_ProjectCommand):
         self.create(west_dir)
         os.chdir(topdir)
         update_config('manifest', 'path', os.fspath(rel_manifest))
+        update_config('manifest', 'file', manifest_filename, topdir=topdir)
 
         return topdir
 
@@ -257,9 +262,11 @@ class Init(_ProjectCommand):
             raise
 
         # Verify the manifest file exists.
-        temp_manifest = tempdir / 'west.yml'
+        temp_manifest_filename = args.manifest_file or 'west.yml'
+        temp_manifest = tempdir / temp_manifest_filename
         if not temp_manifest.is_file():
-            log.die(f'can\'t init: no "west.yml" found in {tempdir}\n'
+            log.die(f'can\'t init: no {temp_manifest_filename} found in '
+                    f'{tempdir}\n'
                     f'  Hint: check --manifest-url={manifest_url} and '
                     f'--manifest-rev={manifest_rev}\n'
                     f'  You may need to remove {west_dir} before retrying.')
@@ -290,6 +297,8 @@ class Init(_ProjectCommand):
             log.die(e)
         log.small_banner('setting manifest.path to', manifest_path)
         update_config('manifest', 'path', manifest_path, topdir=topdir)
+        update_config('manifest', 'file', temp_manifest_filename,
+                      topdir=topdir)
 
         return topdir
 
@@ -640,15 +649,15 @@ class Update(_ProjectCommand):
     def __init__(self):
         super().__init__(
             'update',
-            'update projects described in west.yml',
+            'update projects described in west manifest',
             textwrap.dedent('''\
             Updates each project repository to the revision specified in
-            the manifest file, west.yml, as follows:
+            the manifest file, as follows:
 
               1. Fetch the project's remote to ensure the manifest
                  revision is available locally
               2. Reset the manifest-rev branch to the revision in
-                 west.yml
+                 the manifest
               3. Check out the new manifest-rev commit as a detached HEAD
                  (but see "checked out branches")
 
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -17,7 +17,7 @@ import shlex
 import subprocess
 import sys
 from typing import Any, Callable, Dict, Iterable, List, NoReturn, \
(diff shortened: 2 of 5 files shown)


mbolivar-nordic:
Thanks for this! I will review as soon as I can. I'm working on some other tasks this week.

src/west/app/project.py
```diff
@@ -161,6 +161,8 @@ def do_add_parser(self, parser_adder):
         parser.add_argument('--mr', '--manifest-rev', dest='manifest_rev',
                             help='''manifest revision to check out and use;
                             cannot be combined with -l''')
+        parser.add_argument('--mf', '--manifest-file', dest='manifest_file',
+                            help='''manifest file name to use''')
```
mbolivar-nordic, line 165: A minor point, but if you have a chance, there's no need to triple-quote a single line string, so this could just use single quotes.

src/west/app/project.py
```diff
@@ -212,20 +214,22 @@ def local(self, args) -> Path:
...
-        manifest_file = manifest_dir / 'west.yml'
+        manifest_filename = args.manifest_file or 'west.yml'
+        manifest_file = manifest_dir / manifest_filename
         topdir = manifest_dir.parent
         rel_manifest = manifest_dir.name
         west_dir = topdir / WEST_DIR
 
         if not manifest_file.is_file():
-            log.die(f'can\'t init: no "west.yml" found in {manifest_dir}')
+            log.die(f'can\'t init: no manifest found in {manifest_dir}')
```
mbolivar-nordic, line 224: I think it would be better to say exactly the file that's being searched for:

```suggestion
            log.die(f'can\'t init: no {manifest_filename} found in {manifest_dir}')
```

src/west/app/project.py
```diff
@@ -257,9 +261,10 @@ def bootstrap(self, args) -> Path:
             raise
 
         # Verify the manifest file exists.
-        temp_manifest = tempdir / 'west.yml'
+        temp_manifest_filename = args.manifest_file or 'west.yml'
+        temp_manifest = tempdir / temp_manifest_filename
         if not temp_manifest.is_file():
-            log.die(f'can\'t init: no "west.yml" found in {tempdir}\n'
+            log.die(f'can\'t init: no manifest found in {tempdir}\n'
```
mbolivar-nordic, line 267:
```suggestion
            log.die(f'can\'t init: no {temp_manifest_filename} found in {tempdir}\n'
```

src/west/manifest.py
```diff
@@ -301,10 +304,11 @@ def manifest_path() -> str:
         - `MalformedConfig` if the configuration file has no
           ``manifest.path`` key
 
-        - ``FileNotFoundError`` if no ``west.yml`` exists in
+        - ``FileNotFoundError`` if no manifest file exists in
           ``manifest.path``
```
mbolivar-nordic, line 308:
```suggestion
        - ``FileNotFoundError`` if no manifest file exists as determined by
          ``manifest.path`` and ``manifest.file``
```

src/west/manifest.py
```diff
@@ -1288,6 +1292,13 @@ def as_frozen_yaml(self, **kwargs) -> str:
     def projects(self) -> List[Project]:
         return self._projects
 
+    @property
+    def filename(self):
```
mbolivar-nordic, line 1296: Since this is API, it needs a docstring, even if it's just a short one.

M1cha:
Thanks for the comments and suggestions. I should be able to make these changes by the end of this week.

src/west/manifest.py
M1cha: I actually thought about removing this since it's currently unused. I added it initially before I realized that every potential user has that information in a variable already anyway.

Are there tools/plugins which might benefit from such APIs or is it all internal?

mbolivar-nordic: > Are there tools/plugins which might benefit from such APIs or is it all internal?

The west APIs are separately documented and maintained, yes: https://docs.zephyrproject.org/latest/guides/west/west-apis.html#west-apis

There are various users both upstream (e.g. zephyr/scripts/zephyr_module.py) and in downstream extension commands.

> I actually thought about removing this since it's currently unused.

Go for removing it, we can always add things when we need them later.

mbolivar-nordic:
> Thanks for the comments and suggestions. I should be able to make these changes by the end of this week.

Awesome, thanks for that.

M1cha:
I applied your changes and added tests. Feel free to leave additional feedback :)

M1cha:
Windows seems to fail to delete the protected .git directory. I don't know much about windows and if you don't know the solution either I could just use a different directory instead of deleting the old one.

tests/test_project.py
```diff
@@ -562,6 +563,70 @@ def test_init_local_missing_west_yml_failure(repos_tmpdir):
...
+        f.write('[')
+
+    # fails because of the syntax error
+    cwd = os.getcwd()
+
+    with pytest.raises(subprocess.CalledProcessError):
+        cmd(f'init -l "{zephyr_install_dir}"')
+        # init with a local manifest doesn't parse the file, so let's access it
+        workspace.chdir()
+        cmd('list')
```
mbolivar-nordic, line 592: This looks like dead code to me. The line `cmd(f'init -l "{zephyr_install_dir}"')` should raise the exception. Am I missing something?

tests/test_project.py
```diff
@@ -562,6 +563,70 @@ def test_init_local_missing_west_yml_failure(repos_tmpdir):
...
+
+    os.chdir(cwd)
+    shutil.rmtree(workspace / '.west')
```
mbolivar-nordic, line 595: I would just hack around the windows issue by using `shutil.move()` to put them somewhere harmless inside the tmpdir, instead of `shutil.rmtree()`.

tests/test_project.py
M1cha: I don't know if this is intended behavior or not, but as the comment above says, it looks like when using a local manifest west doesn't seem to parse that file at all because even syntax errors result in successful initialization.

mbolivar-nordic: Yep, this is intended behavior. We only need to parse the manifest when initializing from the network, because `self: path: foo` in the manifest file means "put this repository in directory `foo` in the initialized workspace when you're done". That's what this comment is hinting at: https://github.com/zephyrproject-rtos/west/blob/master/src/west/app/project.py#L267

By contrast, `west init -l foo` means "`foo` is already exactly where I want it; initialize a workspace around it". There's no need to parse the manifest at init time in that case.

mbolivar-nordic: @M1cha any plans to update this? I think I've answered outstanding questions. Please let me know if I missed anything and thanks again for your work.

M1cha: I got sick but am feeling better now. I'll update the PR today :)

tests/test_project.py
```diff
@@ -562,6 +563,70 @@ def test_init_local_missing_west_yml_failure(repos_tmpdir):
...
+    cwd = os.getcwd()
+
+    with pytest.raises(subprocess.CalledProcessError):
+        cmd(f'init -l "{zephyr_install_dir}"')
+        # init with a local manifest doesn't parse the file, so let's access it
+        workspace.chdir()
+        cmd('list')
+
+    os.chdir(cwd)
+    shutil.move(workspace / '.west', workspace / '.west-syntaxerror')
```
mbolivar-nordic, line 595: Thanks for updating this! I'm glad to hear you're feeling better.

```suggestion
```

We are almost there. I just think this part of the test should be removed and then this is ready to go in. Sorry I was unclear about the intended behavior. What I meant was that since `west init -l` isn't parsing the file, there's no need for this part of the test.

You also need to rebase now.

If you can do those two things I am ready to approve. Thanks again :+1:

M1cha: well the reason I wrote that code is so I can verify that west is indeed using the correct manifest file. The syntax error is the way I detect if it's using the west.yml or the correct name that was specified.

M1cha: if you want me to apply your change anyway, just confirm that and I'll do it.

src/west/manifest.py
```diff
@@ -152,7 +152,10 @@ def _mpath(cp: Optional[configparser.ConfigParser] = None,
     cfg.read_config(configfile=cfg.ConfigFile.LOCAL, config=cp, topdir=topdir)
 
     try:
-        return cp.get('manifest', 'path')
+        path = cp.get('manifest', 'path')
+        filename = cp.get('manifest', 'file', fallback=_WEST_YML)
```
tejlmand, line 156: How will this behave if manifest project uses `myname.yml` but we import another project manifest that uses `west.yml` or `othername.yml` for their manifest ?

tejlmand (changes requested):

src/west/manifest.py
M1cha: the `import` key already supports specifying a relative path:
https://docs.zephyrproject.org/latest/guides/west/manifest.html#west-manifest-import-path

mbolivar-nordic: @tejlmand please revisit

tejlmand: thanks for the clarification.

tejlmand (approved):

PR #415, opened 2020-07-13
Make sure west update can be used to recover from a bad import

mbolivar-nordic:
Consider the following situation:

1. `m` is the manifest repository, `p` is a project
2. `m/west.yml` imports `p` at revision `rbad`; `p/west.yml` at `rbad` contains an invalid manifest
3. user runs `west update`, setting `p`'s `manifest-rev` to `rbad` (and failing the update)
4. user updates `m/west.yml` to point at `p` revision `rgood`, which contains good manifest data
5. user runs `west update` again

The `west update` in the last step should fix `p`'s manifest-rev, pointing it at `rgood`, and should succeed.

It doesn't, because west refuses to run `update` and just allows the command to crash because there was no valid manifest data, leaving the workspace in an unusable state: even though the top-level manifest points at a valid configuration, you can't run west update, or list, or ....

Whoops. Turns out `west update` is perfectly smart enough to fix this situation, so just let it run.

This opens us up to running `west update` even with an invalid manifest and no imports. We handle this in main.py.

Reported by @plskeggs

### DIFF
--- a/src/west/app/main.py
+++ b/src/west/app/main.py
@@ -140,7 +140,12 @@ class WestApp:
         # from the manifest itself, which we have failed to load.)
 
         # A few commands are always safe to run without a manifest.
-        no_manifest_ok = ['help', 'config', 'topdir', 'init', 'manifest']
+        # The update command is sometimes safe and sometimes not, but
+        # we need to include it in this list because it's the only way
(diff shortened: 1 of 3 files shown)


plskeggs:
Looks good to me.

Issue #420, opened 2020-08-07
west list presents wrong path for manifest project.


tejlmand:
`west list` presents the wrong path for the manifest project.

This has the implication that `zephyr_module.py` (https://github.com/zephyrproject-rtos/zephyr/blob/master/scripts/zephyr_module.py) may fail, in case it cannot locate the right folder.

This can happen when a manifest project is ALSO a Zephyr module, for example: https://github.com/nrfconnect/sdk-nrf

If a user initialize this repo manually, and not locate it in the path stated here:
https://github.com/nrfconnect/sdk-nrf/blob/master/west.yml#L117-L119

For example:
```
$ mkdir nrf-test
$ cd nrf-test/
$ git init .
Initialized empty Git repository in /<path>/nrf-test/.git/
$ git remote add ncs https://github.com/nrfconnect/sdk-nrf
$ git fetch ncs
remote: Enumerating objects: 24, done.
...
$ git checkout ncs/master 
Note: checking out 'ncs/master'.
$ cd ..
$ west init -l nrf-test/
=== Initializing from existing manifest repository nrf-test
--- Creating /projects/github/ncs/.west and local configuration file
=== Initialized. Now run "west update" inside /projects/github/ncs.
$ west list
manifest     nrf                          HEAD                                     N/A
...
```
as seen, the path printed by `west list` is `nrf`, but the correct path of the manifest repo is in fact: `nrf-test`
Also seen with:
```
$ west topdir
/projects/github/ncs
$ west config manifest.path
nrf-test
```

There could be several reason for wanting to locate the manifest repo at custom folder.
One could be for additional testing, when there is also a `nrf` folder in use.

If the `nrf` repo is not there, then `zephyr_module.py` will fail, as it cannot lookup a `nrf/zephyr/module.yml` file.
If there is ALSO an nrf repo available, then things might become even harder to spot, cause in that case `nrf/zephyr/module.yml` will be loaded as a module, and NOT `nrf-test/zephyr/module.yml`

Fix:
`west list` should print `west config manifest.path` instead of `self: path: <path>`.

PR #421, opened 2020-08-07
west list: fix manifest path

mbolivar-nordic:
This should come from the manifest.path configuration option, which
may disagree with "self: path:" in the manifest file itself.

Add a --manifest-path-from-yml option as a chicken bit to keep the old
behavior in case anybody is relying on it.

Fixes: #420

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -388,6 +388,11 @@ class List(_ProjectCommand):
             '''))
         parser.add_argument('-a', '--all', action='store_true',
                             help='ignored for backwards compatibility'),
+        parser.add_argument('--manifest-path-from-yaml', action='store_true',
+                            help='''print the manifest repository's path
+                            according to the manifest file YAML, which may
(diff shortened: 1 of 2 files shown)


src/west/app/project.py
```diff
@@ -427,11 +432,22 @@ def delay(func, project):
             # as SHAs, unless they are specifically requested, and then
             # ensures they are only computed once.
             try:
+                if (isinstance(project, ManifestProject) and not
+                        args.manifest_path_from_yml):
+                    # Special-case the manifest repository while it's
+                    # still showing up in the 'projects' list. Yet
+                    # more evidence we should tackle #327.
+                    path = config.get('manifest', 'path')
```
tejlmand, line 440: guess this is one more .... manifest project is not a real project 😄

mbolivar-nordic:
Great catch. Fixed and regression tests added. Thanks for review.

PR #422, opened 2020-08-11
manifest: adding manifest_path to kwargs when calling Manifest.__init__

tejlmand:
If `Manifest.from_file(topdir=<topdir>)` is called with topdir then
Manifest is instantiated with topdir, source_file, and manifest_path.

However, if Manifest.from_file() is called without providing
topdir, then Manifest is instantiated with topdir and source_file only.

This causes the path to manifest repo to become None in later calls and
results in: https://github.com/nrfconnect/sdk-zephyr/pull/341

This is now fixed by ensure `manifest_path` is used for instantiating
Manifest class when using Manifest.from_file().

### DIFF
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -908,10 +908,11 @@ class Manifest:
                 # neither source_file nor topdir: search the filesystem
                 # for the workspace and use its manifest.path.
                 topdir = util.west_topdir()
+                mpath = _mpath(topdir=topdir)
                 kwargs.update({
                     'topdir': topdir,
-                    'source_file': os.path.join(topdir, _mpath(topdir=topdir),
-                                                _WEST_YML)
+                    'source_file': os.path.join(topdir, mpath, _WEST_YML),
+                    'manifest_path': mpath
                 })
             else:
                 # Just source_file: find topdir starting there.
--- a/tests/test_manifest.py
+++ b/tests/test_manifest.py
@@ -1051,12 +1051,21 @@ def test_as_dict_and_yaml(manifest_repo):
     for p in frozen_expected['manifest']['projects']:
         p['revision'] = fake_sha
 
+    # Manifest.from_file() and Manifest.from_file(topdir=<topdir>) shall
+    # produce result when given topdir is identical to what util.west_topdir()
+    # produces.
     manifest = MF()
+
+    manifest_topdir = MF(topdir=os.path.dirname(manifest_repo))
 
     # We can always call as_dict() and as_yaml(), regardless of what's
     # cloned.
 
     as_dict = manifest.as_dict()
+
+    as_dict_topdir = manifest_topdir.as_dict()
+    assert as_dict == as_dict_topdir
+
     yaml_roundtrip = yaml.safe_load(manifest.as_yaml())
     assert as_dict == content_dict
     assert yaml_roundtrip == content_dict
--- a/tests/test_manifest.py
+++ b/tests/test_manifest.py
@@ -1041,7 +1041,7 @@ def test_as_dict_and_yaml(manifest_repo):
                        'path': 'project-two',
                        'clone-depth': 1,
                        'west-commands': 'commands.yml'}],
-                     'self': {}}}
+                     'self': {'path': os.path.basename(manifest_repo)}}}
 
     with open(manifest_repo / 'west.yml', 'w') as f:
         f.write(content_str)


tejlmand:
@tomchy this fixes the issue reported here: https://github.com/nrfconnect/sdk-zephyr/pull/341 in the right place.

tejlmand:
@mbolivar-nordic this should be considered to trigger a west release.

tejlmand:
@mbolivar-nordic I have updated the test to verify that `Manifest.from_file()` produces same result as `Manifest.from_file(topdir=<topdir>)`, https://github.com/zephyrproject-rtos/west/pull/422/commits/061bcbff0c71db655bb97166620289286a15c542.

Along this process, I noticed this test `test_parse_multiple_manifest_files` and comment:
https://github.com/zephyrproject-rtos/west/blob/master/tests/test_manifest.py#L839-L842
```
   # The project hierarchy in its return value should still be rooted
    # in the topdir, but the resulting ManifestProject does not have a
    # path, because it's not set in the file, and we're explicitly not
    # comparing its path to manifest.path.
```
I personally don't see the rationale that the Manifest object should not contain the path to the manifest file, so I changed that as that is needed to ensure `Manifest.from_file()` behaves the same as `Manifest.from_file(topdir=<topdir>)`

mbolivar-nordic:
> @mbolivar-nordic this should be considered to trigger a west release.

Can you explain why? This seems to be handling an edge case. Cutting a new release takes about a day of work to test on all the different platforms so I want to understand if and how users are running into this in the wild from command line use cases.

It looks like the test case commit is being added before the changes which makes it pass. This breaks bisectability if it's a test for changes made later in the series. Please commit regression tests with or after the commits that fix the issue.

src/west/manifest.py
```diff
@@ -908,18 +908,20 @@ def from_file(source_file: Optional[PathType] = None,
...
                 })
             else:
                 # Just source_file: find topdir starting there.
                 # We need source_file in kwargs as that's what gets used below.
                 kwargs.update({
                     'source_file': source_file,
                     'topdir':
-                    util.west_topdir(start=os.path.dirname(source_file))
+                    util.west_topdir(start=os.path.dirname(source_file)),
+                    'manifest_path': os.path.dirname(source_file)
```
mbolivar-nordic, line 924: This assumes the manifest directly contains the source file. That is true for foo/west.yml but it is not true in the general case the API is trying to solve. Why isn't this calling `_mpath()` too?

tests/test_manifest.py
```diff
@@ -837,19 +837,18 @@ def test_parse_multiple_manifest_files(manifest_repo):
...
     # However, the project hierarchy *must* be rooted at topdir.
     manifest = Manifest.from_file(source_file=another_yml)
     assert len(manifest.projects) == 3
     assert manifest.topdir is not None
     assert PurePath(manifest.topdir) == PurePath(topdir)
     mproj = manifest.projects[0]
-    assert mproj.path is None
-    assert mproj.abspath is None
-    assert mproj.posixpath is None
+    assert mproj.path == topdir
```
mbolivar-nordic, line 849: This doesn't look right. How could the manifest project have the same path as the top level directory of the workspace?

By definition, it is always in a directory of topdir.

tejlmand:
> > @mbolivar-nordic this should be considered to trigger a west release.
> 
> Can you explain why? This seems to be handling an edge case. Cutting a new release takes about a day of work to test on all the different platforms so I want to understand if and how users are running into this in the wild from command line use cases.
> 

This is far from an edge case.
Everyone creating their own manifest WITHOUT a `self: path: <path>` section in the manifest file will face this issue, example:

Manifest file:
```
manifest:
  version: 0.7
  remotes:
    - name: zephyrproject

  defaults:
    remote: zephyrproject

  projects:
    - name: zephyr
      revision: 3dba8cc8509b1c52d5bebf6eafcf95c0eb8b2cbe
      import: true
```

Let's init, update, and build:
```
$ west init -l manifest_422
=== Initialized. Now run "west update" inside /projects/github/ncs.

$ west update
=== updating zephyr (zephyr):
....

$ west build -bnrf52840dk_nrf52840 zephyr/samples/hello_world/
-- west build: generating a build system
...
CMake Error at /projects/github/ncs/zephyr/cmake/zephyr_module.cmake:47 (message):

    File "/projects/github/ncs/zephyr/scripts/zephyr_module.py", line 316, in <module>
      main()
    File "/projects/github/ncs/zephyr/scripts/zephyr_module.py", line 257, in main
      meta = process_module(project)
    File "/projects/github/ncs/zephyr/scripts/zephyr_module.py", line 99, in process_module
      module_path = PurePath(module)
    File "/usr/lib/python3.7/pathlib.py", line 633, in __new__
      return cls._from_parts(args)
    File "/usr/lib/python3.7/pathlib.py", line 665, in _from_parts
      drv, root, parts = self._parse_args(args)
    File "/usr/lib/python3.7/pathlib.py", line 649, in _parse_args
      a = os.fspath(a)

  TypeError: expected str, bytes or os.PathLike object, not NoneType
```

This happens because manifest path is `None` when `self: path: <path>` is not present in the manifest file.
This can also be seen with:
```
Python 3.7.5 (default, Nov  7 2019, 10:50:52) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from west.manifest import Manifest
>>> manifest = Manifest.from_file()
>>> print(manifest.get_projects([])[0].path)
None
>>> print(manifest.get_projects([])[1].path)
zephyr
```

About cutting a new release, I don't say it has to be done right away, cause people can work around the issue by adding:
```
self:
  path: mypath
```
to the manifest.
What I mean is that this issue in itself is enough to trigger a release, but i'm fine with waiting to see if more things should be included in the release.

> It looks like the test case commit is being added before the changes which makes it pass. This breaks bisectability if it's a test for changes made later in the series. Please commit regression tests with or after the commits that fix the issue.

How do you understand bisectability ?
Finding the point in time when the system broke or when all test cases passed ?

Bisectability to me is for finding when the system broke, and the system was already broken.
Creating the test case proves that system is already broken, and even if we add the test case commit AFTER the fix, then the system is still broken. The test case just reveals this.

This is common TDD, make a test that shows that things are broken before fixing them.
But if you prefer, I can swap the two commits, it just feels wrong to fix something, before it has been proved to be broken :wink:

tests/test_manifest.py
tejlmand: If you look at the code:
https://github.com/zephyrproject-rtos/west/blob/95cd8ff2782c694c78f29391890216b2fef9f0db/tests/test_manifest.py#L804
https://github.com/zephyrproject-rtos/west/blob/95cd8ff2782c694c78f29391890216b2fef9f0db/tests/test_manifest.py#L844
```
    another_yml = topdir / 'another.yml'
...
    manifest = Manifest.from_file(source_file=another_yml)
....
```
so as you see, this test case uses a specific yaml file located directly in the topdir, and NOT a manifest file located in `manifest.path`.
Because this is provided directly as `Manifest.from_file(source_file=another_yml)` then the manifest path to `another_yml` is identical to `topdir` in this particular case.

src/west/manifest.py
tejlmand: because calling `_mpath` would return the west config value `manifest.path`.

But when calling `Manifest.from_file(source_file=<source_file>)` , then that file can be located anywhere, and in that case the value of `manifest.path` which is returned by `_mpath` might be wrong.

So as soon source file is provided directly, then `_mpath` cannot be used reliable.

tejlmand: as example, changing this in `_mapth` would cause the mpath in this comment: https://github.com/zephyrproject-rtos/west/pull/422#discussion_r470450611 to become `mp`, as that is the value of `manifest.path`.

But `another_yaml` is located directly at `topdir`, therefore in this case it is more correct to use `topdir`.

**note** if `source_file` is not provided, as example: `Manifest.from_file()` or `Manifest.from_file(topdir=<topdir>)` then `_mpath` is of course used.

tests/test_manifest.py
mbolivar-nordic: The results are nonsense in that case. The manifest.path attribute should be `None` because there is no manifest repository. There can't be a path to a manifest repository if there is no such repository.

Taking another look at the comments in this test, it even documents this intent:

```
    # [...] the resulting ManifestProject does not have a
    # path, because it's not set in the file, and we're explicitly not
    # comparing its path to manifest.path.
```

mbolivar-nordic:
> How do you understand bisectability ?
> Finding the point in time when the system broke or when all test cases passed ?

What I mean here is that we should not deliberately introduce points in our history where our own tests are failing. I am lumping "tests pass at each commit" in the same category as "code builds at each commit" here.

The TDD red/green/refactor cycle is about development, but the history of the repository should have passing tests all the time in my opinion. That's why I always add regression tests in either the commit that fixes the feature or in a following commit.

mbolivar-nordic:
> Let's init, update, and build:

I appreciate this explanation. I now understand the issue more clearly. Thank you.

This particular bug can be fixed with just the following changes to manifest.py:

```diff
diff --git a/src/west/manifest.py b/src/west/manifest.py
index 798e61f..372c4f5 100644
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -932,10 +932,12 @@ class Manifest:
                 # neither source_file nor topdir: search the filesystem
                 # for the workspace and use its manifest.path.
                 topdir = util.west_topdir()
+                mpath = _mpath(topdir=topdir)
                 kwargs.update({
                     'topdir': topdir,
-                    'source_file': os.path.join(topdir, _mpath(topdir=topdir),
-                                                _WEST_YML)
+                    'source_file': os.path.join(topdir, mpath,
+                                                _WEST_YML),
+                    'manifest_path': mpath
                 })
             else:
                 # Just source_file: find topdir starting there.
```

All that is necessary to keep the existing test suite passing is to update `content_dict` in `test_as_dict_and_yaml()` as you have already done. In other words, we avoid this hunk:

```diff
                 kwargs.update({
                     'source_file': source_file,
                     'topdir':
-                    util.west_topdir(start=os.path.dirname(source_file))
+                    util.west_topdir(start=os.path.dirname(source_file)),
+                    'manifest_path': os.path.dirname(source_file)
                 })
```

And we do not have to make the another.yml case have a path attribute in the manifest. The additional test in 061bcbff0c71db655bb97166620289286a15c542 passes too. I think this makes the most sense semantically.

Would you be willing to go with that?

tejlmand:
> In other words, we avoid this hunk:
> 
> ```diff
>                  kwargs.update({
>                      'source_file': source_file,
>                      'topdir':
> -                    util.west_topdir(start=os.path.dirname(source_file))
> +                    util.west_topdir(start=os.path.dirname(source_file)),
> +                    'manifest_path': os.path.dirname(source_file)
>                  })
> ```

What would be the purpose ?
Removing that would mean that `manifest_path` becomes `None` in that case.
To me this seems like just closing your eyes on manifest_path being None in the `Manifest.from_file(source_file=<file>)` case.
If manifest_path is None, it will not be possible for code to obtain the path to the manifest folder.

There will also be an inconsistency between `Manifest.from_file()` and `Manifest.from_file(source_file=<default_west_yml>), as one will have the right manifest path, and hence also be correct when listing project paths, and the second will not.
Example
```
$ west config manifest.path 
manifest_422
$ python3 
Python 3.7.5 (default, Nov  7 2019, 10:50:52) 
[GCC 8.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> from west.manifest import Manifest
>>> from west import util
>>> import os
>>> manifest = Manifest.from_file()
>>> print(manifest.get_projects([])[0].path)
manifest_422
>>> manifest = Manifest.from_file(os.path.join(util.west_topdir(), 'manifest_422', 'west.yml'))
>>> print(manifest.get_projects([])[0].path)
None
```
Note: `from_file(os.path.join(util.west_topdir(), 'manifest_422', 'west.yml'))` is the same file that is being used when calling `from_file()`, and therefore those two ways of calling `from_file` should be identical, which they are not.

With my current proposal, the print of path is different in the sense that one is full path, and the other is relative to topdir.
But none of them becomes `None`, which may lead to unintended errors later.

```
>>> manifest = Manifest.from_file()
>>> print(manifest.get_projects([])[0].path)
manifest_422
>>> manifest = Manifest.from_file(os.path.join(util.west_topdir(), 'manifest_422', 'west.yml'))
>>> print(manifest.get_projects([])[0].path)
/projects/github/ncs/manifest_422
```

tests/test_manifest.py
tejlmand: > . The manifest.path attribute should be None because there is no manifest repository.

Well, the manifest path is a path. You don't need the manifest project/folder to be a repository, but it is convenient.

But as soon as `Manifest.from_file()` and `Manifest.from_file(source_file=<default_west_yml>)` produces different results on the same input, then I think consistency breaks.
See: https://github.com/zephyrproject-rtos/west/pull/422#issuecomment-675321128

tejlmand:
This was also the reason for my suggestion on extending my solution a bit more with this from chat:
> I was thinking, maybe I should do some extra check to see if manifest (source file) has common path with config::manifest.path and if true, then strip the common path part ?

tejlmand:
> Would you be willing to go with that?

done

PR #425, opened 2020-08-14
tests: bandit support

mbolivar-nordic:
This is an experimental integration of the bandit tool for security
analysis. See https://bandit.readthedocs.io/ for more information on
bandit and its configuration.

The skipped tests in bandit.yaml are essentially saying "yes, we use
the subprocess module, and yes, we know what we are doing". There's
one case (shell=True) where we want the check in general but it's safe
to use in this case.


mbolivar-nordic:
Closing as I don't want this merged

PR #423, opened 2020-08-14
Add 'path-prefix' key to import feature

mbolivar-nordic:
By popular demand :smile: 

---------------------------------
**NOTE** CI says this isn't working on Windows; I'll figure out why tomorrow
---------------------------------

This series starts with some simple cleanups then adds the ability to stick imported projects (and the parent project you're importing from) in a subdirectory.

I may be missing something, but this turned out to be pretty easy to do. The groundwork laid in previous PRs generalized well.

Fixes: https://github.com/zephyrproject-rtos/west/issues/367

PR #426, opened 2020-08-14
manifest: use typing.NamedTuple

mbolivar-nordic:
While we were working on adding typing annotations, I was unaware that
typing.NamedTuple existed and did not use it.  This is a more
convenient alternative to collections.namedtuple since it provides
type checking for fields in the relevant tuples. Switch to it.

Suggested-by: Carles Cufí <carles.cufi@nordicsemi.no>

### DIFF
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -7,7 +7,6 @@
 Parser and abstract data types for west manifests.
 '''
 
-import collections
 import configparser
 import enum
 import errno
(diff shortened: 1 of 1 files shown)


mbolivar-nordic:
As suggested in https://github.com/zephyrproject-rtos/west/pull/423#discussion_r470725942

mbolivar-nordic:
Re-ran CI, looks like there were problems getting an available macOS node last time. All green now.

PR #428, opened 2020-08-18
add __main__.py

mbolivar-nordic:
This allows west to be executed as a module with "python3 -m west".

This can be useful if it's desired to execute west under a particular
Python interpreter, e.g. the same one that's used elsewhere in a
larger system.

### DIFF
--- a/src/west/__main__.py
+++ b/src/west/__main__.py
@@ -1,0 +1,3 @@
+from west.app.main import main
+
+main()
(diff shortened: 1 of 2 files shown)


tejlmand (approved):
Nice.
Tested with `python3 -mwest` with `west build -o=<flag> -- --cmake-flag` and `west list <flags>`
all working as expected.

PR #431, opened 2020-09-05
MAINTAINERS.rst: cleanups, clarifications

mbolivar-nordic:
Just some cleanups and clarifications on how I'd like to do this going
forward. The main clarifications are around SCHEMA_VERSION, since this
is changing in 0.8 and I want to make sure we're all on the same page.

### DIFF
--- a/MAINTAINERS.rst
+++ b/MAINTAINERS.rst
@@ -1,7 +1,15 @@
-Notes for maintainers cutting a release.
+Notes for maintainers.
 
 Pre-release test plan
 ---------------------
+
+0. If no release branch exists, fork the version numbers in the release branch
+   (vX.Y-branch) and the master branch. See "Cutting a release branch", below,
+   for details.
+
+   The rest of these steps should be done in the release branch::
+
+     git checkout vX.Y-branch
 
 1. Make tox happy on the following first-party platforms:
 
@@ -19,14 +27,8 @@ Pre-release test plan
    - Fedora
 
 3. Build alpha N (N=1 to start, then N=2 if you need more commits, etc.) and
-   upload to pypi.
-
-   Make sure to check if west.manifest.SCHEMA_VERSION also needs an update
-   before uploading. ::
-
-     git clean -ffdx
-     WEST_VERSION=X.YaN python3 setup.py sdist bdist_wheel
-     twine upload -u zephyr-project dist/*
+   upload to pypi. See "Building and uploading the release wheels" below for
+   a procedure.
 
 4. Install the alpha on test platforms. ::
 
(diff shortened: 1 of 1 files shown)


mbolivar-nordic:
(I've gone through these steps locally in a v0.8-branch I have yet to push, but I want to get some review on this before moving forward.)

MAINTAINERS.rst
```diff
@@ -1,8 +1,14 @@
-Notes for maintainers cutting a release.
+Notes for maintainers.
 
 Pre-release test plan
 ---------------------
 
+0. If no release branch exists, fork the version numbers in the release branch
+   (vX.Y-branch) and the master branch. See "Cutting a release branch", below,
```
tejlmand, line 7: >  fork the version numbers in the release branch (vX.Y-branch) and the master branch. 

I'm not sure exactly what you mean by this. I cannot fork from two branches.
Did you means something like:
> fork the version number from the master branch into a new release branch (vX.Y-branch)
   (vX.Y-branch).

MAINTAINERS.rst
```diff
@@ -1,8 +1,14 @@
...
+   for details.
+
+   The rest of these steps should be done in the release branch.
```
tejlmand, line 10: I know what you mean, but should we make it even clearer ?
Switch to the newly created release branch:
```
git checkout vX.Y-branch`
```

MAINTAINERS.rst
```diff
@@ -19,14 +25,8 @@ Pre-release test plan
...
 
 3. Build alpha N (N=1 to start, then N=2 if you need more commits, etc.) and
-   upload to pypi.
-
-   Make sure to check if west.manifest.SCHEMA_VERSION also needs an update
-   before uploading. ::
-
-     git clean -ffdx
-     WEST_VERSION=X.YaN python3 setup.py sdist bdist_wheel
-     twine upload -u zephyr-project dist/*
```
tejlmand, line 29: Why are the example commands removed ?

tejlmand (changes requested):
Nice improvement.
Just one sentence where i'm not clear what you actually mean.

And an additional suggestion.
For all places where branches, commits, etc. are used, then it would be nice to have the `git <command>` available, as it makes it easier to just copy and adjust the command.
Also, if someone else joins, they can see what commands was actually used in each step.

MAINTAINERS.rst
mbolivar-nordic: The two branches fork from each other.

Precondition: `vX.Y-branch` does not exist, `master` is at version `X.(Y-1)`.
Postcondition: `v.X.Y-branch` exists and is at version `vX.Y.0a1`, `master` is at version `vX.Y.99`.

MAINTAINERS.rst
mbolivar-nordic: >  Why are the example commands removed ?

They weren't. There are two copies. I eliminated the redundant copy, leaving the single point of truth in the details section.

mbolivar-nordic:
@tejlmand thanks for review, now addressed.

MAINTAINERS.rst
tejlmand: much clearer, could you write that instead ?

mbolivar-nordic: >  could you write that instead ?

Sure. Added to the details section.

tejlmand (approved):

Issue #435, opened 2020-09-14
"west init -l ." is broken in v0.8.0a1 and master


mbolivar-nordic:
To reproduce:

```bash
mkdir workspace
cd workspace
git clone https://github.com/zephyrproject-rtos/zephyr
cd zephyr
west init -l .
west update
```

Expected: works

Actual: 

```
(west-dev) mbolivar@ubuntu:/workdir/zephyr$ west init -l .
=== Initializing from existing manifest repository 
--- Creating .west and local configuration file
=== Initialized. Now run "west update" inside ..
(west-dev) mbolivar@ubuntu:/workdir/zephyr$ west update
Traceback (most recent call last):
  File "/home/mbolivar/.virtualenvs/west-dev/bin/west", line 8, in <module>
    sys.exit(main())
  File "/home/mbolivar/.virtualenvs/west-dev/lib/python3.8/site-packages/west/app/main.py", line 780, in main
    app.run(argv or sys.argv[1:])
  File "/home/mbolivar/.virtualenvs/west-dev/lib/python3.8/site-packages/west/app/main.py", line 106, in run
    self.run_command(argv)
  File "/home/mbolivar/.virtualenvs/west-dev/lib/python3.8/site-packages/west/app/main.py", line 337, in run_command
    cmd.run(args, unknown, self.topdir, manifest=self.manifest)
  File "/home/mbolivar/.virtualenvs/west-dev/lib/python3.8/site-packages/west/commands.py", line 129, in run
    self.do_run(args, unknown)
  File "/home/mbolivar/.virtualenvs/west-dev/lib/python3.8/site-packages/west/app/project.py", line 722, in do_run
    self.update_all(args)
  File "/home/mbolivar/.virtualenvs/west-dev/lib/python3.8/site-packages/west/app/project.py", line 739, in update_all
    manifest = Manifest.from_file(importer=self.update_importer,
  File "/home/mbolivar/.virtualenvs/west-dev/lib/python3.8/site-packages/west/manifest.py", line 977, in from_file
    return Manifest(**kwargs)
  File "/home/mbolivar/.virtualenvs/west-dev/lib/python3.8/site-packages/west/manifest.py", line 1102, in __init__
    source_data = source_file.read_text()
  File "/usr/lib/python3.8/pathlib.py", line 1227, in read_text
    with self.open(mode='r', encoding=encoding, errors=errors) as f:
  File "/usr/lib/python3.8/pathlib.py", line 1213, in open
[...]

PR #436, opened 2020-09-14
init: fix 'west init -l .'

mbolivar-nordic:
This isn't working properly without resolving the manifest directory.

Fixes: #435

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -213,7 +213,12 @@ class Init(_ProjectCommand):
         if args.manifest_rev is not None:
             log.die('--mr cannot be used with -l')
 
-        manifest_dir = Path(args.directory or os.getcwd())
+        # We need to resolve this to handle the case that args.directory
+        # is '.'. In that case, Path('.').parent is just Path('.') instead of
+        # Path('..').
+        #
+        # https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.parent
+        manifest_dir = Path(args.directory or os.getcwd()).resolve()
         manifest_filename = args.manifest_file or 'west.yml'
         manifest_file = manifest_dir / manifest_filename
         topdir = manifest_dir.parent
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -627,6 +627,19 @@ def test_init_local_with_manifest_filename(repos_tmpdir):
     cmd('update')
 
 
+def test_init_local_with_empty_path(repos_tmpdir):
+    # Test "west init -l ." + "west update".
+    # Regression test for:
+    # https://github.com/zephyrproject-rtos/west/issues/435
+
+    local_manifest_dir = repos_tmpdir / 'workspace' / 'zephyr'
+    clone(str(repos_tmpdir / 'repos' / 'zephyr'), str(local_manifest_dir))
+    os.chdir(local_manifest_dir)
+    cmd('init -l .')
+    cmd('update')
+    assert (repos_tmpdir / 'workspace' / 'subdir' / 'Kconfiglib').check(dir=1)
+
+
 def test_init_with_manifest_filename(repos_tmpdir):
     # Test 'west init --mf' on a normal repo
 


tejlmand:
I never used `west init -l .` before, only `west init -l <dir>`.

I just tried the `west init -l .` with older `west` and I see that it works, but I would like to challenge if this is correct behavior.

If I execute a command in a folder, then I would usually not expect that command to enter parent directory and from there create stuff.

Imagine someone (because not understanding `west` well enough) could end up doing:
```
$ cd ~
$ git clone https://github.com/zephyrproject-rtos/zephyr.git
$ cd zephyr
$ west init -l .
```
suddenly gets their home directory populated with `.west`, and if they also do `west update`, then they get a bunch of stuff in their home.
I would never expect `git init .` or `git clone <url> .` to start populating the parent dir.


`west help init` states:
```
usage: west init [-h] [-m MANIFEST_URL] [--mr MANIFEST_REV] [-l] [directory]

Creates a west workspace as follows:

  1. Creates a .west directory and clones a manifest repository
     from a git URL to a temporary subdirectory of .west,
     .west/<tmpdir>.
...
positional arguments:
  directory             Directory to create the workspace in (default: current
                        directory). Missing intermediate directories will be
                        created. If -l is given, this is the path to the
                        manifest repository to use instead.
```
but when using `-l` only the manifest repo is specified, not where the workspace will be created.
Apparently the workspace will be created in `<manifest-repo>/..`, but that is actually not clear from the help text.

Should there have been two dir arguments ?
One for manifest repo, and one for the workspace.
Like:
```
usage: west init [-h] [-m MANIFEST_URL] [--mr MANIFEST_REV] [-l manifest-dir] [directory]

...
  directory             Directory to create the workspace in (default: current
                        directory). Missing intermediate directories will be
                        created. 
  -l manifest-dir, --local manifest-dir  use an existing local manifest repository instead of
                        cloning one; cannot be combined with -m or --mr.

```

options
- extend to ` [-l manifest-dir] [directory]`
- Improve the help text
- Issue warning when user does `west init -l .`,as he might not be aware it will go up one folder.
- Block `-l.`

mbolivar-nordic:
> Imagine someone (because not understanding `west` well enough) could end up doing:


I don't have to imagine it. Questions about this behavior have happened on Slack more than once during Zephyr's transition to west.

People had `~/src/zephyr` directories and didn't understand why west had created `~/src/.west` and `~/src/net-tools` directories, etc.

I actually advocated against having a `west init -l` option because I thought this would be too confusing. But @carlescufi insisted and he was right. It's a very useful option to have, and people rely on it in various important situations.

Just because it has the potential to confuse people doesn't make it wrong. All this PR is doing is making the behavior the same whether you run `west init -l ~/zephyrproject/zephyr` or `west init -l .` (assuming you are in `~/zephyrproject/zephyr`).

There should be no difference between these two things. The shorthand `.` means current working directory. Whether or not you specify that as a full path or relative path makes no difference. I would strongly argue that treating the two cases differently is *more* surprising and incorrect.

tejlmand:
> I actually advocated against having a west init -l option because I thought this would be too confusing.

I think the `-l` is very valuable, but i'm not sure the `-l.` behavior is correct.

> Just because it has the potential to confuse people doesn't make it wrong. 

No, but if it confuses people, and also results in manually having to do cleanup (== wasted time), then it should be considered if the behavior is correct, or if it should be adjusted.

> Just because it has the potential to confuse people doesn't make it wrong. All this PR is doing is making the behavior the same whether you run `west init -l ~/zephyrproject/zephyr` or `west init -l .` (assuming you are in `~/zephyrproject/zephyr`).

But if you have the following manifest in `/tmp/manifest-repo`:
```
...
  self:
    path: subdir/manifest-repo
```

and then do:
```
$ mkdir /tmp/west-test-defpath -p
$ cd  /tmp/west-test-defpath
$ west init -m file:///tmp/manifest-repo/
```
then `.west` will be placed in `/tmp/west-test-defpath/.west` and manifest repo in `/tmp/west-test-defpath/subdir/manifest-repo`.
As expected, and specified by the manifest file.
See:
```
$ west init -m file:///tmp/manifest-repo/
=== Initializing in /tmp/west-test-defpath
--- Cloning manifest repository from file:///tmp/manifest-repo/, rev. master
Initialized empty Git repository in /tmp/west-test-defpath/.west/manifest-tmp/.git/
...
--- setting manifest.path to subdir/manifest-repo
=== Initialized. Now run "west update" inside /tmp/west-test-defpath.
```

But if you do similar, but using `west init -l <dir>`, you actually end up with `.west` under `subdir`.
```
$ mkdir /tmp/west-test-l/subdir/manifest-repo -p
$ cd /tmp/west-test-l/subdir/manifest-repo
$ git clone file:///tmp/manifest-repo/ .
$ $ cd /tmp/west-test-l/
$ west init -l subdir/manifest-repo
=== Initializing from existing manifest repository manifest-repo
--- Creating /tmp/west-test-l/subdir/.west and local configuration file
=== Initialized. Now run "west update" inside /tmp/west-test-l/subdir.
```
Whoops, that was not expected, and the same happens if I do:
```
$ west init -l .
=== Initializing from existing manifest repository manifest-repo
--- Creating /tmp/west-test-l/subdir/.west and local configuration file
=== Initialized. Now run "west update" inside /tmp/west-test-l/subdir.
```

So there is some inconsistency in how `west init -l` works and `west init -m`.

tejlmand (approved):
approved, as this PR simply restores old behavior.

But the comments on `west init -l .` is still valid.

mbolivar-nordic:
> So there is some inconsistency in how `west init -l` works and `west init -m`.
> But the comments on west init -l . is still valid.

Yes, but as far as I can tell, all of the differences you mention are by design and working as documented.

> approved, as this PR simply restores old behavior.

Thanks.

tejlmand:
I think a nice improvement could be to check that `$PWD` is only dir in new workspace when using `-l.` and if not (such as the ~/zephyr case), then issue an error, and provide a `--force` mechanism.
Just so the `~/zephyr` and `~/src` situations doesn't require additional manual clean-up.

But i'm fine with waiting for post `0.8`

mbolivar-nordic:
> I think a nice improvement could be to check that `$PWD` is only dir in new workspace when using `-l.` and if not (such as the ~/zephyr case), then issue an error, and provide a `--force` mechanism.

I don't think this is worth the effort. At first, `west init -l` was introduced as a migration path for existing zephyr users. Its remaining users are more advanced (such as CI system designers and maintainers).

The recommended way to use `west init` in the Zephyr GSG is with `-m`, not `-l`, so I think it is OK if `-l` is a little trickier, since it's for more advanced users at this point.

tejlmand:
> > So there is some inconsistency in how `west init -l` works and `west init -m`.
> 
> Yes, but as far as I can tell, all of the differences you mention are by design and working as documented.

I don't see where it is documented that `west init -l subdir/manifest-repo` would result in the `.west` folder / west workspace will be the parent directory of the manifest repo, in this case `/tmp/west-test-defpath/subdir/.west`.

And not two levels above the manifest repo, when the manifest path in `west.yml` has the value: `self::path subdir/manifest-repo`

tejlmand:
> I don't think this is worth the effort. 

I will not make the mistake, so I can safely agree with you.
But someone having to clean-up `~` for garbage folders might disagree.
I think you are most likely right, that those users are very few by now.

mbolivar-nordic:
> I don't see where it is documented that `west init -l subdir/manifest-repo` would result in the `.west` folder / west workspace will be the parent directory of the manifest repo, in this case `/tmp/west-test-defpath/subdir/.west`.

I think it's implied, but I agree this should be clearer:

https://github.com/zephyrproject-rtos/west/pull/438

PR #438, opened 2020-09-15
Improve west init -l docs

mbolivar-nordic:
Let's make it clearer what happens in this case.

For the sake of not making the command line help too long, let's drop
the point that it can't be combined with -m or --mr and just make it
clear that -l doesn't use MANIFEST_URL. People will find out right
away if they try it that it doesn't work if they combine these
options.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -164,9 +164,11 @@ class Init(_ProjectCommand):
         parser.add_argument('--mf', '--manifest-file', dest='manifest_file',
                             help='manifest file name to use')
         parser.add_argument('-l', '--local', action='store_true',
-                            help='''use an existing local manifest repository
-                            instead of cloning one; cannot be combined with
-                            -m or --mr.''')
+                            help='''use "directory" as an existing local
+                            manifest repository instead of cloning one from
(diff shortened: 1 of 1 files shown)


src/west/app/project.py
```diff
@@ -164,9 +164,11 @@ def do_add_parser(self, parser_adder):
...
                             help='manifest file name to use')
         parser.add_argument('-l', '--local', action='store_true',
-                            help='''use an existing local manifest repository
-                            instead of cloning one; cannot be combined with
-                            -m or --mr.''')
+                            help='''use "directory" as an existing local
+                            manifest repository instead of cloning one from
+                            MANIFEST_URL; .west is created next to "directory"
+                            in this case, and manifest.path is set to
+                            "directory"''')
```
tejlmand, line 171: > and manifest.path is set to "directory"

Not completely true.
If I do `west init -l /very/log/path/manifest-repo`, then `manifest.path` will be set to `manifest-repo`, and not `/very/log/path/manifest-repo`.

mbolivar-nordic: > Not completely true.

How about now? I'm using vaguer language.

tejlmand (approved):

Issue #437, opened 2020-09-15
Manifest._load_project() uses escapes_directory() in ways that break zephyr CI


mbolivar-nordic:
**Background**

- west generally assumes full control of the workspace and makes no promises about what happens if you use symlinks inside
- for west 0.8, we added a `path-prefix:` key to an `import:` map, allowing projects to be placed in subdirectories
- to make this safe from directory traversal attacks that attempt to overwrite files outside the workspace with attacker chosen content on `west update`, we use `west.util.escapes_directory()` to force a project's path to live inside the workspace and error out if it doesn't

**Issue details**

Let's say I make some project directory `~/workspace/myproject` a symlink to something outside the workspace, like so:

```bash
rm -rf /my/workspace/myproject
ln -s /project/cache/myproject /my/workspace/myproject
west update
```

**Actual behavior**

I will get something like this:

```
FATAL ERROR: Malformed manifest file: /my/workspace/west.yml
  Schema file: .../west/manifest-schema.yml
Hint: project myproject absolute path /my/workspace/myproject is not a subdirectory of topdir /my/workspace
```

This is because internally we use west.util.escapes_directory(), which compares paths using `Path.resolve()`, which eliminates symlinks.

Additionally, Zephyr's CI in `.buildkite/run-ci.sh` is doing something very similar to the toy example in the issue details above: https://github.com/zephyrproject-rtos/zephyr/blob/c6cf8d16736798f1209dbc55cf30dd72d205fa34/.buildkite/run.sh#L39

This is blowing up testing with west 0.8 due to this fatal error: https://github.com/zephyrproject-rtos/zephyr/pull/28373

**Expected behavior 1**

This is, at the very least, misleading because `/my/workspace/myproject` is definitely a subdirectory of `/my/workspace`, so at least we should fix the error message to say something that makes more sense, maybe like this:

```
FATAL ERROR:  project myproject has real path /project/cache/myproject, which is is not a subdirectory of topdir /my/workspace
```

**Expected behavior 2**

We need to get west 0.8 out the door and keep Zephyr's CI working. I see these different options:

1. try to make this check work with absolute paths in a way that does not resolve symlinks
2. add a new `west update --no-workspace-escape-check` option that keeps the check in by default, but turns it off when requested, and adds this to Zephyr's CI 

I favor option 2 because I think forcing people to jump through a hoop here and there when they are using west in unexpected ways and prioritizing the integrity of the rest of the file system makes the most sense.

[...]

PR #439, opened 2020-09-18
manifest: Use the parent folder for checking escape path

tejlmand:
Fixes: #437

The project itself will always have a directory name, `<project>`.
It only needs to be ensure that the path leading up to `<project>` must
not escape the west workspace.

We only want to ensure that any paths specified in the manifest doesn't
escape the workspace.

Any user should have the freedom to decide if the project itself should
be a symlink to another location.

This commit provides freedom to the user, and at the sametime ensures
that path-prefix cannot escape the workspace.


tejlmand:
closed in favor of: https://github.com/zephyrproject-rtos/west/pull/441

PR #440, opened 2020-09-18
replace whitelist/blacklist with includelist/excludelist

coderbyheart:
This provides alternatives to the racially laden terms `whitelist`,
used for a list of imports which are included, and `blacklist`, used
for a list of imports which are excluded with `includelist` and
respectively `excludelist`.

### DIFF
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -72,7 +72,7 @@ ImportedContentType = Optional[Union[str, List[str]]]
 ImporterType = Callable[['Project', str], ImportedContentType]
 
 # Type for an import map filter function, which takes a Project and
-# returns a bool. The various whitelists and blacklists are used to
+# returns a bool. The various allowlists and blocklists are used to
 # create these filter functions. A None value is treated as a function
 # which always returns True.
 ImapFilterFnType = Optional[Callable[['Project'], bool]]
@@ -214,15 +214,15 @@ def _manifest_content_at(project: 'Project', path: PathType,
 
 class _import_map(NamedTuple):
     file: str
-    name_whitelist: List[str]
-    path_whitelist: List[str]
-    name_blacklist: List[str]
-    path_blacklist: List[str]
+    name_allowlist: List[str]
+    path_allowlist: List[str]
+    name_blocklist: List[str]
+    path_blocklist: List[str]
     path_prefix: str
 
 def _is_imap_list(value: Any) -> bool:
-    # Return True if the value is a valid import map 'blacklist' or
-    # 'whitelist'. Empty strings and lists are OK, and list nothing.
+    # Return True if the value is a valid import map 'blocklist' or
+    # 'allowlist'. Empty strings and lists are OK, and list nothing.
 
     return (isinstance(value, str) or
             (isinstance(value, list) and
@@ -232,8 +232,8 @@ def _imap_filter(imap: _import_map) -> ImapFilterFnType:
     # Returns either None (if no filter is necessary) or a
     # filter function for the given import map.
 
-    if any([imap.name_whitelist, imap.path_whitelist,
-            imap.name_blacklist, imap.path_blacklist]):
+    if any([imap.name_allowlist, imap.path_allowlist,
+            imap.name_blocklist, imap.path_blocklist]):
         return lambda project: _is_imap_ok(imap, project)
     else:
         return None
@@ -251,18 +251,18 @@ def _is_imap_ok(imap: _import_map, project: 'Project') -> bool:
     # and False otherwise.
 
     nwl, pwl, nbl, pbl = [_ensure_list(lst) for lst in
-                          (imap.name_whitelist, imap.path_whitelist,
-                           imap.name_blacklist, imap.path_blacklist)]
+                          (imap.name_allowlist, imap.path_allowlist,
+                           imap.name_blocklist, imap.path_blocklist)]
     name = project.name
     path = Path(project.path)
-    blacklisted = (name in nbl) or any(path.match(p) for p in pbl)
-    whitelisted = (name in nwl) or any(path.match(p) for p in pwl)
-    no_whitelists = not (nwl or pwl)
+    blocked = (name in nbl) or any(path.match(p) for p in pbl)
+    allowed = (name in nwl) or any(path.match(p) for p in pwl)
+    no_allowlists = not (nwl or pwl)
 
-    if blacklisted:
-        return whitelisted
+    if blocked:
+        return allowed
     else:
-        return whitelisted or no_whitelists
+        return allowed or no_allowlists
 
 class _import_ctx(NamedTuple):
     projects: Dict[str, 'Project']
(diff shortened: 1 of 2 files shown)


coderbyheart:
@mbolivar-nordic @tejlmand ready for review, all checks pass.

src/west/manifest.py
```diff
@@ -1774,28 +1774,28 @@ def _load_imap(self, imp: Dict, src: str) -> _import_map:
...
         copy = dict(imp)
         ret = _import_map(copy.pop('file', _WEST_YML),
-                          copy.pop('name-whitelist', []),
-                          copy.pop('path-whitelist', []),
-                          copy.pop('name-blacklist', []),
-                          copy.pop('path-blacklist', []),
+                          copy.pop('name-allowlist', []),
+                          copy.pop('path-allowlist', []),
+                          copy.pop('name-denylist', []),
+                          copy.pop('path-denylist', []),
```
mbolivar-nordic, line 1780: I'm glad to see you getting ahead of this, but we can't implement it this way. We have to maintain backwards compatibility or NCS will break. Users (including NCS) have manifests in their history with the white/black terms and I think it's not OK to break those.

Can you please rework this so that the old values are still used as a fallback, and leave a comment explaining why?

We can then update the west documentation to only document the new way of doing things.

mbolivar-nordic (changes requested):

src/west/manifest.py
coderbyheart: I'd say west is a `0.x` library so breaking changes are accepted, but you are right: this can be handled in a non-breaking fashion, by keeping the support for the deprecated terms and printing a warning.

I'll look into it.

mbolivar-nordic: > I'd say west is a 0.x library so breaking changes are accepted,

The version number not being 1.0 is about the APIs being unstable, as is stated in the docs. The command line interface and manifest parsing needs to stay compatible with what people are already using unless it is impossible to do otherwise. Remember that west.yml is part of git history and people can do bisections across large commit ranges. It's not acceptable to introduce a break like this.

> but you are right: this can be handled in a non-breaking fashion, by keeping the support for the deprecated terms and printing a warning.

A warning is also not acceptable, for the same reason. Just silently falling back on the old names.

Thanks!

coderbyheart: I don't get why changing a file in a repo's `west.yaml` breaks git bisect operations for this repository.

However I believe you that it does and therefore adding support for both variants is the only way to do it right.

tejlmand: > I don't get why changing a file in a repo's `west.yaml` breaks git bisect operations for this repository.
> 

Manifest repo, as example `nrf-repo`.

> HEAD: bad (manifest: `allow-list`)
> ...
> HEAD\~25
> ...
> HEAD\~50: unknown (manifest: `white-list`)
> ...
> HEAD\~75
> ...
> HEAD\~100: good (manifest: `white-list`)

Starting a bisect:
```
git bisect start
git bisect bad
git bisect good HEAD~100
Bisecting: 49 revisions left to test after this (roughly 6 steps)
[HEAD~75] <text>
west update (this fails, cause latest `west` doesn't support `white-list` --> downgrade west)
pip install west==0.7.3
<test>
git bisect good
Bisecting: 24 revisions left to test after this (roughly 5 steps)
[HEAD~25] <text>
west update (this fails, cause old `west` doesn't support `allow-list` --> upgrade west)
pip install west==0.x.y
west update
<test>
git bisect bad
... --> maybe west needs to be downgraded this time, maybe not ?
```

you start to see how annoying this is to users, cause they won't know when a SHA that requires a different `west` version is needed, until `west update` fails.

tejlmand: > A warning is also not acceptable, for the same reason. Just silently falling back on the old names is what I am asking for.

Agree.

tejlmand: I think `include`/`exclude` would be more correct.
We are in fact including a project list/path, or excluding it from import.

```suggestion
                          copy.pop('name-includelist', []),
                          copy.pop('path-includelist', []),
                          copy.pop('name-excludelist', []),
                          copy.pop('path-exludelist', []),
```

tejlmand (changes requested):

src/west/manifest.py
mbolivar-nordic: Honestly if we're going to bikeshed this I would rather use 'allow' and 'block' for consistency with what other major projects seem to be doing: https://www.theregister.com/2019/09/03/chromium_microsoft_offensive/

I don't think there's any correctness issue, though.

tejlmand: > Honestly if we're going to bikeshed this I would rather use 'allow' and 'block' for consistency with what other major projects seem to be doing

Consistency is king. Accepted.

Wonder why `include`/`exclude` is not used, maybe `exclude` is too offensive for all those poor software developers (like myself) that sucked when playing football/soccer as child so we were always excluded from playing :thinking:

coderbyheart: If used `includelist`/`excludelist` and kept support for `whitelist`/`blacklist` in the latest push.

coderbyheart:
@tejlmand all tests pass. Are the changes ok for you, now?

src/west/manifest.py
```diff
@@ -214,15 +214,23 @@ def _manifest_content_at(project: 'Project', path: PathType,
...
+    '''@deprecated use name_includelist'''
     name_whitelist: List[str]
+    path_includelist: List[str]
+    '''@deprecated use path_includelist'''
     path_whitelist: List[str]
+    name_excludelist: List[str]
+    '''@deprecated use name_excludelist'''
     name_blacklist: List[str]
+    path_excludelist: List[str]
+    '''@deprecated use path_excludelist'''
```
mbolivar-nordic, line 227: There's no need for this tuple to contain these "deprecated" items. Instead just have the following fields:

- file
- name_allowlist
- path_allowlist
- name_denylist
- path_denylist
- path_prefix

(I believe @tejlmand is still asking for that change and has agreed to move to those names, but correct me if I'm wrong.)

We can then populate the tuple from the parsed YAML object as follows:

```py
# Here we fall back on old names for these fields which we've since
# renamed. The old names must still work to keep manifests written
# against west 0.7 and 0.8. Those are still in people's git histories
# and we need to keep 'west update' working when those revisions are
# checked out.
name_allowlist = copy.pop('name-allowlist', copy.pop('name-whitelist', []))
path_allowlist = copy.pop('path-allowlist', copy.pop('path-whitelist', []))
name_denylist = copy.pop('name-denylist', copy.pop('name-whitelist', []))
path_denylist = copy.pop('path-denylist', copy.pop('path-whitelist', []))

ret = _import_map(copy.pop('file', _WEST_YML),
                  name_allowlist, path_allowlist,
                  name_denylist, path_denylist,
                  copy.pop('path-prefix', ''))

```

The `_import_map` type is not public API as indicated by the leading `_` and the absence of any official documentation about it here: https://docs.zephyrproject.org/latest/guides/west/west-apis.html#module-west.manifest

We are therefore free to change its implementation details as needed; there is no "deprecation" necessary because it's not API. I don't think it makes sense to keep these here if we are making this change.

tests/test_manifest.py
```diff
@@ -1896,7 +1896,79 @@ def test_import_flags_ignore(tmpdir):
     ''', import_flags=ImportFlag.IGNORE)
     assert m.get_projects(['foo'])
 
-def test_import_map_name_whitelist(manifest_repo):
+def test_import_map_name_includelist(manifest_repo):
+    # This tests an example from the documentation which uses
+    # name-includelist.
```
mbolivar-nordic, line 1901: There is no such documentation :smiley: 

The previous version of this comment is referring to this example: https://docs.zephyrproject.org/latest/guides/west/manifest.html#example-3-1-downstream-with-name-whitelist

Can you please update the comment until we have a chance to update the documentation to reflect this change? We don't have any such documentation merged yet.

mbolivar-nordic (changes requested):
Thanks very much for the update. I have these blocking requests:

- remove extra fields from `_import_map` type
- settle on consistent naming with other projects doing this same change
- minor comment update

With those, I'm +1. Thanks again.

src/west/manifest.py
coderbyheart: Ok, I wasn't certain that _import_map is not used outside of the package. I'll remove the fallback fields.

@tejlmand [wrote](https://github.com/zephyrproject-rtos/west/pull/440#discussion_r494119991):
> I think `include`/`exclude` would be more correct.

tests/test_manifest.py
coderbyheart: Will do.

src/west/manifest.py
mbolivar-nordic: >  @tejlmand wrote:
>>  I think include/exclude would be more correct.

Yes, but then I responded:

>>  Honestly if we're going to bikeshed this I would rather use 'allow' and 'block' for consistency with what other major projects seem to be doing

And @tejlmand replied:

> Consistency is king. Accepted.

coderbyheart: Sure, changed.

tejlmand (approved):

mbolivar-nordic (approved):

mbolivar-nordic:
Thanks @coderbyheart

PR #441, opened 2020-09-24
manifest: rework workspace escape checks

mbolivar-nordic:
Let's use a lexical check instead of escapes_directory(). This
continues to let some users who are symlinking outside the workspace
to keep doing that, even though we don't officially support it in
west.

Also throw in a check for an absolute project path, which should never
happen and is similarly a sign of an attempt to escape the workspace.

### DIFF
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -24,7 +24,7 @@ import pykwalify.core
 import yaml
 
 from west import util
-from west.util import PathType, escapes_directory
+from west.util import PathType
 import west.configuration as cfg
 
 #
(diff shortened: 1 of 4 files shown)


mbolivar-nordic:
Example error messages now:

```
FATAL ERROR: can't load west manifest
  Malformed manifest file: /home/mbolivar/src/west-sandbox/escape/m/west.yml 
  Schema file: /home/mbolivar/src/west/src/west/manifest-schema.yml
  Hint: project "nope" path /this/is/wrong is absolute (and normalizes to /this/is/wrong); this should be relative to the workspace topdir(/home/mbolivar/src/west-sandbox/escape)
```

```
FATAL ERROR: can't load west manifest
  Malformed manifest file: /home/mbolivar/src/west-sandbox/escape/m/west.yml 
  Schema file: /home/mbolivar/src/west/src/west/manifest-schema.yml
  Hint: project "nope" path this/is/../../../../not-ok normalizes to ../../not-ok, which escapes the workspace topdir
```

src/west/manifest.py
```diff
@@ -1648,14 +1648,28 @@ def _load_project(self, pd: Dict, url_bases: Dict[str, str],
...
+        # symbolic links.
+        ret_norm = os.path.normpath(ret.path)
+
+        if os.path.isabs(ret_norm):
+            self._malformed(f'project "{ret.name}" has absolute path '
+                            f'{ret.path}; this must be relative to the '
+                            f'workspace topdir' +
+                            (f' ({self.topdir})' if self.topdir else ''))
+
+        if ret_norm.startswith('..'):
```
tejlmand, line 1669: so what if user does: `path-prefix: dummy/../../../i_escaped` ?

mbolivar-nordic: This case is shown above in the example errors.

tejlmand (approved):

src/west/manifest.py
mbolivar-nordic: (And now has an explicit test case apart from the existing path-prefix escape check)

PR #444, opened 2020-09-29
West v0.8.0

mbolivar-nordic:
Maureen asked us to hold off on this until after Zephyr 2.4.0. @carlescufi I held off yesterday since I saw you were off, but I realized you are still off now. I think we should just go ahead and release this, @tejlmand .


tejlmand (approved):
I've been looking forward to see this PR.

PR #446, opened 2020-10-05
An empty "self: path:" is an error

mbolivar-nordic:
A manifest like this is not well-formed:

```yaml
manifest:
  projects: [...]
  self:
    path:
```

Currently the `Manifest` parser accepts this data and gives back a `None` value in `self.projects[MANIFEST_PROJECT_INDEX].path`, which is wrong. The entire manifest should just be rejected as ill-formed.

### DIFF
--- a/tests/test_manifest.py
+++ b/tests/test_manifest.py
@@ -586,7 +586,6 @@ def test_manifest_project():
     assert mp.revision == 'HEAD'
     assert mp.clone_depth is None
 
-@pytest.mark.xfail()
 def test_self_tag():
     # Manifests may contain a self section describing the manifest
     # repository. It should work with multiple projects and remotes as
--- a/tests/test_manifest.py
+++ b/tests/test_manifest.py
@@ -586,6 +586,7 @@ def test_manifest_project():
     assert mp.revision == 'HEAD'
     assert mp.clone_depth is None
 
+@pytest.mark.xfail()
 def test_self_tag():
     # Manifests may contain a self section describing the manifest
     # repository. It should work with multiple projects and remotes as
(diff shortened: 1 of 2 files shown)


mbolivar-nordic:
@sdalu please speak up if you want this turned into a warning instead.

sdalu:
I'm fine with rejecting the manifest, as a reason is given in the error message.

Speaking about documentation, what would be the recommended form for T3? the one with `self: path: manifest-repo`, or `without the self:` section? Could be both form mentioned in https://docs.zephyrproject.org/latest/guides/west/repo-tool.html ?

tests/test_manifest.py
```diff
@@ -586,6 +586,7 @@ def test_manifest_project():
     assert mp.revision == 'HEAD'
     assert mp.clone_depth is None
 
+@pytest.mark.xfail()
```
tejlmand: I like that :+1:

tejlmand (approved):

mbolivar-nordic:
> I'm fine with rejecting the manifest, as a reason is given in the error message.

Thanks for confirming.

> Speaking about documentation, what would be the recommended form for T3? the one with `self: path: manifest-repo`, or without the `self:` section? Could both form be mentioned in https://docs.zephyrproject.org/latest/guides/west/repo-tool.html ?

This is really up to the use case.

If you are doing `west init -m https://example.com/some/path/foo`, and `foo/west.yml` does not have a `self: path:`, the default manifest repository path on the file system is `foo`. If that is OK, then there is no need to set `self: path:`.

If it's not OK, you can use `self: path:` to get consistent behavior regardless of the manifest repository fetch URL.

If you're doing `west init -l some_repo`, it doesn't matter either way, as any `self: path:` in `some_repo/west.yml` would just get ignored anyway.

PR #454, opened 2020-11-24
Project groups

mbolivar-nordic:
The purpose of this pull request is to allow projects to declare groups they belong to, like this:

```yaml
  manifest:
    # ...
    projects:
    - name: foo
      groups:
      - YouAreIt
    - name: bar
      groups:
      - YouAreIt
      - MeToo
    - name: baz
```

The above projects have these groups lists:

- foo: ['YouAreIt']
- bar: ['YouAreIt', 'MeToo']
- baz: []

Paired with this is a new top-level `update-groups` key in the manifest, which lets you specify allowlists and blocklists for groups:

```yaml
    manifest:
      # ...
      groups:
          - some-group
          - another-group
      no-groups:
          - not-this-one
```

By default, `west update` will no longer update any project whose group is blocked (if there is a `no-groups` blocklist), or not explicitly allowed (if there is a `groups` allowlist). Allowlists override blocklists: if a project is blocked and allowed, it will be updated. You can also configure allowlists and blocklists in the local configuration file, via new `manifest.groups` and `manifest.no-groups` configuration options.

In order to allow for one-time overrides, let the user specify additional groups to add to the allowlist via a new "--groups" argument to `west update`, and additional groups to add to the blocklist via "--no-groups".

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -376,35 +376,38 @@ class List(_ProjectCommand):
         default_fmt = '{name:12} {path:28} {revision:40} {url}'
         parser = self._parser(
             parser_adder,
-            epilog=textwrap.dedent(f'''\
-            FORMAT STRINGS
-            --------------
+            epilog=f'''\
+{ACTIVE_PROJECTS_HELP}
 
-            Projects are listed using a Python 3 format string. Arguments
-            to the format string are accessed by name.
+FORMAT STRINGS
+--------------
 
-            The default format string is:
+Projects are listed using a Python 3 format string. Arguments
+to the format string are accessed by name.
 
-            "{default_fmt}"
+The default format string is:
 
-            The following arguments are available:
+"{default_fmt}"
 
-            - name: project name in the manifest
-            - url: full remote URL as specified by the manifest
-            - path: the relative path to the project from the top level,
-              as specified in the manifest where applicable
-            - abspath: absolute and normalized path to the project
-            - posixpath: like abspath, but in posix style, that is, with '/'
-              as the separator character instead of '\\'
-            - revision: project's revision as it appears in the manifest
-            - sha: project's revision as a SHA. Note that use of this requires
-              that the project has been cloned.
-            - cloned: "cloned" if the project has been cloned, "not-cloned"
-              otherwise
-            - clone_depth: project clone depth if specified, "None" otherwise
-            '''))
+The following arguments are available:
+
+- name: project name in the manifest
+- url: full remote URL as specified by the manifest
+- path: the relative path to the project from the top level,
+  as specified in the manifest where applicable
+- abspath: absolute and normalized path to the project
+- posixpath: like abspath, but in posix style, that is, with '/'
+  as the separator character instead of '\\'
+- revision: project's revision as it appears in the manifest
+- sha: project's revision as a SHA. Note that use of this requires
+  that the project has been cloned.
+- cloned: "cloned" if the project has been cloned, "not-cloned"
+  otherwise
+- clone_depth: project clone depth if specified, "None" otherwise
+- groups: project groups, as a comma-separated list
+''')
         parser.add_argument('-a', '--all', action='store_true',
-                            help='ignored for backwards compatibility'),
+                            help='include inactive projects'),
         parser.add_argument('--manifest-path-from-yaml', action='store_true',
                             help='''print the manifest repository's path
                             according to the manifest file YAML, which may
@@ -414,7 +417,9 @@ class List(_ProjectCommand):
                             help='''format string to use to list each
                             project; see FORMAT STRINGS below.''')
 
-        self._add_projects_arg(parser)
+        parser.add_argument('projects', metavar='PROJECT', nargs='*',
+                            help='''projects (by name or path) to operate on;
+                            see ACTIVE PROJECTS below''')
 
         return parser
 
@@ -441,6 +446,13 @@ class List(_ProjectCommand):
         self._setup_logging(args)
 
         for project in self._projects(args.projects):
+            # Skip inactive projects unless the user said
+            # --all or named some projects explicitly.
+            if not (args.all or args.projects or
+                    self.manifest.is_active(project)):
+                log.dbg(f'{project.name}: skipping, inactive')
+                continue
+
             # Spelling out the format keys explicitly here gives us
             # future-proofing if the internal Project representation
             # ever changes.
@@ -471,7 +483,8 @@ class List(_ProjectCommand):
                     revision=project.revision or 'N/A',
                     clone_depth=project.clone_depth or "None",
                     cloned=delay(cloned_thunk, project),
-                    sha=delay(sha_thunk, project))
+                    sha=delay(sha_thunk, project),
+                    groups=','.join(project.groups))
             except KeyError as e:
                 # The raised KeyError seems to just put the first
                 # invalid argument in the args tuple, regardless of
@@ -1420,6 +1433,24 @@ WEST_DIR = '.west'
 MANIFEST_URL_DEFAULT = 'https://github.com/zephyrproject-rtos/zephyr'
 # Default revision to check out of the manifest repository.
 MANIFEST_REV_DEFAULT = 'master'
+
+#
+# Other shared globals.
+#
+
+ACTIVE_PROJECTS_HELP = '''\
+ACTIVE PROJECTS
+---------------
+
+Default output is limited to "active" projects as determined by the:
+
+- "groups" manifest file section
+- "manifest.groups" local configuration option in .west/config
+
+To include inactive projects as well, use "--all" or give an explicit
+list of projects (by name or path). See the west documentation for
+more details on active projects.
+'''
 
 #
 # Helper class for creating format string keys that are expensive or
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -365,7 +365,7 @@ class List(_ProjectCommand):
     def __init__(self):
         super().__init__(
             'list',
-            'print information about projects in the west manifest',
+            'print information about projects',
             textwrap.dedent('''\
             Print information about projects in the west manifest,
             using format strings.'''))
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -862,11 +862,6 @@ class Update(_ProjectCommand):
         else:
             return 'smart'
 
-    def fetch_missing_imports(self, args):
-        self.fs = 'always'      # just to be safe -- TODO needed?
-        self.manifest = Manifest.from_file(topdir=self.topdir,
-                                           importer=self.update_importer)
-
     def update(self, project):
         if self.args.stats:
             stats = dict()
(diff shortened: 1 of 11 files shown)


mbolivar-nordic:
@carlescufi the basics are working for me with some simple manual `west update` testing, so I'm pushing this to open it up for feedback on the CLI while I'm improving test coverage.

I guess we'll probably want to add some configuration options so users can set a workspace default for the new west command line options.

tejlmand (changes requested):
still needs some discussions.
As example, what should be presented on `west list` which is used in build system for inclusion of Zephyr modules ?

A user could have done `west update --group <some-group>` just to see what's there, and then later simply do `west update` and no longer care for group.

How should build system handle `<some-group>` if it is also a Zephyr module.

mbolivar-nordic:
Pushed an update with the following changes:

- `manifest: groups:` and `manifest: no-groups:` instead of `manifest: update-groups: ...`, i.e. the manifest file allowlists and blocklists are at top level now
- the manifest will consult the `manifest.groups` and `manifest.no-groups` **local** configuration file options to extend these lists if there is a workspace
- renamed `should_update` to `is_updatable`
- `west list` no longer prints projects that aren't updatable unless you give `--all`
- a stale comment in conftest.py was found and removed in a new cleanup commit

src/west/manifest.py
```diff
@@ -1422,6 +1483,85 @@ def _load(self, manifest: Dict[str, Any],
...
+
+        if cfg_groups or manifest_groups:
+            cfg_groups = cfg_groups or []
+            manifest_groups = manifest_groups or []
+            self.groups = cfg_groups + manifest_groups
+
+        if cfg_no_groups or manifest_no_groups:
+            cfg_no_groups = cfg_no_groups or []
+            manifest_no_groups = manifest_no_groups or []
+            self.no_groups = cfg_no_groups + manifest_no_groups
```
tejlmand, line 1505: So if manifest contains:
```
    groups:
        - groupA
        - groupB
      no-groups:
        - groupC
```
and user configures:
```
[manifest]
#  I want group C
    groups = groupC
#  I don't want group A and B
    no-groups = groupA,groupB
```
then the `self.groups` and `self.no_groups` will both contain:
```
['groupA', 'groupB', 'groupC']
```

I don't think this is the intention.

When fixing this, I think user defined group configs should have precedence over manifest defaults.

tejlmand (changes requested):

tejlmand:
@mbolivar-nordic Thanks for the update.

src/west/app/project.py
```diff
@@ -401,9 +401,24 @@ def do_add_parser(self, parser_adder):
             - cloned: "cloned" if the project has been cloned, "not-cloned"
               otherwise
             - clone_depth: project clone depth if specified, "None" otherwise
+            - groups: project groups, as a comma-separated list
+
+            INCLUDED PROJECTS
+            -----------------
```
tejlmand, line 407: I have the suspicion that more users will execute `west help list` to find out why `<someproject>` is not in the list of printed project, compared to users wanting to know print formatting.

So consider if this section should move above `FORMAT STRINGS` section.

src/west/app/project.py
```diff
@@ -401,9 +401,24 @@ def do_add_parser(self, parser_adder):
...
+
+            By default, output does not include information for any project
+            that "west update" would not update by default due to:
+
+            - the "groups" / "no-groups" manifest file keys
+            - the "manifest.groups" / "manifest.no-groups" local configuration
+              options in .west/config
+
+            To override this, give an explicit list of projects (by name or
+            path), or use "west list --all".
```
tejlmand, line 417: Guessing that some users might come here to understand why `project-off` is not printed with `west list`, then maybe we should give a small hint, like:
> Additional groups can be added to manifest.groups / manifest.no-groups using west config.

tejlmand (commented):
A couple of small thoughts.
Feel free to ignore if you disagree.

src/west/manifest.py
mbolivar-nordic: > I don't think this is the intention.

It is the intention; what's the issue?

src/west/app/project.py
mbolivar-nordic: It's right there in the bullet list, isn't it?

```
            - the "manifest.groups" / "manifest.no-groups" local configuration
              options in .west/config
```

src/west/app/project.py
mbolivar-nordic: > So consider if this section should move above `FORMAT STRINGS` section.

I've done this as part of a more general rework related to the other options.

src/west/app/project.py
tejlmand: yes, I did notice, I was more thinking that a reference to `west config` would be nice, so that users would be hinted on how to update `.west/config`.

The bullet only talks about `.west/config`, thus assuming the user already has knowledge about `west config`.

src/west/manifest.py
tejlmand: I have made a small manifest example with groups here:
https://github.com/tejlmand/zephyr/tree/testing/west_454_groups

The manifest projects are divided into three groups: `groupA`, `groupB`, `groupC`.

`groupA` and `groupB` has a manifest setup, and `groupC` is just default, as:
https://github.com/tejlmand/zephyr/blob/7f42bb41c77d4d6d48930b211ca65ca8da983a65/west.yml#L17-L21
```
manifest:
  groups:
    - groupA
  no-groups:
    - groupB
```

doing a `west list` I get:
```
$ west list
manifest     zephyr                       HEAD                                     N/A
cmsis        modules/hal/cmsis            542b2296e6d515b265e25c6b7208e8fea3014f90 https://github.com/zephyrproject-rtos/cmsis
```
just as expected.

Now, I don't want `groupA` listed (or updated), so let me disable that:
```
$ west config manifest.no-groups groupA
$ west list
manifest     zephyr                       HEAD                                     N/A
cmsis        modules/hal/cmsis            542b2296e6d515b265e25c6b7208e8fea3014f90 https://github.com/zephyrproject-rtos/cmsis
```
Hooh, it's still there. As a user I want to be able to overrule the manifest file and disable a group.

Let me see if I can enable `groupB`.
```
$ west config manifest.groups groupB
$ west list
manifest     zephyr                       HEAD                                     N/A
cmsis        modules/hal/cmsis            542b2296e6d515b265e25c6b7208e8fea3014f90 https://github.com/zephyrproject-rtos/cmsis
hal_atmel    modules/hal/atmel            9f0a699a2f0253dc9688211ef696daa7b8fadb75 https://github.com/zephyrproject-rtos/hal_atmel
```
working as expected.

(Same happens for `west update` of course)

Also:
```
$ west help update
...
advanced options:
  ....
  -G GROUPS, --no-groups GROUPS
                        update as if GROUPS are in manifest.no-groups
```
Cool, let me try this:
```
$ west update -G groupA
=== updating cmsis (modules/hal/cmsis):
HEAD is now at 542b229 DSP: Integrate CMSIS-DSP 1.8.0 (CMSIS 5.7.0)
```
Damn, `groupA` is still being updated.

Note, it's not only the commented line with this, but both those lines:
https://github.com/mbolivar-nordic/west/blob/project-groups/src/west/manifest.py#L1506
https://github.com/mbolivar-nordic/west/blob/project-groups/src/west/manifest.py#L1511

I believe the user should be able to overrule settings from `west.yml` or `.west/config`

carlescufi: >Hooh, it's still there. As a user I want to be able to overrule the manifest file and disable a group.

+1 to that.

>Damn, groupA is still being updated.

That's a bug IMHO. `-G` should take precedence to configs and manifest, since the user is overriding explicitly.

src/west/app/project.py
mbolivar-nordic: There's only so much we can do in the --help output. We have pretty good documentation on configuration linked to right from the front page of the west docs:

https://docs.zephyrproject.org/latest/guides/west/index.html

Do we really need to repeat it all here?

tejlmand: not necessarily

tejlmand (changes requested):
Tested with:
https://github.com/tejlmand/zephyr/tree/testing/west_454_groups

where west.yml contains:
```
manifest:
  groups:
    - groupA
    - -groupB
    - -groupC
```
works.
Only `groupA` is listed:
```
$ west list
manifest     zephyr                       HEAD                                     N/A
cmsis        modules/hal/cmsis            542b2296e6d515b265e25c6b7208e8fea3014f90 https://github.com/zephyrproject-rtos/cmsis
```

Enabling `groupB` also works:
```
$ west config manifest.groups "groupB"
$ west list 
manifest     zephyr                       HEAD                                     N/A
cmsis        modules/hal/cmsis            542b2296e6d515b265e25c6b7208e8fea3014f90 https://github.com/zephyrproject-rtos/cmsis
hal_atmel    modules/hal/atmel            9f0a699a2f0253dc9688211ef696daa7b8fadb75 https://github.com/zephyrproject-rtos/hal_atmel
```

but enabling multiple groups fails:
```
$ west config manifest.groups "groupB,groupC"
$ west list 
manifest     zephyr                       HEAD                                     N/A
cmsis        modules/hal/cmsis            542b2296e6d515b265e25c6b7208e8fea3014f90 https://github.com/zephyrproject-rtos/cmsis
```

my `.west/config`:
```
$ west config -l
manifest.path=zephyr
manifest.groups="groupB,groupC"
zephyr.base=zephyr
```

mbolivar-nordic:
> manifest.groups="groupB,groupC"

That's the problem. If you manually edit it to

> manifest.groups=groupB,groupC

Then it works.

The root cause seems to be that we are using the standard library `configparser` to *read* values, but the third party `configobj` library to *write* them. They aren't working the same way: `configparser` returns the value with the quotes, which don't get stripped, and end up unblocking the groups `"groupB` and `groupC"`.

mbolivar-nordic:
@tejlmand I'll need to think about the right way to fix this bug. This ties in to #149 because the current `west.configuration` API is stateful and awkward to use anyway.

mbolivar-nordic:
Fixed it by dropping the configobj dependency completely. See commit log for further justification and details.

mbolivar-nordic:
Now changed to OR together the project group allow status instead of ANDing it together as discussed offline.

src/west/manifest.py
```diff
@@ -1317,6 +1391,45 @@ def as_frozen_yaml(self, **kwargs) -> str:
...
+            return True
+
+        if extra_groups:
+            blocked_groups = set(self._blocked_groups)
+            _update_blocked_groups(blocked_groups, extra_groups)
+        else:
+            blocked_groups = self._blocked_groups
+
+        # A project with groups is active if and only if any one of its
+        # groups are blocked.
```
tejlmand, line 1430: Did you mean, `groups are not blocked.` ?

tejlmand (approved):
Looks good to me.
Thanks for the good effort.

btw. I got some commented out lines deleted in my `.west/config` that I had kept for future reference when testing this :dizzy_face:

src/west/manifest.py
mbolivar-nordic: > Did you mean, `groups are not blocked.` ?

Yes. I just deleted the comment since the docstring has all the details anyway. Thanks.

Issue #365, opened 2020-01-31
Support git submodules


mbolivar:
This issue tracks adding support for projects with git submodules to west.

The overall goal is to clone and update submodules on `west update` in addition to the projects that contain them.

finikorg:
We have now project we want to add as a Zephyr module. It has submodule. At the moment we are investigating what is the best way of handling it. Support git submodules in west would help.

carlescufi:
I suggest something like this in the manifest schema:

```
  projects:
    required: true
    type: seq
    sequence:
      - type: map
        mapping:
          # Project's unique name.
          name:
            required: true
            type: str
          submodules:
             required: false
             type: any
```

and then do something similar to what we have in the `import:` field so you could do:

```
submodules: true
```
or
```
submodules:
    name-allowlist:
        - submodule_1
        - submodule_2
```
or
```
submodules:
    name-denylist:
        - submodule_1
        - submodule_2
```

mbolivar-nordic:
I think I would prefer to start with a `submodules: true` (which defaults to `false`) to update the submodules as well with `git submodule update` or so, before we go crazy with allow or blocklists.

marc-hb:
@finikorg would you have time to give #459 a try?

tejlmand:
> I think I would prefer to start with a `submodules: true` (which defaults to `false`) to update the submodules as well with `git submodule update` or so, before we go crazy with allow or blocklists.

There is already a use-case where `west` is used to clone a single repo, which also happens to be a one submodule out of two.
https://github.com/nrfconnect/sdk-nrf/blob/master/west.yml#L107-L114
```
    - name: cmock
      path: test/cmock
      revision: c243b9a7a7b3c471023193992b46cf1bd1910450
      remote: throwtheswitch
    - name: unity
      path: test/cmock/vendor/unity
      revision: 031f3bbe45f8adf504ca3d13e6f093869920b091
      remote: throwtheswitch
```
where the `cmock` repo actually contains two submodules:
- `c_exception`
- `unity`

so I think there is a use-case for specific list of submodules to clone, in addition to all. (when `true` is given)

PR #459, opened 2020-12-03
update: add support for git submodules in west update.

kkasperczyk-no:
The submodules were added to the manifest in order to enable updating projects submodules when performing update operation.

Description of the `west update` command behaviour regarding submodules:
* Submodules can be defined in the west manifest file for the specific project, by adding submodules: field.
* Submodules field is not required for every project and should be added only if it is desired.
* Submodules field may have boolean values (true/false) or be the list of specific submodules.
* Assigning false value to the submodules field means that submodules are not used for given project. Example presented below:
    - name: example_project
      submodules: false
* Assigning true value to the submodules field means that all submodules attached by git to the project repository will be updated. Example presented below:
    projects:
    - name: example_project
      submodules: true
* Assigning true value to the submodules field means that all submodules attached by git to the project repository will be updated. Example presented below:
    projects:
    - name: example_project
      submodules: true
* Assigning list of submodules to the submodules field means that only those submodules present on the list will be updated (no matter that there might be any other submodules attached in git). Every element of the submodules list should have the name and the path fields (which is an absolute path to the directory where submodule should be updated). Example presented below:
    - name: example_project
      submodules:
        - name: first_submodule
          path: first_submodule_dir/repo
        - name: second_submodule
          path: second_submodule_dir/repo
* Submodules for each project are defined independently, so in some cases they might have assigned true/false values, lists or not be defined at all.
* When calling `west update` command, submodules for each project are detected and updated with the `recursive` option.
* By default submodules are updated using `--checkout` strategy, however it is possible to update them using `--rebase` strategy, by calling `west update -r` command.

What `west update` command does not do with the submodules:
* Updating submodules does not support `west update -k` option. In case of using such command, `-k` option is ignored and submodules are updated using default `--checkout` strategy.
* Assigning true or false value to the submodules field does not apply to the all projects and their submodules, but only to the all submodules of a specific project for which submodules field was defined.

Fixes: #365

Signed-off-by: Kamil Kasperczyk <kamil.kasperczyk@nordicsemi.no>

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -924,6 +924,28 @@ class Update(_ProjectCommand):
         else:
             return 'smart'
 
+    def update_submodules(self, project):
+        # Updates given project submodules by using
+        # 'git submodule update --init --checkout --recursive' command
+        # from the project.path location.
+        if not project.submodules:
+            return
+
+        submodules = project.submodules
+        call_location = os.path.join(self.topdir, project.path)
+        submodules_update_strategy = ('--rebase' if self.args.rebase
+                                      else '--checkout')
+        # For the list type, update given list of submodules.
+        if isinstance(submodules, list):
+            for submodule in submodules:
+                project.git(['-C', call_location, 'submodule', 'update',
+                             '--init', submodules_update_strategy,
+                             '--recursive', submodule.path])
+        # For the bool type, update all project submodules
+        elif isinstance(submodules, bool):
+            project.git(['-C', call_location, 'submodule', 'update', '--init',
+                         submodules_update_strategy, '--recursive'])
+
     def update(self, project):
         if self.args.stats:
             stats = dict()
@@ -988,6 +1010,9 @@ class Update(_ProjectCommand):
             if take_stats:
                 stats['checkout new manifest-rev'] = perf_counter() - start
             _post_checkout_help(project, current_branch, sha, is_ancestor)
+
+        # Update project submodules, if it has any.
+        self.update_submodules(project)
 
         # Print performance statistics.
         if take_stats:
--- a/src/west/manifest-schema.yml
+++ b/src/west/manifest-schema.yml
@@ -103,6 +103,10 @@ mapping:
           path:
             required: false
             type: str
+          # Submodules to checkout within the project
+          submodules:
+            required: false
+            type: any
           # Limits the clone depth of the project if given.
           clone-depth:
             required: false
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -87,6 +87,14 @@ ManifestDataType = Union[str, Dict]
 
 _logger = logging.getLogger(__name__)
 
+# Type for the submodule value passed through the manifest file.
+class Submodule(NamedTuple):
+    name: str
+    path: str
+
+# Submodules might be the list of values, bool or None.
+SubmodulesType = Optional[Union[List[Submodule], bool]]
+
 # Manifest locating, parsing, loading, etc.
 
 class _defaults(NamedTuple):
@@ -577,6 +585,7 @@ class Project:
     def __init__(self, name: str, url: str,
                  revision: Optional[str] = None,
                  path: Optional[PathType] = None,
+                 submodules: Optional[SubmodulesType] = None,
                  clone_depth: Optional[int] = None,
                  west_commands: Optional[WestCommandsType] = None,
                  topdir: Optional[PathType] = None,
@@ -591,6 +600,7 @@ class Project:
         :param url: fetch URL
         :param revision: fetch revision
         :param path: path (relative to topdir), or None for *name*
+        :param submodules: submodules to pull within the project
         :param clone_depth: depth to use for initial clone
         :param west_commands: path to a west commands specification YAML
             file in the project, relative to its base directory,
@@ -604,6 +614,7 @@ class Project:
 
         self.name = name
         self.url = url
+        self.submodules = submodules
         self.revision = revision or _DEFAULT_REV
         self.clone_depth = clone_depth
         self.path = os.fspath(path or name)
@@ -911,6 +922,7 @@ class ManifestProject(Project):
 
         # Pretending that this is a Project, even though it's not (#327)
         self.url: str = ''
+        self.submodules = False
         self.revision: str = 'HEAD'
         self.clone_depth: Optional[int] = None
         self.groups = []
@@ -1074,6 +1086,16 @@ class Manifest:
         '''
         kwargs.update({'source_data': source_data})
         return Manifest(**kwargs)
+
+    @staticmethod
+    def is_submodule_ok(value: Any) -> bool:
+        if isinstance(value, dict):
+            # Check whether value contain Submodule fields of proper types.
+            if 'name' in value.keys() and 'path' in value.keys():
+                if isinstance(value['name'], str) and isinstance(value['path'],
+                                                                 str):
+                    return True
+        return False
 
     def __init__(self, source_file: Optional[PathType] = None,
                  source_data: Optional[ManifestDataType] = None,
@@ -1308,6 +1330,42 @@ class Manifest:
         if unknown or (only_cloned and uncloned):
             raise ValueError(unknown, uncloned)
         return ret
+
+    def _get_submodules(self, submodules) -> SubmodulesType:
+        '''Gets a list of `Submodules` objects or boolean from the manifest
+        *submodules*.
+
+        For submodules of list type, method checks format of all elements
+        and converts dictionaries into the Submodule objects.
+
+        For submodules set to None or submodules of bool type, method
+        does nothing, as both boolean values are valid (True - means
+        that all project submodules should be considered and False - means
+        that are not any submodules).
+
+        :param submodules: content of the manifest submodules value.
+        '''
+        self.submodules = submodules
+        self.submodules_type = type(self.submodules)
+        # For the boolean type or submodules set to None just do nothing.
+        if self.submodules is None or self.submodules_type == bool:
+            return self.submodules
+
+        # For the list type, check whether each element has proper
+        # structure and convert it to Submodules object.
+        if self.submodules_type == list:
+            for index in range(len(self.submodules)):
+                if self.is_submodule_ok(self.submodules[index]):
+                    self.submodules[index] =  \
+                        Submodule(self.submodules[index]['name'],
+                                  self.submodules[index]['path'])
+                else:
+                    self._malformed(f'Invalid submodule structure \
+                        {self.submodules[index]}')
+        else:
+            self._malformed(f'Invalid submodules {self.submodules} '
+                            f'type: {self.submodules_type}')
+        return self.submodules
 
     def _as_dict_helper(
             self, pdict: Optional[Callable[[Project], Dict]] = None) \
@@ -1846,8 +1904,11 @@ class Manifest:
             self._malformed(
                 f'project {name}: "groups" cannot be combined with "import"')
 
+        submodules = self._get_submodules(pd.get('submodules'))
+
         ret = Project(name, url, pd.get('revision', defaults.revision),
-                      path, clone_depth=pd.get('clone-depth'),
+                      path, submodules=submodules,
+                      clone_depth=pd.get('clone-depth'),
                       west_commands=pd.get('west-commands'),
                       topdir=self.topdir, remote_name=remote,
                       groups=groups)
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -477,6 +477,390 @@ def test_update_some_with_imports(repos_tmpdir):
     manifest = Manifest.from_file(topdir=ws)
     assert manifest.get_projects(['Kconfiglib'])[0].is_cloned()
 
+def test_update_submodules_list(repos_tmpdir):
+    # The west update command should not only update projects,
+    # but also its submodules. Test uses two pairs of project
+    # and submodule checking out submodule in the default
+    # and custom location to verify both cases.
+
+    # Repositories paths
+    remotes = repos_tmpdir / 'repos'
+    # tagged_repo is zephyr's submodule and will be updated
+    # in the default location (i.e. zephyr/tagged_repo).
+    zephyr = remotes / 'zephyr'
+    tagged_repo = remotes / 'tagged_repo'
+    # Kconfiglib is net_tools's submodule and will be updated
+    # in the custom location (i.e. net-tools/third_parties/Kconfiglib).
+    net_tools = remotes / 'net-tools'
+    kconfiglib = remotes / 'Kconfiglib'
+    kconfiglib_submodule = 'third_parties/Kconfiglib'
+
+    # Creating west workspace and manifest repository.
+    ws = repos_tmpdir / 'ws'
+    create_workspace(ws, and_git=True)
+    manifest_repo = ws / 'mp'
+    create_repo(manifest_repo)
+
+    # Commit west.yml describing projects and submodules dependencies.
+    add_commit(manifest_repo, 'manifest repo commit',
+               files={'west.yml':
+                      f'''
+                      manifest:
+                        projects:
+                        - name: zephyr
+                          url: {zephyr}
+                          submodules:
+                            - name: tagged_repo
+                              path: tagged_repo
+                        - name: net-tools
+                          url: {net_tools}
+                          submodules:
+                            - name: Kconfiglib
+                              path: {kconfiglib_submodule}
+                       '''})
+    cmd(f'init -l {manifest_repo}')
+
+    # Make tagged_repo to be zephyr project submodule.
+    subprocess.check_call(
+        [GIT, 'submodule', 'add', str(tagged_repo), 'tagged_repo'], cwd=zephyr)
+    # Commit changes to the zephyr repo.
+    add_commit(zephyr, 'zephyr submodule change commit')
+
+    # Make Kconfiglib to be net-tools project submodule.
+    subprocess.check_call(
+        [GIT, 'submodule', 'add', str(kconfiglib), kconfiglib_submodule],
+        cwd=net_tools)
+    # Commit changes to the net-tools repo.
+    add_commit(net_tools, 'net-tools submodule change commit')
+
+    # Get parsed data from the manifest.
+    manifest = Manifest.from_file(topdir=ws, import_flags=MIF.IGNORE_PROJECTS)
+    projects = manifest.get_projects(['zephyr', 'net-tools'])
+    zephyr_project = projects[0]
+    net_tools_project = projects[1]
+
+    # Verify if tagged_repo submodule data are correct.
+    assert zephyr_project.submodules
+    assert zephyr_project.submodules[0].name == 'tagged_repo'
+    assert zephyr_project.submodules[0].path == 'tagged_repo'
+    # Verify if Kconfiglib submodule data are correct.
+    assert net_tools_project.submodules
+    assert net_tools_project.submodules[0].name == 'Kconfiglib'
+    assert net_tools_project.submodules[0].path == kconfiglib_submodule
+
+    # Check if projects repos are cloned - should not be.
+    assert not zephyr_project.is_cloned()
+    assert not net_tools_project.is_cloned()
+
+    # Update only zephyr project.
+    cmd('update zephyr', cwd=ws)
+
+    # Verify if only zephyr project was cloned.
+    assert zephyr_project.is_cloned()
+    assert not net_tools_project.is_cloned()
+
+    # Verify if tagged-repo submodule was also cloned
+    res = zephyr_project.git('rev-parse --show-cdup', check=False,
+                             cwd=os.path.join(ws, 'zephyr', 'tagged_repo'),
+                             capture_stderr=True, capture_stdout=True)
+    assert not (res.returncode or res.stdout.strip())
+
+    # Update all projects
+    cmd('update', cwd=ws)
+
+    # Verify if both projects were cloned
+    assert zephyr_project.is_cloned()
+    assert net_tools_project.is_cloned()
+
+    # Verify if Kconfiglib submodule was also cloned
+    res = net_tools_project.git('rev-parse --show-cdup', check=False,
+                                cwd=os.path.join(ws, 'net-tools',
+                                                 kconfiglib_submodule),
+                                capture_stderr=True, capture_stdout=True)
+    assert not (res.returncode or res.stdout.strip())
+
+def test_update_all_submodules(repos_tmpdir):
+    # The west update command should not only update projects,
+    # but also its submodules. Test verifies whether setting submodules
+    # value to boolean True results in updating all project submodules
+    # (even without specifying them implicitly as a list). Moreover it checks
+    # if submodules are updated recursively.
+
+    # Repositories paths
+    remotes = repos_tmpdir / 'repos'
+    # tagged_repo and net_tools are zephyr's submodules
+    zephyr = remotes / 'zephyr'
+    tagged_repo = remotes / 'tagged_repo'
+    net_tools = remotes / 'net-tools'
+    # Kconfiglib is net-tools submodule
+    kconfiglib = remotes / 'Kconfiglib'
+
+    # Creating west workspace and manifest repository.
+    ws = repos_tmpdir / 'ws'
+    create_workspace(ws, and_git=True)
+    manifest_repo = ws / 'mp'
+    create_repo(manifest_repo)
+
+    # Commit west.yml describing projects and submodules dependencies.
+    add_commit(manifest_repo, 'manifest repo commit',
+               files={'west.yml':
+                      f'''
+                      manifest:
+                        projects:
+                        - name: zephyr
+                          url: {zephyr}
+                          submodules: true
+                       '''})
+    cmd(f'init -l {manifest_repo}')
+
+    # Make tagged_repo to be zephyr project submodule.
+    subprocess.check_call(
+        [GIT, 'submodule', 'add', str(tagged_repo), 'tagged_repo'], cwd=zephyr)
+    # Commit changes to the zephyr repo.
+    add_commit(zephyr, 'zephyr submodule tagged_repo commit')
+
+    # Make Kconfiglib to be net_tools submodule.
+    subprocess.check_call(
+        [GIT, 'submodule', 'add', str(kconfiglib), 'Kconfiglib'],
+        cwd=net_tools)
+    # Commit changes to the net_tools repo.
+    add_commit(net_tools, 'net_tools submodule Kconfiglib commit')
+
+    # Make net_tools to be zephyr project submodule.
+    subprocess.check_call(
+        [GIT, 'submodule', 'add', str(net_tools), 'net-tools'], cwd=zephyr)
+    # Commit changes to the zephyr repo.
+    add_commit(zephyr, 'zephyr submodule net-tools commit')
+
+    # Get parsed data from the manifest.
+    manifest = Manifest.from_file(topdir=ws, import_flags=MIF.IGNORE_PROJECTS)
+    projects = manifest.get_projects(['zephyr'])
+    zephyr_project = projects[0]
+
+    # Verify if zephyr has submodules.
+    assert zephyr_project.submodules
+
+    # Check if project repo is cloned - should not be.
+    assert not zephyr_project.is_cloned()
+
+    # Update zephyr project.
+    cmd('update zephyr', cwd=ws)
+
+    # Verify if zephyr project was cloned.
+    assert zephyr_project.is_cloned()
+
+    # Verify if tagged-repo submodule was also cloned.
+    res = zephyr_project.git('rev-parse --show-cdup', check=False,
+                             cwd=os.path.join(ws, 'zephyr', 'tagged_repo'),
+                             capture_stderr=True, capture_stdout=True)
+    assert not (res.returncode or res.stdout.strip())
+
+    # Verify if net-tools submodule was also cloned.
+    res = zephyr_project.git('rev-parse --show-cdup', check=False,
+                             cwd=os.path.join(ws, 'zephyr', 'net-tools'),
+                             capture_stderr=True, capture_stdout=True)
+    assert not (res.returncode or res.stdout.strip())
+
+    # Verify if Kconfiglib submodule was also cloned, as a result of recursive
+    # update.
+    res = zephyr_project.git('rev-parse --show-cdup', check=False,
+                             cwd=os.path.join(ws, 'zephyr', 'net-tools',
+                                              'Kconfiglib'),
+                             capture_stderr=True, capture_stdout=True)
+    assert not (res.returncode or res.stdout.strip())
+
+def test_update_no_submodules(repos_tmpdir):
+    # Test verifies whether setting submodules value to boolean False does not
+    # result in updating project submodules.
+
+    # Repositories paths
+    remotes = repos_tmpdir / 'repos'
+    # tagged_repo and net_tools are zephyr's submodules
+    zephyr = remotes / 'zephyr'
+    tagged_repo = remotes / 'tagged_repo'
+    net_tools = remotes / 'net-tools'
+
+    # Creating west workspace and manifest repository.
+    ws = repos_tmpdir / 'ws'
+    create_workspace(ws, and_git=True)
+    manifest_repo = ws / 'mp'
+    create_repo(manifest_repo)
+
+    # Commit west.yml describing projects and submodules dependencies.
+    add_commit(manifest_repo, 'manifest repo commit',
+               files={'west.yml':
+                      f'''
+                      manifest:
+                        projects:
+                        - name: zephyr
+                          url: {zephyr}
+                          submodules: false
+                       '''})
+    cmd(f'init -l {manifest_repo}')
+
+    # Make tagged_repo to be zephyr project submodule.
+    subprocess.check_call(
+        [GIT, 'submodule', 'add', str(tagged_repo), 'tagged_repo'], cwd=zephyr)
+    # Commit changes to the zephyr repo.
+    add_commit(zephyr, 'zephyr submodule tagged_repo commit')
+
+    # Make net_tools to be zephyr project submodule.
+    subprocess.check_call(
+        [GIT, 'submodule', 'add', str(net_tools), 'net-tools'], cwd=zephyr)
+    # Commit changes to the zephyr repo.
+    add_commit(zephyr, 'zephyr submodule net-tools commit')
+
+    # Get parsed data from the manifest.
+    manifest = Manifest.from_file(topdir=ws, import_flags=MIF.IGNORE_PROJECTS)
+    projects = manifest.get_projects(['zephyr'])
+    zephyr_project = projects[0]
+
+    # Check if project repo is cloned - should not be.
+    assert not zephyr_project.is_cloned()
+
+    # Update zephyr project.
+    cmd('update zephyr', cwd=ws)
+
+    # Verify if zephyr project was cloned.
+    assert zephyr_project.is_cloned()
+
+    # Verify if tagged-repo submodule was also cloned (should not be).
+    res = zephyr_project.git('rev-parse --show-cdup', check=False,
+                             cwd=os.path.join(ws, 'zephyr', 'tagged_repo'),
+                             capture_stderr=True, capture_stdout=True)
+    assert (res.returncode or res.stdout.strip())
+
+    # Verify if net-tools submodule was also cloned (should not be).
+    res = zephyr_project.git('rev-parse --show-cdup', check=False,
+                             cwd=os.path.join(ws, 'zephyr', 'net-tools'),
+                             capture_stderr=True, capture_stdout=True)
+    assert (res.returncode or res.stdout.strip())
+
+def test_update_submodules_strategy(repos_tmpdir):
+    # The west update command is able to update submodules using default
+    # checkout strategy or rebase strategy, selected by adding -r argument
+    # to the invoked command. Test verifies if both strategies are working
+    # properly when updating submodules.
+
+    # Repositories paths
+    remotes = repos_tmpdir / 'repos'
+    # tagged_repo is zephyr's submodule
+    zephyr = remotes / 'zephyr'
+    tagged_repo = remotes / 'tagged_repo'
+    # Kconfiglib is net_tools's submodule
+    net_tools = remotes / 'net-tools'
+    kconfiglib = remotes / 'Kconfiglib'
+
+    # Creating west workspace and manifest repository.
+    ws = repos_tmpdir / 'ws'
+    create_workspace(ws, and_git=True)
+    manifest_repo = ws / 'mp'
+    create_repo(manifest_repo)
+
+    tagged_repo_dst_dir = os.path.join(ws, 'zephyr', 'tagged_repo')
+    kconfiglib_dst_dir = os.path.join(ws, 'net-tools', 'Kconfiglib')
+
+    # Commit west.yml describing projects and submodules dependencies.
+    add_commit(manifest_repo, 'manifest repo commit',
+               files={'west.yml':
+                      f'''
+                      manifest:
+                        projects:
+                        - name: zephyr
+                          url: {zephyr}
+                          submodules:
+                            - name: tagged_repo
+                              path: tagged_repo
+                        - name: net-tools
+                          url: {net_tools}
+                          submodules:
+                            - name: Kconfiglib
+                              path: Kconfiglib
+                       '''})
+    cmd(f'init -l {manifest_repo}')
+
+    # Make tagged_repo to be zephyr project submodule.
+    subprocess.check_call(
+        [GIT, 'submodule', 'add', str(tagged_repo), 'tagged_repo'], cwd=zephyr)
+    # Commit changes to the zephyr repo.
+    add_commit(zephyr, 'zephyr submodule change commit')
+
+    # Make Kconfiglib to be net-tools project submodule.
+    subprocess.check_call(
+        [GIT, 'submodule', 'add', str(kconfiglib), 'Kconfiglib'],
+        cwd=net_tools)
+    # Commit changes to the net-tools repo.
+    add_commit(net_tools, 'net-tools submodule change commit')
+
+    # Get parsed data from the manifest.
+    manifest = Manifest.from_file(topdir=ws, import_flags=MIF.IGNORE_PROJECTS)
+    projects = manifest.get_projects(['zephyr', 'net-tools'])
+    zephyr_project = projects[0]
+    net_tools_project = projects[1]
+
+    # Check if projects repos are cloned - should not be.
+    assert not zephyr_project.is_cloned()
+    assert not net_tools_project.is_cloned()
+
+    # Update only zephyr project using checkout strategy (selected by default).
+    cmd('update zephyr', cwd=ws)
+
+    # Verify if only zephyr project was cloned.
+    assert zephyr_project.is_cloned()
+    assert not net_tools_project.is_cloned()
+
+    # Verify if tagged-repo submodule was also cloned
+    res = zephyr_project.git('rev-parse --show-cdup', check=False,
+                             cwd=tagged_repo_dst_dir, capture_stderr=True,
+                             capture_stdout=True)
+    assert not (res.returncode or res.stdout.strip())
+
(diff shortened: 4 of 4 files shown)


kkasperczyk-no:
@carlescufi @mbolivar-nordic could you review it and share your comments?

mbolivar-nordic:
Hi @kkasperczyk-no , thanks so much for taking this effort on. I know several people have been looking for this feature and I will work with you to get it merged.

I think this is a good start, and we'll just need to discuss and analyze the behavior in various cases to make sure it's doing what we need.

CC'ing @nashif and @marc-hb as well.

kkasperczyk-no:
@mbolivar-nordic thanks, that would be great to have your support. I wanted to help improving west, but actually this is my first touch of its source code (so far I was only end user), so any remarks from the west community are welcome.

nashif:
@sylvioalves FYI

src/west/manifest-schema.yml
```diff
@@ -95,6 +95,21 @@ mapping:
           path:
             required: false
             type: str
+          # Submodules to checkout within the project
+          submodules:
+            required: false
+            type: seq
```
mbolivar-nordic, line 101: I think it would be good to have this be a boolean, too. That would let people do something simple, like:

```yaml
- name: myproject
  submodules: true
```

To get all the submodules without having to specify them by hand.

You can do this by setting the schema type to `any` and checking the type at runtime using `isinstance`. You can see the example code for `import:` handling in manifest.py for details.

src/west/manifest.py
```diff
@@ -66,6 +66,9 @@
 # though it's just a str in each individual file right now.
 WestCommandsType = Union[str, List[str]]
 
+# Type for the submodules value passed through the manifest file.
+SubmodulesType = List[Dict]
```
mbolivar-nordic, line 70: The Project class is meant to be a high level view, so exposing unparsed manifest dictionary data like this feels a bit out of character for the API to me. Is it possible to make this a list of structured objects, like a `List[submomdule]`, where `submodule` is a `typing.namedtuple` that specifies things about the submodule itself, like its name and path within the superproject?

README.rst
```diff
@@ -43,6 +43,7 @@ What just happened:
 
 - ``west update`` clones the other repositories named in the manifest file,
   creating working trees in the installation directory ``zephyrproject``.
+  It also checkouts submodules if cloned projects have defined any.
```
mbolivar-nordic, line 46: It's OK to leave this out of the PR. I will handle writing the documentation (which lives in the zephyr repository) once it's merged, before the next west release. It's not quite true as written, because submodules aren't always checked out. It's opt-in.

src/west/app/project.py
```diff
@@ -867,6 +867,15 @@ def fetch_missing_imports(self, args):
         self.manifest = Manifest.from_file(topdir=self.topdir,
                                            importer=self.update_importer)
 
+    def update_submodules(self, project, submodules, call_location):
+        # Updates given list of project submodules by using
+        # 'git submodule update --init' command from the location passed
+        # through the call_location parameter value.
+        if submodules:
+            for submodule in submodules:
+                project.git(['-C', call_location, 'submodule', 'update',
```
mbolivar-nordic, line 876: A few questions here:

- Don't you need a `--checkout` here to make sure the submodule isn't rebased if `submodule.<name>.update` is `rebase`, for example?
- What about recursion into submodules of submodules? Should this be optional, somehow?
- How can we ensure the submodule is checked out as a detached HEAD?

mbolivar-nordic (commented):
Here are some initial comments and questions from my side. Thanks!

README.rst
kkasperczyk-no: Ok, done.

src/west/manifest-schema.yml
kkasperczyk-no: I did, as you suggested and I also added another test to check whether using submodules: true results in updating all project submodules.

src/west/manifest.py
kkasperczyk-no: I agree with approach you suggested. I tried to follow it, please verify if current version looks better.

src/west/app/project.py
kkasperczyk-no: Honestly I don't think I'm able to answer it, as I wasn't assuming that it is needed. As far as I understand, you suggest that any local changes done in submodules should be overridden by the update and current code version will not do this. I'm really not very into detailed west operation to judge whether it's good or not, so I believe that I should rely on your statement. Could you share your vision about this functionality to let me understand it better? If it comes to the recursion - I think that could be useful think, but do you think that it should be selected by setting some value in the manifest or e.g. by calling west update command with some additional argument (-r is already taken by the rebase option).

marc-hb: > As far as I understand, you suggest that any local changes done in submodules should be overridden by the update and current code version will not do this. I'm really not very into detailed west operation to judge whether it's good or not,

I think consistency between submodules and west "modules" is very important and should always be the default unless the user very explicitly and clearly requests `west` and `submodules` to behave differently. Whatever `west update` does, `git submodule` should do the  same otherwise it will cause a lot of confusion.

(Just FYI I almost always use `west update -r`, I find `west update` without `-r` crazy :-)

> What about recursion into submodules of submodules? Should this be optional, somehow?

Maybe optional (then good luck for extensive test coverage) but again it should have the exact same **default** behaviour as `west`. Imagine a complex project with more than one level of `west` manifests and then multiple levels of git submodules somewhere underneath. Now imagine that `west update` imports all the manifests by default but for some strange reason updates only one level of submodules by default. That line would feel arbitrary and be quite confusing.

kkasperczyk-no: @marc-hb thanks, now I see potential problems more clearly. Then I assume that probably better option would be to make recursion on by default, to make sure that none of the levels was skipped, and allow disabling it in manifest, right?

marc-hb: > Then I assume that probably better option would be to make recursion on by default,

I think yes, recursive by default. However I've used west imports only once for pure testing purposes so far. So for that and other reasons please wait for @mbolivar-nordic

> and allow disabling it in manifest, right?

Depends what you mean by "disabling". Disabling this entire new submodules feature completely for fine-grained user control and reverting to older behaviours sounds nice to me. Not sure where it should be done: in the top level/other manifest, user config file, other, all of the above with well defined precedence...? The devil is in the details.

Disabling recursive updates which would leave only the first level of submodule seems like a confusing and bad setting to me because as I wrote before this would make the feature "recursive until the first levels of submodules". 

Here's how a "normal" recursive option works:
- yes: works on all levels.
- no: works on 1 level

Here's how a weird, recursive submodule setting would work in west:
- yes: works on all west imports + all submodule levels
- no: works on all west imports + 1 submodule level ?!?

I don't think anyone would enjoy this feature.

mbolivar-nordic: Hi,

> As far as I understand, you suggest that any local changes done in submodules should be overridden by the update and current code version will not do this. 

Not exactly. By default, west uses `git checkout --detach` to update each project to the new manifest revision. This does not affect any uncommitted working tree changes as long as they don't conflict with the checked out revision.

It does leave behind any commits that aren't part of the new manifest-rev unless you use `-r`. It leaves a detached HEAD unless you use `-r` or sometimes when you use `-k`.

I agree with @marc-hb that "west with submodules" should try to at least keep the default behavior:

1. leave a detached HEAD checked out
2. leave any uncommitted working tree changes alone
3. error out if uncommited changes conflict with the new submodule revision

Is that possible to do with git submodules?

> I think yes, recursive by default. However I've used west imports only once for pure testing purposes so far. So for that and other reasons please wait for @mbolivar-nordic

I think recursing into submodules of submodules is a good idea if it is possible / practical to get "everything" by default.

We can always add additional configuration to do more complicated stuff:

```yaml
manifest:
  projects:
  - name: foo
    submodules:
      recurse-behavior: whatever
```

Note I am **not** suggesting we add a `recurse-behavior` key right now. I'm just saying we could do it later if we need finer grained control.

But I am not a submodules user (hey, I just use west...) so I don't know if that addresses people's use cases and I want the users to chime in here. Thanks.

kkasperczyk-no: @marc-hb @mbolivar-nordic great tips, I really appreciate it. I added `--recursive` option as a default one, because for now it seems to be the best option and I also added test case for that in one of tests.

> It does leave behind any commits that aren't part of the new manifest-rev unless you use -r. It leaves a detached HEAD unless you use -r or sometimes when you use -k.

I agree with @marc-hb that "west with submodules" should try to at least keep the default behavior:

    leave a detached HEAD checked out
    leave any uncommitted working tree changes alone
    error out if uncommited changes conflict with the new submodule revision

Is that possible to do with git submodules?
>

Actually current behavior of submodules with `--checkout` option is almost the same, as described. I wrote almost, because it leaves at detached HEAD state, leaves uncommitted changes alone, prints error in case of conflicts and overrides new commits which are not part of new manifest-rev, but it doesn't care about `-r` or `-k` option. I tried to fix it using `--rebase` and `--checkout` options, but few problems occurred:

- Current west update strategy assumes that we should make rebase or keep descendants only when project is not in the detached state. That implies problem when some changes are done in the project's submodule and it should be rebased, but project itself is not modified, so rebase will not be selected. That's not a big problem at all, because new method introducing exact same behavior for submodules could be implemented to solve it.

- Real problem is using `--recursive` option, because it implies that all submodules of submodule will be updated the same way, as submodule is (with `--rebase`, `--checkout` or anything else). That means that when using `-r` submodules on the first level will behave same way as projects (so rebase when modified or checkout otherwise) and on other levels will always do the same as parent submodule did, without possibility to select other strategy. This will raise confusion mentioned by @marc-hb.  

What are your opinions on that? Do you think making assumption that submodules are always checked out and don't support rebase is acceptable? Summing up - this way default `west update` should work properly, but command extensions unfortunately not.

mbolivar-nordic: > What are your opinions on that? Do you think making assumption that submodules are always checked out and don't support rebase is acceptable?

I'll try to summarize to check my understanding: I think this means the west project update respects `-k` and `-r`, but submodules are always checked out as detached HEADs to their "gitlink" revisions. Submodule update behavior is not affected by `-k` or `-r`.

I find this acceptable: it's consistent, and treats west projects as first class citizens, with submodules just being additional "bags on the side" that follow the west project strictly, which is easy to understand and explain.

@marc-hb ?

mbolivar-nordic: Pinging @sylvioalves on this question also, since @nashif brought him in. I think this comment thread is the last major unresolved question.

marc-hb: > I think this means the west project update respects -k and -r, but submodules are always checked out as detached HEADs to their "gitlink" revisions. Submodule update behavior is not affected by -k or -r.

That's not what I understood. What I understood in a nutshell  (I could be wrong): the behavior of `west --rebase` and `west --keep-descendants` is project-specific (depending on whether there is a branch) whereas `submodule --rebase` is unconditional and this can creates funny inconsistencies.

> treats west projects as first class citizens, with submodules just being additional "bags on the side" that follow the west project strictly, which is easy to understand and explain.

I'm sure there will be `west/submodule` inconsistencies unfortunately impossible to address because these are two different tools with two different designs. There should just be as few inconsistencies as possible. I'm afraid `--rebase` and `--keep-descendants` might indeed end in that minimal list of unavoidable inconsistencies.  In such cases I think it's best to keep at least all submodules behaving the same, no special treatment for  the first submodule level for instance. Warnings are also useful for such complex situations and they save code comments.

I think the documentation of important inconsistencies like this potential one should be documented/specified in this PR already. I find that writing documentation is a great way to take some distance from the implementation details and make sure the high-level design is sound _and_ user friendly.

kkasperczyk-no: > I'll try to summarize to check my understanding: I think this means the west project update respects -k and -r, but submodules are always checked out as detached HEADs to their "gitlink" revisions. Submodule update behavior is not affected by -k or -r.
>

@mbolivar-nordic Yes, you are absolutely right. I will try to clarify it even more, to dispel any doubts. Currently for **every** project `west update` simplified behavior looks like this:

`
if current_branch != HEAD:
    if k_option_enabled:
        do_nothing()
    else if r_option_enabled:
        rebase()
    else:
        checkout()
else:
    checkout()
`

Update method cares about `-k` and `-r` options only when current branch is not in the HEAD state and otherwise they are ignored.

> That's not what I understood. What I understood in a nutshell (I could be wrong): the behavior of west --rebase and west --keep-descendants is project-specific (depending on whether there is a branch) 
>
That's true.

> whereas submodule --rebase is unconditional and this can creates funny inconsistencies.
>
@marc-hb I'm not 100% sure what you meant, but submodule --rebase currently is not used and submodules **always** do checkout, no matter of what parent project does.

I could make submodules to behave the same as projects and respect `-r` and `-k` options, but only at first level, because of the `--recursive` option, but I think that would be wrong, as you wrote:

>  In such cases I think it's best to keep at least all submodules behaving the same, no special treatment for the first submodule level for instance.
>

If it comes to the documentation - of course I can add it, but @mbolivar-nordic declared that he would like to do it on his own.

marc-hb: > I could make submodules to behave the same as projects and respect -r and -k options, but only at first level, because of the --recursive option

I just performed a quick test and `git submodule update --rebase --recursive` does rebase recursively:
```
git submodule update --rebase --recursive
First, rewinding head to replay your work on top of it...
Applying: mycommit
Submodule path 'rimage': rebased into 'fccf52133104987d9a532e6a9c332ba5c7a1bd91'
First, rewinding head to replay your work on top of it...
Applying: mycommit
Submodule path 'rimage/tomlc99': rebased into 'e3a03f5ec7d8d33be705c5ce8a632d998ce9b4d1'
```
Note there was a branch in both repos; not sure it matters for submodules.

According to `git help submodule`, `--rebase` and `--checkout` seem mutually exclusive.

> If it comes to the documentation - of course I can add it, but @mbolivar-nordic declared that he would like to do it on his own.

I certainly don't want to get into "who should do what" while I'm only offering incomplete reviews from a safe distance but I think the high-level logic should be somewhat documented or in other words "specified" before this gets merged because I don't think it's possible to review the code without the complete knowledge of how it's supposed to behave. That documentation could be just a draft, maybe not in this PR then. I mean something looking exactly like what you just did here:

> if k_option_enabled:
>        do_nothing()

This particular bit seems a bit too simplified :-)

marc-hb: > This means that when using -r submodules on the first level will behave same way as projects (so rebase when modified or checkout otherwise) and on other levels will always do the same as parent submodule did, without possibility to select other strategy. 

OK, I think I finally understood this. I got confused by "only on the first submodule level". 

I think what you're saying is: `west --rebase` can be repo-specific while `submodule --rebase --recursive` always performs a rebase, it's never repo-specific. That's one unavoidable inconsistency between `west` and `submodule`.

Considering `submodule --recursive` always performs the same action when traversing down, I think we all agree that all `submodule` invocations from west should all be given the same parameters, not depending on decisions `west` made in west repos. This keeps all submodules behaving the same no matter what.

However I think I disagree on never passing `--rebase` to `submodule`. I think it's more intuitive (and IMHO safer) to always pass `--rebase` through `west` and to `submodule` when and only when the user ask `west` to `--rebase`[*] 

Either way, the existence of some unavoidable inconsistency probably means  a "don't touch submodule" flag is needed sooner than later.

[*] I forgot why `west update --rebase` does not rebase detached HEADs. Feature or limitation?

kkasperczyk-no: I'm not west specialist and I build my understanding of project policy on your comments and reading code, but what I understood is that the worst thing is making inconsistent features and in my opinion we are just starting to go this way.
> I think what you're saying is: west --rebase can be repo-specific while submodule --rebase --recursive always performs a rebase, it's never repo-specific. That's one unavoidable inconsistency between west and submodule.

Considering submodule --recursive always performs the same action when traversing down, I think we all agree that all submodule invocations from west should all be given the same parameters, not depending on decisions west made in west repos. This keeps all submodules behaving the same no matter what.
> 
I agree with this, making submodules update strategy dependent only on first level of submodules and forcing another levels to behave way as first does is not good idea and we should keeps all submodules behaving the same no matter what.

> However I think I disagree on never passing --rebase to submodule. I think it's more intuitive (and IMHO safer) to always pass --rebase through west and to submodule when and only when the user ask west to --rebase.
>
But that one is for me the field to make another inconsistency. We have two options (at least I see two):

- Do, as you suggest, what is for sure fine, because we will support `-r`option for submodules, but we can't support `-k` option, as I think we won't be able to check ancestors of all submodules passed recursively and make decision whether we should do the checkout or not. That raises another inconsistency between projects and submodules update strategy.

- Do not support `-r` and `-k` options for submodules, what means that they will behave like less advanced version of update for projects, which supports only default behavior resulting by invoking `west update` command without any additional arguments. In my opinion this one has got less functionality, but is more clearer.

> [*] I forgot why west update --rebase does not rebase detached HEADs. Feature or limitation?
>
I don't know, but it looks like in the code someone made an assumption that only when branch is not in the detached HEAD state, any other options than default checkout might be considered.

marc-hb: I think it's intuitive and fairly consistent to pass-through a `--rebase` feature that has the exact same name and almost same behavior while dropping the `--keep-descendants` feature for the simple and intuitive reason that the latter does not exist in submodules. AFAIK the `--rebase` behavior difference is only for "careless" people who forget to save their precious work in a branch, correct? All other, careful users will not see any `--rebase` difference?

But more importantly: `--rebase` is not any random feature, it is a critical feature, because it's the only `--do-not-delete-all-my-work` feature available in both `west` and `submodule`. Consistency considerations aside, restricting it to `west` would be a great loss and block everyone from doing any work in submodules. Users who do not work in submodules and only use them as "bags on the side" don't care about `--rebase` etc. because they have nothing to rebase and won't be affected whatsoever.

In fact `--rebase` should have been a safer default in both. I digress.

mbolivar-nordic: > > 
> 
> If it comes to the documentation - of course I can add it, but @mbolivar-nordic declared that he would like to do it on his own.

@ kkasperczyk-no  Feel free to send a patch to the zephyr repository with documentation for this feature! I would certainly appreciate it.

I was just trying to make less work for you, since we can't merge that patch until we're ready to release the next version of west. That is something I have to do, because it also involves rolling up API changes, writing a changelog, etc., so I was going to be doing docs anyway.

If you write a docs patch, I'll just cherry-pick it into the other docs work I will do for the next release, with gratitude :smile:.

However, I think simplified pseudocode descriptions of the feature's behavior are good enough for getting the PR reviewed and merged, with more polished docs to be written later.

> [*] I forgot why west update --rebase does not rebase detached HEADs. Feature or limitation?

@marc-hb Limitation, I suppose.

> Do not support -r and -k options for submodules, what means that they will behave like less advanced version of update for projects, which supports only default behavior resulting by invoking west update command without any additional arguments. In my opinion this one has got less functionality, but is more clearer.

I agree.

I propose a compromise: can we add a separate `--rebase-submodules` option that rebases the submodules if given, and otherwise we just continue to use `--checkout`?  This makes it clear that `-k` and `-r` are for west projects, while submodules have different behavior with their own set of options.

marc-hb: > However, I think simplified pseudocode descriptions of the feature's behavior are good enough for getting the PR reviewed and merged, with more polished docs to be written later.

+1

> I propose a compromise: can we add a separate --rebase-submodules option that rebases the submodules if given, and otherwise we just continue to use --checkout? This makes it clear that -k and -r are for west projects, while submodules have different behavior with their own set of options.

I think this is an extra complication and hassle to highlight a rebase difference/west limitation that is not visible in "normal" usage. But it's certainly much better than not having the feature at all and being forced to drop all local commits in submodules.

If it's not the default then it will lose the users' commits but only the first times: after having tediously restored branches one by one or worse: inspected hard-to-read reflogs a few times to recover their work, the pain experienced should help them remember about the existence of `--rebase-submodules` for the next update and teach them the new lesson to always use branches even with submodules (in case they forget about the `--rebase-submodules` option again). Or they can configure `.gitmodules` but that's much less flexible: sometimes you do want to revert to some exact, untainted version.

> which supports only default behavior resulting by invoking west update command without any additional arguments. In my opinion this one has got less functionality, but is more clearer.

I understand that "-k and -r are not passed" is shorter and clearer than "-r is passed and -k is not" but that's a very small difference. It's not like the latter is difficult to remember considering submodules only support the latter (better than west).

In `git help submodule` you can see that `--rebase` is not considered like "an extra feature", it's shown at the exact same level than `--checkout` and `--rebase`. `--checkout` is just the default value, for backward compatibility I guess.

kkasperczyk-no: > But more importantly: --rebase is not any random feature, it is a critical feature, because it's the only --do-not-delete-all-my-work feature available in both west and submodule. 
>
I think it convinces me. All in all saving our work from losing is one of the elementary repository use case and I might be more important than some structural assumptions.

> I understand that "-k and -r are not passed" is shorter and clearer than "-r is passed and -k is not" but that's a very small difference.
>
To be honest I don't like it in 100%, but I must admit that considering all aspects, which we discussed, each has got some disadvantages and following @marc-hb suggestion is probably the best one. It prevents user from losing work, has some inconsistency, but still is quite simple and doesn't require adding new commands not known for users (that for sure would result in wasting some users time on learning it). Anyway I think that we should decide something together to bring it to the end, so after reflection I'm ok with @marc-hb version. I will try to implement, test and introduce it, but if anybody is not ok with that, or we should try another approach, because of some reason, please share your opinion.

kkasperczyk-no: Rebasing option and test for both checkout and rebase update strategies were added.

mbolivar-nordic: > I'm ok with @marc-hb version. I will try to implement, test and introduce it, but if anybody is not ok with that, or we should try another approach, because of some reason, please share your opinion.

OK, I can see the arguments on either side and I don't have a strong opinion here, so I want the users' voices to carry the largest weight. I am fine with this.

@marc-hb if you have a chance to look at this and give it a +1 I'd appreciate it, same for @sylvioalves.

sylvioalves: Sure, let me review the solution approach and test locally..

sylvioalves: For the purpose I was looking for, it works.

tests/test_project.py
```diff
@@ -477,6 +477,323 @@ def test_update_some_with_imports(repos_tmpdir):
...
+                        - name: zephyr
+                          url: {zephyr}
+                          submodules:
+                            - name: tagged_repo
+                              path: tagged_repo
+                        - name: net-tools
+                          url: {net_tools}
+                          submodules:
+                            - name: Kconfiglib
+                              path: {kconfiglib_submodule}
```
marc-hb, line 519: Plural starts at two - is there a test with at least 2 submodules in one west project? If not then there's lack of test coverage.

src/west/app/project.py
```diff
@@ -924,6 +924,29 @@ def fetch_strategy(self, args):
         else:
             return 'smart'
 
+    def update_submodules(self, project, submodules, call_location):
+        # Updates given project submodules by using
+        # 'git submodule update --init --checkout --recursive' command
+        # from the location passed through the call_location parameter value.
+        if submodules is not None:
```
marc-hb, line 931: You can save one indentation level with an early return:
```python
if submodules is None:
   return
```

src/west/app/project.py
```diff
@@ -924,6 +924,29 @@ def fetch_strategy(self, args):
...
+            # For the list type, update given list of submodules.
+            if isinstance(submodules, list):
+                for submodule in submodules:
+                    if isinstance(submodule, Submodule):
+                        project.git(['-C', call_location, 'submodule',
+                                     'update', '--init',
+                                     submodules_update_strategy, '--recursive',
+                                     submodule.path])
+            # For the bool type, update all project submodules
+            elif isinstance(submodules, bool):
```
marc-hb, line 945: https://www.google.com/search?q=isinstance+anti+pattern

You're making the internal implementation follow the user interface/configuration file exactly; it doesn't have to. You can have one `project.all_submodules` boolean on one hand and a possibly empty list of `project.submodules` as _another_ field. Combining them together in a Python `Union` unnecessarily mimics the user interface and makes IMHO the code more complicated.

src/west/manifest.py
```diff
@@ -911,6 +931,7 @@ def __init__(self, path: Optional[PathType] = None,
 
         # Pretending that this is a Project, even though it's not (#327)
         self.url: str = ''
+        self.submodules = None
```
marc-hb, line 934: Why not `False`?

marc-hb (changes requested):
Short of a complete documentation, I still think this PR misses at least a good few examples. I had to go and find them in the test code which was not the most efficient review process. These examples could be in the commit message or in (now out of date) #365 or in any other place convenient for design discussion and review. The review for such a major new feature like this should start "from the top" = from the user interface. The user interface was in fact what was discussed so far but anchored to some place deep down in the implementation; that's backwards.

For instance here's one relatively basic user interface question I could not quickly find the answer to: let's say I want to update submodules of every west project except for one (too expensive) west project. Can I do this? If yes how? Should I entirely remove the west project from my west manifest or is there some other way? This is just one random user interface question that should be answered easily.

BTW I think `submodules: all/none` would look better than `submodules: true/false`.

Please don't get me wrong:  @kkasperczyk-no has done a tremendous amount of great work in this PR and prototyping very early is very important too, however for such a major new feature the bar should IMHO be high, not the least because there is no second chance with user interface design.

src/west/app/project.py
marc-hb: With a `all_submodules` boolean and a possibly empty submodules list you would not even need this `if` at all: iterating over a possibly empty list is perfectly fine.

src/west/app/project.py
```diff
@@ -989,6 +1012,10 @@ def update(self, project):
                 stats['checkout new manifest-rev'] = perf_counter() - start
             _post_checkout_help(project, current_branch, sha, is_ancestor)
 
+        # Update project submodules, if it has any.
+        self.update_submodules(project, project.submodules,
+                               os.path.join(self.topdir, project.path))
```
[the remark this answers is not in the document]
kkasperczyk-no: Right, fixed.

src/west/app/project.py
```diff
@@ -924,6 +924,29 @@ def fetch_strategy(self, args):
...
 
+    def update_submodules(self, project, submodules, call_location):
+        # Updates given project submodules by using
+        # 'git submodule update --init --checkout --recursive' command
+        # from the location passed through the call_location parameter value.
+        if submodules is not None:
+            if self.args.rebase:
+                submodules_update_strategy = '--rebase'
+            else:
+                submodules_update_strategy = '--checkout'
```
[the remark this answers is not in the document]
kkasperczyk-no: Done.

src/west/manifest.py
```diff
@@ -87,6 +87,23 @@
...
 
+# Type for the submodule value passed through the manifest file.
+class Submodule(NamedTuple):
+    name: str
+    path: str
+
+# Submodules might be the list of values or bool.
+SubmodulesType = Union[List[Submodule], bool]
+
+def _is_submodule_ok(value: Any) -> bool:
```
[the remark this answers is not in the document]
kkasperczyk-no: Done.

src/west/manifest.py
kkasperczyk-no: Done.

tests/test_project.py
kkasperczyk-no: The `test_update_all_submodules` test is having 2 submodules in one project and updates them, by assigning true value to the submodules. Do you think it covers mentioned use case, or it should be verified in some other test if passing at least two elements to the submodules list works?

src/west/manifest.py
```diff
@@ -1309,6 +1330,44 @@ def get_projects(self,
...
+
+        For submodules of list type, method checks format of all elements
+        and converts dictionaries into the Submodule objects.
+
+        :param submodules: content of the manifest submodules value.
+        '''
+        self.submodules = submodules
+        self.submodules_type = type(self.submodules)
+        if self.submodules is not None:
+            # For the bool type, make sure that value is equal to True
```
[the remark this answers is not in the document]
kkasperczyk-no: You are right, false value is checked then in the updating method, so here false value, should not be rejected. Fixed.

src/west/app/project.py
kkasperczyk-no: I applied the version from the first comment.

src/west/app/project.py
kkasperczyk-no: I think that this comment and the following concern the similar thing:
> BTW I think submodules: all/none would look better than submodules: true/false

I know that the code is a bit more complicated, however I wanted to keep it consistent with already existing fields in the manifest like `import`, for which we can assign boolean true/false values and a list. Isn't it better solution than adding new method of enabling/disabling the field in the manifest, which specific for only one case and different from the others?

kkasperczyk-no:
@marc-hb I updated commit message to contain some highlights about the feature. I have a hope that it will be easier to review the code now.

mbolivar-nordic:
> Please don't get me wrong: @kkasperczyk-no has done a tremendous amount of great work in this PR and prototyping very early is very important too, however for such a major new feature the bar should IMHO be high, not the least because there is no second chance with user interface design.

I think we have time to complete the documentation and the implementation in parallel.

I've started work on other 0.9 release documentation which I want to include this feature. I'm not ready to push that quite yet, but I did get a prep work PR together here:

https://github.com/zephyrproject-rtos/zephyr/pull/31119

I'll add documentation for the new west 0.9 features (project groups and submodules being the main line items) as a follow up to that PR if we can get it merged quickly (please help review if you can).

I remain optimistic we can make ourselves happy about both implementation and documentation relatively soon.

I disagree with @marc-hb that we only have one chance to get this right, though. West's manifest file format has versioned schema support, as mentioned here:

https://docs.zephyrproject.org/latest/guides/west/manifest.html#manifest-files

We can bump the schema version if we make incompatible changes, and fall back to 0.9 behavior if the manifest explicitly requests it or does not provide a `manifest: version:`.

Therefore I don't think we need to let the perfect be the enemy of the good here. We have room to adjust.

src/west/manifest.py
```diff
@@ -1309,6 +1331,45 @@ def get_projects(self,
...
+
+        For submodules of bool type, method does nothing, as both boolean
+        values are valid (True - means that all project submodules should be
+        considered and False - means that are not any submodules).
+
+        :param submodules: content of the manifest submodules value.
+        '''
+        self.submodules = submodules
+        self.submodules_type = type(self.submodules)
+        if self.submodules is not None:
```
marc-hb, line 1353: This can be an early return too (which would avoid the double negation as a bonus)

src/west/manifest.py
```diff
@@ -1309,6 +1331,45 @@ def get_projects(self,
...
+                    if self.is_submodule_ok(self.submodules[index]):
+                        self.submodules[index] =  \
+                            Submodule(self.submodules[index]['name'],
+                                      self.submodules[index]['path'])
+                    else:
+                        self._malformed(f'Invalid submodule structure \
+                            {self.submodules[index]}')
+            # For the bool type, just do nothing.
+            elif self.submodules_type == bool:
+                pass
```
marc-hb, line 1367: ... and a second early return
```python
if self.submodules_type == bool:
   return self.submodules
```

tests/test_project.py
marc-hb: Thanks, I barely reviewed the test code and this was not the easiest to spot because as you wrote there's no _explicit_ list with more than 1 submodule in the test inputs.

> or it should be verified in some other test if passing at least two elements to the submodules list works?

I don't think that is necessary because that would add coverage only for the parsing / input parts of the code and west relies on YAML for most of that.

src/west/app/project.py
marc-hb: I missed the consistency aspect, makes sense. If @mbolivar-nordic is happy with this then I'm happy.

marc-hb:
Nit: the commit subject seems convoluted. I suggest: `add  support for git submodules in west update`

kkasperczyk-no:
> Nit: the commit subject seems convoluted. I suggest: `add support for git submodules in west update`

Ok, updated.

src/west/app/project.py
```diff
@@ -924,6 +924,30 @@ def fetch_strategy(self, args):
...
+
+        if submodules is None:
+            return
+
+        submodules_update_strategy = ('--rebase' if self.args.rebase
+                                      else '--checkout')
+        # For the list type, update given list of submodules.
+        if isinstance(submodules, list):
+            for submodule in submodules:
+                if isinstance(submodule, Submodule):
```
marc-hb, line 942: Is there any valid situation where `isinstance(submodules, list)` is True but `sinstance(submodule, Submodule)` is False for some submodule? If yes this deserves a short comment mentioning that situation. If not then this code should crash, not silently ignore a bug. The easiest is to just remove this test and let `submodule.path` crash below. If not enough for some reason then manually throw an exception in a new `else` clause.

(Sorry I didn't spot this earlier)

src/west/manifest.py
```diff
@@ -1309,6 +1331,45 @@ def get_projects(self,
             raise ValueError(unknown, uncloned)
         return ret
 
+    def _get_submodules(self, submodules) -> SubmodulesType:
+        '''Gets a list of `Submodules` objects or boolean from the manifest
+        *submodules*.
+
+        If submodules is empty, a copy of ``self.submodules`` is returned.
```
marc-hb, line 1338: `self.modules` is returned in any case. I don't get this particular comment.

src/west/app/project.py
```diff
@@ -924,6 +924,30 @@ def fetch_strategy(self, args):
...
 
+    def update_submodules(self, project):
+        # Updates given project submodules by using
+        # 'git submodule update --init --checkout --recursive' command
+        # from the project.path location.
+        submodules = project.submodules
+        call_location = os.path.join(self.topdir, project.path)
+
+        if submodules is None:
+            return
```
marc-hb, line 935: Optimization nit: you could do this earlier. In fact you can probably do it before even calling this method which would then become the `update_submodules_list()` method (no need actually to change the name, just to get the point across).

src/west/app/project.py
```diff
@@ -924,6 +924,30 @@ def fetch_strategy(self, args):
...
+                                      else '--checkout')
+        # For the list type, update given list of submodules.
+        if isinstance(submodules, list):
+            for submodule in submodules:
+                if isinstance(submodule, Submodule):
+                    project.git(['-C', call_location, 'submodule', 'update',
+                                 '--init', submodules_update_strategy,
+                                 '--recursive', submodule.path])
+        # For the bool type, update all project submodules
+        elif isinstance(submodules, bool):
```
marc-hb, line 947: Does this run when submodules is False?

marc-hb (changes requested):
Is there a global `submodules` boolean that would apply to all projects that don't define their own? I think there should be for convenience. If not in this version then maybe in the next one; adding a global later may not even require bumping the manifest version number that @mbolivar-nordic mentioned.

Bugs aside, a global setting seems to be the only good reason for allowing `project.submodules` to be `None`: so it can "inherit" the global setting. 

I'm conservatively choosing "Request changes" because I think I found a bug with `None`, apologies if I didn't.

mbolivar-nordic:
> Is there a global `submodules` boolean that would apply to all projects that don't define their own? I think there should be for convenience. If not in this version then maybe in the next one; adding a global later may not even require bumping the manifest version number that @mbolivar-nordic mentioned.

This could be done by adding support for something like `manifest: defaults: submodules: foo`. But I agree that's a backwards compatible schema change that could be done as an enhancement in a later version.

src/west/app/project.py
marc-hb: Just submitted a longer description of `--keep-descendants` in https://github.com/zephyrproject-rtos/zephyr/pull/31201

src/west/app/project.py
kkasperczyk-no: I moved it a bit higher, but I'm not sure about moving it outside of the method, as I would like to make sure that nobody will not use if in some other place without checking if submodules are not None.

src/west/app/project.py
kkasperczyk-no: Ok, that's right, done. I think that this check was even not needed, as submodules objects integrity is previously checked, while getting submodules list from the manifest in `_get_submodules` method.

src/west/app/project.py
kkasperczyk-no: Thanks, that was serious bug, I didn't spot it earlier, as it occurred after introducing some change suggested in our discussions. I changed check mentioned in this comment:

> Optimization nit: you could do this earlier. In fact you can probably do it before even calling this method which would then become a do_update_submodules() method (no need actually to change the name, just a sample name to get the point across).
>

from `if submodules is None:` to the `if not submodules` in order to consider also boolean False value. Moreover I also added new test, checking if submodules are not update when value is set to false, as such bug should not have happened.

tejlmand:
> > Is there a global `submodules` boolean that would apply to all projects that don't define their own? I think there should be for convenience. If not in this version then maybe in the next one; adding a global later may not even require bumping the manifest version number that @mbolivar-nordic mentioned.
> 
> This could be done by adding support for something like `manifest: defaults: submodules: foo`. But I agree that's a backwards compatible schema change that could be done as an enhancement in a later version.

We should be careful here.
Cause today, it is possible to nest repos in `west.yml` and no complaints by `git` if they are indeed submodules of each other.

One such use can be seen in this manifest: https://github.com/nrfconnect/sdk-nrf/blob/a523fb1a71d9c45a75f8f26f17643d5c81115662/west.yml#L107-L114
```
    - name: cmock
      path: test/cmock
      revision: c243b9a7a7b3c471023193992b46cf1bd1910450
      remote: throwtheswitch
    - name: unity
      path: test/cmock/vendor/unity
      revision: 031f3bbe45f8adf504ca3d13e6f093869920b091
      remote: throwtheswitch
```
where `unity` is in fact a submodule inside `cmock`.

This works, because `git` knows about the submodule, and accepts that `west` handles the checkout.

Now, imagine someone using this manifest through an import, like:
```
- name: nrf
  repo-path: sdk-nrf
  revision: <SHA>
  import: true
```

and then decides to add:
```
manifest:
  defaults:
    submodules: true

  projects:
      ....
    - name: nrf
      repo-path: sdk-nrf
      revision: <SHA>
      import: true
```
inheriting such global setting could result in some strange behavior, so we need to be careful.

src/west/app/project.py
tejlmand: > > I think what you're saying is: west --rebase can be repo-specific while submodule --rebase --recursive always performs a rebase, it's never repo-specific. That's one unavoidable inconsistency between west and submodule.
> 
> > Considering submodule --recursive always performs the same action when traversing down, I think we all agree that all submodule invocations from west should all be given the same parameters, not depending on decisions west made in west repos. This keeps all submodules behaving the same no matter what.
> 
> I agree with this, making submodules update strategy dependent only on first level of submodules and forcing another levels to behave way as first does is not good idea and we should keep all submodules behaving the same no matter what.

Jumping in here.

Why not use the existing infrastructure of `west` ?

Using `git submodule status` here, instead of `git submodule update --init --recursive` means we can internally create `west` projects.

Git stores the remote in `.gitmodules`, as example here:
https://github.com/ThrowTheSwitch/CMock/blob/master/.gitmodules#L7
```
[submodule "vendor/c_exception"]
	path = vendor/c_exception
	url = https://github.com/throwtheswitch/cexception.git
	branch = master
```
and a `git submodule status` will give you the SHA to checkout.
```
/.../cmock $ git submodule status
-dce9e8b26f2179439002e02d691429e81a32b6c0 vendor/c_exception
-031f3bbe45f8adf504ca3d13e6f093869920b091 vendor/unity
```
so you should be able to create those two projects a regular `west` projects, in which case the `-k` and `-r` will work.
Once the first level of submodules have been checked out and detached, you then enter the folder if `west` manifest has recursive == true.

As example:
```
# This would be done by `west` if the submodule is added as a project:
/.../cmock$ git -C vendor/c_exception/ init 
/.../cmock$ git -C vendor/c_exception/ fetch https://github.com/throwtheswitch/cexception.git master
 # SHA from above `git submodule status` command
/.../cmock$ git -C vendor/c_exception/ checkout dce9e8b26f2179439002e02d691429e81a32b6c0

# which means we can just do the same again, until no more submodules.
/.../cmock$ cd vendor/c_exception/
/.../cmock/vendor/c_exception$ git status
HEAD detached at dce9e8b
/.../cmock/vendor/c_exception$  git submodule status
-2c7629a0ae90ffe991b5fd08e4db8672f72ed64c vendor/unity

/.../cmock/vendor/c_exception$  cat .gitmodules 
[submodule "vendor/unity"]
        path = vendor/unity
        url = https://github.com/throwtheswitch/unity.git
```

The `.gitmodules` file can easily be read using `configparser`, and creating `west` projects for each recursive submodule allows `west` to use exactly the same strategy on submodules as it does on any other `west` project.

I ackknowledge, this is quite a change to the current approach, but it would ensure `west` behaves identical bewteen `west` projects and submodules regarding `-k` and `-r`.

mbolivar-nordic:
> inheriting such global setting could result in some strange behavior, so we need to be careful.

How could that cause problems? The `defaults` thing does not propagate across `import`. We made this design decision specifically to avoid weird behavior like this, e.g. overriding a default remote.

Each manifest is independently resolved, and the projects are inserted into the importing manifest, without caring about its own defaults.

tejlmand:
> > inheriting such global setting could result in some strange behavior, so we need to be careful.
> 
> How could that cause problems? The `defaults` thing does not propagate across `import`. We made this design decision specifically to avoid weird behavior like this, e.g. overriding a default remote.

ahh, right.
No worries then :)

marc-hb:
> This works, because git knows about the submodule, and accepts that west handles the checkout.

Wow, wow that feels both useful and totally scary. This means the submodule is there but not initialized in `.git/config`, correct?

I just tested blindsiding submodules like this locally (with pure git commands, no west involved) and this half-initialized state is... "fascinating": `git status` detect changes correctly but `git submodule status` does not have a clue and still think the submodule is missing. Now the `absorbgitdirs` section in `git help submodule` sounds like `git` has more or less anticipated this scenario but I would still NOT recommend west manifests creating half initialized submodules like this, if safe it's at best very confusing and more confusion is not exactly what you want to inflict to submodules users...

I also noted how @mbolivar-nordic has carefully avoided the topic of nested git clones in general, even before looking at submodules: https://github.com/zephyrproject-rtos/zephyr/pull/31119#discussion_r553667456 . Can't blame him :-)

src/west/app/project.py
marc-hb: I failed to parse the latest sentence so I'm not sure about the concern. However this check is the very first statement now so it will be trivial to move out if needed or wanted later for any reason. So I'm happy.

mbolivar-nordic:
> Wow, wow that feels both useful and totally scary.

It is definitely useful. It may be scary. Until this PR gets merged, it's the accepted workaround for submodules in west.

mbolivar-nordic:
Notice how `unity` is actually inside the `cmock` repository here: https://github.com/nrfconnect/sdk-nrf/blob/master/west.yml#L107

From the `cmock` perspective; `unity` is actually a submodule: https://github.com/nrfconnect/sdk-nrf/blob/master/west.yml#L107

marc-hb (approved):
LGTM with the following caveat: I didn't review the test code. Speaking of test code
thanks for adding a test for `False`, I didn't dare asking but you maybe you read my thoughts ;-)

mbolivar-nordic (approved):

kkasperczyk-no:
@mbolivar-nordic @marc-hb I'm happy to see it approved, thanks for your involvement in reviewing and tremendous support!

marc-hb:
Linking to @mbolivar-nordic documentation WIP: https://github.com/zephyrproject-rtos/zephyr/pull/31241

mbolivar-nordic:
> @mbolivar-nordic @marc-hb I'm happy to see it approved, thanks for your involvement in reviewing and tremendous support!

Me too, thanks for your contribution!

src/west/manifest.py
```diff
@@ -87,6 +87,14 @@
 
 _logger = logging.getLogger(__name__)
 
+# Type for the submodule value passed through the manifest file.
+class Submodule(NamedTuple):
+    name: str
+    path: str
```
marc-hb, line 93: So I started actually trying this new feature and got confused by the purpose of the `name` atrribute. Then I realized why I got confused: `name` is not anywhere.
Then I asked whether `name` could be removed... while  west 0.9 was released. Too late?

So in #474 I'm making `name` optional, please help review and/or test.

PS: `name` is barely used in `git submodule` too.

Issue #463, opened 2021-01-08
Unable to use nested folders on Windows


mfiumara:
Hi I'm seeing a problem with a manifest that's nested, it seems to only happen on Windows environments and not on Linux.

I have a reproduction path:
```
mkdir test-west
cd test-west
west init -m https://github.com/mfiumara/west-test.git --mr master
```

I encounter the following error:
```
C:\Users\Mattia\evalan>mkdir test-west

C:\Users\Mattia\evalan>cd test-west

C:\Users\Mattia\evalan\test-west>west init -m https://github.com/mfiumara/west-test.git --mr master
=== Initializing in C:\Users\Mattia\evalan\test-west
--- Cloning manifest repository from https://github.com/mfiumara/west-test.git, rev. master
Initialized empty Git repository in C:/Users/Mattia/evalan/test-west/.west/manifest-tmp/.git/
remote: Enumerating objects: 3, done.
remote: Counting objects: 100% (3/3), done.
remote: Compressing objects: 100% (2/2), done.
remote: Total 3 (delta 0), reused 3 (delta 0), pack-reused 0
Unpacking objects: 100% (3/3), 498 bytes | 19.00 KiB/s, done.
From https://github.com/mfiumara/west-test
 * branch            master     -> FETCH_HEAD
 * [new branch]      master     -> origin/master
3cdaef778b1c58d67f90dff48df3314e05046f58 refs/remotes/origin/master
Already on 'master'
Branch 'master' set up to track remote branch 'master' from 'origin'.
Traceback (most recent call last):
  File "c:\program files\python39\lib\shutil.py", line 806, in move
    os.rename(src, real_dst)
FileNotFoundError: [WinError 3] The system cannot find the path specified: 'C:\\Users\\Mattia\\evalan\\test-west\\.west\\manifest-tmp' -> 'C:\\Users\\Mattia\\evalan\\test-west\\applications\\some-folder'

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "c:\program files\python39\lib\runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "c:\program files\python39\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\Mattia\AppData\Roaming\Python\Python39\Scripts\west.exe\__main__.py", line 7, in <module>
[...]

PR #464, opened 2021-01-12
West 0.9 fixups

mbolivar-nordic:
Various standalone fixups and improvements to be finalized before cutting west 0.9.

@marc-hb  and @kkasperczyk-no this does include some changes to APIs and implementation details for the new submodules feature, but it is not expected to change any command line behavior. It's mostly nitpicks that I figured would be easier to get done as follow-up commits than slowing down the review with lots of back-and-forth comments.


mbolivar-nordic:
Hmm, I forgot that I also need to add something for https://github.com/zephyrproject-rtos/west/issues/463 to this PR.

mbolivar-nordic:
> Hmm, I forgot that I also need to add something for #463 to this PR.

Just for the record, that's done and this should be good to go.

tejlmand (approved):

src/west/manifest.py
```diff
@@ -89,11 +89,13 @@
 
 # Type for the submodule value passed through the manifest file.
 class Submodule(NamedTuple):
+    '''Represents a Git submodule within a project.'''
+
     name: str
     path: str
 
 # Submodules might be the list of values, bool or None.
```
kkasperczyk-no, line 97: If `SubmodulesType` is not Optional anymore `None` value should be removed from the comment.

src/west/app/project.py
```diff
@@ -932,18 +920,17 @@ def update_submodules(self, project):
...
 
         submodules = project.submodules
-        call_location = os.path.join(self.topdir, project.path)
         submodules_update_strategy = ('--rebase' if self.args.rebase
                                       else '--checkout')
         # For the list type, update given list of submodules.
         if isinstance(submodules, list):
             for submodule in submodules:
-                project.git(['-C', call_location, 'submodule', 'update',
+                project.git(['submodule', 'update',
```
kkasperczyk-no, line 928: Just to clarify `call_location` was used to allow calling git command from the project location and update submodules there, so submodules paths could be relative to this place. Did you remove it, because by default command is called from the project directory, or submodules paths should be absolute?

mbolivar-nordic: > . Did you remove it, because by default command is called from the project directory, or submodules paths should be absolute?

The first one. Project.git() runs in the project's abspath by default:

https://docs.zephyrproject.org/latest/guides/west/west-apis.html#west.manifest.Project.git

PR #470, opened 2021-01-27
Group filters

mbolivar-nordic:
As a result of review of the project groups feature documentation, change some of the names to be more clear, and be a bit more forgiving in the error handling.

A project's 'groups' list remains the same. No changes there.

However, we make the following changes:

- 'manifest: groups:' is now 'manifest: group-filter:'
- we add a west.manifest.Manifest.group_filter value, exposing 'manifest: group-filter:' after validation and string conversion
- the config option 'manifest.groups' is now 'manifest.group-filter'
- invalid values in manifest.group-filter are downgraded to warnings
- the west update '--groups' option becomes '--group-filter', with a slightly shorter name '--gf' if you can't handle that

I tacked on another fixup to the tests I noticed.

### DIFF
--- a/src/west/manifest-schema.yml
+++ b/src/west/manifest-schema.yml
@@ -123,9 +123,9 @@ mapping:
             required: false
             include: groups-schema
 
-  # If present, an allowlist/blocklist of project groups. Prefix a
-  # group name with "-" to block it. Give a group name as-is to
-  # allow it.
+  # If present, a list of project groups to enable and disable. Prefix
+  # a group name with "-" to disable it. Give a group name as-is to
+  # enable it.
   group-filter:
     required: false
     include: groups-schema
--- a/src/west/manifest-schema.yml
+++ b/src/west/manifest-schema.yml
@@ -19,7 +19,7 @@
 #
 # --------------------------------------------------------------
 
-schema;groups:
+schema;groups-schema:
   type: seq
   sequence:
     - type: str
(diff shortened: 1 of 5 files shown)


mbolivar-nordic:
cc @marc-hb

src/west/manifest-schema.yml
```diff
@@ -121,14 +121,14 @@ mapping:
             type: any
           groups:
             required: false
-            include: groups
+            include: groups-schema
 
   # If present, an allowlist/blocklist of project groups. Prefix a
```
tejlmand, line 126: still mentioning `allowlist/blocklist` at this line

src/west/manifest-schema.yml
```diff
@@ -121,14 +121,14 @@ mapping:
...
   # group name with "-" to block it. Give a group name as-is to
```
tejlmand, line 127:
```suggestion
  # group name with "-" to disable it, and a "+" to enable it. Give a group name as-is to
```

carlescufi (approved):
Reviewed the first commit, delta from the previous implementation of `groups` looks good to me.

tejlmand (approved):
Changes looks good.
Two places spotted that are still using allow/block.

src/west/manifest-schema.yml
```diff
@@ -121,14 +121,14 @@ mapping:
...
-            include: groups
+            include: groups-schema
 
-  # If present, an allowlist/blocklist of project groups. Prefix a
-  # group name with "-" to block it. Give a group name as-is to
-  # allow it.
-  groups:
+  # If present, a list of project groups to enable and disable. Prefix
+  # a group name with "-" to disable it. Give a group name as-is to
+  # enable it.
```
marc-hb, line 128: Didn't this change to `+` prefix? At least an optional one?

mbolivar-nordic: Yes, it did. I'll send a fix for master but I might be lazy and leave as-is for the release branch; I'm almost done with all the testing and I don't want to have to start over.

mbolivar-nordic: (Note the docs are correct)

PR #472, opened 2021-01-28
manifest-schema.yml: fix a comment

mbolivar-nordic:
This is stale from before the move to group-filter and changing its
semantics to requiring a + and -.


mbolivar-nordic:
@marc-hb

marc-hb (approved):

PR #474, opened 2021-02-04
manifest: make optional the 'name' attribute unused by Submodule

marc-hb:
2 commits, see commit messages

Forcing users to define something unused is not just tedious, it's also confusing.

### DIFF
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -90,8 +90,8 @@ _logger = logging.getLogger(__name__)
 class Submodule(NamedTuple):
     '''Represents a Git submodule within a project.'''
 
-    name: str
     path: str
+    name: Optional[str] = None
 
 # Submodules may be a list of values or a bool.
 SubmodulesType = Union[List[Submodule], bool]
@@ -353,7 +353,8 @@ def _is_submodule_dict_ok(subm: Any) -> bool:
 
     try:
         _assert(isinstance(subm, dict))
-        _assert(len(subm) == 2)
+        # Required key
+        _assert('path' in subm)
         # Allowed keys
         for k in subm:
             _assert(k in ['path', 'name'])
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -340,18 +340,29 @@ def _update_disabled_groups(disabled_groups: Set[str],
                  "along with as much information as you can, such as the "
                  "stack trace that preceded this message.")
 
-def _is_submodule_dict_ok(value: Any) -> bool:
-    # Check whether value is a dict that contains the expected
+def _is_submodule_dict_ok(subm: Any) -> bool:
+    # Check whether subm is a dict that contains the expected
     # submodule fields of proper types.
 
-    if not isinstance(value, dict):
+    class _failed(Exception):
+        pass
+
+    def _assert(cond):
+        if not cond:
+            raise _failed()
+
+    try:
+        _assert(isinstance(subm, dict))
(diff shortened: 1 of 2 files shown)


src/west/manifest.py
```diff
@@ -87,11 +87,12 @@
 _logger = logging.getLogger(__name__)
 
 # Type for the submodule value passed through the manifest file.
-class Submodule(NamedTuple):
+class Submodule:
     '''Represents a Git submodule within a project.'''
 
-    name: str
-    path: str
+    def __init__(self, path, name=None):
```
mbolivar-nordic, line 93: I think it would be nicer to keep this a NamedTuple. There is support for default arguments with this type:

https://docs.python.org/3/library/typing.html#typing.NamedTuple

Moving to a class like this without type annotations sacrifices checking that mypy  can do for us. Having runtime type checking is essential, but it's also nice to have the type information available during static analysis.

mbolivar-nordic (changes requested):
Thanks for the cleanup. Just one change requested. Would like to get @kkasperczyk-no to ACK as well.

kkasperczyk-no (commented):
It looks good for me. I agree that `name` could be optional, as it is not necessary to make submodules work. I added it only to allow describing submodules in the west.yml in some clear way (not only by paths) and some testing purposes, but yes it should not be mandatory. I'm not giving approve yet, as I would like to first make sure if we don't need any documentation changes? I think it would be good to mention here: https://github.com/zephyrproject-rtos/zephyr/blob/master/doc/guides/west/manifest.rst#option-2-list-of-mappings that names are not mandatory.

marc-hb:
Thanks @kkasperczyk-no . Documentation was on my mind but it's in a different repo so I was planning to fix it later. This change is backwards-compatible.

mbolivar-nordic:
>  I'm not giving approve yet, as I would like to first make sure if we don't need any documentation changes?

Part of the problem with having west's documentation in the zephyr repository is that we have to wait until we've already released the next version to get the documentation for it merged. If we change this documentation now, it will be incorrect, because this PR won't be part of any release, so users can't `pip install` any version of west with an optional `name`.

That's why I always write a docs PR with all the changes for a new west release when it's ready, and usually not before.

No objections to sending a PR to the zephyr repository marked DNM for now, if you want to have it ready to go when the time comes. I did that for the 'import' feature and I think it was good to have all the docs written before merging it into master.

src/west/manifest.py
```diff
@@ -90,8 +90,8 @@
 class Submodule(NamedTuple):
     '''Represents a Git submodule within a project.'''
 
-    name: str
     path: str
+    name: str = ''
```
mbolivar-nordic, line 94:
```suggestion
    name: Optional[str] = None
```

Why not this instead? We can still write `if foo.name:`, but having a nullable type seems nicer to me.

mbolivar-nordic (approved):
LGTM. This:

```py
         _assert('path' in subm and isinstance(subm['path'], str))
```

could be:

```py
         _assert('path' in subm)
```

since all keys must be strings due to the "allowed keys" loop that follows, but it's not a big deal.

mbolivar-nordic (changes requested):
Looks like the signed-off-by lines are wrong, actually. Please fix those and I'll merge.

mbolivar-nordic (approved):

marc-hb:
Documentation update in draft PR https://github.com/zephyrproject-rtos/zephyr/pull/32106

PR #482, opened 2021-02-19
Import group filters, too

mbolivar-nordic:
The west 0.9 documentation for group-filter states:

    Only the top level manifest file’s manifest: group-filter:
    value has any effect. The manifest: group-filter: values in any
    imported manifests are ignored.

See [Project Groups and Active Projects](https://docs.zephyrproject.org/latest/guides/west/manifest.html#project-groups-and-active-projects) for details on this feature.

This turns out to have been a mistake, which we are reversing course on.

It's a mistake because all users who import a manifest which makes some projects inactive will have to also manually disable the same groups just to get the same default list of projects.

Example manifest fragments showing the issue:

```yaml
   # my/west.yml
   manifest:
     projects:
       - name: upstream
         import: true
```

and:

```yaml
   # upstream/west.yml
   manifest:
     group-filter: [-disabled-group]
     projects:
       - name: i-want-this
       - name: i-do-not-want-this
         groups:
         - disabled-group
```

As written, my/west.yml will include 'i-do-not-want-this' as an active project.

This kind of leaky faucet has proven to be an unintuitive and poor user experience. People expect to get the default projects without having to copy the group-filter for everything they import.

We can't reverse this behavior without pushing out a new schema version, however: manifest schema 0.9 is doomed to repeat this behavior.

What we can do is create a new schema version that does the right thing, and get the word out to users to upgrade as soon as we can.

This PR updates the behavior for the new schema version (0.10) after some prep work to keep things tidy.

Additional notes:

- supersedes #481 by implementing the new behavior, and including test cases for all relevant scenarios.
- the original use case which showed the mistake is described in nrfconnect/sdk-nrf#3884
- the corresponding documentation update is in https://github.com/zephyrproject-rtos/zephyr/pull/32381

### DIFF
--- a/src/west/__init__.py
+++ b/src/west/__init__.py
@@ -1,3 +1,6 @@
-# Copyright (c) 2020, Nordic Semiconductor ASA
+# Copyright (c) 2021 Nordic Semiconductor ASA
 
-# nothing here
+import logging
+
+# https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library
+logging.getLogger('west').addHandler(logging.NullHandler())
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -704,6 +704,8 @@ class Project:
         if self.west_commands:
             ret['west-commands'] = \
                 _west_commands_maybe_delist(self.west_commands)
+        if self.groups:
+            ret['groups'] = self.groups
 
         return ret
 
@@ -1381,6 +1383,8 @@ class Manifest:
         # and Python 3.7+ guarantee.
         r: Dict[str, Any] = {}
         r['manifest'] = {}
+        if self.group_filter:
+            r['manifest']['group-filter'] = self.group_filter
         r['manifest']['projects'] = project_dicts
         r['manifest']['self'] = self.projects[MANIFEST_PROJECT_INDEX].as_dict()
 
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -49,7 +49,7 @@ QUAL_REFS_WEST = 'refs/west/'
 #:
 #: This value changes when a new version of west includes new manifest
 #: file features not supported by earlier versions of west.
-SCHEMA_VERSION = '0.9'
+SCHEMA_VERSION = '0.10'
 # MAINTAINERS:
 #
 # If you want to update the schema version, you need to make sure that
@@ -1835,6 +1835,9 @@ class Manifest:
         # Load projects and add them to the list, returning
         # information about which ones have imports that need to be
         # processed next.
+
+        if 'projects' not in manifest:
+            return
 
         have_imports = []
         names = set()
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -277,35 +277,36 @@ def _is_imap_ok(imap: _import_map, project: 'Project') -> bool:
 
 class _import_ctx(NamedTuple):
     projects: Dict[str, 'Project']
-    filter_fn: ImapFilterFnType
+    imap_filter: ImapFilterFnType
     path_prefix: Path
 
-def _new_ctx(ctx: _import_ctx, imap: _import_map) -> _import_ctx:
+def _compose_ctx_and_imap(ctx: _import_ctx, imap: _import_map) -> _import_ctx:
     # Combine the map data from "some-map" in a manifest's
     # "import: some-map" into an existing import context type,
     # returning the new context.
     return _import_ctx(ctx.projects,
-                       _and_filters(ctx.filter_fn, _imap_filter(imap)),
+                       _compose_imap_filters(ctx.imap_filter,
+                                             _imap_filter(imap)),
                        ctx.path_prefix / imap.path_prefix)
 
-def _filter_ok(filter_fn: ImapFilterFnType,
-               project: 'Project') -> bool:
-    # filter_fn(project) if filter_fn is not None; True otherwise.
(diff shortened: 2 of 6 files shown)


marc-hb:
Nit-picking the description of the bug:

```yaml
# upstream/west.yml
   manifest:
     group-filter: [-opt-in-group]
     projects:
       - name: core-project
       - name: restricted-access-project
         groups:
         - opt-in-group
```

I find "I-want-this" confusing because it looks like "I" wrote it there. If "I" had written the upstream manifest then I would have not inflicted this problem to myself in the first place.

"I-do-not-want-this" is not a problem; people cloning upstream Zephyr routinely download stuff they don't need.

src/west/__init__.py
```diff
@@ -1,3 +1,6 @@
-# Copyright (c) 2020, Nordic Semiconductor ASA
+# Copyright (c) 2021 Nordic Semiconductor ASA
 
-# nothing here
+import logging
+
+# https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library
+logging.getLogger('west').addHandler(logging.NullHandler())
```
marc-hb, line 6: > This results in just the ProjectCommandLogHandler output being shown

Is that on stdout ? Many CI engines and maybe others color stderr differently, extremely useful.

mbolivar-nordic: > Is that on stdout ? Many CI engines and maybe others color stderr differently, extremely useful.

No, it's on stderr.

mbolivar-nordic:
> "I-do-not-want-this" is not a problem; people cloning upstream Zephyr routinely download stuff they don't need.

Guess we will have to agree to disagree on this one. As discussed offline, we designed this feature with opt-in groups for zephyr repositories that are not restricted access also.

marc-hb:
I understand but then it's really just a minor annoyance/bug. A "bug description" is easier to understand when it focuses on the worst case. Never mind; this is not even in a commit message.

carlescufi (approved):
Approved based on reviewing `manifest: re-work group filter handling` and its description, as well as the discussion in #481.

tejlmand (commented):
Looking good.

But when testing this, I used the following syntax in my test manifest:
```
manifest:
  version: 0.10
```
according to this: https://docs.zephyrproject.org/latest/guides/west/manifest.html#version

However, this results in the following error message which may not be easy for
everyone to relate to quoting.
```
$ west list
FATAL ERROR: can't load west manifest
  Malformed manifest file: /projects/github/ncs/nrf/west.yml
  Schema file: /home/tora/.local/lib/python3.8/site-packages/west/manifest-schema.yml
  Hint: invalid version 0.1; must be one of: 0.6.99, 0.7, 0.8, 0.9, 0.10
```

However this is working (with quoting):
```
manifest:
  version: '0.10'
```

would it be possible to improve user feedback in this situation, to make them
aware they would need to quote the version ?

Also the Zephyr example should be updated to use quoted from.

mbolivar-nordic:
Great comments @tejlmand; now addressed.

> would it be possible to improve user feedback in this situation,

This now prints:

```
  Hint: invalid version 0.1; must be one of: 0.6.99, 0.7, 0.8, 0.9, 0.10. Do you need to quote the value (e.g. "0.10" instead of 0.10)?
```

src/west/manifest.py
```diff
@@ -2224,3 +2302,31 @@ def _check_paths_are_unique(self, mp: ManifestProject,
...
+            # if group filters were used anywhere.
+            #
+            # Hopefully no users ever actually see this warning.
+
+            if self_group_filter or ctx.group_filter:
+                _logger.warning(
+                    'providing legacy group-filter semantics '
+                    'due to explicit "manifest: version: 0.9"; '
+                    'please upgrade to west 0.10 or later and '
+                    'update this manifest.')
```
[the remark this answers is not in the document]
mbolivar-nordic: > when user sees this message he is already running a west version that supports the new behavior

Changed to:

```py
                _logger.warning(
                    "providing deprecated group-filter semantics "
                    "due to explicit 'manifest: version: 0.9'; "
                    "for the new semantics, use "
                    "'manifest: version: \"0.10\"' or later")
```

tejlmand (approved):

src/west/manifest.py
```diff
@@ -1836,6 +1836,9 @@ def _load_projects(self, manifest: Dict[str, Any],
         # information about which ones have imports that need to be
         # processed next.
 
+        if 'projects' not in manifest:
+            return
```
marc-hb, line 1921: I think this was a bit of a hack. Dropping this in #825, see commit messages there.

src/west/manifest.py
```diff
@@ -2224,3 +2309,31 @@ def _check_paths_are_unique(self, mp: ManifestProject,
...
+
+                # Set attribute for white-box testing the above warning.
+                self._legacy_group_filter_warned = True
+
+            _update_disabled_groups(self._disabled_groups, self_group_filter)
+            self.group_filter = self_group_filter
+
+        else:
+            _update_disabled_groups(self._disabled_groups, ctx.group_filter)
+            self.group_filter = [f'-{g}' for g in self._disabled_groups]
```
marc-hb, line 2339: Because `self._disabled_groups` is a non-deterministic `set()`, this made the output of `manifest --resolve` and `manifest --freeze` non-deterministic. This was just reported in #824. Fix and test coverage submitted in
- #827

PR #485, opened 2021-03-11
west init: get default manifest-rev from remote

mbolivar-nordic:
Thanks to @marc-hb for pointing me in the right direction on this `ls-remote` incantation.

Instead of hard-coding a default revision to fetch from the remote,
get it from the remote URL itself if it's not provided by the user.

If that fails, we ask for the --manifest-rev option to be provided
explicitly.

This makes the code more robust by making its behavior match git
clone, which checks out whatever the remote wants you to use by
default instead of making assumptions.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -124,10 +124,12 @@ With -l, creates a workspace around an existing local repository;
 without -l, creates a workspace by cloning a manifest repository
 by URL.
 
-If cloning a repository:
+With -m, clones the repository at that URL and uses it as the
+manifest repository. If --mr is not given, the remote's default
+branch will be used, if it exists.
 
-- the default REVISION to check out is "{MANIFEST_REV_DEFAULT}"
-- the default URL is "{MANIFEST_URL_DEFAULT}"''',
+With neither, -m {MANIFEST_URL_DEFAULT} is assumed.
+''',
             requires_workspace=False)
 
     def do_add_parser(self, parser_adder):
@@ -231,9 +233,11 @@ If cloning a repository:
         return topdir
 
     def bootstrap(self, args) -> Path:
-        manifest_url = args.manifest_url or MANIFEST_URL_DEFAULT
-        manifest_rev = args.manifest_rev or MANIFEST_REV_DEFAULT
         topdir = Path(abspath(args.directory or os.getcwd()))
+        log.banner('Initializing in', topdir)
+
+        manifest_url = args.manifest_url or MANIFEST_URL_DEFAULT
+        manifest_rev = args.manifest_rev or self.get_head_branch(manifest_url)
         west_dir = topdir / WEST_DIR
 
         try:
@@ -242,7 +246,6 @@ If cloning a repository:
         except util.WestNotFound:
             pass
 
-        log.banner('Initializing in', topdir)
         if not topdir.is_dir():
             self.create(topdir, exist_ok=False)
 
@@ -314,6 +317,48 @@ If cloning a repository:
         log.dbg(f"running '{cmd_str}' in {cwd or os.getcwd()}",
                 level=log.VERBOSE_VERY)
         subprocess.check_call(args, cwd=cwd)
+
+    def check_output(self, args, cwd=None):
+        cmd_str = util.quote_sh_list(args)
+        log.dbg(f"running '{cmd_str}' in {cwd or os.getcwd()}",
+                level=log.VERBOSE_VERY)
+        return subprocess.check_output(args, cwd=cwd)
+
+    def get_head_branch(self, url: str) -> str:
+        # Get the branch which url's HEAD points to. Requires git 2.8.0
+        # or later. Errors out if it can't, prints a banner if it can.
+        err_msg = (f'failed getting the default branch from {url}; '
+                   'please provide the --manifest-rev option')
+
+        # The '--quiet' option disables printing the URL to stderr.
+        try:
+            output = self.check_output(
+                ('git', 'ls-remote', '--quiet', '--symref', url, 'HEAD')
+            ).decode('utf-8')
+        except subprocess.CalledProcessError:
+            log.die(err_msg)
+
+        for line in output.splitlines():
+            if not line.startswith('ref: '):
+                continue
+            # The output looks like this:
+            #
+            # ref: refs/heads/foo	HEAD
+            # 6145ab537fcb3adc3ee77db5f5f95e661f1e91e6	HEAD
+            #
+            # So this is the 'ref: ...' case.
+            #
+            # Per git-check-ref-format(1), references can't have tabs
+            # in them, so this doesn't have any weird edge cases.
+            without_ref = line[len('ref: '):]
+            if not without_ref:
+                continue
+            ret = without_ref.split('\t')[0]
+            log.small_banner('no --manifest-rev was given; '
+                             f"using remote's default branch: {ret}")
+            return ret
+
+        log.die(err_msg)
 
     def clone_manifest(self, url: str, rev: str, dest: str,
                        exist_ok=False) -> None:
@@ -1465,8 +1510,6 @@ WEST_DIR = '.west'
 
 # Default manifest repository URL.
 MANIFEST_URL_DEFAULT = 'https://github.com/zephyrproject-rtos/zephyr'
-# Default revision to check out of the manifest repository.
-MANIFEST_REV_DEFAULT = 'master'
 
 #
 # Other shared globals.
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -3,7 +3,6 @@
 import collections
 import os
 import re
-import shlex
 import shutil
 import subprocess
 import textwrap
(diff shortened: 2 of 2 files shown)


src/west/app/project.py
```diff
@@ -124,10 +124,12 @@ def __init__(self):
 without -l, creates a workspace by cloning a manifest repository
 by URL.
 
-If cloning a repository:
+With -m, clones the repository at that URL and uses it as the
+manifest repository. If --mr is not given, the remote's currently
+active branch will be used, if it exists.
```
henrikbrixandersen, line 129: Should we refer to HEAD instead "active"?

marc-hb: HEAD feels like an implementation detail to me. "active"  is used in`git help clone` and `git help glossary` but is nowhere to be found in `git help checkout` where HEAD is used instead :-( git as usual...

As far as the server side is concerned, I believe the most usual name is neither: https://www.google.com/search?q=git+default+branch

mbolivar-nordic: Right, I'm imitating git clone here.

marc-hb: I searched more and I think we have a clear winner. `git clone` is surprisingly "wrong".
"Active" is much better than "default" for a local, developer clone. Not so much for a server.

https://docs.github.com/en/github/setting-up-and-managing-your-github-user-account/managing-the-default-branch-name-for-your-repositories
https://docs.github.com/en?query=%22default+branch%22

https://docs.gitlab.com/ee/user/project/repository/branches/#default-branch
https://docs.gitlab.com/search/?query=%22default+branch%22


https://gerrit-review.googlesource.com/Documentation/project-configuration.html#default-branch

https://gerrit.googlesource.com/git-repo/+/50a81de2bc9f20%5E%21/
```python
def ResolveRemoteHead(self, name=None):
    """Find out what the default branch (HEAD) points to.

     `output = self.bare_git.ls_remote('-q', '--symref', '--exit-code', name, 'HEAD')`
```


Git itself uses a variety of names in the same paragraph to confuse the reader and "default" is one of them:


git help remote
```
       set-head
           Sets or deletes the default branch (i.e. the target of the symbolic-ref refs/remotes/<name>/HEAD) for the named
           remote. Having a default branch for a remote is not required, but allows the name of the remote to be specified in
           lieu of a specific branch. For example, if the default branch for origin is set to master, then origin may be
           specified wherever you would normally specify origin/master.
```

https://github.com/git/git/blob/master/Documentation/gitrepository-layout.txt

```

HEAD::
	A symref (see glossary) to the `refs/heads/` namespace
	describing the currently active branch.  It does not mean
	much if the repository is not associated with any working tree
	(i.e. a 'bare' repository), but a valid Git repository
	*must* have the HEAD file; some porcelains may use it to
	guess the designated "default" branch of the repository
	(usually 'master'). 
	
```

src/west/app/project.py
```diff
@@ -315,6 +317,43 @@ def check_call(self, args, cwd=None):
...
+        cmd_str = util.quote_sh_list(args)
+        log.dbg(f"running '{cmd_str}' in {cwd or os.getcwd()}",
+                level=log.VERBOSE_VERY)
+        return subprocess.check_output(args, cwd=cwd)
+
+    def get_head_branch(self, url: str) -> str:
+        # Get the branch which url's HEAD points to. Requires git 2.8.0
+        # or later. Errors out if it can't, prints a banner if it can.
+        err_msg = (f'failed getting the active branch from {url}; '
+                   'please provide the --manifest-rev option')
```
marc-hb, line 330: See above.

src/west/app/project.py
```diff
@@ -315,6 +317,43 @@ def check_call(self, args, cwd=None):
...
+        # The output looks like this:
+        #
+        # ref: refs/heads/foo	HEAD
+        # 6145ab537fcb3adc3ee77db5f5f95e661f1e91e6	HEAD
+        #
+        # The '--quiet' option disables printing the URL to stderr.
+        try:
+            output = self.check_output(
+                ('git', 'ls-remote', '--quiet', '--symref', url, 'HEAD')
+            ).decode('utf-8')
```
marc-hb, line 341: Matching gerrit looks good.

src/west/app/project.py
```diff
@@ -315,6 +317,43 @@ def check_call(self, args, cwd=None):
...
+            log.die(err_msg)
+
+        for line in output.splitlines():
+            if not line.startswith('ref: '):
+                continue
+            # Per git-check-ref-format(1), references can't have tabs
+            # in them, so this doesn't have any weird edge cases.
+            ret = line[len('ref: '):].split('\t')[0]
+            log.small_banner('no --manifest-rev was given; '
+                             f"using remote's active branch: {ret}")
```
marc-hb, line 352: See above.

src/west/app/project.py
```diff
@@ -232,7 +234,7 @@ def local(self, args) -> Path:
 
     def bootstrap(self, args) -> Path:
         manifest_url = args.manifest_url or MANIFEST_URL_DEFAULT
-        manifest_rev = args.manifest_rev or MANIFEST_REV_DEFAULT
+        manifest_rev = args.manifest_rev or self.get_head_branch(manifest_url)
```
marc-hb, line 237: To be clear using `_head_` _in the source code_ looks good to me. Implementation details in the code are a good thing.

mbolivar-nordic:
@marc-hb I read through your sources and I agree. Comments addressed, thanks.

I wrote this code independently, but it's good to know that Google is using the same implementation as us.

src/west/app/project.py
```diff
@@ -315,6 +317,43 @@ def check_call(self, args, cwd=None):
                 level=log.VERBOSE_VERY)
         subprocess.check_call(args, cwd=cwd)
 
+    def check_output(self, args, cwd=None):
+        cmd_str = util.quote_sh_list(args)
+        log.dbg(f"running '{cmd_str}' in {cwd or os.getcwd()}",
+                level=log.VERBOSE_VERY)
+        return subprocess.check_output(args, cwd=cwd)
```
marc-hb, line 324: I'm surprised `west` had nothing like this already.

src/west/app/project.py
```diff
@@ -315,6 +317,43 @@ def check_call(self, args, cwd=None):
...
+            ).decode('utf-8')
+        except subprocess.CalledProcessError:
+            log.die(err_msg)
+
+        for line in output.splitlines():
+            if not line.startswith('ref: '):
+                continue
+            # Per git-check-ref-format(1), references can't have tabs
+            # in them, so this doesn't have any weird edge cases.
+            ret = line[len('ref: '):].split('\t')[0]
```
marc-hb, line 350: Single-line parsing is impressive but a bit cryptic and hard to... parse! Rather than making it longer with local variables or what not, how about adding the typical parsed line as a comment?
```python
# Parsing something like: ref: refs/heads/main	HEAD
# Per git-check-ref-format(1), references can't have tabs
# in them, so this doesn't have any weird edge cases.
ret = line[len('ref: '):].split('\t')[0]
```

src/west/app/project.py
```diff
@@ -315,6 +317,43 @@ def check_call(self, args, cwd=None):
...
+            log.small_banner('no --manifest-rev was given; '
+                             f"using remote's default branch: {ret}")
+            return ret
```
marc-hb, line 353: Just to be on the safe side:
```suggestion
            if ret.strip():
                   return ret
```

`'ref:    '[len('ref: '):].split('\t')[0]` is pure whitespace and whitespace in an error message is very puzzling.

src/west/app/project.py
mbolivar-nordic: > I'm surprised `west` had nothing like this already.

We would normally use Project.git(), but we can't, because we haven't bootstrapped anything yet.

src/west/app/project.py
mbolivar-nordic: > typical parsed line as a comment?

I'm a bit confused because this is already in the same function, but I'll move it closer to its point of use.

marc-hb: Doh! Sorry I missed it.

mbolivar-nordic:
> still lets 'ref:       ' through.

Uhm, is that possible output?

mbolivar-nordic:
> This new `without_ref` adds more lines than my `if ret.strip():` trick

That part is on purpose, to avoid doing too much on one line.

mbolivar-nordic:
Extra tests commit is to get this to pass on Windows.

marc-hb:
> Uhm, is that possible output?

No, the idea of the one-line `if ret.strip():` is to deal with something unexpected from a somehow broken git server: _not_  any garbage from that server but only with that special case of whitespace garbage because that's the only case that produces output and an empty error message very hard to relate to the output of the `git ls-remote` command run manually.

Admittedly paranoid, never mind.

marc-hb (approved):
> That part is on purpose, to avoid doing too much on one line.

The previous version was fine by me, I just missed the example.

tejlmand (approved):
Looks good.
Did some manual testing with remotes using different HEADs, and also where remote HEAD would be detached.

Everything worked as expected.

Nice improvement :+1:

mbolivar-nordic:
Thanks for reviews!

PR #494, opened 2021-04-05
Move git checks to WestCommand

mbolivar-nordic:
This makes it easier for commands to check the git version.

### DIFF
--- a/src/west/commands.py
+++ b/src/west/commands.py
@@ -229,6 +229,14 @@ class WestCommand(ABC):
                 level=log.VERBOSE_VERY)
         return subprocess.check_output(args, cwd=cwd)
 
+    def die_if_no_git(self):
+        '''Abort if git is not installed on PATH.
+        '''
+        if not hasattr(self, '_git'):
+            self._git = shutil.which('git')
+        if self._git is None:
+            log.die("can't find git; install it or ensure it's on your PATH")
+
     @property
     def git_version_info(self):
         '''Returns git version info as a tuple of ints in (major,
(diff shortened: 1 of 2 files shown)


src/west/commands.py
```diff
@@ -224,6 +226,29 @@ def check_output(self, args, cwd=None):
         log.dbg(f"running '{cmd_str}' in {cwd or os.getcwd()}",
                 level=log.VERBOSE_VERY)
         return subprocess.check_output(args, cwd=cwd)
+
+    def git_version_at_least(self, version):
```
tejlmand, line 238: minor, should we call it `git_version_minimum_required` / `git_version_minimum`?

tejlmand (approved):
Looks good. Nice cleanup.
Just a minor suggestion.

marc-hb (commented):
Why are `check_call` and `check_output` methods at all?

```
pylint
R0201: Method could be a function (no-self-use)
```

src/west/commands.py
mbolivar-nordic: > minor, should we call it `git_version_minimum_required` / `git_version_minimum`?

I changed it to a property, `git_version_info`, following along with `sys.version_info`.

mbolivar-nordic:
> Why are `check_call` and `check_output` methods at all?

Because I hate `west.log` and want to convert all the output routines (`log.die`, etc.) to WestCommand methods (`WestCommand.die`, etc.) at some point.

That will remove the need for global state in `west.log` (continuing work towards resolving #149) without sacrificing the ability to do things like `west --quiet update` by adding a per-instance "be quiet" flag or so.

Moved to staticmethod for now to silence the linter.

marc-hb (approved):
> Because I hate west.log and want to convert all the output routines ...

I don't get all the details but it's more than enough for me to know you thought about it.

> Moved to staticmethod for now to silence the linter.

It's not just the linter, I find `@staticmethod` more visible. I often skip the list of arguments.

The new `git_version_info()` looks better and great; much more flexible.

PR #496, opened 2021-04-06
Update optimizations

mbolivar-nordic:
This implements new `west update` options that can be used to optimize performance:

- `--name-cache`, `--path-cache`: allows cloning project repositories from a known place on the file system before hitting the network to fetch the current revisions (improves performance by not fetching if possible)
- `--narrow`: fetches the project revision directly even if it is a SHA, and does not fetch tags (improves performance by fetching fewer objects than the default `west update`)
- `--fetch-opt`: allows specifying additional options to each `git fetch` used by `west update`, such as `--depth=1` (which may improve performance if many objects in history are not needed)

It also adds some new configuration options:

- `update.narrow`: boolean, default false. If true, `west update` uses `--narrow` always.
- `update.path-cache`: string, default no value. If nonempty, west update uses this as its `--path-cache` if not otherwise specified.
- `update.name-cache`: same as `update.path-cache`, but for the `--name-cache` option.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -1231,10 +1231,19 @@ class Update(_ProjectCommand):
         project.git(['fetch', '-f'] + tags + clone_depth +
                     self.args.fetch_opt +
                     ['--', project.url, refspec])
-        _update_manifest_rev(project, next_manifest_rev)
 
         if take_stats:
-            stats['fetch and set manifest-rev'] = perf_counter() - start
+            stats['fetch'] = perf_counter() - start
+
+        # Update manifest-rev, leaving an entry in the reflog.
+        if take_stats:
+            start = perf_counter()
+
+        new_ref = project.sha(next_manifest_rev)
+        _update_manifest_rev(project, new_ref)
+
+        if take_stats:
+            stats['set manifest-rev'] = perf_counter() - start
 
     @staticmethod
     def clean_refs_west(project, stats, take_stats):
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -804,6 +804,11 @@ class Update(_ProjectCommand):
 
         self.group_filter: List[str] = []
 
+        if args.narrow:
+            self.narrow = True
+        else:
+            self.narrow = config.getboolean('update', 'narrow', fallback=False)
+
         def handle(group_filter_item):
             item = group_filter_item.strip()
             if not item.startswith(('-', '+')):
@@ -1201,7 +1206,7 @@ class Update(_ProjectCommand):
         # non-commit object" error when the revision is an annotated
         # tag. ^{commit} type peeling isn't supported for the <src> in a
         # <src>:<dst> refspec, so we have to do it separately.
-        if _maybe_sha(rev) and not self.args.narrow:
+        if _maybe_sha(rev) and not self.narrow:
             # We can't in general fetch a SHA from a remote, as some hosts
             # forbid it for security reasons. Let's hope it's reachable
             # from some branch.
@@ -1222,7 +1227,7 @@ class Update(_ProjectCommand):
         log.small_banner(f'{project.name}: fetching, need revision {rev}')
         # --tags is required to get tags if we're not run as 'west
         # update --narrow', since the remote is specified as a URL.
-        tags = (['--tags'] if not self.args.narrow else [])
+        tags = (['--tags'] if not self.narrow else [])
         clone_depth = (['--depth', str(project.clone_depth)] if
                        project.clone_depth else [])
         # -f is needed to avoid errors in case multiple remotes are
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -1163,11 +1163,7 @@ class Update(_ProjectCommand):
         # the latest value it should be.
 
         if self.fs == 'always' or _rev_type(project) not in ('tag', 'commit'):
-            if take_stats:
-                start = perf_counter()
-            _fetch(project)
-            if take_stats:
-                stats['fetch and set manifest-rev'] = perf_counter() - start
+            self.fetch(project, stats, take_stats)
         else:
             log.dbg('skipping unnecessary fetch')
             if take_stats:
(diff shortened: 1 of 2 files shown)


mbolivar-nordic:
@marc-hb how do you think these options should play together with submodules?

marc-hb:
First I would like to say the new options and command line interface seem to capture perfectly all the sometimes contradicting requirements that were expressed in old, long, convoluted and confusing discussions scattered across multiple github issues (and I plead guilty for some of that). Of course the devil is always in the details and time will tell but the first impression of the user interface is: wow, what still stands between west and world domination?

Now back to your question, the long story short is: I'm not sure. After months (years?) discussing git optimizations in west, you just made me realize I never asked myself these same questions about submodules. Maybe because I try to avoid them as much as possible?

I'm not sure how caching will work because I haven't looked at this PR yet and will probably not have any time this week. Note git submodules have for quite some time been using their parent's .git/ directory instead of their own, not sure whether that will help. Worst case west caching should cache nothing submodule related.

I just had a look `git help submodule` and the `update` command supports a `--depth` argument. So I guess you could carry this one forward to submodules. However it would require filtering out other, unsupported `--fetch-opt` arguments. I would recommend leaving this as potential future improvement, waiting a fair amount of time after the dust of these new features has settled.

Submodules don't seem to have any `--narrow` equivalent. There is a `branch` attribute in `.gitmodules` that defaults to the remote default branch but it does not seem to have any `--narrow` or `revision-in` effect, all remote branches are fetched anyway. I think it's only meant for `git submodule update --remote` and for initial creation convenience. Again, I would recommend to leave anything here as a potential future improvement.


Sorry for not having much more than only superficial compliments, I'll try to get back to my usual negative self ASAP!

carlescufi:
@czeslawmakarski and @KaSroka please test and review

tejlmand:
Looks good, but I think it would be valuable if the fetch / narrow / cache options can be specified with `west config`.

mbolivar-nordic:
> Looks good, but I think it would be valuable if the fetch / narrow / cache options can be specified with `west config`.

OK, let's bikeshed that before I implement:

- `update.narrow`: boolean, default false. If true, west update uses `--narrow` always.
- `update.path-cache`: string, default empty. If nonempty, west update uses this as its `--path-cache` if not otherwise specified.
- update.name-cache`: same as `update.path-cache`, but for the `--name-cache` option.

mbolivar-nordic:
>  fetch [...] options can be specified with west config.

I prefer to leave `--fetch-opt` out of configuration for now if that's OK.

mbolivar-nordic:
> Again, I would recommend to leave anything here as a potential future improvement.

I think so too. Thanks.

tejlmand:
> > fetch [...] options can be specified with west config.
> 
> I prefer to leave `--fetch-opt` out of configuration for now if that's OK.

sure, let's start simple so feel free to leave out `--fetch-opt` for now.
Can always be added later, but in long run I guess any human user running `west update` with `--fetch-opt` in a workspace would probably want to use that particular option on each invocation in the workspace.

mbolivar-nordic:
> Can always be added later, but in long run I guess any human user running `west update` with `--fetch-opt` in a workspace would probably want to use that particular option on each invocation in the workspace.

I kind of disagree, especially for the `--depth 1` case. So let's keep this out for now as agreed, then, thanks.

marc-hb:
> I kind of disagree, especially for the --depth 1 case

I firmly disagree for the `--depth 1` case.  From `git help fetch`:

```
        --depth=<depth>
           Limit fetching to the specified number of commits from
           the tip of each remote branch history. If fetching to a
           shallow repository created by git clone with
           --depth=<depth> option (see git-clone(1)), deepen or
           __SHORTEN__ the history to the specified number of commits.
           Tags for the deepened commits are not fetched.
```
Emphasis mine.

BTW this highlights an important mismatch between `west` and `git` interfaces:
-  `git` had both `clone` and `fetch`. If you want to use `--depth`, you most likely want to use it only when cloning, not when fetching.
- `west` has only `update` for both. In fact it does not even invoke `git clone`.

 -------

Besides poor performance and [breaking basic git features unexpectedly](https://github.com/thesofproject/linux/issues/2556), misunderstandings is yet another reason to avoid shallow clones. They're really a hack.

As a coincidence I just experienced this error today after long minutes of git silence. No idea what it means:
```
git fetch --deepen=500 myremote
fatal: shallow file has changed since we read it
```

mbolivar-nordic:
> * `update.narrow`: boolean, default false. If true, west update uses `--narrow` always.
> 
> * `update.path-cache`: string, default empty. If nonempty, west update uses this as its `--path-cache` if not otherwise specified.
> 
> * `update.name-cache`: same as `update.path-cache`, but for the `--name-cache` option.

Now added.

mbolivar-nordic:
Rebased to fix merge conflicts.

marc-hb:
>  Note git submodules have for quite some time been using their parent's .git/modules/ directory instead of their own, not sure whether that will help.

This could have really helped. Unfortunately it does not. `git clone` does not copy `.git/modules/`. In other words the new way git submodules are implemented offers a cache extremely easy to use but `git clone` does not take advantage of it! :-(

I disconnected the network, ran `git clone` locally _and_ copied `.git/modules` _manually_. Then ta-da: I could initialize and update submodules without a network connection and recursively! That's a hack though, I would consider `.git/modules/` as a private implementation detail that should not be accessed. Especially not in the very first version of these large new features.

carlescufi (approved):
This looks great to me already.

czeslawmakarski:
I'm not on the approvers list, but I checked and it works great by us. So I do approve.

KaSroka (approved):
Looks good. As I have changed project I'm in I have no much room to test this but it should greatly speed up CI that only cares about given version and doesn't need history - this is what we needed in my old project.

mbolivar-nordic:
Thanks for the reviews and testing!

lucsegers:
@mbolivar-nordic could you tell me how this should be used in a CI environment?
Should I just use "west update --narrow"? 
Not sure how I should use the path-cache and name-cache in a ci environment. Do we have some examples for this?

mbolivar-nordic:
@lucsegers I don't have any general advice; the results are always going to depend on the details.

Example: `west update --narrow` should generally be a win, but it depends on your manifest. Not all git servers support fetching a SHA directly, which means `--narrow` won't always work in all CI environments.

In terms of using a cache, you need to set it up so your CI environment has a prepopulated cache available and mounted on the file system somewhere. The specifics for this will depend on your CI environment. Are you running CI in a bare metal server somewhere? Set up the cache in the file system and update it periodically. Are you using docker? Try to figure out a way to do a volume mount from the host machine, if you have control over the file system of the host machine. Etc.

marc-hb:
I just added a brand new "performance" label to about 20 issues and pull requests:
https://github.com/zephyrproject-rtos/west/issues?q=label%3Aperformance+

AFAIK the most recent discussion happened in #638, have a look to at least this one.

PR #503, opened 2021-05-01
commands: fix git version detection on macOS

mbolivar-nordic:
I ran into this testing 0.11.0a1. Our current code matches against the
last word in the 'git --version' output. Unfortunately, for 'git
version 2.24.3 (Apple Git-128)', that's incorrectly matching against
'Git-128)', which doesn't work.

Generalize the version detection a bit to handle this.

### DIFF
--- a/src/west/commands.py
+++ b/src/west/commands.py
@@ -239,24 +239,70 @@ class WestCommand(ABC):
 
     @property
     def git_version_info(self):
-        '''Returns git version info as a tuple of ints in (major,
-        minor, patch) format, like (2, 31, 1) for git version 2.31.1.
+        '''Returns git version info as a tuple of ints, usually in
+        (major, minor, patch) format, like (2, 29, 1) for git version
+        2.29.1.
 
         Aborts the program if there is no git installed.
+
+        In rare circumstances, you may get a (major, minor) tuple,
+        like (2, 29).
         '''
+        # It's perfectly safe to compare 2-tuples against 3-tuples.
+        # For example, '(2, 29) > (2, 28, 0)' is True.
+        # https://docs.python.org/3/reference/expressions.html#comparisons
+
         if not hasattr(self, '_git_ver'):
             self.die_if_no_git()
-            # raw_ver usually looks like '2.31.1'
-            raw_ver = self.check_output(
-                [self._git, '--version']).decode().strip().split()[-1]
-            # match just the numeric prefix, in case we get something like
-            # '2.31.1-EXTRAVERSION' in some situations. The major/minor/patch
-            # group names are purely for clarity.
-            match = re.match(
-                r'^(?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)', raw_ver)
-            self._git_ver = tuple(int(group) for group in match.groups())
+            self._git_ver = self._parse_git_version(
+                self.check_output([self._git, '--version']))
             log.dbg(f'git version: {self._git_ver}', level=log.VERBOSE_VERY)
         return self._git_ver
+
+    @staticmethod
+    def _parse_git_version(raw_version):
+        # Convert the raw 'git --version' output to a tuple.
+        #
+        # This is a @staticmethod so it can be white box tested.
+        #
+        # Usually the resulting tuple looks like (major, minor,
+        # patch).
+        #
+        # We get a length 2 tuple in obscure situations like git
+        # built from a development source tree created using 'git
+        # archive'. (See GIT-VERSION-GEN in the git sources if you're
+        # curious about details.)
+        #
+        # Downstream distributors sometimes tweak the results by
+        # adding to the end of 'x.y.z' in the 'git version x.y.z'
+        # string, but git itself always prints "git version %s", where
+        # the %s is the version.
+        #
+        # https://github.com/git/git/blob/7e391989789db82983665667013a46eabc6fc570/help.c#L646
+        #
+        # Some example possibilities:
+        #
+        # git version 2.25.1
+        # git version 2.28.0.windows.1
+        # git version 2.24.3 (Apple Git-128)
+        # git version 2.29.GIT
+        #
+        # We handle this by matching the first bit in the
+        # whitespace-separated output that has a prefix that looks
+        # like a semver.
+
+        match = re.search(
+            r'\s(?P<major>\d+)\.(?P<minor>\d+)(\.(?P<patch>\d+))?',
+            raw_version.decode(), flags=re.ASCII)
+        if not match:
+            log.die(f"can't get git version from {raw_version!r}")
+
+        major, minor, patch = (match.group('major'), match.group('minor'),
+                               match.group('patch'))
+        version = int(major), int(minor)
+        if patch is None:
+            return version
+        return version + (int(patch),)
 
 #
 # Private extension API
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -1,0 +1,16 @@
+# Copyright (c) 2021, Nordic Semiconductor ASA
+
+import os
+
+from west.commands import WestCommand
+
+assert 'TOXTEMPDIR' in os.environ, "you must run these tests using tox"
+
+gv = WestCommand._parse_git_version
+
+def test_parse_git_version():
+    # White box test for git parsing behavior.
+    assert gv(b'git version 2.25.1\n') == (2, 25, 1)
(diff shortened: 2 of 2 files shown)


mbolivar-nordic:
This is blocking v0.11. I'll backport once it's merged.

marc-hb:
Just for the record:
```
$ /usr/bin/git --version
git version 2.24.3 (Apple Git-128)

$ /usr/local/bin/git --version # homebrew
git version 2.28.0
```

marc-hb:
This sort of typical parsing issues is why I tend to prefer "trial and error" approaches like this one:
https://github.com/zephyrproject-rtos/west/compare/3c233fd14a7..339fe1a9ffa18e4a596#diff-8ab679fb4cff42c12eb05fbefdcebcebef2004540bf2b42bcb71ed5cede86a93L1083
```python
            try:
                # Silence the verbose and repetitive init.defaultBranch
                # warning. The branch 'placeholder' will never have any
                # commit so it will never actually exist.
                project.git(['init', '--initial-branch=placeholder',
                             project.abspath],
                            capture_stdout=True, capture_stderr=True,
                            cwd=util.west_topdir())
                # If git < 2.28 then the above failed because of
                # --initial-branch. If git >= 2.28 then it failed for
                # some other reason and will fail again for the same
                # reason which will be printed this time:
                project.git(['init', project.abspath], cwd=util.west_topdir())
```

src/west/commands.py
```diff
@@ -246,15 +246,32 @@ def git_version_info(self):
...
+            #
+            # We handle this by matching the first string in the output
+            # that has a prefix that looks like a semver. Hopefully this
+            # is general enough.
+
+            raw_ver = self.check_output([self._git, '--version']).decode()
+
+            # Match just the numeric prefix, to handle '.windows.1'
+            # type suffixes.
+            ver_re = re.compile(r'^(\d+)\.(\d+)\.(\d+)')
```
marc-hb, line 264: Looks like this `^` has no effect when using `match()`, only when using `search()` (ignoring MULTILINE).

src/west/commands.py
```diff
@@ -246,15 +246,32 @@ def git_version_info(self):
...
+
+            for word in raw_ver.strip().split():
```
marc-hb, line 266: Have you considered `\b`oundary as in `r'\b(\d+)\.(\d+)\.(\d+)'` instead of the `.split()`? It's not equivalent as it would match `(2.8.1)` for instance but it would make the code much shorter.

src/west/commands.py
marc-hb: Should the last digit be optional (`?`) to be on the safe side?

marc-hb (commented):
Have you considered `from packaging import version` ?

mbolivar-nordic:
> This sort of typical parsing issues is why I tend to prefer "trial and error" approaches like this one (PR #495 v1)

Heard, but this never hit any users since it was caught in pre-release testing, so I think it's okay to do it this way too.

mbolivar-nordic:
> Have you considered `from packaging import version` ?

I hadn't, but we shouldn't be using that. The version.parse method is for parsing PEP 440 style versions, which are not guaranteed here.

src/west/commands.py
mbolivar-nordic: > Looks like this `^` has no effect when using `match()`, only when using `search()` (ignoring MULTILINE).
> Should the last digit be optional (?) to be on the safe side?

Good catch on both fronts. I did some more digging in the git sources and pushed a new version that IMO handles all the possibilities we can reasonably expect.

src/west/commands.py
mbolivar-nordic: > Have you considered `\b`oundary as in `r'\b(\d+)\.(\d+)\.(\d+)'` instead of the `.split()`? It's not equivalent as it would match `(2.8.1)` for instance but it would make the code much shorter.

I've golfed it down a few lines.

src/west/commands.py
```diff
@@ -239,25 +239,79 @@ def die_if_no_git(self):
...
+        # git version 2.28.0.windows.1
+        # git version 2.24.3 (Apple Git-128)
+        # git version 2.29.GIT
+        #
+        # We handle this by matching the first bit in the
+        # whitespace-separated output that has a prefix that looks
+        # like a semver.
+
+        match = re.search(
+            r'\s(?P<major>\d+)\.(?P<minor>\d+)(\.(?P<patch>\d+))?',
```
marc-hb, line 302: I guess you made the `\s` is non-optional based on what you learned in the sources? More precisely, you found it was easy to customize the version to `git version 6.6.6-whatever` but not to get rid of the `git version ` prefix?

src/west/commands.py
```diff
@@ -239,25 +239,79 @@ def die_if_no_git(self):
...
+            raw_version.decode(), flags=re.ASCII)
+        if not match:
+            log.die(f"can't get git version from {raw_version!r}")
+
+        major, minor, patch = (match.group('major'), match.group('minor'),
+                               match.group('patch'))
+
+        if patch is not None:
```
marc-hb, line 310: Tiny nit, useless negation:
```python
if patch is None:
   return int(major), int(minor)
```

Alternatively:
```python
version = (major, minor)
if patch is not None
    version += (patch,)

return version
```

src/west/commands.py
```diff
@@ -239,25 +239,79 @@ def die_if_no_git(self):
...
+        #
+        # We handle this by matching the first bit in the
+        # whitespace-separated output that has a prefix that looks
+        # like a semver.
+
+        match = re.search(
+            r'\s(?P<major>\d+)\.(?P<minor>\d+)(\.(?P<patch>\d+))?',
+            raw_version.decode(), flags=re.ASCII)
+        if not match:
+            log.die(f"can't get git version from {raw_version!r}")
```
marc-hb, line 305: Nice.

src/west/commands.py
```diff
@@ -239,25 +239,79 @@ def die_if_no_git(self):
...
+        # For example, '(2, 29) > (2, 28, 0)' is True.
+        #
+        # Language reference on comparisons:
+        #
+        #    Collections that support order comparison are ordered
+        #    the same as their first unequal elements (for example,
+        #    [1,2,x] <= [1,2,y] has the same value as x <= y). If a
+        #    corresponding element does not exist, the shorter
+        #    collection is ordered first (for example, [1,2] < [1,2,3]
+        #    is true).
```
marc-hb, line 261: I think this comment is useful but very long. The example and link are enough.

src/west/commands.py
```diff
@@ -239,25 +239,79 @@ def die_if_no_git(self):
...
-            match = re.match(
-                r'^(?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)', raw_ver)
-            self._git_ver = tuple(int(group) for group in match.groups())
+            self._git_ver = self._parse_git_version(
+                self.check_output([self._git, '--version']))
             log.dbg(f'git version: {self._git_ver}', level=log.VERBOSE_VERY)
         return self._git_ver
 
+    @staticmethod
+    def _parse_git_version(raw_version):
```
marc-hb, line 273: Just curious about the extra method. It's not that much code, even less without the comments.

marc-hb (approved):
As you spent time looking at the git sources I trust you didn't find any easier `git --parsable-version` ? Or at least none available in... old git versions :-D

PS: regular expressions are a programming language on their own but Python (and Perl?) bring that to a different level.

src/west/commands.py
mbolivar-nordic: > Is the extra method to ease testing? Just curious.

Yes, exactly

src/west/commands.py
mbolivar-nordic: > I guess you made the `\s` is non-optional based on what you learned in the git sources?

You made me triple check, and yes, not only is this safe, but it's safe regardless of the l10n (not i18n) settings: https://github.com/git/git/blob/7e391989789db82983665667013a46eabc6fc570/help.c#L646

I thought this could vary by locale, but I was misreading the gettext configuration files.

mbolivar-nordic:
@marc-hb thanks for the approval. I've cleaned this up a bit per your review and will merge when CI passes.

mbolivar-nordic:
> As you spent time looking at the git sources I trust you didn't find any easier `git --parsable-version` ? Or at least none available in... old git versions :-D

I addressed this in a response to a per-line comment, but I'll repeat it here. No, I can't find anything like that, but the help.c file makes it clear that we can expect git not to play tricks on us in the future by changing the way this works.

PR #509, opened 2021-06-23
west update: fix --fetch=smart for lightweight tags

mbolivar-nordic:
When a project revision in west.yml is a lightweight tag,
'git cat-file -t <revision>' prints 'commit' instead of 'tag'.

Hilarious as this may be, it confuses the _rev_type() helper used by
west update. The helper doesn't know what kind of revision it is,
since it's not a branch and doesn't look like a commit either. We then
fetch the tag unnecessarily since we don't know what we're dealing
with. That works, but we don't need to hit the network in this case.

Fix this by handling the lightweight tag case as well.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -1512,7 +1512,7 @@ def _rev_type(project, rev=None):
     elif stdout != 'commit':    # just future-proofing
         return 'other'
 
-    # to tell branches apart from commits, we need rev-parse.
+    # to tell branches, lightweight tags, and commits apart, we need rev-parse.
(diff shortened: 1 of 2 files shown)


marc-hb (approved):
> When a project revision in west.yml is a lightweight tag,
> 'git cat-file -t ' prints 'commit' instead of 'tag'.

> Hilarious as this may be, ...

Well it's not funnier than printing 'commit' for a branch too. Pretty sure this is because git considers neither branch nor lightweight tags as "real" objects, they're just aliases.

Git is a tremendous success and making fun of it like this is unacceptable, please fix the commit message.

(just kidding)

PR #514, opened 2021-07-24
Detached HEAD in manifest repository and master fixes

mbolivar-nordic:
The main point of this PR is to fix an issue where initializing from a URL leaves the manifest repository in a detached HEAD state. 

To get here, though, I needed to include other fixes so I could test it on my Arch machine, which is no longer passing tox, due to changes in the way git works.

While I'm here, I decided to grep for 'master' throughout the tree and remove it where possible or no longer applicable, to keep up as much as practical with where git is going in the time I have to work on this. I'm leaving some work for the future; see commit logs for details.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -375,7 +375,7 @@ With neither, -m {MANIFEST_URL_DEFAULT} is assumed.
         else:
             # Fetch the ref-space similar to git clone plus the ref
             # given by user.  Redundancy is ok, for example if the user
-            # specifies 'heads/master'. This allows users to specify:
+            # specifies 'heads/my-branch'. This allows users to specify:
             # pull/<no>/head for pull requests
             self.check_call(('git', 'fetch', 'origin', '--tags', '--',
                              rev, 'refs/heads/*:refs/remotes/origin/*'),
@@ -1486,15 +1486,13 @@ class Update(_ProjectCommand):
             # we called Update.init_project() above), check out
             # 'manifest-rev' in a detached HEAD state.
             #
-            # Otherwise, the initial state would have nothing checked
-            # out, and HEAD would point to a non-existent
-            # refs/heads/master branch (that would get created if the
-            # user makes an initial commit). Among other things, this
-            # ensures the rev-parse --abbrev-ref HEAD which happens
-            # later in the update() will always succeed.
+            # Otherwise, it's possible for the initial state to have
+            # nothing checked out and HEAD pointing to a non-existent
+            # branch. This causes the 'git rev-parse --abbrev-ref HEAD'
+            # which happens later in the update to fail.
             #
             # The --detach flag is strictly redundant here, because
-            # the refs/heads/<branch> form already detaches HEAD, but
+            # the qualified manifest-rev form detaches HEAD, but
             # it avoids a spammy detached HEAD warning from Git.
             if take_stats:
                 start = perf_counter()
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -9,7 +9,7 @@ import argparse
 from functools import partial
 import logging
 import os
-from os.path import abspath, relpath, exists
+from os.path import abspath, relpath
 from pathlib import PurePath, Path
 import multiprocessing
 import shutil
@@ -239,7 +239,11 @@ With neither, -m {MANIFEST_URL_DEFAULT} is assumed.
         log.banner('Initializing in', topdir)
 
         manifest_url = args.manifest_url or MANIFEST_URL_DEFAULT
-        manifest_rev = args.manifest_rev or self.get_head_branch(manifest_url)
+        if args.manifest_rev:
+            # This works with tags, too.
+            branch_opt = ['--branch', args.manifest_rev]
+        else:
+            branch_opt = []
         west_dir = topdir / WEST_DIR
 
         try:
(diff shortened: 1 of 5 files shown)


tests/conftest.py
```diff
@@ -45,6 +57,27 @@
...
+
+    tmpdir = tmpdir_factory.mktemp("west-check-git-caps-tmpdir")
+
+    try:
+        subprocess.run([GIT, 'init', '--initial-branch', 'foo',
+                        os.fspath(tmpdir)],
+                       stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL,
+                       check=True)
+        GIT_INIT_HAS_BRANCH = True
+    except subprocess.CalledProcessError:
```
marc-hb, line 78: I had a `CalledProcessError` for the same feature in the first version of #495 and you asked me to change it to `self.git_version_info >= (2, 28, 0)`. Is it because this is test code, can you elaborate?

marc-hb (commented):
A lot less code, nice. Who knows, maybe getting closer to "bare git" may even save some long discussions in the future like this one https://github.com/zephyrproject-rtos/west/issues/319#issuecomment-786210081

PS: I haven't really looked at the code yet.

tests/conftest.py
mbolivar-nordic: >  Is it because this is test code, can you elaborate?

Yep, see the comment:

```py
    # We aren't using WestCommand._parse_git_version() here just to
    # try to keep the conftest behavior independent of the code being
    # tested.
```

src/west/app/project.py
```diff
@@ -239,7 +239,11 @@ def bootstrap(self, args) -> Path:
         log.banner('Initializing in', topdir)
 
         manifest_url = args.manifest_url or MANIFEST_URL_DEFAULT
-        manifest_rev = args.manifest_rev or self.get_head_branch(manifest_url)
+        if args.manifest_rev:
+            # This works with tags, too.
+            branch = ['--branch', args.manifest_rev]
```
marc-hb, line 244: Nit: would you mind calling this variable `branch_opt` instead?

src/west/app/project.py
```diff
@@ -266,10 +275,13 @@ def bootstrap(self, args) -> Path:
...
         if not temp_manifest.is_file():
+            if args.manifest_rev:
+                and_rev = f' and --manifest-rev={args.manifest_rev}'
+            else:
+                and_rev = ''
             log.die(f'can\'t init: no {temp_manifest_filename} found in '
                     f'{tempdir}\n'
-                    f'  Hint: check --manifest-url={manifest_url} and '
-                    f'--manifest-rev={manifest_rev}\n'
+                    f'  Hint: check --manifest-url={manifest_url}{and_rev}\n'
```
marc-hb, line 284: Nesting an f-string into another one makes  me miss the simplicity of concatenation with `+`

marc-hb (commented):
An actual code  review this time.

> This fixes an issue reported by Carles Cufí where the manifest
repository is left at a detached HEAD when using recent versions of
git (I can reproduce on Arch with git v2.32.0, and in Ubuntu 21.04).

Was that bug filed somewhere and is there some explanation about why it was noticed only now?

The code changes look great and I will not cry about the lost feature, it was certainly not worth all the code that's getting removed and it's very easy to replicate manually using the two steps in the commit message. However it is still a lost feature so these two steps should IMHO be not just buried in a commit message and much more visible, probably in a future doc update?

My main frustration here is the lack of bug description. The great code simplification aside, we don't know why this fixes what bug...

marc-hb:
cc: @M1cha (couldn't add him as reviewer)

carlescufi (approved):
This fixes the issue completely for me. Thanks!

tests/conftest.py
```diff
@@ -295,11 +295,20 @@ def create_workspace(workspace_dir, and_git=True):
     if and_git:
         create_repo(mp)
 
-def create_repo(path):
-    # Initializes a Git repository in 'path', and adds an initial commit to it
+def create_repo(path, initial_branch='master'):
+    # Initializes a Git repository in 'path', and adds an initial
+    # commit to it in a new branch 'initial_branch'. We're currently
+    # keeping the old default initial branch to keep assumptions made
+    # elsewhere in the test code working with newer versions of git.
```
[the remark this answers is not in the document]
M1cha: no, arch is currently at `2.32.0` without any patches.

M1cha (approved):
LGTM but it took me some time to realize that `grepo` is not a typo.

tests/conftest.py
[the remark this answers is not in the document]
mbolivar-nordic: > > If not specified, fall back to the default name (currently master, but this is subject to change in the future;
> 
> So is Arch shipping some early 2.33 version?

Right -- and my systemwide default is `main`.

Going back to the commit log:

> the test cases assume that newly created repositories
> have 'master' branches, which isn't true in general anymore unless you
> ask for it explicitly.

The "in general" part is key here.

mbolivar-nordic:
> Was that bug filed somewhere and is there some explanation about why it was noticed only now?

You got me on this. I completely mischaracterized the source of the bug. Let me file a proper issue and fix the commit logs. Thanks for calling me out.

mbolivar-nordic:
> Let me file a proper issue and fix the commit logs.

https://github.com/zephyrproject-rtos/west/issues/522. The fix is the same, but the commit logs are adjusted to reflect what really happened.

> is there some explanation about why it was noticed only now?

There's nothing in the release testing process that checks for that. I added a commit which adds a step to the release process documented in MAINTAINERS.rst to avoid this happening again.

mbolivar-nordic:
> it took me some time to realize that `grepo` is not a typo.

That's a shorthand that @marc-hb and I use to disambiguate the Google tool from the general concept of a repository, yeah...

marc-hb (approved):
Approving on the assumption that nothing has changed besides the commit message and review nits (= tried to look at the range-diff but -  github + not much time)

> Let me file a proper issue and fix the commit logs.

OK so it was a relatively straight-forward regression after all... ?

> > is there some explanation about why it was noticed only now?

Answering myself: the regression was introduced in March which is not that long ago.

> That's a shorthand that @marc-hb and I use to disambiguate the Google tool from the general concept of a repository, yeah...

Not sure how new it is but the name "git-repo" is finally... googl-able (can't resist the irony)

PR #515, opened 2021-07-25
fix buffered update of projects with `/` in their name

M1cha:


src/west/app/project.py
```diff
@@ -865,10 +865,15 @@ def _update_concurrent_thread(args):
             saved_stdout = os.dup(sys.stdout.fileno())
             saved_stderr = os.dup(sys.stderr.fileno())
 
+            logfile = os.path.join(logdir, project.name)
+
+            # the name might contain `/`
+            os.makedirs(os.path.dirname(logfile), exist_ok=True)
```
marc-hb, line 871: There is a newer  https://docs.python.org/3/library/pathlib.html#pathlib.Path.mkdir . I believe `pathlib` is preferred in general for various reasons and `west` already uses it.

According to `setup.py` west requires python>=3.6

src/west/app/project.py
```diff
@@ -865,10 +865,15 @@ def _update_concurrent_thread(args):
...
+
             # the dup2s replace stderr and stdout to a file.
             # unlike just replacing the stream objects this also affects writes
             # by child processes.
-            logfile = open(os.path.join(logdir, project.name), "wb")
+            logfile = open(logfile, "wb")
```
marc-hb, line 876: No big deal but I'm not a fan of the same variable changing type like this, I think it's a bit confusing. `logfilepath` instead? `path` also highlights the fact that it's not just a basename.

marc-hb:
Is there a simple way to add a test for this?

mbolivar-nordic:
The docs are silent about whether slashes are allowed or forbidden characters in project names:

https://docs.zephyrproject.org/latest/guides/west/manifest.html#id6

However, I am surprised that we would want to support `/`. I think that's just asking for trouble. Linux doesn't support files with slashes in their names, does it? So why should west support projects with slashes in their names?

M1cha:
I think the reason is that the remote path to the repo relative to the base url defaults to the name.
So naturally, people use the path for the name to prevent having to name the project.

Combine that with the (probably repo-inspired) style of having a `github` remote instead of one remote per user/org and you got yourself a lot of slashes in your project names.

mbolivar-nordic:
>  So naturally, people use the path for the name to prevent having to name the project.

IMO, that's wrong; such cases should be using the `repo-path` key instead.

> Combine that with the (probably repo-inspired) style of having a github remote instead of one remote per user/org and you got yourself a lot of slashes in your project names.

I don't get this, sorry. Both west and repo have as many remotes as you want. If someone has

```yaml
manifest:
  remotes:
    - name: github
      url-base: https://github.com
```

That's their choice, but west doesn't mandate that -- what am I missing here?

M1cha:
you're 100% right but either we should deny `/` for project names or the buffer should support them.
The way it is right now it's too inconsistent IMO.

mbolivar-nordic:
I'd be fine with denying '/' (and '\', for Windows) in project names. Would you mind sending a PR?

M1cha:
here  you go: https://github.com/zephyrproject-rtos/west/pull/521

mbolivar-nordic:
> here you go: #521

Thanks! Can this (#515) be closed, then?

PR #517, opened 2021-07-27
Fix west.manifest.Project.sha() for tag revisions

mbolivar-nordic:
We need to use commit peeling to ensure that a revision which is a tag
is converted to a SHA. The current way Project.sha() works only
handles branches.

### DIFF
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -832,8 +832,8 @@ class Project:
         # Though we capture stderr, it will be available as the stderr
         # attribute in the CalledProcessError raised by git() in
         # Python 3.5 and above if this call fails.
-        cp = self.git(f'rev-parse {rev}', capture_stdout=True, cwd=cwd,
-                      capture_stderr=True)
+        cp = self.git(f'rev-parse {rev}^{{commit}}', capture_stdout=True,
+                      cwd=cwd, capture_stderr=True)
         # Assumption: SHAs are hex values and thus safe to decode in ASCII.
         # It'll be fun when we find out that was wrong and how...
         return cp.stdout.decode('ascii').strip()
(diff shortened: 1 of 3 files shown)


mbolivar-nordic:
@marc-hb I will file a bug for the issue fixed in #514 when I get back to west this week, but hopefully you'll agree that the commit logs and test cases are clear enough here to not bother with that for this one.

src/west/manifest.py
```diff
@@ -832,8 +832,8 @@ def sha(self, rev: str, cwd: Optional[PathType] = None) -> str:
         # Though we capture stderr, it will be available as the stderr
         # attribute in the CalledProcessError raised by git() in
         # Python 3.5 and above if this call fails.
-        cp = self.git(f'rev-parse {rev}', capture_stdout=True, cwd=cwd,
-                      capture_stderr=True)
+        cp = self.git(f'rev-parse {rev}^{{commit}}', capture_stdout=True,
```
carlescufi, line 835: Why is that?

```
carles@ubuntu ~/src/zephyr/zephyr (main)
$ git rev-parse v2.6.0
d477a71920b3fd4491dbef3f29906f2421ad2329
```

marc-hb: > git rev-parse v2.6.0
> d477a71920b3fd4491dbef3f29906f2421ad2329

whereas:
```
git rev-parse v2.6.0^{commit}
79a6c07536bc14583198f8e3555df6134d8822cf
```

carlescufi: Ah right. This is only for annotated tags then. Thanks!

carlescufi (approved):

src/west/manifest.py
mbolivar-nordic: > $ git rev-parse v2.6.0
> d477a71920b3fd4491dbef3f29906f2421ad2329

To expand on what @marc-hb said,

```
$ git cat-file -t d477a71920b3fd4491dbef3f29906f2421ad2329
tag
$ git cat-file -t 79a6c07536bc14583198f8e3555df6134d8822cf
commit
```

d4777a7 is the SHA of the tag object, not the commit object.

PR #518, opened 2021-07-27
west status: only print output for projects with nonempty status

mbolivar-nordic:
The 'west status' output is no longer useful with many projects in the
mix. Apply a similar cleanup to 'west status' that 'west diff' got in
a53ec10cf28b4e500b5095eec33086cac1a6d839 ("west diff: only print
output for projects with nonempty diffs").

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -697,12 +697,76 @@ class Status(_ProjectCommand):
 
         failed = []
         for project in self._cloned_projects(args, only_active=not args.all):
-            log.banner(f'status of {project.name_and_path}:')
+            # 'git status' output for all projects is noisy when there
+            # are lots of projects.
+            #
+            # We avoid this problem in 2 steps:
+            #
+            #   1. Check if we need to print any output for the
+            #      project.
+            #
+            #   2. If so, run 'git status' on the project. Otherwise,
+            #      skip output for the project entirely.
+            #
+            # In verbose mode, we always print output.
+
             try:
+                if not (log.VERBOSE or self.should_print_for(project)):
+                    continue
+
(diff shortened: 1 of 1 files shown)


mbolivar-nordic:
cc @sjg20

marc-hb:
> The 'west status' output is no longer useful with many projects in the mix

What I would prefer is some kind of `west update --if-present` option because not cloning at all the 95% of the projects I don't use is an older and also pretty good way to solve this problem

carlescufi (approved):
Looks good!

mbolivar-nordic:
> What I would prefer is some kind of `west update --if-present` option because not cloning at all the 95% of the projects I don't use is an older and also pretty good way to solve this problem

Sounds like a nice future extension -- feel free to file an issue. I think that since `west status` already ignores inactive projects by default, and the upstream zephyr manifest now has groups, you may have the ingredients for most of what you want already by just cloning a subset of active projects.

marc-hb:
> Sounds like a nice future extension -- feel free to file an issue.

Done in #519

src/west/app/project.py
```diff
@@ -697,13 +697,77 @@ def do_run(self, args, user_args):
...
+                                 stderr=subprocess.PIPE,
+                                 cwd=project.abspath)
+
+        def has_output():
+            # 'git status --porcelain' prints nothing if there
+            # are no notable changes, so any output at all
+            # means we should run 'git status' on the project.
+            stdout, stderr = None, None
+            try:
+                stdout, stderr = popen.communicate(timeout=0.1)
```
marc-hb, line 748: ~This is racy~, new bug filed:
- #981

PR #520, opened 2021-07-28
manifest: add project 'userdata' key (was: metadata)

mbolivar-nordic:
Add a new optional 'userdata' key to projects.

The value will be parsed and stored in the userdata attribute of the
corresponding west.manifest.Project object.

Example manifest file:

```yaml
  manifest:
    projects:
      - name: foo
      - name: bar
        userdata: a-string
      - name: baz
        userdata:
          key: value
    defaults:
      remote: r
    remotes:
      - name: r
        url-base: base
```

Example Python usage:

```py
  manifest = west.manifest.Manifest.from_file()

  foo, bar, baz = manifest.get_projects(['foo', 'bar', 'baz'])

  print(foo.userdata) # None
  print(bar.userdata) # prints 'a-string'
  print(baz.userdata) # prints {'key': 'value'}
```

### DIFF
--- a/src/west/manifest-schema.yml
+++ b/src/west/manifest-schema.yml
@@ -126,6 +126,10 @@ mapping:
           groups:
             required: false
             include: groups-schema
+          # Arbitrary project user data.
+          userdata:
+            required: false
+            type: any
 
   # If present, a list of project groups to enable and disable. Prefix
   # a group name with "-" to disable it; prefix with "+" to enable it.
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -645,6 +645,7 @@ class Project:
       an empty list.
     - ``submodules``: the project's submodules configuration; either
       a list of Submodule objects, or a boolean.
+    - ``userdata``: the parsed 'userdata' field in the manifest, or None
     '''
 
     def __eq__(self, other):
@@ -656,7 +657,8 @@ class Project:
                 f'clone_depth={self.clone_depth}, '
                 f'west_commands={self.west_commands}, '
                 f'topdir={repr(self.topdir)}, '
-                f'groups={self.groups})')
+                f'groups={repr(self.groups)}, '
+                f'userdata={self.userdata})')
 
     def __str__(self):
         path_repr = repr(self.abspath or self.path)
@@ -670,7 +672,8 @@ class Project:
                  west_commands: Optional[WestCommandsType] = None,
                  topdir: Optional[PathType] = None,
                  remote_name: Optional[str] = None,
-                 groups: Optional[GroupsType] = None):
+                 groups: Optional[GroupsType] = None,
+                 userdata: Optional[Any] = None):
         '''Project constructor.
 
         If *topdir* is ``None``, then absolute path attributes
@@ -702,6 +705,7 @@ class Project:
         self.topdir = os.fspath(topdir) if topdir else None
         self.remote_name = remote_name or 'origin'
         self.groups: GroupsType = groups or []
+        self.userdata: Any = userdata
 
     @property
     def path(self) -> str:
@@ -2026,13 +2030,16 @@ class Manifest:
             self._malformed(
                 f'project {name}: "groups" cannot be combined with "import"')
 
+        userdata = pd.get('userdata')
+
         ret = Project(name, url, pd.get('revision', defaults.revision), path,
                       submodules=self._load_submodules(pd.get('submodules'),
                                                        f'project {name}'),
                       clone_depth=pd.get('clone-depth'),
                       west_commands=pd.get('west-commands'),
                       topdir=self.topdir, remote_name=remote,
-                      groups=groups)
+                      groups=groups,
+                      userdata=userdata)
 
         # Make sure the return Project's path does not escape the
         # workspace. We can't use escapes_directory() as that
--- a/tests/test_manifest.py
+++ b/tests/test_manifest.py
@@ -534,7 +534,7 @@ def test_project_repr():
       west-commands: some-path/west-commands.yml
     ''')
     assert repr(m.projects[1]) == \
-        'Project("zephyr", "https://foo.com", revision="r", path=\'zephyr\', clone_depth=None, west_commands=[\'some-path/west-commands.yml\'], topdir=None, groups=[])'  # noqa: E501
+        'Project("zephyr", "https://foo.com", revision="r", path=\'zephyr\', clone_depth=None, west_commands=[\'some-path/west-commands.yml\'], topdir=None, groups=[], userdata=None)'  # noqa: E501
 
 def test_project_sha(tmpdir):
     tmpdir = Path(os.fspath(tmpdir))
@@ -547,6 +547,26 @@ def test_project_sha(tmpdir):
                       path=tmpdir.name,
                       topdir=tmpdir.parent)
     assert project.sha(project.revision) == expected_sha
+
+def test_project_userdata(tmpdir):
+    m = M('''\
+    defaults:
+      remote: r
+    remotes:
+      - name: r
+        url-base: base
+    projects:
+    - name: foo
+    - name: bar
+      userdata: a-string
+    - name: baz
+      userdata:
+        key: value
+    ''')
+    foo, bar, baz = m.get_projects(['foo', 'bar', 'baz'])
+    assert foo.userdata is None
+    assert bar.userdata == 'a-string'
+    assert baz.userdata == {'key': 'value'}
 
 def test_no_projects():
     # An empty projects list is allowed.


carlescufi (approved):
The hands-off approach is what I like about the proposal. No objections.

mbolivar-nordic:
Since this isn't urgently needed at the moment I'm going to let this sit until @marc-hb gets back to the office and has a chance to comment.

marc-hb:
The concept and implementation look both very clean so I'm going to nitpick the name. I assume that's what is expected  for me :-) I would call this `privatedata`.

"metadata" is data about data. In my experience it has well predefined types that describe a specific piece of other data. This has neither.

"private data" is a very common concept and name in C which is a programming language fairly popular in Zephyr. The private data concept is to "pass-through" data from A to C through B where A and C have well defined data type expectations and B none (hi, void *). This looks pretty much like this PR with west in the 'B' role, the manifest author in the 'A' role and the manifest reader in the 'C' role. Does it not?

Is there any public use case for this feature already?

mbolivar-nordic:
> I'm going to nitpick the name. I assume that's what is expected from me :-)

:joy: 

> "metadata" is data about data. In my experience it has well predefined types that describe a specific piece of other data. This has neither.

I'm not sure I agree. The assumption here is that in fact the contents of this field will satisfy some sort of schema; west just doesn't know what it is and it will vary by use case.

> "private data"

I am reluctant about "private" here because API users could be forgiven for assuming that `Project.privatedata` is something they should **not** touch (it's data that is private to the Project object), when in fact the opposite is true (it's private to west itself).

I thought about alternate wording for longer than I care to admit and I think that `metadata` is still best, but what do you think about `userdata`? If I see `Project.userdata`, I personally think "oh yeah, that's for me", not "nope, I can't touch that".

> Is there any public use case for this feature already?

I originally raised this as a possible way to declare maintainers for Zephyr modules, but it hasn't been accepted for this purpose since putting that into the MAINTAINERS YAML file is also a good option.

Whether or not upstream Zephyr wants to use it for that, though, I can imagine downstream Zephyr based distributions wanting to track distro-specific project data in the manifest itself so it won't go stale. (I actually don't have to imagine this very hard as I am pushing for this for a Nordic use case.)

marc-hb:
> The assumption here is that in fact the contents of this field will satisfy some sort of schema; west just doesn't know what it is and it will vary by use case.

Yes, I meant "private" as in "west doesn't know".

> I am reluctant about "private" here because API users could be forgiven for assuming that Project.privatedata is something they should not touch (it's data that is private to the Project object), when in fact the opposite is true (it's private to west itself).

You're right: private is a relative term and that sucks. `private_to_west` is obviously too long.

> maintainers for Zephyr modules, but it hasn't been accepted

Thanks, any example is useful.

> I thought about alternate wording for longer than I care to admit

/me thought my nitpicking expertise was indispensable...

> but what do you think about userdata? If I see Project.userdata, I personally think "oh yeah, that's for me", not "nope, I can't touch that".

I find `userdata` much better!


Mandatory reference: https://martinfowler.com/bliki/TwoHardThings.html

marc-hb (approved):

gregshue:
How will this field in the manifest be validated?

What happens when there are multiple users of a particular field in the project (e.g., a zephyr script needs module-specific metadata, like maintainer, and module-provided scripts specify new manifest-provided configuration settings, analogous to path, for their execution)?

It seems to me that within each west project the userdata should be a list of schema ID/content.  Perhaps the schema ID should be a URI or URL that can be used for validation?

mbolivar-nordic:
> How will this field in the manifest be validated?

It must be a valid YAML map value or the manifest file won't parse. Beyond that, west doesn't care, by design.

> What happens when there are multiple users of a particular field in the project (e.g., a zephyr script needs module-specific metadata, like maintainer, and module-provided scripts specify new manifest-provided configuration settings, analogous to path, for their execution)?

That's up to the user to figure out, not west.

>  It seems to me that within each west project the userdata should be a list of schema ID/content. Perhaps the schema ID should be a URI or URL that can be used for validation?

You are free to put a list of whatever you want in the userdata field, since it can take any value.

marc-hb:
@mbolivar-nordic 's approach where you either 'inside' west or 'outside' with no middle ground and no compromise in between sounds like the more maintainable and sustainable to me.

gregshue:
> > How will this field in the manifest be validated?
> 
> It must be a valid YAML map value or the manifest file won't parse. Beyond that, west doesn't care, by design.

"Doesn't care" sounds like it:
- ignores any collisions on multiple imports, which seems inappropriate.  What does it really do?
- does not report if there exists non-validated fields, which also seems inappropriate.  What does it really do?

marc-hb:
> ignores any collisions on multiple imports, which seems inappropriate. What does it really do?

Good point. Should be easy to test.

> does not report if there exists non-validated fields, which also seems inappropriate. What does it really do?

Can you give an example of a "non-validated" field? I wonder what is the validation.

gregshue:
> Can you give an example of a "non-validated" field? I wonder what is the validation.
'userdata' is not validated other than it parses correctly.  Validation should include making sure it has content/structure meaningful to the consumer(s).  For a field like `userdata`, the client would have to provide the schema for field validation schema along with a tag identifying the originator (for private use) or purpose/content (for shared use).

The questions come from looking at the native functionality of west, which includes:
- manifest --resolve, --freeze, and --validate.
- support for nested importing of manifest files

At the very least the behavior of conflict resolution during --resolve and --freeze need to be useful to the top-level integrator (who may know nothing about any of multiple clients that expect to consume their relevant portion of `userdata`.

marc-hb:
> Validation should include making sure it has content/structure meaningful to the consumer(s). 

That's really not how I understand this new feature. I understand this feature as a blind "pass-through", a mere pipe carrying whatever with validation being performed at either the input or the output or both or neither but never inside the dumb west pipe.

> For a field like userdata, 

AFAIK there is no other field "like userdata", it's very unique because it's the only dumb pipe through west. 

> the client would have to provide the schema for field validation schema along with a tag identifying the originator (for private use) or purpose/content (for shared use).

Who are the "client" and "originator", are these defined somewhere?

Again I believe the idea of this feature is to have validation and interpretation entirely outside of west so I'm afraid we are talking across each other.

> manifest --resolve, --freeze, and --validate.

I wonder about --resolve, --freeze and imports too. @mbolivar-nordic ?

--validate should IMHO ignore userdata content entirely, merely checking that the YAML syntax valid.

gregshue:
@marc-hb 
> That's really not how I understand this new feature. 

Understood.  @mbolivar-nordic described what the feature does (implementation).  I fully expect he built the functionality right (passing verification).

I am talking about what I, as a user, expect it to do to be useful in a scaled up system (tool "requirements"). It seems to me he has not build the right functionality yet (failing validation).

> AFAIK there is no other field "like userdata", it's very unique because it's the only dumb pipe through west.

As I see it, "userdata" is the first time the manifest is opened up for extensibility (including downstream-specified content without modifying `west`).  "Dumb passthrough" is how the current design has chosen to _support_ the `--resolve, --freeze, and --validate` native commands and `import` mechanism.

> Who are the "client" and "originator", are these defined somewhere?

`west` should not care _who_ the clients are, but it should care _that_ :
- there can be an unbounded number of clients
- clients are likely unaware of each other
- userdata for different clients needs to be retained and not considered a conflict

The above items needs to be solved in a common way to have a workable solution.  The only common place that it can be solved is within `west`.

marc-hb:
> west should not care who the clients are, but it should care that :
>
> there can be an unbounded number of clients
> clients are likely unaware of each other
> userdata for different clients needs to be retained and not considered a conflict

I think what you describe here can be summarized as "demultiplexing the upper layer". In this context it would typically be achieved with a registry of keys reserved at the top-level. There is nothing that stops you from creating such a registry inside your organization. Are you suggesting the zephyr project or the west project even should maintain such a registry?

gregshue:
> I think what you describe here can be summarized as "demultiplexing the upper layer". 

That is only one technique for supporting extensibility.  The issue exists because extensibility itself was introduced.

> In this context it would typically be achieved with a registry of keys reserved at the top-level.

It does not need to be a central registry.  As far as `west` is concerned a URI would suffice.

> There is nothing that stops you from creating such a registry inside your organization.

True, but we also need to resolve conflicts with content for tools from outside my organization.  Remember, my need is to REUSE the released manifest settings via `import` and extend the project `userdata` with content for my own purposes.

marc-hb:
> > I think what you describe here can be summarized as "demultiplexing the upper layer".

> That is only one technique for supporting extensibility.

Sure but this is the west bug tracker and in case you haven't noticed yet there's fairly strong desire not to get into things that can be dealt with outside of west. From a protocol perspective Ethernet does not need to know anything about IPv4 or IPv6, it only needs to be told which one is which and that's all. I think this has been demonstrated to be extensible enough.

> It does not need to be a central registry. As far as west is concerned a URI would suffice.

... and URIs are unique because the DNS is already a central registry, got it. How about sketching that in a new github issue? Keep in mind my first sentence in this comment :-)

gregshue:
> From a protocol perspective Ethernet does not need to know anything about IPv4 or IPv6, it only needs to be told which one is which and that's all.

From a _protocol_ perspective, Ethernet does not even need to be told IPv4 vs IPv6.  Ethernet simply sends and receives frames of a higher layer.  (It must have it's own way of indicating large vs standard MTU sizes.) The agent receiving the validated frame data must decode which protocol it contains. Any agent combining packets or handling decoding of frame data is doing work above the Ethernet protocol.

Following the analogy, when `west` processes `import` directives or executes `--resolve` it is working above the manifest file syntax level.  Aggregating the `userdata` from multiple manifests for a specific project is exactly the responsibility of `west`.  If west silently discards some values then the feature is limited to a single client and is likely to never get used.  If west aborts the `import` when multiple values exist, then `import` becomes impractical. If west requires `userdata` to be a list of dictionaries where each entry requires a URI as a consumer/format tag along with the data itself, and an optional URL for a validation schema, then aggregating content is straight forward and clients don't have to be enhanced.

> > > Are you suggesting the zephyr project or the west project even should centrally maintain such a registry?
> > 
> > It does not need to be a central registry. As far as west is concerned a URI would suffice.
> 
> URIs are unique because the DNS is already a central registry

Note my earlier response was to needing a central registry maintained by either west or the Zephyr Project.  It said nothing about not using a central registry maintained elsewhere.  :-)

> How about sketching that in a new github issue?

We could, but this has so much context captured already!  :-)

How do we get this discussion chain added/migrated/referenced in a new issue?

marc-hb:
> From a protocol perspective, Ethernet does not even need to be told IPv4 vs IPv6.

https://en.wikipedia.org/wiki/EtherType#Values

> Aggregating the userdata from multiple (imported) manifests for a specific project is exactly the responsibility of west

I think we all agree on that. Did you try it? How does it behave now?

> We could, but this has so much context captured already! :-)

Captured but diluted in numerous clarifications and digressions. Plus you can't re-open a merged PR.

> If west requires userdata to be a list of dictionaries where each entry requires a URI as a consumer/format tag along with the data itself, and an optional URL for a validation schema, then aggregating content is straight forward and clients don't have to be enhanced.
> ...
> How do we get this discussion chain added/migrated/referenced in a new issue?

By extracting, cleaning up and summarizing into a new issue linked to this merged PR:
- the problem statement
- a (supposedly failing) example
- one possible recommended solution/behavior.

gregshue:
> https://en.wikipedia.org/wiki/EtherType#Values
OK, I'm out-of-date.  Isn't this still just a type indicator used to route to the higher layer and possibly to imply data size for validation?

> Did you try it? How does it behave now?
I have not had time to try it.  Please remember that the question is not _can_ it work, but will it _always_ work.

Please feel free to do so. It will be awhile before I can.

marc-hb:
> Please remember that the question is not can it work, but will it always work.

Another important question is: who wants or needs this.

gregshue:
> Another important question is: who wants or needs this.

Great question!  I haven't heard any requirement for the `userdata` feature.  Now that it's in we need to make it broadly useful.

mbolivar-nordic:
Marc Herbert ***@***.***> writes:

>
> I wonder about --resolve, --freeze and imports too. @mbolivar-nordic ?
>
> --validate should IMHO ignore userdata content entirely, merely checking that the YAML syntax valid.

Yep, that's an issue. #536.

mbolivar-nordic:
>> Aggregating the userdata from multiple (imported) manifests for a specific project is exactly the responsibility of west
>
> I think we all agree on that.

No; this is not the model.

West doesn't do any aggregation whatsoever along these lines, by design.

If you define a project named 'foo' in your top level manifest, and
another manifest you import also defines 'foo', the latter is ignored.

This is well documented in the section in the documentation which
introduces manifest imports:

> During resolution, west ignores projects which have already been
> defined in other files.

PR #521, opened 2021-07-28
manifest: deny using `/` and `\` as project names

M1cha:


src/west/manifest.py
```diff
@@ -348,6 +348,7 @@ def _compose_imap_filters(imap_filter1: ImapFilterFnType,
         return imap_filter1 or imap_filter2
 
 _RESERVED_GROUP_RE = re.compile(r'(^[+-]|[\s,:])')
+_INVALID_PROJECT_NAME_RE = re.compile(r'(^[+-]|[/\\])')
```
mbolivar-nordic, line 351: What's up with the `^[+-]` part of the regular expression?

mbolivar-nordic: Is this just copy/pasted from _RESERVED_GROUP_RE? If so, it should just be removed; you can use `r'[/\\]'` as the regular expression.

The `^[-+]` is in _RESERVED_GROUP_RE because we forbid groups from starting with - or +.

M1cha: I was sure I removed this :D 
sorry about that

mbolivar-nordic (approved):

Issue #523, opened 2021-07-30
west.manifest.Manifest.get_projects() is doing too much work, incorrectly


mbolivar-nordic:
This method is used to look up projects by "project IDs". Project IDs may be either the names of, or paths to, projects defined in the manifest.

The Manifest class also already has these dicts as instance methods:

- _projects_by_name: a map from project names to Project instances, with the special 'manifest' key pointing at the ManifestProject
- _projects_by_rpath: a map from resolved Path objects to Project instances; this also maps the ManfiestProject object's path to the object itself

So get_projects() can just rely on these dicts to do its lookups. However, it's currently got extra special-case code it doesn't need to handle

- a 'manifest' project ID
- the manifest project's absolute path
 
And what's more, it's doing the absolute path work incorrectly. Specifically:

```py
mpath: Optional[Path] = Path(mp.path).resolve()
```

Here `mp` is the ManifestProject instance. This is trying to calculate a resolved path to the manifest, but it only works if you're in the workspace root, since `mp.path` is relative, so it's wrong.

We got lucky and can still find the manifest project by its relative path because we fall back on _projects_by_rpath, which is probably why nobody noticed until https://github.com/zephyrproject-rtos/zephyr/issues/37319. But it can lead to returning the ManifestProject incorrectly in some rare edge cases, so it's still a bug.

PR #524, opened 2021-07-30
manifest: simplify and fix get_projects()

mbolivar-nordic:
Fixes: #523

The Manifest class has these instance attributes:

- _projects_by_name: maps project names -> Projects, has 'manifest' as
  a key and the ManifestProject as a value (this is one reason why
  'manifest' is a reserved project name)

- _projects_by_rpath: maps resolve()d Paths -> Projects, has the
  manifest repository's path as a key, pointing to the ManifestProject

So get_projects() ought to just rely on these dicts to do its job.

However, it's currently got extra special-case code it doesn't need,
and the code is buggy. Specifically this part:

mpath: Optional[Path] = Path(mp.path).resolve()

This is trying to get the ManifestProject's resolved path, but it only
works if you happen to be in the workspace topdir, since mp.path is
relative.

We got "lucky" and can still find the manifest project by its relative
path because we fall back on _projects_by_rpath.

But this is still a correctness problem. When the current
directory *isn't* the topdir, but has a subdirectory or file inside it
whose name happens to match mp.path, we could return the
ManifestProject as a result, incorrectly.

Just remove all the unnecessary extra code and use _projects_by_name /
_projects_by_rpath to do the work.

### DIFF
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -1384,35 +1384,23 @@ class Manifest:
 
         # Otherwise, resolve each of the project_ids to a project,
         # returning the result or raising ValueError.
-        mp = self.projects[MANIFEST_PROJECT_INDEX]
-        if mp.path is not None:
-            mpath: Optional[Path] = Path(mp.path).resolve()
-        else:
-            mpath = None
         for pid in project_ids:
+            project: Optional[Project] = None
+
             if isinstance(pid, str):
-                if pid == 'manifest':
-                    project: Optional[Project] = mp
-                else:
-                    project = self._projects_by_name.get(pid)
-            else:
-                project = None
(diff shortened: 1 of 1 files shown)


mbolivar-nordic:
@gregshue this removes the buggy `mpath` line you mentioned in https://github.com/zephyrproject-rtos/zephyr/issues/37319, but does not make the `Path(self.topdir).joinpath(pid).resolve()` change you applied to your tree.

That change is not correct because any path arguments to get_projects are relative to the current working directory (they may also be absolute).

That API design decision was made so commands like `west list .` work properly when you are inside a project directory (`west list` and other commands that may take projects on the command line all use `get_projects` internally).

gmarull (approved):

carlescufi (approved):

PR #533, opened 2021-08-12
project: overwrite stdout/stderr objects as well

M1cha:


mbolivar-nordic (commented):
Thanks for the quick fix. I'm going to give it some more thorough testing tomorrow before merging, but it looks good to me.

mbolivar-nordic:
This is no longer crashing, but I'm not getting the right output with `color.ui` set to its default value. The control characters are coming through, so I see e.g.:

```
←[92m=== updating hal_cypress (modules/hal/cypress):
```

mbolivar-nordic:
@M1cha any updates? I really would like to have this feature, but I will need to revert it if we can't support it on all platforms.

M1cha:
sry I forgot about this.
I might have missed it because I was using the windows VM through SSH from linux.
Does windows usually support colors or should they always be stripped on windows?

mbolivar-nordic:
Windows supports colors; colorama is cross platform.

marc-hb:
I think it's really not obvious why the file objects _and_ the file descriptors must BOTH be saved and restored. So I think parts of the commit message should be transferred to the source code.

To be honest I don't understand the explanation in the commit message either but I'm assuming that's just me.

> This is no longer crashing, but I'm not getting the right output with color.ui set to its default value. The control characters are coming through, so I see e.g.:

Same with both `cmd` and Powershell? Just curious.

marc-hb:
> @M1cha any updates? I really would like to have this feature, but I will need to revert it if we can't support it on all platforms.

Reverting even Linux because the colors are broken on Windows would sound a bit harsh...

Reverting Linux because Windows crashes: still a bit harsh if it crashes only when explicitly using the new -j option but: OK.

mbolivar-nordic:
I have reason to believe the majority of our users are on Windows, so as much as it pains me, I feel an obligation to only let in features that work well cross platform. If we start down the path of "meh, Windows is a pain to program for, let me just fix it for Linux", then we'll end up in a situation where west's developers and maintainers have a much better experience than its users :disappointed:

mbolivar-nordic:
> Same with both `cmd` and Powershell? Just curious.

I only tested with `cmd`.

marc-hb:
If the situation had been reversed I would have written the same:
> Reverting even Windows because the colors are broken on Linux would sound a bit harsh...

In other words, I would personally prefer to have this (optional and non-default!) feature with broken colours rather than not at all. This is really a "killer" feature.

Releasing it widely also makes it more likely to land in the hands of some Windows expert who might have ideas on how to fix it.

No offence but west is a development tool, not a consumer product. What I'm trying to say is raw performance trumps polish, especially when the default behaviour is still "slow and beautiful".

> I only tested with cmd.

If the colors are broken only with the obsolete and unmaintained cmd that no one at Microsoft still uses then... "ship it?"

M1cha:
valid discussion but in this case it's an easy fix.
The fix was not related to the buffering though. let me know if you want a separate PR for that.

mbolivar-nordic:
I gave this a try but now:

1. I'm not seeing any color at all on Windows with `west update -j3`
2. I'm getting an exception when I try that

Traceback is weird:

```
=== updating trusted-firmware-m (modules/tee/tfm):
HEAD is now at c74be389 trusted-firmware-m: platform: lpcxpresso55s69: Update SDK
Exception ignored in: <finalize object at 0x777ab8; dead>
  File "C:\Users\Marti\AppData\Local\Programs\Python\Python39-32\lib\weakref.py", line 580, in __call__
    return info.func(*info.args, **(info.kwargs or {}))
  File "C:\Users\Marti\AppData\Local\Programs\Python\Python39-32\lib\tempfile.py", line 816, in _cleanup
    cls._rmtree(name)
  File "C:\Users\Marti\AppData\Local\Programs\Python\Python39-32\lib\tempfile.py", line 812, in _rmtree
    _shutil.rmtree(name, onerror=onerror)
  File "C:\Users\Marti\AppData\Local\Programs\Python\Python39-32\lib\shutil.py", line 737, in rmtree
    return _rmtree_unsafe(path, onerror)
  File "C:\Users\Marti\AppData\Local\Programs\Python\Python39-32\lib\shutil.py", line 615, in _rmtree_unsafe
    onerror(os.unlink, fullname, sys.exc_info())
  File "C:\Users\Marti\AppData\Local\Programs\Python\Python39-32\lib\tempfile.py", line 804, in onerror
    cls._rmtree(path)
  File "C:\Users\Marti\AppData\Local\Programs\Python\Python39-32\lib\tempfile.py", line 812, in _rmtree
    _shutil.rmtree(name, onerror=onerror)
  File "C:\Users\Marti\AppData\Local\Programs\Python\Python39-32\lib\shutil.py", line 737, in rmtree
    return _rmtree_unsafe(path, onerror)
  File "C:\Users\Marti\AppData\Local\Programs\Python\Python39-32\lib\shutil.py", line 596, in _rmtree_unsafe
    onerror(os.scandir, path, sys.exc_info())
  File "C:\Users\Marti\AppData\Local\Programs\Python\Python39-32\lib\shutil.py", line 593, in _rmtree_unsafe
    with os.scandir(path) as scandir_it:
NotADirectoryError: [WinError 267] The directory name is invalid: 'C:\\Users\\Marti\\AppData\\Local\\Temp\\tmp0reutx4l\\trusted-firmware-m'
```

Python 3.9.0, Windows 10, cmd.exe. This is a 100% fresh zephyr workspace,

```
west init zephyrproject
cd zephyrproject
west update # works fine, i get color
west update -j3 # crashes with the above traceback
```

marc-hb:
Also, I insist:
> I think it's really not obvious why the file objects and the file descriptors must BOTH be saved and restored. So I think parts of the commit message should be transferred to the source code.

M1cha:
@mbolivar-nordic why do you expect to see colors in buffered mode? In linux there shouldn't be any colors either.

The stacktrace looks like a destructor for one of the tempdirs tries to delete the directory recursively. No idea why it's failing though.

mbolivar-nordic:
> Also keep in mind that I never questioned support for either powershell or cmd. I find those discussions interesting but they have nothing to do with this PR and they don't reflect or affect my willingness to work on this.

@M1cha You're totally right: you never questioned cmd support. I'm sorry to have implied otherwise by using the word "we" earlier. I was mainly reacting to this comment:

> If the colors are broken only with the obsolete and unmaintained cmd that no one at Microsoft still uses then... "ship it?"

This I think is a pretty clear questioning of that ;).

mbolivar-nordic:
Did you mean #551 to replace this? Please close if so.

marc-hb:
Just for record, another colorama issue on Windows: https://github.com/zephyrproject-rtos/west/issues/634

marc-hb:
... and of course switching to PowerShell, away from cmd.exe avoids colorama issue #634 :-)

PR #535, opened 2021-08-19
Support the MSYS2 platform

jeffrizzo:
It's pretty much like Cygwin as far as these things go, so
do the same things.

### DIFF
--- a/src/west/configuration.py
+++ b/src/west/configuration.py
@@ -233,7 +233,7 @@ def _location(cfg: ConfigFile, topdir: Optional[PathType] = None) -> str:
         if 'BSD' in plat:
             return '/etc/westconfig'
 
-        if 'CYGWIN' in plat:
+        if 'CYGWIN' in plat or 'MSYS_NT' in plat:
             # Cygwin can handle windows style paths, so make sure we
             # return one. We don't want to use os.path.join because
             # that uses '/' as separator character, and the ProgramData


src/west/configuration.py
```diff
@@ -244,6 +244,11 @@ def _location(cfg: ConfigFile, topdir: Optional[PathType] = None) -> str:
             pd = PureWindowsPath(os.environ['ProgramData'])
             return os.fspath(pd / 'west' / 'config')
 
+        if 'MSYS_NT' in plat:
+            # MSYS2 is supported like Cygwin
```
mbolivar-nordic, line 248: OK, but then please write this as follows to avoid duplication:

```py
if 'CYGWIN' in plat or 'MSYS_NT' in plat:
```

mbolivar-nordic (changes requested):
No objections, but I'd like to see a small code change first.

@tejlmand you use MSYS2, right? Or maybe I'm misremembering. FYI if so.

src/west/configuration.py
jeffrizzo: I made the suggested change, and force-pushed.

mbolivar-nordic (approved):

tejlmand:
> @tejlmand you use MSYS2, right? Or maybe I'm misremembering. FYI if so.

Nope.

Issue #536, opened 2021-09-10
project userdata bugs


mbolivar-nordic:
I found 2 bugs playing around with project userdata:

1. It doesn't round trip in e.g. `west manifest --freeze` invocations
2. It's missing a `repr()` in `Project.__repr__`

PR #537, opened 2021-09-10
manifest: fix some Project.userdata issues

mbolivar-nordic:
Make sure the userdata round-trips properly through 'west manifest
--freeze' by adding it to the project's dict representation.

Add a missing repr() inside the userdata related part of
Project.__repr__.

Fixes: #536

### DIFF
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -658,7 +658,7 @@ class Project:
                 f'west_commands={self.west_commands}, '
                 f'topdir={repr(self.topdir)}, '
                 f'groups={repr(self.groups)}, '
-                f'userdata={self.userdata})')
+                f'userdata={repr(self.userdata)})')
 
     def __str__(self):
         path_repr = repr(self.abspath or self.path)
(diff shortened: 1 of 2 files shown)


marc-hb:
Is there no simple round-trip test that would automatically detect any future mismatch? On any field.

marc-h38 (approved):

mbolivar-nordic:
> Is there no simple round-trip test that would automatically detect any future mismatch? On any field.

Sorry, I don't understand the question. The test function being changed already has some round trip tests. The part that was missing was making sure they are also reflected in the project's as_dict() return value.

The Project.as_dict() API is used by the Manifest class when forming the data structure that eventually gets serialized as a YAML string representing the entire manifest.

mbolivar-nordic:
> **[marc-h38](/marc-h38) ** approved these changes [ 12 hours ago](#pullrequestreview-751015370)

Looks like you used the wrong account here; mind using @marc-hb?

marc-hb (approved):

marc-hb:
I mean is there no test that simply parses the entire YAML file, spits it back and compares?

If there is one, then it should have already failed before this commit.

mbolivar-nordic:
test_manifest_freeze() in project.py does roughly what you are asking for, but it didn't use the new userdata feature.

marc-hb:
> test_manifest_freeze() in test_project.py does roughly what you are asking for, but it didn't use the new userdata feature.

So how about adding every existing (and future) field as its mock input?

Issue #539, opened 2021-09-16
west does not follow a moving tag


dkalowsk:
As requested on Discord.

Our local build system generates and updates a git tag on some regularly cadence, for example a "nightly" tag that moves forward when the daily CI runs pass.  Under previous versions of west (I believe 0.10.1 was what I was at), a `west update` would correctly follow the tag, and pull down the now latest version of that repository.

With west v 0.11.1 this pattern no longer works.  It appears that west determines it already has the tag and therefore does not need to run git fetch on it.

This can be worked around by running `west update -f always` (which can be made default with a `west config update.fetch always`} but that appears to pull down a lot more detail than needed.  

Example west.yml:
```
    - name: our_module
      remote: github
      repo-path: some/path/our_module
      revision: NIGHTLY
```

mbolivar-nordic:
I'm guessing this is a lightweight tag. If so, west used to treat those as commits, which would make this work, but that was due to a bug which was fixed in https://github.com/zephyrproject-rtos/west/commit/121dfd8ac41a333c0127c6f29c299e1bbe05240b. So I would expect -falways to be required for all sorts of tags now.

marc-hb:
>  for example a "nightly" tag that moves forward when the daily CI runs pass.

Why use a tag instead of a branch? The only difference between a (light weight) tag and a branch and the latter moves while the former does not.

https://git-scm.com/book/en/v2/Git-Basics-Tagging
> A lightweight tag is very much like a branch that **does not change** — it’s just a pointer to a specific commit.

https://git-scm.com/docs/git-tag#_on_re_tagging

> 1. The sane thing. Just admit you screwed up, and use a different name.
> 2. The INSANE thing. You really want to call the new version "X" too, even though others have already seen the old one.

dkalowsk:
@marc-hb diving into a debate about how to best use git is likely beyond the scope here.  I don't disagree with your general thought, but the system I have to work with is already established and not eligible for a change at this time.  The key point is that a behavior that did exist, worked "correctly" before, is likely used and depended upon by someone else (Hyrum's law, some xkcd, etc etc).  The release notes in west don't appear to highlight this behavioral change, which means it will come as a surprise to those using it.

@mbolivar-nordic thanks for the pointer to the commit.  Why was a lightweight tag being treated as a commit considered a bug?  Was there some larger issue attached to it beyond the `cat-file` output?

marc-hb:
[...]

PR #541, opened 2021-09-16
update: add --fetch=tags option

mbolivar-nordic:
Like '--fetch=smart', this makes 'west update' skip fetches for SHAs
that are already available locally. But unlike '--fetch=smart', it
forces any tags to get fetched again.

Although tags "shouldn't" change, in practice they do sometimes, and
this makes it a little faster to update the workspace in a way that
doesn't re-fetch SHAs unnecessarily.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -701,6 +701,8 @@ class Status(_ProjectCommand):
 
 class Update(_ProjectCommand):
 
+    FETCH_STRATEGIES = ['always', 'smart', 'tags']
+
     def __init__(self):
         super().__init__(
             'update',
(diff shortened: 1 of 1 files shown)


marc-hb:
Isn't `--always --narrow` a faster and better solution for #539?

The code change is relatively small and clear but I'm afraid I'm getting more or more lost in the different fetch options.

marc-hb:
> Isn't --fetch=always --narrow a faster and better solution for #539?

BTW `update: fix --tags explanation in --narrow help string` #542

mbolivar-nordic:
> Isn't `--fetch=always --narrow` a faster and better solution for #539?
> 
> 
> 

It's not faster in the usual case where a small number of projects have tags, and a large number have SHAs.

marc-hb:
> It's not faster in the usual case where a small number of projects have tags, and a large number have SHAs.

OK. If it's not slower either then maybe there's no value in this new `--fetch=tags`. The submitter of #539 seemed happy with `--fetch=always --narrow`. I mean happy with the performance, obviously not happy with having to use _any_ option.

tejlmand:
I don't think I like this feature.

git is very flexible and allows you to mess up pretty bad, but `west` is a tool that should help users.
Make multi-repo easier, but also be a shining star guiding users to better git workflows (or at least avoid the most terrible workflows out there :wink: )

And part of such help / guide is to have some kind of good / recommended workflow while still allow sufficient flexibility.
Users with very specialized workflows can still resort to plain git.

Introducing a `--fetch=tags` is like saying, `Yeah, we know moving tags is indication of bad workflow but we want to support it easily.`
I don't think `west` should go down that path, and as it has already been shown, `--fetch=always --narrow` accomplishes the same thing.

But I will not block this feature if others think it is of high value.

tejlmand:
> Although tags "shouldn't" change, in practice they do sometimes, and
this makes it a little faster to update the workspace in a way that
doesn't re-fetch SHAs unnecessarily.

also from link provided earlier my @marc-hb:
https://git-scm.com/docs/git-tag#_on_re_tagging

> Does this seem a bit complicated? It should be. There is no way that it would be correct to just "fix" it automatically. People need to know that their tags might have been changed.

This feature would allow workflows that change tags without people knowing it.

mbolivar-nordic:
I'm convinced by the resistance to this patch.

PR #542, opened 2021-09-16
update: fix --tags explanation in --narrow help string

marc-hb:
For 'git fetch', neither --tags nor --no-tags is the default value. The
default value is a middle ground that fetches "any tag that points into
the histories being fetched is also fetched" - second paragraph in "git
help fetch". Rather than duplicating this relatively non-intuitive git
help in such a small space, be completely transparent about what the
west code does and lead the reader to the original git documentation.


mbolivar-nordic (approved):

marc-hb:
More `git fetch --tags` adventures: https://github.com/thesofproject/sof/pull/8114/files

PR #546, opened 2021-10-06
project: Add `origin` remote

### LINKED ISSUE #545: west update fails on submodules with relative paths
When west initializes project repository it only adds remote with name specified in `west.yml` while git assumes that default remote is `origin`. Relative submodule paths are resolved to default remote. 
Without `origin` remote, `west update` fails while trying to update repository that contains submodules with relative paths, with following warning
 ```
warning: could not look up configuration 'remote.origin.url'. Assuming this repository is its own authoritative upstream.
```
git logic for resolving relative paths can be found here https://github.com/git/git/blob/v2.20.1/builtin/submodule--helper.c#L135
Manually adding `origin` remote solves the issue.
i'd propose west should also add `origin` remote to projects by default.

### PR DESCRIPTION
**Pull request by Mierunski**
Git submodules with relative paths fail to update when repository
does not contain `origin` remote. This commit adds `origin` remote
ass addition to current remotes.

fixes #545

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -1243,6 +1243,14 @@ class Update(_ProjectCommand):
             # However, west always fetches project data by URL, not name.
             # The user is therefore free to change the URL of this remote.
             project.git(f'remote add -- {project.remote_name} {project.url}')
+
+            # Git submodules that use relative paths resolve them based
+            # on `origin` remote, make sure we have it
+            try:
+                project.git(f'remote get-url -- origin')
+            except subprocess.CalledProcessError:
+                project.git(f'remote add -- origin {project.url}')
+
         else:
             log.small_banner(f'{project.name}: cloning from {cache_dir}')
             # Clone the project from a local cache repository. Set the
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/west/app/project.py [hunk 1]
```diff
@@ -1243,6 +1243,15 @@ def init_project(self, project):
...
             project.git(f'remote add -- {project.remote_name} {project.url}')
+
+            # Git submodules that use relative paths resolve them based
+            # on `origin` remote. Adding `origin` allows for usage of
+            # relative submodule paths.
+            try:
+                project.git(f'remote add -- origin {project.url}')
+            except subprocess.CalledProcessError as e:
+                if e.returncode != 128: # Remote already exists
+                    raise(e)
```
marc-hb, line 1254: I'm concerned this may be trusting `git` exit codes a bit too much. I didn't find anything about exit codes in git man pages. This also triggers an exception in the most common case which is counter-intuitive. How about something that does not check `returncode` instead?
```python
# ... based on `origin` remote, make sure we have it
try:
          project.git(f'remote get-url -- origin')
          project.git(f'remote add -- origin {project.url}')
```

Issue #527, opened 2021-08-05
west update -jN fails on windows


mbolivar-nordic:
With west [v0.12.0a2](https://github.com/zephyrproject-rtos/west/commit/06b063556418312bb447a56ccf8a81b131d004f6), on my windows 10 box running python 3.9.0:

```
C:\Users\Marti\zephyrproject>west update -j5
multiprocessing.pool.RemoteTraceback:
"""
Traceback (most recent call last):
  File "c:\users\marti\v\west-0.12\lib\site-packages\west\app\project.py", line 874, in _update_concurrent_thread
    self.update(project)
  File "c:\users\marti\v\west-0.12\lib\site-packages\west\app\project.py", line 1132, in update
    log.banner(f'updating {project.name_and_path}:')
  File "c:\users\marti\v\west-0.12\lib\site-packages\west\log.py", line 91, in banner
    inf('===', *args, colorize=True)
  File "c:\users\marti\v\west-0.12\lib\site-packages\west\log.py", line 82, in inf
    print(*args)
OSError: [WinError 6] The handle is invalid

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "C:\Users\Marti\AppData\Local\Programs\Python\Python39-32\lib\multiprocessing\pool.py", line 125, in worker
    result = (True, func(*args, **kwds))
  File "c:\users\marti\v\west-0.12\lib\site-packages\west\app\project.py", line 880, in _update_concurrent_thread
    sys.stdout.flush()
OSError: [WinError 6] The handle is invalid
"""

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "C:\Users\Marti\AppData\Local\Programs\Python\Python39-32\lib\runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "C:\Users\Marti\AppData\Local\Programs\Python\Python39-32\lib\runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "C:\Users\Marti\v\west-0.12\Scripts\west.exe\__main__.py", line 7, in <module>
  File "c:\users\marti\v\west-0.12\lib\site-packages\west\app\main.py", line 779, in main
    app.run(argv or sys.argv[1:])
  File "c:\users\marti\v\west-0.12\lib\site-packages\west\app\main.py", line 106, in run
    self.run_command(argv)
[...]

PR #549, opened 2021-10-13
Revert parallel update

mbolivar-nordic:
Parallel updates are a great feature, but they break color.ui and attempts to fix that have caused unexpected exceptions to be thrown on Windows, so we need to back the feature out for now until we can get it done in a way that works with both color and Windows.

Fixes: #527
Is an alternative to: #533

Given what appears like limited cycles to test this feature out all around right now, I'm just proposing a revert. I want to get 0.12 out the door so the userdata feature is available more widely, and I don't want the main branch to keep diverging from v0.12-branch, so my plan is to just make v0.12-branch match main after merging this, using some [git magic](https://lwn.net/Articles/872094/).

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -9,7 +9,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: [3.6, 3.7, 3.8, 3.9]
+        python-version: [3.7, 3.8, 3.9]
     steps:
       - uses: actions/checkout@v1
       - name: Set up Python


carlescufi:
> > EDIT: It's not as trivial as I thought, so approving this right now.
> 
> Care to elaborate a bit?

I meant that it's not just `if! windows`, because the default UI coloring is affected as well, which means that this is an issue on all platforms.

mbolivar-nordic:
> > The colors problem is not fixable though.
> > git itself (and basically every proper CLI tool) checks if stderr is a tty and disables colors if it's not.

This is not correct.

lmapii:
Sorry for posting on this closed issue, but I was looking for this exact feature. Would it be acceptable to keep the original behavior (not using `multiprocessing` at all) for `jobs=1`? I don't care much about the output formatting but much more about the performance of, e.g., my CI pipeline ...

marc-hb:
I would personally love and approve a "best effort" `-j` that does the best it can and gracefully degrades (no color, no parallelism, etc.) in problematic situations (e.g.: colorama issue on cmd.ext, other).

That degradation would of course have to be "graceful": e.g. user-friendly warning or error message but not a big stack trace as we've seen before.

To make it "even more graceful" you could _hide_ the `-j` option, I mean NOT document it or advertise it at all until it's bug-free. Maybe even print an "NOT YET SUPPORTED" warning when using `-j` in situations that already work. "Hiding" experimental features like this would be the best way to get help from everyone and fix the last remaining issues.

Would all this be user-friendly enough for @mbolivar-nordic, @carlescufi ? I'm not sure. The best place to make your own opinion is the past discussion in #533.

marc-hb:
> I don't care much about the output formatting but much more about the performance of, e.g., my CI pipeline ...

Find OTHER, unrelated optimizations for CI at https://github.com/zephyrproject-rtos/west/issues?q=label%3Aperformance+

My favorite is: `--fetch-opt=--filter=tree:0`, see #638

lmapii:
I am already using that.

PR #551, opened 2021-10-16
Parallel update

M1cha:


mbolivar-nordic:
Hey again, I'm picking up this thread once more. I wonder if you've got time to rebase this so that the "fixup" commits end up in the right places?

Were you able to reproduce the crashes on windows I reported? Did you have a fix? I don't see this information anywhere in the PR description or commit logs.

mbolivar-nordic:
> Were you able to reproduce the crashes on windows I reported? Did you have a fix? I don't see this information anywhere in the PR description or commit logs.

Ping

marc-hb:
:'-(

M1cha:
@mbolivar-nordic I'm sorry about this. This branch doesn't fix anything, it was just intended as a base to start doing that. Then I lost motivation, then forgot to reply to your questions, :roll_eyes: 

The current situation is that I have no motivation whatsoever to fix it since haven't used windows since 7 and there's a steep learning curve for me to be able to even start to debug this.

Additionally, the projects where I wanted to speed up the CI using parallel updates take WAY longer now anyway due to hardware-in-the-loop tests, so saving 30 seconds during checkout doesn't really make that much of a difference anymore.

marc-hb:
> The current situation is that I have no motivation whatsoever to fix it since haven't used windows since 7 and there's a steep learning curve for me to be able to even start to debug this.


It is very rare to find a developer comfortable with **both** Linux and Windows, especially for "user interface" / terminal issues like this. This is one of the reasons why I insisted in #533 to (temporarily) drop parity between Linux and Windows for a non-functional, non-feature optimization like this one.

If the Windows-specific, display issues found in #533 are deemed really not acceptable then another option is to simply make this optimization not available on Windows at all. Until one envious, Windows developer comes and tries to restore parity; this has been observed many times before. I believe this is how cross-platform development frequently happens: splitting the work between Linux and Windows developers. Fixing Windows-specific issues would be much, much less work than adding it from scratch (which @M1cha already did!) and to be "envious" the feature needs to be working on Linux in the first place.

As it looks like it contributed, reverting this very useful optimization because of its lack of a `--color=always` option was also throwing the baby with the bathwater.

Issue #550, opened 2021-10-15
`west list --format="{sha}"` fails


tejlmand:
When working in Zephyr, the command `west list --format="{sha}"` prints all sha's for projects, like this:
```
$ west list --format="{sha}"
c429d1ad354d655909f0869b071f3d3b3c278860
f167efe85c8c7de886f1bc47f9173cfb8a346bb5
094aeb41bb93e9199d24d665ee43e9e05d6d7b1c
....
```

but when Zephyr is not the manifest project, this command fails, for example when using Nordic Semiconductors nRF Connect SDK release from here: https://github.com/nrfconnect/sdk-nrf

`west list` is working fine:
```
$ west list
manifest     nrf                          HEAD                                     N/A
zephyr       zephyr                       c429d1ad354d655909f0869b071f3d3b3c278860 https://github.com/nrfconnect/sdk-zephyr
mcuboot      bootloader/mcuboot           7606d9cad32c95a0d6fb7e872b070c71e85f931d https://github.com/nrfconnect/sdk-mcuboot
nrfxlib      nrfxlib                      1c906daa5ea70b6f8796611c586fc010c78e4fea https://github.com/nrfconnect/sdk-nrfxlib
trusted-firmware-m modules/tee/tfm              d2e93153863e7add9fae3849faf7de430258c0de https://github.com/nrfconnect/sdk-trusted-firmware-m
...
```

but using `--format="{sha}" fails:
```
$ west list --format="{sha}"
FATAL ERROR: subprocess failed while listing manifest
```

PR #552, opened 2021-10-18
west list: fix "{sha}" for manifest project

mbolivar-nordic:
This isn't working properly because the ManifestProject's revision is
"HEAD", so it looks like an ordinary project and we try to get
manifest-rev from it. That is obviously incorrect. Fix it by detecting
the ManifestProject explicitly and doing the right thing there.

Fixes: #550

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -389,10 +389,10 @@ The following arguments are available:
             if not project.is_cloned():
                 log.die(f'cannot get sha for uncloned project {project.name}; '
                         f'run "west update {project.name}" and retry')
-            elif project.revision:
-                return project.sha(MANIFEST_REV)
-            else:
+            elif isinstance(project, ManifestProject):
                 return f'{"N/A":40}'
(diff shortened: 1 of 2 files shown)


tejlmand:
> Then it sound like this commit is also fixing a totally different, non-manifest bug for revision-less projects. That other bug must be described in the commit message too

What bug ?
I don't see a problem related to revision-less projects when running
```
$ west --version
West version: v0.11.1

$ west list --format="{name:20} {sha:20} {revision:20}"
manifest             7b6343a709b5631f785fdb910f736fa67b0f166e HEAD                
canopennode          f167efe85c8c7de886f1bc47f9173cfb8a346bb5 master              
civetweb             094aeb41bb93e9199d24d665ee43e9e05d6d7b1c 094aeb41bb93e9199d24d665ee43e9e05d6d7b1c
....
```
I made `canopenmode` revision less, and it defaults nicely to master.

I don't see a bug.

PR #555, opened 2021-11-03
Bump schema version

mbolivar-nordic:
The addition of project userdata was a breaking change for older
versions of west. This means we should have bumped the schema version
when it was added, but this was missed. Do it now.


carlescufi (approved):

mbolivar-nordic:
I will backport to v0.12-branch along with the reversion of the parallel update feature.

marc-hb:
> The addition of project userdata was a breaking change for older versions of west. 

Just for the record and links this was #520 and #537

PR #557, opened 2021-11-12
Regression test for submodules with relative paths

### LINKED ISSUE #545: west update fails on submodules with relative paths

mbolivar-nordic:
The goal of this PR is to add a regression test for #545.

I'm starting with a regression test that I think captures all the cases:

- a project with `submodules: true`
- a project with `submodules: <list-of-mappings>`
- a submodule with both `name:` and `path:`
- a submodule with only `path:`
- a submodule with submodules of its own

All submodule paths are relative. The test is currently failing as expected and is marked xfail as such.

I'm hoping to get some confirmation that this addresses the problem and I haven't missed any use cases as the discussion in the issue is quite long at this point.

### DIFF
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -871,6 +871,110 @@ def test_update_submodules_strategy(repos_tmpdir):
     assert net_tools_project.sha('HEAD', cwd=kconfiglib_dst_dir) \
            == kconfiglib_new_sha
 
+@pytest.mark.xfail
+def test_update_submodules_relpath(tmpdir):
+    # Regression test for
+    # https://github.com/zephyrproject-rtos/west/issues/545.
+    #
+    # We need to make sure that submodules with relative paths are
+    # initialized correctly even when ther is no "origin" remote.
+    #
+    # Background: unless the config variable git.$SUBMODULENAME.url is
+    # set, "git submodule init" relies on the "default remote" for
+    # figuring out how to turn that relative path into something
+    # useful. The "default remote" ends up being "origin" in the
+    # situation where we are initializing a submodule from a
+    # superproject which is a west project.
+    #
+    # However, west neither makes nor expects any guarantees
+    # about any particular git remotes being present in a project, so
+    # that only works if we set the config variable before
+    # initializing the module.
+
+    # The following paths begin with 'project' if they are west projects,
+    # and 'submodule' if they are direct or transitive submodules of a
+    # west project. Numeric suffixes reflect the tree structure of the
+    # workspace:
+    #
+    # workspace
+    # ├── manifest
+    # ├── project-1
+    # │   └── submodule-1-1
+    # └── project-2
+    #     ├── submodule-2-1
+    #     └── submodule-2-2
+    #         └── submodule-2-2-1
+    #
+    # Some of them are declared to west directly, some implicitly.
+    pseudo_remotes = tmpdir / 'remotes'  # parent directory for 'remote' repos
+    remote_repositories = [
+        pseudo_remotes / path for path in
+        ['manifest',
+         'project-1',
+         'submodule-1-1',
+         'project-2',
+         'submodule-2-1',
+         'submodule-2-2',
+         'submodule-2-2-1',
+         ]
+    ]
+    for remote_repo in remote_repositories:
+        create_repo(remote_repo)
+
+    add_commit(pseudo_remotes / 'manifest', 'manifest west.yml',
+               files={'west.yml': f'''
+manifest:
+  remotes:
+    - name: not-origin
+      url-base: file://{pseudo_remotes}
+  defaults:
+    remote: not-origin
+  projects:
+    - name: project-1
+      submodules: true
+    - name: project-2
+      submodules:
+        - path: submodule-2-1
+        - name: sub-2-2
+          path: submodule-2-2
+'''})
+
+    def add_submodule(superproject, submodule_name,
+                      submodule_url, submodule_path):
+        subprocess.check_call([GIT, '-C', os.fspath(superproject),
+                               'submodule', 'add',
+                               '--name', submodule_name,
+                               submodule_url, submodule_path])
+        add_commit(superproject, f'add submodule {submodule_name}')
+
+    add_submodule(pseudo_remotes / 'project-1', 'sub-1-1',
+                  '../submodule-1-1', 'submodule-1-1')
+    add_submodule(pseudo_remotes / 'project-2', 'sub-2-1',
+                  '../submodule-2-1', 'submodule-2-1')
+    add_submodule(pseudo_remotes / 'project-2', 'sub-2-2',
+                  '../submodule-2-2', 'submodule-2-2')
+    add_submodule(pseudo_remotes / 'project-2' / 'submodule-2-2', 'sub-2-2-1',
+                  '../submodule-2-2-1', 'submodule-2-2-1')
+
+    workspace = tmpdir / 'workspace'
+    remote_manifest = pseudo_remotes / 'manifest'
+    cmd(f'init -m "{remote_manifest}" "{workspace}"')
+
+    workspace.chdir()
+    cmd('update')
+    expected_dirs = [
+        workspace / expected for expected in
+        ['project-1',
+         'project-1' / 'submodule-1-1',
+         'project-2',
+         'project-2' / 'submodule-2-1',
+         'project-2' / 'submodule-2-2',
+         'project-2' / 'submodule-2-2' / 'submodule-2-2-1',
+         ]
+    ]
+    for expected_dir in expected_dirs:
+        expected_dir.check(dir=1)
+
 def test_update_recovery(tmpdir):
     # Make sure that the final 'west update' can recover from the
     # following turn of events:


tests/test_project.py
```diff
@@ -871,6 +871,106 @@ def test_update_submodules_strategy(repos_tmpdir):
...
+@pytest.mark.xfail
+def test_update_submodules_relpath(tmpdir):
+    # Regression test for
+    # https://github.com/zephyrproject-rtos/west/issues/545.
+    #
+    # We need to make sure that submodules with relative paths are
+    # initialized correctly even when ther is no "origin" remote.
+    #
+    # Background: unless the config variable git.$SUBMODULENAME.url is
+    # set, git relies on the "default remote" for figuring out how to
```
marc-hb, line 883: Very briefly mention that this is what `git submodule init` does: initializing this variable from the .gitmodules file.

tests/test_project.py
```diff
@@ -871,6 +871,106 @@ def test_update_submodules_strategy(repos_tmpdir):
...
+    #
+    # workspace
+    # ├── manifest
+    # ├── project-1
+    # │   └── submodule-1-1
+    # └── project-2
+    #     ├── submodule-2-1
+    #     └── submodule-2-2
+    #         └── submodule-2-2-1
+    remotes = tmpdir / 'remotes'  # parent directory for 'remote' repos
```
marc-hb, line 905:
```suggestion
    pseudo_remotes = tmpdir / 'remotes'  # parent directory for 'remote' repos
```
At least for the very first variable, no need to change them all.

tests/test_project.py
```diff
@@ -871,6 +871,106 @@ def test_update_submodules_strategy(repos_tmpdir):
...
+    remote_repositories = [
+        remotes / path for path in
+        ['manifest',
+         'project-1',
+         'submodule-1-1',
+         'project-2',
+         'submodule-2-1',
+         'submodule-2-2',
+         'submodule-2-2-1',
```
marc-hb, line 914: I'm trying to think of different names for "top" submodules known to west versus sub-submodules not know to west. This is obviously a big difference. How about `top-submodule-x-y` and `rec-submodule-x-y-z`? Or `project-1-submodule-1` for the top ones?

tests/test_project.py
```diff
@@ -871,6 +871,106 @@ def test_update_submodules_strategy(repos_tmpdir):
...
+  defaults:
+    remote: not-origin
+  projects:
+    - name: project-1
+      submodules: true
+    - name: project-2
+      submodules:
+        - path: submodule-2-1
+        - name: sub-2-2
+          path: submodule-2-2
```
marc-hb, line 935: Add a one-line comment stating that there are more not known to west.

tests/test_project.py
```diff
@@ -871,6 +871,106 @@ def test_update_submodules_strategy(repos_tmpdir):
...
+'''})
+
+    def add_submodule(superproject, submodule_name,
+                      submodule_url, submodule_path):
+        subprocess.check_call([GIT, '-C', os.fspath(superproject),
+                               'submodule', 'add',
+                               '--name', submodule_name,
+                               submodule_url, submodule_path])
```
marc-hb, line 943: When testing with pseudo remotes I always use a `file:///` prefix otherwise cloning using hardlinks and refs are all shared and always found no matter what you're trying to test fetching.

tests/test_project.py
```diff
@@ -871,6 +871,106 @@ def test_update_submodules_strategy(repos_tmpdir):
...
+        add_commit(superproject, f'add submodule {submodule_name}')
+
+    add_submodule(remotes / 'project-1', 'sub-1-1',
+                  '../submodule-1-1', 'submodule-1-1')
+    add_submodule(remotes / 'project-2', 'sub-2-1',
+                  '../submodule-2-1', 'submodule-2-1')
+    add_submodule(remotes / 'project-2', 'sub-2-2',
+                  '../submodule-2-2', 'submodule-2-2')
+    add_submodule(remotes / 'project-2' / 'submodule-2-2', 'sub-2-2-1',
+                  '../submodule-2-2-1', 'submodule-2-2-1')
```
marc-hb, line 953: I think you need to add these from the bottom up, not top down otherwise the commits at the top will be lagging behind (git status should show that)

marc-hb (commented):
Thanks for doing this! Just an early look.

marc-hb:
> I'm working on a fix in parallel along the lines of what we've been discussing, and will update the PR when I have it.

I really recommend getting the test merged test first before any tentative fix: much lower risk of controversy and will help anyone suggesting anything actually test their suggestions (and even submit test fixes if needed)

tests/test_project.py
mbolivar-nordic: I think the names are clear enough as-is, especially given the generous comments.

tests/test_project.py
mbolivar-nordic: I don't follow, sorry. How can I add a sub-submodule for a submodule that isn't added yet?

mbolivar-nordic:
> I really recommend getting the test merged test first

Very well, let's go that way then. Removed wip and addressed comments.

mbolivar-nordic:
I am cordially inviting @tejlmand to the party :)

tests/test_project.py
marc-hb: `submodule-2-2` and `submodule-2-2-1` look the same. They look like one is "only one level down" and that's it. But from a west perspective they are totally different. The last one is _invisible_ to west.

tests/test_project.py
marc-hb: OK so let's say we want A->B->C from parent to child. You certainly don't need A to add C to B (B->C). Plus it's faster to link B->C first because when you add B to A later then C comes with it "for free" without having to go back and perform another commit.

If you do A->B first, then after doing B->C then you need to go back to A to update B to the newer and "bigger" B that includes C.

BTW this test code does not seem to have any commit yet?

tests/test_project.py
mbolivar-nordic: I am sorry, but I find this comment very nitpicky. This is just a test case, and a heavily commented test case at that. It isn't like this is user-facing documentation. Can we please just leave this be?

tests/test_project.py
mbolivar-nordic: > You certainly don't need A to add C to B (B->C).

I must be missing something. I think I definitely do need A to add C to B.

I need somewhere to add C. That place is B. I cannot create B without creating A, since B lives inside A.

What am I missing? Can you phrase this suggestion in the form of a diff?

mbolivar-nordic:
Ping on this one

Mierunski (approved):
Looks good to me

tests/test_project.py
marc-hb: I'm not familiar with the repo initialization routines and I'm not sure I have the best solution, but the initialization problem I noticed when reading the code seems confirmed when trying to run it:
```
tox -- -k submodules_rel --runxfail
cd /tmp/blah/pytest-6/test_update_submodules_relpath0/remotes/

tree project-2

project-2/
├── submodule-2-1
└── submodule-2-2
    └── submodule-2-2-1

git -C project-2/ status

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	modified:   submodule-2-2 (new commits)

git clone --recursive project-2 incomplete

tree incomplete/
incomplete/
├── submodule-2-1
└── submodule-2-2
```
Also,  the remotes should all be at the same level. In a real-world example they would all be bare repos at the same level. Relative paths are complicated enough, better not add small divergences. It may be possible to simply move the remotes after initialization. As long as they're only used as remotes, only their `.git/` will be used so it should hopefully not matter whether the checkout is in some inconsistent state.

marc-hb:
Another problem I just noticed is that submodule-1-1, submodule-2-1 etc. all have the same initial commit with the same SHA1. Probably because they're created in the same second. This is a  risk of false negatives where a test fetches from the wrong place and passes anyway. Adding a few random characters in the commit message should be enough to address this.

tests/test_project.py
marc-hb: > I cannot create B without creating A, since B lives inside A

I think you're confusing or conflating initial repo creation with adding a submodule.

```
$ tree
├── A
├── B
└── C

$ git -C B submodule add ../C
Cloning into 'submtest/test/B/C'...

$ git -C B commit -m 'add ../C to B'
[master 616d32033d4a] add ../C to B
 2 files changed, 4 insertions(+)
 create mode 100644 .gitmodules
 create mode 160000 C

$ git -C A submodule add ../B
Cloning into 'submtest/test/A/B'...
done.

$ tree A/
A/
└── B
    └── C

```

carlescufi (approved):
LGTM

PR #566, opened 2022-02-17
Allow externally supplied default manifest URL

RyanLindemanCAE:
The purpose of these changes is to allow for the west tool to be used for non-Zephyr projects more easily by allowing for an externally specified manifest URL to be used instead of the Zephyr default manifest URL. This way a company could have their employees set the MANIFEST_URL environment variable to their own URL and make it easier for their developers to simply use `west init` or `west init --mr <some-revision>` without providing a complicated URL on the command line each time.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -128,7 +128,8 @@ With -m, clones the repository at that URL and uses it as the
 manifest repository. If --mr is not given, the remote's default
 branch will be used, if it exists.
 
-With neither, -m {MANIFEST_URL_DEFAULT} is assumed.
+With neither, -m {MANIFEST_URL_DEFAULT} is assumed unless
+MANIFEST_URL is set in your environment then it will be used.
 ''',
             requires_workspace=False)
 
@@ -236,7 +237,8 @@ With neither, -m {MANIFEST_URL_DEFAULT} is assumed.
         topdir = Path(abspath(args.directory or os.getcwd()))
         log.banner('Initializing in', topdir)
 
-        manifest_url = args.manifest_url or MANIFEST_URL_DEFAULT
+        mu = os.environ.get('MANIFEST_URL')
+        manifest_url = args.manifest_url or mu or MANIFEST_URL_DEFAULT
         if args.manifest_rev:
             # This works with tags, too.
             branch_opt = ['--branch', args.manifest_rev]


src/west/app/project.py
```diff
@@ -128,7 +128,8 @@ def __init__(self):
 manifest repository. If --mr is not given, the remote's default
 branch will be used, if it exists.
 
-With neither, -m {MANIFEST_URL_DEFAULT} is assumed.
+With neither, -m {MANIFEST_URL_DEFAULT} is assumed unless
+MANIFEST_URL is set in your environment then it will be used.
```
mbolivar-nordic, line 132:
```suggestion
With neither, the WEST_MANIFEST_URL environment variable will be used
if it is set. If it is unset, the default value is {MANIFEST_URL_DEFAULT}.
```

src/west/app/project.py
```diff
@@ -236,7 +237,8 @@ def bootstrap(self, args) -> Path:
         topdir = Path(abspath(args.directory or os.getcwd()))
         log.banner('Initializing in', topdir)
 
-        manifest_url = args.manifest_url or MANIFEST_URL_DEFAULT
+        mu = os.environ.get('MANIFEST_URL')
```
mbolivar-nordic, line 240:
```suggestion
        mu = os.environ.get('WEST_MANIFEST_URL')
```

mbolivar-nordic (commented):
Sure, seems like a straightforward change. I have two change requests:

- please rename the environment variable to `WEST_MANIFEST_URL` to match the existing convention used by `WEST_CONFIG_SYSTEM` and friends, and in general to keep this nicely namespaced
- grammar nitpicks

Other than that, I think this is a justified change and I'm happy to merge it.

mbolivar-nordic (changes requested):
changes requested above

marc-hb:
Environment variables are "super-globals" that are not version controlled and difficult to capture.

https://www.google.com/search?q=environment+variables+antipattern

I don't think it's a great idea to have two places running the exact same command and getting totally different results, bad for https://reproducible-builds.org/

I've never seen `git clone` or any equivalent have anything similar. I guess one reason is: you clone / init only once.

> without providing a complicated URL on the command line each time.

If you really need to run `west init` very often for some unknown reason, what's wrong with a short `west init --mr $MYMF`?

mbolivar-nordic:
> I've never seen `git clone` or any equivalent have anything similar. I guess one reason is: you clone / init only once.

OTOH, git does respect `GIT_DIR` in many commands, so it's not like there's no precedent for environment variables in git (or, as I pointed out above, in west).

mbolivar-nordic:
and actually... 

> I've never seen `git clone` or any equivalent have anything similar.

https://github.com/git/git/blob/45fe28c951c3e70666ee4ef8379772851a8e4d32/builtin/clone.c#L953

marc-hb:
I didn't know about GIT_DIR and GIT_WORK_TREE, thanks. None seems to affect `git clone`. 

> I've never seen git clone or any equivalent have anything similar. I guess one reason is: you clone / init only once.

I meant `svn checkout`, `repo init`, etc. Disclaimer: I didn't verify this, it's just I've never seen an envvar documented or recommended anywhere for this, so I think west will be the outlier. I think most people clone rarely and the rest just does `git clone $MYREPO` if they really need to; short and more flexible.

This being said, it's not a big, disrupting change either and you're the boss!

RyanLindemanCAE:
Do you want me to add some type of unit test for this? Also, how will this get documented in the Zephyr west tool instructions? Is this something I can submit a PR to help with?

mbolivar-nordic:
> I didn't know about GIT_DIR and GIT_WORK_TREE, thanks. None seems to affect `git clone`.

Are you sure? The GIT_WORK_TREE link was in the git clone implementation.

marc-hb:
> Are you sure? The GIT_WORK_TREE link was in the git clone implementation.

Too short and very bad phrasing sorry. I meant: none affects where `git clone` **downloads from**, which is the purpose of this PR.

RyanLindemanCAE:
> Sure, if you don't mind adding a test in tests/test_project.py that uses this, I'd appreciate it. It's a small enough and easy enough to maintain feature that I wasn't going to ask for that, but since you are offering, I will gladly accept.

Sorry for the delay, the unit tests have been pushed. Hopefully, you agree with how they were implemented. Let me know if there are any fixes/changes you would like me to make.

mbolivar-nordic (approved):

mbolivar-nordic:
Code changes look good -- please click on the details in the CI failures and make the linter happy. Here's a copy/paste of the failing tests so you know what to look for in the future:

```
 py3 run-test: commands[0] | python -m flake8 --config=/home/runner/work/west/west/tox.ini /home/runner/work/west/west
./tests/test_project.py:65:80: E501 line too long (88 > 79 characters)
./tests/test_project.py:1127:13: F541 f-string is missing placeholders
./tests/test_project.py:1133:9: F541 f-string is missing placeholders
./tests/test_project.py:1143:80: E501 line too long (80 > 79 characters)
./tests/test_project.py:1144:80: E501 line too long (90 > 79 characters)
```

RyanLindemanCAE:
> Code changes look good -- please click on the details in the CI failures and make the linter happy. Here's a copy/paste of the failing tests so you know what to look for in the future:
> 
> ```
>  py3 run-test: commands[0] | python -m flake8 --config=/home/runner/work/west/west/tox.ini /home/runner/work/west/west
> ./tests/test_project.py:65:80: E501 line too long (88 > 79 characters)
> ./tests/test_project.py:1127:13: F541 f-string is missing placeholders
> ./tests/test_project.py:1133:9: F541 f-string is missing placeholders
> ./tests/test_project.py:1143:80: E501 line too long (80 > 79 characters)
> ./tests/test_project.py:1144:80: E501 line too long (90 > 79 characters)
> ```

These errors should be fixed. I didn't realize you were testing the test python scripts for Python compliance.

mbolivar-nordic:
All good -- @marc-hb are you willing to go with me on this one and merge it, or do you want to discuss further?

I know environment variables are the enemy of reproducibility, but they're also super useful, and I think the existence of a default manifest URL is kind of a wart in west's design to begin with. But it's a wart we cannot change without breaking everyone's scripts at this point, so I'm inclined to allow workarounds like this

marc-hb:
I don't like it but I don't want to block it either, it's a very small code change and this is a feature easy to ignore. Please just don't advertise it too loudly in the --help / documentation  :-) 

> the existence of a default manifest URL is kind of a wart in west's design to begin with

Very good point. So, two wrongs make it right? :-D

> perhaps we can add a configuration option to specify the default manifest URL instead?

Yes I think a configuration file would be better. Configuration files provide much much better "forensics" than environment variables. Environment variables on Linux are an absolute disaster, here's a short quote from https://superuser.com/a/1452523/111302

```
ssh fedora29
 └─ -bash # login shell
      ├── /etc/profile
      |    ├─ /etc/profile.d/*.sh
      |    ├─ /etc/profile.d/sh.local
      |    └─ /etc/bashrc
      ├── ~/.bash_profile
      |    └─ ~/.bashrc
      |          └─ /etc/bashrc
      |
      |
      └─ $ bash  # non-login shell
            └─ ~/.bashrc
                 └─ /etc/bashrc
                       └─ /etc/profile.d/*.sh
```


What would also make _any_ solution much better IMHO is _logging_ where the effective manifest location was found (env, which config file, CLI, etc.). This can save a lot of time when desperately staring at the logs of some distant and messy CI system.

marc-hb:
> Environment variables on Linux are an absolute disaster,

Wait, they're a mess on Windows too. Just a very different kind of mess:
https://superuser.com/questions/544402/can-i-change-the-environment-on-a-per-process-basis-on-windows-like-on-unix
https://superuser.com/questions/867728/user-vs-system-environment-variables-do-system-variables-override-user-variabl

mbolivar-nordic:
> What would also make _any_ solution much better IMHO is _logging_ where the effective manifest location was found (env, which config file, CLI, etc.). This can save a lot of time when desperately staring at the logs of some distant and messy CI system.

OK, I suggest a compromise then, where we add some debug log messages identifying the manifest URL and its origin, which will normally not be printed, but will be printed with `west -v init [...]`. Deal?

RyanLindemanCAE:
> > What would also make _any_ solution much better IMHO is _logging_ where the effective manifest location was found (env, which config file, CLI, etc.). This can save a lot of time when desperately staring at the logs of some distant and messy CI system.
> 
> OK, I suggest a compromise then, where we add some debug log messages identifying the manifest URL and its origin, which will normally not be printed, but will be printed with `west -v init [...]`. Deal?

Does that translate to adding something kind of like this (not quite right, just an example)?
> log.dbg('using manifest url from command line', args.manifest_url) if args.manifest_url else
> log.dbg('using manifest url from environment', mu) if mu else log.dbg('using default manifest url', MANIFEST_URL)

RyanLindemanCAE:
Also, the target for this change is probably wrong, please let me know if I need to resubmit this PR of it the target could be changed instead to "main".

mbolivar-nordic:
Thanks!

PR #567, opened 2022-02-25
Allow userdata to be defined under self in manifest repository

RyanLindemanCAE:
The purpose of this PR is to allow for specifying userdata in the manifest repository. One of the primary reasons I am making the switch from the Google repo tool to the Zephyr west tool is to allow for specifying the firmware version information using the userdata capability in the manifest file. I also wanted to be able to combine my manifest repository with my project repository. I soon discovered that userdata is allowed for projects but excluded for no seemingly good reason for the manifest repository.

### DIFF
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -1001,6 +1001,7 @@ class ManifestProject(Project):
     - ``west_commands``:``west_commands:`` key in the manifest's
       ``self:`` map. This may be a list of such if the self
       section imports multiple additional files with west commands.
+    - ``userdata``: the parsed 'userdata' field under self in the manifest file
 
     Other readable attributes included for Project compatibility:
 
@@ -1015,11 +1016,13 @@ class ManifestProject(Project):
     def __repr__(self):
         return (f'ManifestProject(path={repr(self.path)}, '
                 f'west_commands={self.west_commands}, '
-                f'topdir={repr(self.topdir)})')
+                f'topdir={repr(self.topdir)}, '
+                f'userdata={repr(self.userdata)})')
 
     def __init__(self, path: Optional[PathType] = None,
                  west_commands: Optional[WestCommandsType] = None,
-                 topdir: Optional[PathType] = None):
+                 topdir: Optional[PathType] = None,
+                 userdata: Optional[Any] = None):
         '''
         :param path: Relative path to the manifest repository in the
             west workspace, if known.
@@ -1029,6 +1032,7 @@ class ManifestProject(Project):
         :param topdir: Root of the west workspace the manifest
             project is inside. If not given, all absolute path
             attributes (abspath and posixpath) will be None.
+        :param userdata: User data specified under self in manifest file
         '''
         self.name: str = 'manifest'
 
@@ -1038,6 +1042,8 @@ class ManifestProject(Project):
         self.revision: str = 'HEAD'
         self.clone_depth: Optional[int] = None
         self.groups = []
+        self.userdata: Optional[Any] = userdata
+
         # The following type: ignore is necessary since every Project
         # actually has a non-None _path attribute, so the parent class
         # defines its type as 'str', where here we need it to be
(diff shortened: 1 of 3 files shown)


RyanLindemanCAE:
Let me know if I need to bump the schema version number. I wasn't sure that I needed to since I thought version '0.10' was the last official version. I can see that it was bumped recently so I'm thinking I might be able to slip this change in without bumping it again.

mbolivar-nordic:
Yes, this needs a new schema version. We only bump the schema version when we make backwards-incompatible changes, which is not every release.

RyanLindemanCAE:
Done. Please let me know if I did it wrong.

RyanLindemanCAE:
@mbolivar-nordic - Any suggestions on how to get this PR moving towards a merge?
Thanks!

RyanLindemanCAE:
@mbolivar-nordic - Another ping? Any way I can get this moving?

mbolivar-nordic:
@RyanLindemanCAE sorry, #572 was kind of a hot potato that I had to deal with. Could you please rebase on top of that now that I've merged #571, which fixed #572, resolving the conflicts? I'll take another look and merge once that's done.

RyanLindemanCAE:
@mbolivar-nordic I have done as you asked, please review my changes.

RyanLindemanCAE:
It would be nice to find a way to "merge" the userdata values found in lower level manifest files similar to how you merge west_commands but I wasn't sure how to handle that since userdata can be either a string or key/value pair.
What do you think?

RyanLindemanCAE:
@mbolivar-nordic - The more I think about this, the more I think we need a slightly better solution then what is currently implemented. For example, if project A (the main manifest) has userdata defined under self and includes project B (using self.imports) which also has userdata defined under self then how would someone access both of these userdata definitions?
One thought is to have manifest.userdata be a dictionary where the key is the (relative?) path to the manifest file and the self.userdata in that file. Then project A and B could be distinguished using their relative manifest path as the key but this seems different than the project based userdata. Do you have any suggestions or ideas?

src/west/manifest.py
```diff
@@ -1352,6 +1360,8 @@ def __init__(self, *,  # All arguments are keyword-only.
         # These back the relevant properties.
         self._posixpath: Optional[str] = None
         self._repo_posixpath: Optional[str] = None
+        # This backs the userdata property
```
mbolivar-nordic, line 1363: I think you can just say `self.userdata: Any = None` here, and update `self.userdata` in `_load_self` as needed. There's no need for a property. Also, where is the property definition? I don't see it in this patch.

mbolivar-nordic (commented):
Thanks so much for the update. It looks like you meant to add a `userdata` attribute or property to the `Manifest` class itself, rather than just saving it in the `ManifestProject`. I have no objection to adding it to `ManifestProject` if it is useful to you, but having an attribute in the `Manifest` class makes good sense to me as the "recommended" place to get this, since that will be a nice long-term place to get this information after we successfully phase out the `ManifestProject` class (see #327). Did you forget to add it?

src/west/manifest.py
RyanLindemanCAE: I believe I have addressed your suggestion here.

RyanLindemanCAE:
> Thanks so much for the update. It looks like you meant to add a `userdata` attribute or property to the `Manifest` class itself, rather than just saving it in the `ManifestProject`. I have no objection to adding it to `ManifestProject` if it is useful to you, but having an attribute in the `Manifest` class makes good sense to me as the "recommended" place to get this, since that will be a nice long-term place to get this information after we successfully phase out the `ManifestProject` class (see #327). Did you forget to add it?

When I freeze my manifest file as part of a release it uses the ManifestProject to get the "self" details for the frozen file. When you remove ManifestProject, I expect you will use the Manifest class to obtain the "self" details during a freeze operation. Hopefully that helps explain why I think userdata should be in ManifestProject too for now. It also makes it easier to iterate through all "projects" seeking out userdata information. Thoughts or comments are welcome.

mbolivar-nordic (approved):

PR #571, opened 2022-03-19
Overhaul manifest path handling

mbolivar-nordic:
Fixes: #572 

This is an API break. We are still allowed to do that until west 1.0.

The west.manifest.Manifest class's path handling has become
unmaintainable and we need to overhaul it. The from_data() and
from_file() constructors are too clever and try to allow too many use
cases, which results in the guts of the manifest construction code
often not being able to tell whether it's in a real workspace or not.

This makes it difficult to know when we can safely read configuration
files or not. That in turn makes it difficult to accurately
distinguish between the 'manifest.path' value in the local configuration
file and the 'self: path:' value in the manifest data, especially when
we can take keyword arguments that say "hey, pretend this is your
path".

Additionally, there are a few assumptions in the code that pretend
that the manifest file itself always lives in the top level directory
of the manifest repository. That was a valid assumption up to the
point the 'manifest.file' configuration variable was introduced, but
it's no longer valid anymore, since 'manifest.file' can point to a
subdirectory of the manifest repository. This leaves us with some
hacks to find the git repository we live in that shouldn't be
necessary and which we can now remove with this overhaul.

Rework the whole thing so that we can correctly handle the following
situations:

 - when running in a workspace, you should now use from_file() or
   a newly introduced from_topdir() to make your Manifest

 - when running outside of any workspace, use from_data()

From now on, we forbid creating a manifest from data "as if" it were
in a real workspace. If all you have is data, the abspath attributes
will all be None, reflecting reality.

Accordingly, rework the `Manifest.__init__()` method to either take a
topdir or a bit of data, but not both. Additionally, now that we have
both manifest.path and manifest.file configuration options, it's not
usually the right thing to ask for a Manifest from a *file*: what you
really usually want is a Manifest from a *workspace*, and for that you
just need a topdir. From the topdir, you can create a
west.configuration.Configuration, and from there you can read
manifest.path and manifest.file to get the complete path to the top
level manifest within its repository. For that reason, we remove the
source_file keyword argument from `__init__()` in favor of topdir and a
new 'config' argument.

For backwards compatibility, it's still allowed to use from_file() to
load another manifest file than the one pointed to by
topdir/manifest.path/manifest.file. However, this handling code is now
just a bit of special case trickery in from_file(), and it also
introduces a restriction that the other file is still in a git
repository in the workspace. This moves potential sources for error or
confusion out of `Manifest.__init__()` and the functions it calls. (This
has proven to be a useful feature in practice; I am aware of west
extensions that use it to compare the parsed contents of two different
manifest files within the same workspace.)

To migrate away from the now-outdated from_file(), introduce a new
from_topdir() factory method and use it throughout the tree. This is
clearer and more accurate code, which always avoids the hacks left
behind for compatibility in from_file().

Finally, add various new instance attributes to the Manifest class
that will let us tell the YAML path from the actual path, and the path
to the manifest file from the path to the manifest repository in the
workspace. These in turn let us remove some hacky code from the 'west
list' implementation, and allow us to rework our internals so that
ManifestProject is just a legacy remnant at this point, further
helping along with #327.

Keep tests up to date, removing obsolete code as needed and updating
cases to reflect API breakage.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -23,7 +23,7 @@ from west.configuration import Configuration
 from west import log
 from west import util
 from west.commands import WestCommand, CommandError
-from west.manifest import ImportFlag, Manifest, MANIFEST_PROJECT_INDEX, \
+from west.manifest import ImportFlag, Manifest, \
     ManifestProject, _manifest_content_at, ManifestImportFailed, \
     _ManifestImportDepth, ManifestVersionError, MalformedManifest
 from west.manifest import is_group as is_project_group
@@ -281,15 +281,13 @@ With neither, -m {MANIFEST_URL_DEFAULT} is assumed.
                      if args.manifest_rev else '') +
                     f'  You may need to remove {west_dir} before retrying.')
 
-        # Parse the manifest to get the manifest path, if it declares one.
+        # Parse the manifest to get "self: path:", if it declares one.
         # Otherwise, use the URL. Ignore imports -- all we really
         # want to know is if there's a "self: path:" or not.
-        projects = Manifest.from_file(temp_manifest,
-                                      import_flags=ImportFlag.IGNORE,
-                                      topdir=topdir).projects
-        manifest_project = projects[MANIFEST_PROJECT_INDEX]
-        if manifest_project.path:
-            manifest_path = manifest_project.path
+        manifest = Manifest.from_data(temp_manifest.read_text(),
+                                      import_flags=ImportFlag.IGNORE)
+        if manifest.yaml_path:
+            manifest_path = manifest.yaml_path
         else:
             # We use PurePath() here in case manifest_url is a
             # windows-style path. That does the right thing in that
@@ -421,14 +419,19 @@ The following arguments are available:
             # as SHAs, unless they are specifically requested, and then
             # ensures they are only computed once.
             try:
-                if (isinstance(project, ManifestProject) and not
-                        args.manifest_path_from_yaml):
+                if isinstance(project, ManifestProject):
                     # Special-case the manifest repository while it's
                     # still showing up in the 'projects' list. Yet
                     # more evidence we should tackle #327.
-                    path = self.config.get('manifest.path')
-                    apath = abspath(os.path.join(self.topdir, path))
-                    ppath = Path(apath).as_posix()
+                    if args.manifest_path_from_yaml:
+                        path = self.manifest.yaml_path
+                        apath = (abspath(os.path.join(self.topdir, path))
+                                 if path else None)
+                        ppath = Path(apath).as_posix() if apath else None
+                    else:
+                        path = self.manifest.repo_path
+                        apath = self.manifest.repo_abspath
+                        ppath = self.manifest.repo_posixpath
                 else:
                     path = project.path
                     apath = project.abspath
@@ -528,7 +531,7 @@ class ManifestCommand(_ProjectCommand):
         # errors and printing useful messages. We re-do error checking
         # for manifest-related errors that it won't handle.
         try:
-            manifest = Manifest.from_file(topdir=self.topdir)
+            manifest = Manifest.from_topdir(topdir=self.topdir)
         except _ManifestImportDepth:
             log.die("cannot resolve manifest -- is there a loop?")
         except ManifestImportFailed as mif:
--- a/src/west/app/main.py
+++ b/src/west/app/main.py
@@ -116,7 +116,8 @@ class WestApp:
             return
 
         try:
-            self.manifest = Manifest.from_file(topdir=self.topdir)
+            self.manifest = Manifest.from_topdir(topdir=self.topdir,
+                                                 config=self.config)
         except (ManifestVersionError, MalformedManifest, MalformedConfig,
                 FileNotFoundError, ManifestImportFailed) as e:
             # Defer exception handling to WestCommand.run(), which uses
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -7,7 +7,6 @@
 Parser and abstract data types for west manifests.
 '''
 
-import configparser
 import enum
 import errno
 import logging
@@ -18,7 +17,7 @@ import shlex
 import subprocess
 import sys
 from typing import Any, Callable, Dict, Iterable, List, NoReturn, \
-    NamedTuple, Optional, Set, Tuple, TYPE_CHECKING, Union
+    NamedTuple, Optional, Set, TYPE_CHECKING, Union
 
 from packaging.version import parse as parse_version
 import pykwalify.core
@@ -26,7 +25,7 @@ import yaml
 
 from west import util
 from west.util import PathType
-import west.configuration as cfg
+from west.configuration import Configuration, ConfigFile
 
 #
 # Public constants
@@ -159,28 +158,6 @@ def _west_commands_merge(wc1: List[str], wc2: List[str]) -> List[str]:
     else:
         return wc1 or wc2
 
-def _mpath(cp: Optional[configparser.ConfigParser] = None,
-           topdir: Optional[PathType] = None) -> Tuple[str, str]:
-    # Return the value of the manifest.path configuration option
-    # in *cp*, a ConfigParser. If not given, create a new one and
-    # load configuration options with the given *topdir* as west
-    # workspace root.
-    #
-    # TODO: write a cfg.get(section, key)
-    # wrapper, with friends for update and delete, to avoid
-    # requiring this boilerplate.
-    if cp is None:
-        cp = cfg._configparser()
-    cfg.read_config(configfile=cfg.ConfigFile.LOCAL, config=cp, topdir=topdir)
-
-    try:
-        path = cp.get('manifest', 'path')
-        filename = cp.get('manifest', 'file', fallback=_WEST_YML)
-
-        return (path, filename)
-    except (configparser.NoOptionError, configparser.NoSectionError) as e:
-        raise MalformedConfig('no "manifest.path" config option is set') from e
-
 # Manifest import handling
 
 def _default_importer(project: 'Project', file: str) -> NoReturn:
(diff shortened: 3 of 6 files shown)


mbolivar-nordic:
@carlescufi I can run `west update` using sdk-zephyr as a manifest repository with these patches now. I believe that fixes @permac's bug, but it'd be good to get confirmation on that.

src/west/app/project.py
```diff
@@ -421,14 +419,19 @@ def delay(func, project):
...
-                    path = self.config.get('manifest.path')
-                    apath = abspath(os.path.join(self.topdir, path))
-                    ppath = Path(apath).as_posix()
+                    if args.manifest_path_from_yaml:
+                        path = self.manifest.yaml_path
+                        apath = (abspath(os.path.join(self.topdir, path))
+                                 if path else None)
+                        ppath = Path(apath).as_posix() if apath else None
+                    else:
+                        path = self.config.get('manifest.path')
```
tejlmand, line 432: shouldn't the `project.path` in the `ManifestProject` be identical to the config `manifest.path`, from what I can tell:
https://github.com/zephyrproject-rtos/west/blob/9dbc5cdf13374daa443d4e6b36bc6114d5356fa7/src/west/manifest.py#L1698
https://github.com/zephyrproject-rtos/west/blob/9dbc5cdf13374daa443d4e6b36bc6114d5356fa7/src/west/manifest.py#L1704
https://github.com/zephyrproject-rtos/west/blob/9dbc5cdf13374daa443d4e6b36bc6114d5356fa7/src/west/manifest.py#L1721
https://github.com/zephyrproject-rtos/west/blob/9dbc5cdf13374daa443d4e6b36bc6114d5356fa7/src/west/manifest.py#L1788-L1789

at best, if they are identical, this code is not needed as it simply uses `self.config.get('manifest.path')` instead of `project.path` that is used in line L436 below.
At worst, if `project.path` != `self.config.get('manifest.path')` then we risk that any code doing:
```
from west.manifest import Manifest
manifest = Manifest.from_file()
manifest_project = manifest.get_projects(['manifest'])
```
compared to `west list manifest` will have different path values, which risk leading to confusion.

src/west/app/project.py
```diff
@@ -421,14 +419,19 @@ def delay(func, project):
...
-                        args.manifest_path_from_yaml):
+                if isinstance(project, ManifestProject):
                     # Special-case the manifest repository while it's
                     # still showing up in the 'projects' list. Yet
                     # more evidence we should tackle #327.
-                    path = self.config.get('manifest.path')
-                    apath = abspath(os.path.join(self.topdir, path))
-                    ppath = Path(apath).as_posix()
+                    if args.manifest_path_from_yaml:
+                        path = self.manifest.yaml_path
```
tejlmand, line 427: similar to the comment on `get_option('manifest.path')`.

Shouldn't `project.path` be identical to `yaml_path` when loading from the yaml and not topdir ?
https://github.com/zephyrproject-rtos/west/blob/9dbc5cdf13374daa443d4e6b36bc6114d5356fa7/src/west/manifest.py#L1872-L1885

and
https://github.com/zephyrproject-rtos/west/blob/9dbc5cdf13374daa443d4e6b36bc6114d5356fa7/src/west/manifest.py#L1788-L1789

I share the same concern as described in the `get_option('manifest.path')` comment.

If there is a use-case that I have not noticed, then please provide extra info, but i'm concerned that special treatment of the path from `ManifestProject` bears a risk of unexpected behavior to any usages of `manifest.get_projets()` and confusion (and problem that already exists today, and must be addressed, just hoping not to increase confusion).

For example, Zephyr uses `get_projects()` directly instead of `west list`.
https://github.com/zephyrproject-rtos/zephyr/blob/a4d4384447710e9885c0248385db0d5a5d197d50/scripts/zephyr_module.py#L377-L380

tejlmand (changes requested):
Thanks for this.

Overall looks very good, found a little area of concern in `projects.py` so placed some comments I would like feedback on.
If I don't respond back again in due time, feel free to dismiss this review.

src/west/app/project.py
mbolivar-nordic: > shouldn't the project.path in the ManifestProject be identical to the config manifest.path

Yes, but we are trying to get rid of ManifestProject (#327), so I am trying to slowly phase out any in-tree users.

Let me explain my long term goals for after this PR. They are:

#. add a new `west.manifest.Manifest.__init__` kwarg: `legacy_manifest_projects: Optional[bool] = None`. If `legacy_manifest_projects` is `True`, then `Manifest.projects` will include the `ManifestProject` at index 0. If it is `False`, then `Manifest.projects` will **not** include the `ManifestProject`. For compatibility, if `legacy_manifest_projects` is None, then issue a warning about changing behavior and then work the same as if it were `False`.
#. manually set `legacy_manifest_projects` to `True` in static methods in the Manifest class (i.e. `west.manifest.Manifest.from_topdir()`, `west.manifest.Manifest.from_file()`, and `west.manifest.Manifest.from_data()`), so users see no differences and get no warnings (since people in practice use these methods instead of calling `Manifest.__init__` directly
#. Add new **module-level** functions to `west.manifest`: `from_topdir()`, `from_file()`, and `from_data()`, which set `legacy_manifest_projects` to `False`. (I'm talking about a new `west.manifest.from_topdir()` function here, **not** `west.manifest.Manifest.from_topdir()`, etc.)
#. Move all in-tree users over to the new module-level functions, including this part of project.py. Add special case helpers that retain backwards compatibility so the manifest repository is still considered in `west list` output and any other places where it's currently needed (details TBD)
#. Deprecate the static methods (`west.manifest.Manifest.from_topdir()`, etc.)
#. Eventually, remove both the static methods and `ManifestProject`. Some details TBD; I haven't thought it all the way through yet as I'm just trying to fix this one bug in this PR.

At that point, `west list` will not see a `ManifestProject`, and it will need to be using `self.manifest.repo_abspath` and other related fields to access path information for the manifest itself as is being done in the current re-work. I'm laying the ground for that transition to be a bit easier later on in this change by making `west list` get things from the right place for long-term use.

But I take your point that calling self.config is a little ugly here, so I'll add a new `Manifest.repo_path` attribute which caches the configuration option.

src/west/app/project.py
mbolivar-nordic: No, they are different in this case.

For example, with this diff to `zephyr/west.yml`:

```
index 35cb064ce3..453b2366aa 100644
--- a/west.yml
+++ b/west.yml
@@ -245,6 +245,6 @@ manifest:
     - -ci
 
   self:
-    path: zephyr
+    path: torsten's-house
     import: submanifests
```

We have:

```
$ west list --manifest-path-from-yaml -f {path} manifest
torsten's-house
```

But the `ManifestProject.path` attribute is still `zephyr` because `self._config_path` is `zephyr` when we have a topdir that has `manifest.path=zephyr`.

src/west/app/project.py
tejlmand: > Yes, but we are trying to get rid of ManifestProject (https://github.com/zephyrproject-rtos/west/issues/327), so I am trying to slowly phase out any in-tree users.

Fair enough, and thanks for the detailed explanation.

Then I assume the right thing to do in the future when requesting all possible Zephyr modules, is to do:
```
manifest = Manifest.from_file()

# Get the path for the manifest repo 
manifest_repo_path = manifest.repo_path

# Get all other projects.
projects = manifest.get_projects()
```
as the manifest repo will not be part of the project list in future.

Is this something you expect to complete for `0.13` ?

src/west/app/project.py
tejlmand: Ah right, this is also the where you might do a `west init -l` on a manifest repo that has a path different from the one in the manifest, but you want `west list` to print the `self::path` from that yaml, and not the actual in use.

tejlmand (approved):
thanks for the detailed explanations.

LGTM.

src/west/app/project.py
mbolivar-nordic: > Then I assume the right thing to do in the future when requesting all possible Zephyr modules, is to do:

Not quite. Remember that `west.manifest.Manifest.from_file()` will be the deprecated version that works the same way as it does today, with a `ManifestProject`. The new ways to get a manifest without a `ManifestProject` will be `west.manifest.from_file()` or `west.manifest.from_topdir()`.

Some of the details for zephyr modules are still TBD in my mind. I think one way to handle this cleanly would be to introduce a new `Repository` class, like this:

```py
class Repository:
    '''Attributes:

    - path: relative to topdir
    - abspath: absolute, host form
    - posixpath: absolute, POSIX form
    '''

    def git():
        # run git in the repository

class Project(Repository):
    '''A repository that is also a project in a manifest file,
    but has other attributes as well, like 'url', that Repository
    does not have.'''

class Manifest:
    '''New attribute:

    - repositories: List[Repository], includes all projects
      and the manifest repository
    '''

```

Then in zephyr you could do:

```py
import west.manifest

manifest = west.manifest.from_topdir()
modules = [repository for repository in manifest.repositories if is_zephyr_module(repository)]
for module in modules:
    handle_module(module)
```

Then we could probably keep get_projects() around for backwards compatibility, but return an error unless `legacy_manifest_projects` was `True` at init time. I think the list comprehension style is a little cleaner anyway.

mbolivar-nordic: > Is this something you expect to complete for 0.13 ?

No; I basically only work on my own west PRs when I get a direct request for a specific feature or fix these days, since there's not a lot of review bandwidth

marc-hb:
This makes a HUGE error handling difference after `west init` and before `west update`! From cryptic stack trace to beautiful error message. When is the 0.14 release planned?

mbolivar-nordic:
On Fri, Aug 05 2022, Marc Herbert wrote:
> When is the 0.14 release planned?

I'm going to try to do it tomorrow. I need to do the release testing on
a windows machine I don't normally have with me, in addition to the mac
and linux boxes I have at my usual location.

PR #575, opened 2022-04-15
manifest: fix Manifest.from_file() with fall_back

mbolivar-nordic:
Without a file argument, we should still use fall_back=True to search
for the workspace in the environment. This behavior was removed in the
recent rework of path handling, introducing a regression where the
manifest repository cannot be found when outside the workspace. Fix
it.


nordic-piks:
Tested on environment when this issue was found, working!

carlescufi (approved):
LGTM, reviewed the fix but not the test.

PR #582, opened 2022-05-16
app: fix set_zephyr_base()

mbolivar-nordic:
Commit e8080e9c3f23f3cb7246d6a7aa86586cc5b2c9b9
("manifest: simplify and fix get_projects()") caused a regression in
the west command line interface, which turns out to be due to a bug in
main.py. The fix itself seems fine.

The bug is in main.py's set_zephyr_base() function, where we search
for any project named 'zephyr' like this:

    projects = manifest.get_projects(['zephyr'])

This was always incorrect: when searching for a project by path,
we cannot make any assumptions about the current working directory.
This line of code happens to work if you are running west from the
manifest topdir, which is what we do in testing, but it fails if you
are somewhere else in the file system, e.g. in WEST_TOPDIR/zephyr.

We happened to get bizarrely lucky before this fix to get_projects(),
because prior to that fix, we were comparing

    Path(THE_MANIFEST_REPOSITORY_PATH).resolve()

with

    Path('zephyr').resolve()

in the get_projects() call from set_zephyr_base(), as long as
THE_MANIFEST_REPOSITORY_PATH is 'zephyr', which it is by default when
using the upstream zephyr repository as the manifest repository.

Prior to commit e8080e9c, this combination of bugs always resulted in
get_projects() returning the ManifestProject no matter where you were
on the file system. But with that fix in manifest.py, we need to also
fix main.py to pass a fully resolved WEST_TOPDIR/zephyr path to
get_projects() to try to find a project with path 'zephyr' inside the
current workspace.


lairdjm (approved):
Fixes the warning I was seeing on arch and debian linux

Issue #588, opened 2022-07-04
west fails after init until update has been invoked when using import


tejlmand:
After running `west init -l .` then west commands such as `status` and `list` fails with a stack trace when any project in the manifest contains `import`.

Also extension commands in the manifest repo cannot be executed (more precisely, they are not loaded).

Can be reproduced using: https://github.com/nrfconnect/sdk-nrf/

```
$ cd /<path-to/ncs/nrf
$ west init -l .
$ west status
Traceback (most recent call last):
  File "/home/tora/.local/bin/west", line 8, in <module>
    sys.exit(main())
  File "/home/tora/.local/lib/python3.8/site-packages/west/app/main.py", line 786, in main
    app.run(argv or sys.argv[1:])
  File "/home/tora/.local/lib/python3.8/site-packages/west/app/main.py", line 109, in run
    self.run_command(argv)
  File "/home/tora/.local/lib/python3.8/site-packages/west/app/main.py", line 337, in run_command
    self.handle_builtin_manifest_load_err(args)
  File "/home/tora/.local/lib/python3.8/site-packages/west/app/main.py", line 195, in handle_builtin_manifest_load_err
    p, f = self.mle.project, self.mle.filename
AttributeError: 'ManifestImportFailed' object has no attribute 'filename'

$ west list
Traceback (most recent call last):
  File "/home/tora/.local/bin/west", line 8, in <module>
    sys.exit(main())
  File "/home/tora/.local/lib/python3.8/site-packages/west/app/main.py", line 786, in main
    app.run(argv or sys.argv[1:])
  File "/home/tora/.local/lib/python3.8/site-packages/west/app/main.py", line 109, in run
    self.run_command(argv)
  File "/home/tora/.local/lib/python3.8/site-packages/west/app/main.py", line 337, in run_command
    self.handle_builtin_manifest_load_err(args)
  File "/home/tora/.local/lib/python3.8/site-packages/west/app/main.py", line 195, in handle_builtin_manifest_load_err
    p, f = self.mle.project, self.mle.filename
AttributeError: 'ManifestImportFailed' object has no attribute 'filename'

$ west help
usage: west [-h] [-z ZEPHYR_BASE] [-v] [-V] <command> ...

The Zephyr RTOS meta-tool.

optional arguments:
  -h, --help            get help for west or a command
  -z ZEPHYR_BASE, --zephyr-base ZEPHYR_BASE
                        Override the Zephyr base directory. The default is
[...]

PR #590, opened 2022-07-14
main: update ManifestImportFailed error handling

mbolivar-nordic:
Commit 2327610945323390cc330d23938d3c50a6f596c6
("manifest: ManifestImportFailed: relax argument specification")
changed the attributes in ManifestImportFailed instances, but the
corresponding error handling code in main.py was not updated.

This is incorrect and results in an ugly traceback instead of a
properly formatted error message.

Restore the preferred behavior by using the new attribute 'imp'.

Fixes: #588

### DIFF
--- a/src/west/app/main.py
+++ b/src/west/app/main.py
@@ -192,8 +192,8 @@ class WestApp:
                 if args.command == 'update':
                     return      # that's fine
 
-                p, f = self.mle.project, self.mle.filename
-                ctxt = f'  Missing file: "{f}"'
+                p, imp = self.mle.project, self.mle.imp
+                ctxt = f'  Failed importing "{imp}"'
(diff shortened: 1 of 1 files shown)


mbolivar-nordic:
Example output with this patch from a reproducer I set up on my machine:

```
$ west status
FATAL ERROR: failed manifest import in zephyr (zephyr)
  Failed importing "west.yml" from revision "85a0034809612b332fbdf478357e8a4109c493fd"
  Hint: for this to work:
          - zephyr must be cloned
          - its manifest-rev ref must point to a commit with the import data
        To fix, run:
          west update
```

marc-hb (approved):
You know software has a good design when fixing bugs like these requires changing so few lines and so little logic.

Issue #301, opened 2019-09-05
Support quiet mode for `west flash`


nashif:
Right now when flashing we get lots of output from both west and pyocd, for example:
```
west  flash --board-id OSHW000041114e45002a3006bcf000320571000097969900
-- west flash: rebuilding
ninja: no work to do.
-- west flash: using runner pyocd
-- runners.pyocd: Flashing file: /Users/anashif/zephyr/zephyrproject/zephyr/build/zephyr/zephyr.hex
[====================] 100%
0001035:INFO:loader:Erased 0 bytes (0 sectors), programmed 0 bytes (0 pages), skipped 16384 bytes (4 pages) at 38.84 kB/s
```

When using west with sanitycheck in verbose mode, this is lots of output that disturbs the test progress reporting (might be relevant with sanitycheck -vv, but sanitycheck -v should not have this type of information).

Can we add support  for -q (for quiet) with flash and probably with other commands please?

mbolivar:
Thanks for filing an issue!

This should be feasible to do for the built-in commands as a starting point (`west update -q`, etc.), and then we can do the same thing in our zephyr extensions. Basically all west I/O goes through one module, which can learn lower verbosity levels than info by default, so it's just a matter of also capturing subprocess output and dealing with it appropriately.

It needs to be prioritized, though, so I'm going to leave this on the backlog for now.

mbolivar:
@nashif how would you want this command to behave?

```
west -v update --quiet
```

options:

1. -v overrides --quiet
2. --quiet overrides -v
3. error

sjg20:
I suggest that

west update

is silent, only showing warnings/errors. That is the way normal tools work, at least in linux.

Then west update -v

can spit out more stuff

and perhaps west update -d

can spit out loads of debug stuff also

marc-hb:
> That is the way normal tools work, at least in linux.

Most of the output of `west update` is not from `west` itself but from `git`. Which non-default git verbosity option(s) would you like `west update` to use by default?

I don't think "normal" tools that _download_ something are completely silent _by default_; probably because most interactive users don't want to wait minutes-long timeouts to discover network or server issues.

Note interactive use was not the main purpose of this bug. `git` is less verbose in non-interactive mode (see this with a simple `git fetch 2>fetch.log`). `west update` was not the main purpose of this bug either.

I think version control tools that download new versions are never silent by default, likely because most users want to know if there's new code (to rebase on, etc.)

[...]

PR #598, opened 2022-08-30
Treewide cleanups

mbolivar-nordic:
A variety of changes in this PR that generally fall into the category of cleanups and improvements. I can split this into separate PRs if it's helpful.

- Remove obsolete build.py and cmake.py modules
- Move to python 3.8+, now that zephyr has done that
- Use a python 3.8 only feature to justify doing that for more than CI reasons :)
- Deprecate west.log (part of #149 ) as prep work for parallelizing tests and adding `west --quiet` (#301 )


mbolivar-nordic:
cc @gmarull 

Tests are passing for me locally on 3.10, but I expect CI errors.

src/west/app/main.py
```diff
@@ -154,36 +152,48 @@ def handle_builtin_manifest_load_err(self, args):
         # Handle ManifestVersionError is a special case.
         if isinstance(self.mle, ManifestVersionError):
             if args.command == 'help':
-                log.wrn(mve_msg(self.mle, suggest_upgrade=False) +
-                        '\n  Cannot get extension command help, ' +
-                        "and most commands won't run." +
-                        '\n  To silence this warning, upgrade west.')
+                self.queued_io.append(
```
carlescufi, line 155: Have you considred adding something that automates the whole `self.queued_io.append( lambda)` part?

mbolivar-nordic: Yeah, I couldn't think of anything that would make it simpler. We want to have the real command instance when we do the I/O so we can include WestCommand instance data in there. That implies a need for a new scope, which we can only do with a `lambda` or a `def`. So that's got the stay the way it is IMO.

We could add `self.queue_io(lambda cmd: ...)` to save a bit of typing, but I think the extra level of indirection is less clear, not more clear.

carlescufi: Understood, thanks for the explanation!

carlescufi (approved):

mbolivar-nordic:
Merging since nobody at TSC objected to Zephyr v1.14 being EOL (the buiild and cmake module removals in this PR will for sure break `west build` on that version of zephyr).

PR #599, opened 2022-08-31
treewide: don't call Project.git() with strings

mbolivar-nordic:
Related discussion:

https://github.com/zephyrproject-rtos/zephyr/discussions/48837

I got another report about this on Discord today, so it looks like it
may really be a west bug.

Hypothesis: special characters when doing things like commit 'peeling'
in refs like "v3.0.0^{commit}" are being treated as metacharacters by
the windows cmd.exe on some people's machines:

   You must use quotation marks around the following special
   characters: & < > [ ] | { } ^ = ; ! ' + , ` ~ [white space].

   https://docs.microsoft.com/en-us/windows-server/administration/windows-commands/cmd#syntax

If so, this should be a fix, since we'll no longer be using
shlex.split() in Project.git() for git() calls within west itself.
We definitely do include ^, {, and } in our command strings, and
shlex.split() is a POSIX-specific split, which may not be doing the
right thing on Windows, despite it passing CI.

Some Zephyr users are reporting problems that appear suspiciously like
this being the case:

https://github.com/zephyrproject-rtos/zephyr/discussions/48837

Unclear why this doesn't reproduce in our own testing.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -648,7 +648,7 @@ class Status(_ProjectCommand):
                     continue
 
                 log.banner(f'status of {project.name_and_path}:')
-                project.git('status', extra_args=user_args)
+                project.git(['status'], extra_args=user_args)
             except subprocess.CalledProcessError:
                 failed.append(project)
         self._handle_failed(args, failed)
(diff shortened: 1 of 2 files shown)


mbolivar-nordic:
I'd like confirmation of the fix from at least one of the affected users before merge, so DNM for now.

PR #603, opened 2022-09-08
configuration: fix error handling of keys missing a '.'

### LINKED ISSUE #600: west config does not sanitize option key
The expected use case works correctly, i.e.:
```
[tmon@tmon-laptop ~]$ west config --global build.board
nrf52840dk_nrf52840
```

However, when the option does not contain period, west terminates with an unhandled exception. 
```
[tmon@tmon-laptop ~]$ west config --global board
  File "/usr/bin/west", line 33, in <module>
    sys.exit(load_entry_point('west==0.14.0', 'console_scripts', 'west')())
  File "/usr/lib/python3.10/site-packages/west/app/main.py", line 793, in main
    app.run(argv or sys.argv[1:])
  File "/usr/lib/python3.10/site-packages/west/app/main.py", line 109, in run
    self.run_command(argv)
  File "/usr/lib/python3.10/site-packages/west/app/main.py", line 342, in run_command
    cmd.run(args, unknown, self.topdir, manifest=self.manifest,
  File "/usr/lib/python3.10/site-packages/west/commands.py", line 135, in run
    self.do_run(args, unknown)
  File "/usr/lib/python3.10/site-packages/west/app/config.py", line 150, in do_run
    self.read(args)
  File "/usr/lib/python3.10/site-packages/west/app/config.py", line 181, in read
    value = self.config.get(args.name, configfile=args.configfile or ALL)
  File "/usr/lib/python3.10/site-packages/west/configuration.py", line 173, in get
    return self._get(lambda cf: cf.get(option), default, configfile)
  File "/usr/lib/python3.10/site-packages/west/configuration.py", line 216, in _get
    return getter(cf)
  File "/usr/lib/python3.10/site-packages/west/configuration.py", line 173, in <lambda>
    return self._get(lambda cf: cf.get(option), default, configfile)
  File "/usr/lib/python3.10/site-packages/west/configuration.py", line 77, in get
    return self._get(option, self.cp.get)
  File "/usr/lib/python3.10/site-packages/west/configuration.py", line 89, in _get
    section, key = option.split('.', 1)
ValueError: not enough values to unpack (expected 2, got 1)
```

It would be better to display some actual error message similar to what `git config` does:
```
[tmon@tmon-laptop ~]$ git config --global email
error: key does not contain a section: email
```

marc-hb:
Fixes #600

De-duplicate the `option.split('.', 1)` "parsing" code into a new,
minimalistic and centralized `_InternalCF.key_parse` function. Add two
lines of error handling in this new function in order to provide a
meaningful error message instead of this cryptic and context free exception
when using a "build_board" key name without a dot:

```
    ...
    File "/usr/lib/python3.10/site-packages/west/configuration.py", in _get
    section, key = option.split('.', 1)
    ValueError: not enough values to unpack (expected 2, got 1)
```

The API user now gets this:

```
    ...
    File "/usr/lib/python3.10/site-packages/west/configuration.py", in parse_key
        raise ValueError(f"Invalid key name: '{dotted_name}'")
ValueError: Invalid key name: 'build_board'

```

### DIFF
--- a/src/west/configuration.py
+++ b/src/west/configuration.py
@@ -58,6 +58,13 @@ class _InternalCF:
     # but presenting a west-style section.key = value style API.
 
     @staticmethod
+    def parse_key(dotted_name: str):
+        section_child = dotted_name.split('.', 1)
+        if len(section_child) != 2:
+            raise ValueError(f"Invalid key name: '{dotted_name}'")
+        return section_child
+
+    @staticmethod
     def from_path(path: Optional[Path]) -> Optional['_InternalCF']:
         return _InternalCF(path) if path and path.exists() else None
 
@@ -69,7 +76,7 @@ class _InternalCF:
             raise FileNotFoundError(path)
 
     def __contains__(self, option: str) -> bool:
-        section, key = option.split('.', 1)
+        section, key = _InternalCF.parse_key(option)
 
         return section in self.cp and key in self.cp[section]
 
@@ -86,7 +93,7 @@ class _InternalCF:
         return self._get(option, self.cp.getfloat)
 
     def _get(self, option, getter):
-        section, key = option.split('.', 1)
+        section, key = _InternalCF.parse_key(option)
 
         try:
             return getter(section, key)
@@ -94,7 +101,7 @@ class _InternalCF:
             raise KeyError(option)
 
     def set(self, option: str, value: Any):
-        section, key = option.split('.', 1)
+        section, key = _InternalCF.parse_key(option)
 
         if section not in self.cp:
             self.cp[section] = {}
@@ -105,7 +112,7 @@ class _InternalCF:
             self.cp.write(f)
 
     def delete(self, option: str):
-        section, key = option.split('.', 1)
+        section, key = _InternalCF.parse_key(option)
 
         if section not in self.cp:
             raise KeyError(option)


mbolivar-nordic (changes requested):
I think a class is overkill since we don't need these object saround. It would also IMO be cleaner to do this validation in config.py instead of at the API level in configuration.py, so we don't have to dump stack. https://github.com/zephyrproject-rtos/west/pull/604

If you want to keep this PR, can you please raise ValueError instead, and replace the SectionKey class with a method that just returns a 2-tuple instead? I'd be happy to merge that too.

marc-hb:
I'm really not a fan of OO but the class would have made it easy to add other checks later if needed and right now the class has barely more lines of codes than a function raising the same exception.

I doubt this check is still required after your #604. "Defensive programming" is great... up to a point, maybe we can also assume that `configuration.py` and `config.py` in the same software trust each other? Bugs aside, will there be any code path that could lead to a dot-free `option` here once #604 is merged?

If the check and exception are not required then I will abandon this. I don't think the tiny and localized `option.split` duplication is worth a one-line function.

mbolivar-nordic:
> the class would have made it easy to add other checks later if needed

So would a function.

> right now the class has barely more lines of codes than a function raising the same exception

But it's not idiomatic Python in my opinion. Allocating a temporary object just to run a method on it and then throw it away is just weird when you can just define a function or method and call it.

> maybe we can also assume that configuration.py and config.py in the same software trust each other?

West is designed as a set of libraries that are used by the main command line application, extension commands, and other third party scripts. The config.py file is a built in command implementation. The configuration.py file is an API implementation. Their purposes are different. It's fine for APIs to barf an exception if you violate their contracts. It's not fine for a command to barf an exception on expected errors like invalid inputs.

> Bugs aside, will there be any code path that could lead to a dot-free option here once https://github.com/zephyrproject-rtos/west/pull/604 is merged?

I hope the above explanation clears this up for you.

> I don't think the tiny and localized option.split duplication is worth a one-line function.

... but it is worth a class, somehow?

marc-hb:
> The config.py file is a built in command implementation. The configuration.py file is an API implementation.

Thanks, I forgot about API usage! I will resubmit.

> But it's not idiomatic Python in my opinion. Allocating a temporary object just to run a method on it and then throw it away is just weird when you can just define a function or method and call it.

I think it's common for parsers to construct objects in any language. Constructors ensure the output either exists or not; that you never return something incomplete or inconsistent which is typical with a more imperative style. Throwing objects away does not bother me, it's not different from returning tuples or lists (which are of course objects too)

Anyway this does not matter as long as this is merely looking for one dot '.' in a single string, I will switch to a function. Interesting codestyle discussion still.

> > **If the check and exception are not required then I will abandon this.**  I don't think the tiny and localized option.split duplication is worth a one-line function.

> ... but it is worth a class, somehow?

Neither. You took this sentence totally out of its context ("IF check/exception NOT required...") 

After too much Java exposure I've become a huge fan of http://steve-yegge.blogspot.com/2006/03/execution-in-kingdom-of-nouns.html but maybe I've just met my match :-) Didn't expect that considering the number of (admittedly persistent!) classes in west...

mbolivar-nordic:
> I have never considered Python to be heavily influenced by functional languages, no matter what people say or think. - Guido Van Rossum

That's referring to https://en.wikipedia.org/wiki/Functional_programming, not whether or not we use a function! Any class definition is full of functions in the form of methods.

mbolivar-nordic:
> If you want to keep this PR, can you please raise ValueError instead, and replace the SectionKey class with a method that just returns a 2-tuple instead? I'd be happy to merge that too.

So I'm assuming we're going with something like this in a future update, right?

marc-hb:
> So I'm assuming we're going with something like this in a future update, right?

Sounds good.

> That's referring to https://en.wikipedia.org/wiki/Functional_programming, not whether or not we use a function!

Yes that was even more off-topic that the rest. Well, it's not like we're spamming many people here :-)

My point was just: Python is much more "object-oriented" than "the opposite". Unlike Java, that does not mean everything has to be class.

> Any class definition is full of functions in the form of methods.

... yet that's definitely not what people would refer to as "functional programming", agreed.

mbolivar-nordic (approved):

PR #610, opened 2022-11-03
commands: add change

mariuszpos:
Add command to change project details in the manifest.
Example of usage:

Change revision for openthread  project:
west change openthread --revision 36b2ac9fd02988319e9dcebc578d8f9a9f5ff892

Change path for openthread  project:
west change openthread --path modules/lib/openthread

Signed-off-by: Mariusz Poslinski <mariusz.poslinski@nordicsemi.no>

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -11,6 +11,7 @@ import logging
 import os
 from os.path import abspath, relpath
 from pathlib import PurePath, Path
+import ruamel.yaml
 import shutil
 import shlex
 import subprocess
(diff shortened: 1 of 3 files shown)


marc-hb:
Thanks! Can you please provide a commit message explaining the use case(s)?

src/west/app/project.py
```diff
@@ -1481,6 +1481,66 @@ def do_add_parser(self, parser_adder):
...
+                        p["revision"] = args.revision
+                    if args.path:
+                        print("path")
+                        p["path"] = args.path
+                    if args.url:
+                        print("url")
+                        p["url"] = args.url
+
+            if not project_found:
+                log.wrn('Project {} not found'.format(args.name))
```
jangalda-nsc, line 1539: probably `log.wrn` -> `self.wrn`

src/west/app/project.py
```diff
@@ -1481,6 +1481,66 @@ def do_add_parser(self, parser_adder):
...
+        with open(manifest.path, "r") as fh:
+            x = m.load(fh.read())
+            project_found = False
+            for p in x["manifest"]["projects"]:
+                if p["name"] == args.name:
+                    project_found = True
+                    if args.revision:
+                        p["revision"] = args.revision
+                    if args.path:
+                        print("path")
```
jangalda-nsc, line 1532: please add more meaningful messages like:
"Path of project X changed, Old: ABC New: DEF"

mbolivar-nordic (changes requested):
You still need to read and address my comments from the first time you tried this: https://github.com/zephyrproject-rtos/zephyr/pull/51416#pullrequestreview-1147085329

tests/test_project.py
```diff
@@ -2057,24 +2057,17 @@ def test_import_project_rolling(repos_tmpdir):
                           import: true
                       '''})
 
-    cmd(f'init -l {manifest_repo}')
+    cmd(f'change zephyr --revision pull/123/head')
+    cmd(f'change zephyr --path modules/lib')
```
mbolivar-nordic, line 2061: Why are you removing existing test code to suit the purpose of this new command instead of just adding more?

These changes are not well explained. This is just not OK. You need to write a good commit message justifying why these changes are coming in instead of just making small modifications and not responding to comments with words as well as code.

For example, this still doesn't work when you get the project from a `self: import:`; I requested this in the original attempt in the zephyr repository, pointed you again to this comment when you tried in the west repo, and it's still not addressed, nor have you asked any questions to clarify if you do not understand why this is important and why your current approach doesn't seem to work.

This effort is not going well; if you want this merged, you will need to start being more communicative instead of just posting new code changes with no explanation.

mbolivar-nordic (changes requested):

tests/test_project.py
mariuszpos: Regarding removing existing code it was by mistake.
I'll add more comments and and I'll make the corrections.

marc-hb:
Rather than hardcoding the list of supported keys and duplicating the same code for each, why not make everything (except the tests) more generic like this:

```
west change openthread revision=36b2ac9fd02988319e9dcebc578d8f9a9f5ff892 path=modules/lib/openthread
```

marc-hb:
> west change openthread revision=36b2ac9fd02988319e9dcebc578d8f9a9f5ff892 path=modules/lib/openthread

Or you could use YAML syntax to be consistent with `west.yml` and save even more code:

```
west change openthread  "{path: modules/lib/openthread, revision: abcdef}"
```

Parsing is as simple as:

```shell
echo "{path: modules/lib/openthread, revision: abcdef}" |
  python3 -c 'import sys,yaml; print(yaml.safe_load(sys.stdin.read()))'

=> {'path': 'modules/lib/openthread', 'revision': 'abcdef'}
```

mbolivar-nordic:
@marc-hb 's comments are well worth considering if the basic incorrectness of the current implementation is ever addressed

marc-hb:
@mariuszpos out of consideration for the time spent by reviewers can you please explain why you closed this PR and what your future plans are? If any.

marc-hb:
https://zephyr2022.sched.com/event/10CDW/how-to-get-your-zephyr-patches-merged-carles-cufi-marti-bolivar-nordic-semiconductor-pre-registration-required

https://youtu.be/mnVNLM7ynMQ

(any ressemblance etc. is not coincidental)

mariuszpos:
@marc-hb, to bring this PR to completion a lot of work and solid testing of the new command is needed. We don't have time for that at the moment. We will probably come back to this PR in some time to do it right. At this point, we are considering adding the change command as an west extension to the sdk-nrf, but in a simplified form.

marc-hb:
Thank you @mariuszpos , appreciated!

PR #618, opened 2023-01-04
main.py: Prevent shortened command line arguments

nordicjm:
This prevents python's argparse library from allowing it to generate its own shortened command line options which can cause problems when commands are added in future with similar names to existing arguments.

Not tested with tox since it does not seem to want to run on my PC for some reason, not even in a new virtual environment.

### DIFF
--- a/src/west/app/main.py
+++ b/src/west/app/main.py
@@ -328,7 +328,7 @@ class WestApp:
         parser = WestArgumentParser(
             prog='west', description='The Zephyr RTOS meta-tool.',
             epilog='''Run "west help <command>" for help on each <command>.''',
-            add_help=False, west_app=self)
+            add_help=False, west_app=self, allow_abbrev=False)
 
         # Remember to update zephyr's west-completion.bash if you add or
         # remove flags. This is currently the only place where shell


nordicjm:
@mbolivar-nordic pulled in.

nordicjm:
Seems like https://vielmetti.typepad.com/logbook/2022/10/git-security-fixes-lead-to-fatal-transport-file-not-allowed-error-in-ci-systems-cve-2022-39253.html needs to be pulled in

mbolivar-nordic:
@nordicjm how can you tell? I don't see any CI results at all

mbolivar-nordic:
nvm, I can repro

mbolivar-nordic:
@nordicjm the fix for the new git behavior that addresses CVE-2022-39253 is big enough to deserve its own issue and separate pull request, so I've created https://github.com/zephyrproject-rtos/west/pull/620 and we'll need to get that in first before you can rebase this PR on top.

marc-hb:
Reminds me the infamous https://bugs.python.org/issue9334 which prevents passing arguments like in the https://github.com/thesofproject/sof/commit/9fb7a607ebdaf204 example.

Probably unrelated, sorry for the noise.

mbolivar-nordic (approved):

Issue #619, opened 2023-01-04
submodule update fails in some git versions due to protocol.file.allow=user value


mbolivar-nordic:
This issue is only triggered by new git behavior discussed in:

https://github.blog/2022-10-18-git-security-vulnerabilities-announced/#cve-2022-39253

West should continue to work on older versions of git without the new behavior discussed there.

Specifically, this issue happens because 'west update' uses 'git submodule update --init --recursive' to clone submodules in some cases like when when when users are updating a submodule from a "remote" repository which is actually a file on the local host which may contain symbolic links under its .git directory. In such cases, the 'west update' will fail because the clone using the file protocol is disallowed at the 'git submodule update' step.

mbolivar-nordic:
High prio since it's breaking our own CI -- I expect the real-world impact to users will be low. My expectation is the vast majority of users are updating submodules with 'safe' protocols like https or ssh that are still allowed by default in new git versions.

PR #620, opened 2023-01-04
Fix submodule file protocol error

mbolivar-nordic:
Fixes: #619 

The user-visible fix is in `update: add --submodule-init-config option`: this allows users to clone submodules on a case-by-case basis using `west update --submodule-init-config protocol.file.allow=always ...` without having to globally change their git configuration to allow this unsafe protocol to be used everywhere.

The other fixes are for our CI and are due to fallout from the git change discussed in #619 and one other unrelated mypy issue.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -795,6 +795,11 @@ class Update(_ProjectCommand):
                            help='''proceed as if FILTER was appended to
                            manifest.group-filter; may be given multiple
                            times''')
+        group.add_argument('--submodule-init-config',
+                           action='append', default=[],
+                           help='''git configuration option to set when running
+                           'git submodule init' in '<option>=<value>' format;
+                           may be given more than once''')
 
         group = parser.add_argument_group('deprecated options')
         group.add_argument('-x', '--exclude-west', action='store_true',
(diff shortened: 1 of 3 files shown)


tests/test_project.py
```diff
@@ -601,13 +601,16 @@ def test_update_submodules_list(repos_tmpdir):
 
     # Make tagged_repo to be zephyr project submodule.
     subprocess.check_call(
-        [GIT, 'submodule', 'add', str(tagged_repo), 'tagged_repo'], cwd=zephyr)
+        [GIT, '-c', 'protocol.file.allow=always',
```
marc-hb, line 604: Could you add some new `SUBMODULE_ADD_OPTS` constant to avoid the repetition and make testing local changes easier?

src/west/app/project.py
```diff
@@ -795,6 +795,11 @@ def do_add_parser(self, parser_adder):
                            help='''proceed as if FILTER was appended to
                            manifest.group-filter; may be given multiple
                            times''')
+        group.add_argument('--submodule-init-config',
```
tejlmand, line 798: for the convenience of users who need to do this, should we support this on a west workspace basis.
For example by supporting a:
```
$ west config update.submodule-init-config protocol.file.allow=always
```
setting or similar.

That would allow users to set this on a per-workspace instead of globally / having to repeat on each update command.

Such proposal can be made on a followup PR.

tejlmand (approved):
lgtm.

You may consider to support a west workspace setting for this in addition.

tests/test_project.py
mbolivar-nordic: > Could you add some new SUBMODULE_ADD_OPTS

Done for this change and the subsequent one testing the related west update option.

mbolivar-nordic:
> You may consider to support a west workspace setting for this in addition.

I think that's overkill for now. The error condition that makes this happen seems unlikely to me to happen in practice. If we get real complaints about this, I'll take this up.

marc-hb (approved):

PR #624, opened 2023-02-09
log.py: minor log level pydoc fix

marc-hb:
The message is only printed if the ``level`` parameter is _lower or equal to_ (not "at least) the current verbosity level.

We need a "sign errors" joke similar to this one:
https://www.martinfowler.com/bliki/TwoHardThings.html

### DIFF
--- a/src/west/log.py
+++ b/src/west/log.py
@@ -72,8 +72,9 @@ def dbg(*args, level=VERBOSE_NORMAL):
     :param args: sequence of arguments to print.
     :param level: verbosity level to set, e.g. VERBOSE_VERY.
 
-    The message is only printed if level is at least the current
-    verbosity level.'''
+    The message is only printed if the ``level`` parameter is at most
+    the current verbosity level.
+    '''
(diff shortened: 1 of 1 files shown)


src/west/log.py
```diff
@@ -72,8 +72,9 @@ def dbg(*args, level=VERBOSE_NORMAL):
     :param args: sequence of arguments to print.
     :param level: verbosity level to set, e.g. VERBOSE_VERY.
 
-    The message is only printed if level is at least the current
-    verbosity level.'''
+    The message is only printed if the ``level`` parameter is lower or
+    equal to the current verbosity level.
```
mbolivar-nordic, line 76: Please change "at least" to "at most" instead; this is more readable IMO

mbolivar-nordic (commented):
Please change "at least" to "at most" instead for brevity and grammar

src/west/log.py
marc-hb: Done

PR #626, opened 2023-02-14
add typing information

mbolivar-nordic:
The west package contains type annotations, but mypy doesn't know about them because of a missing metadata file. This prevents us from type-checking calls into the west APIs, which we should be able to do.

Add the necessary metadata to make this work.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -49,4 +49,5 @@ setuptools.setup(
     ],
     python_requires='>=3.8',
     entry_points={'console_scripts': ('west = west.app.main:main',)},
+    zip_safe=False
 )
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -49,4 +49,5 @@
     ],
     python_requires='>=3.8',
     entry_points={'console_scripts': ('west = west.app.main:main',)},
+    zip_safe=False
```
marc-hb, line 52: Is this required by mypy too? Sadly?

mbolivar-nordic: Yes, my understanding is we need the py.typed file on the file system.

mbolivar-nordic:
Thanks for review! I'll get the ball rolling on a new west release as soon as I can; we've accumulated a fair number of patches that should be in people's hands at this point.

PR #628, opened 2023-02-17
commands: fix "no workspace" and other error paths

mbolivar-nordic:
Commit 2ac3e279ff98a4e62ed449430dfdf327881bff84
("WestCommand: use Configuration objects") introduced a configuration attribute to WestCommand instances, and used it to look up configuration options.

However, it stored this attribute at the wrong time in WestCommand.run(). We need to stash the configuration before doing anything else, because otherwise calls like self.die() in the same method will look up self.color_ui, which looks up the configuration, which isn't present, which calls self.die(), leading to an infinite recursion.

If we stash the configuration immediately, we instead have a valid configuration option and we get an error printed instead.

### DIFF
--- a/src/west/commands.py
+++ b/src/west/commands.py
@@ -182,13 +182,13 @@ class WestCommand(ABC):
         :param config: `west.configuration.Configuration` or ``None``,
             accessible as ``self.config`` from `WestCommand.do_run`
         '''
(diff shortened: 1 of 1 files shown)


mbolivar-nordic:
"error paths are never tested"

PR #635, opened 2023-02-27
project.py: add option to pipe forall subcommand output

boeglin:
This allows to avoid one pager invocation per project, for instance when subcommand is "git grep" or "git diff", as it hides the banner and previous projects output.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -1456,6 +1456,8 @@ class ForAll(_ProjectCommand):
                             required=True)
         parser.add_argument('-a', '--all', action='store_true',
                             help='include inactive projects'),
+        parser.add_argument('-p', '--pipe', action='store_true',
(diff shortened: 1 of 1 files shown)


mbolivar-nordic:
Is this just for git? If so, it seems like `git [-P|--no-pager] ...` has you covered. I'm just wondering how useful this is in general, especially since it has the counter-intuitive consequence of unifying stderr into stdout.

boeglin:
Well, the idea was to provide a behavior similar to "repo forall -p", but it seems I forgot that this option also pipes the whole output to a single pager, so not quite what I was looking for either, actually.
And I don't have any examples in mind of other commands that have the same behavior as git, so I'll stick to git -P.
Sorry for the noise.

mbolivar-nordic:
No worries, thanks for sending a patch!

Issue #636, opened 2023-02-27
"west build" exits with code 0 when interrupted with Ctrl-C (SIGINT)


xudongzheng:
If "west build" is used as part of a larger script, it can be helpful to distinguish between a complete build versus an interrupted build.

```
$ west --version
West version: v1.0.0
$ python3 --version
Python 3.8.9
```

mbolivar-nordic:
This is a valid issue, but it will require some changes to west's internals and a new release to fix.

marc-hb:
This was just discussed in
- #633 

Maybe also in much older:
- #135

mbolivar-nordic:
Fix is posted: https://github.com/zephyrproject-rtos/west/pull/639

mbolivar-nordic:
@xudongzheng could you please test the latest main branch fixes your issue?

mbolivar-nordic:
Thank you for testing! I will make a new west release soon.

PR #639, opened 2023-03-01
app: fix control-C behavior

mbolivar-nordic:
We shouldn't be exiting 0 on interrupt, because this means we did not successfully complete. Use better platform-specific exception handling instead. See comments for details.

Fixes: #636

### DIFF
--- a/src/west/app/main.py
+++ b/src/west/app/main.py
@@ -18,7 +18,9 @@ from io import StringIO
 import logging
 import os
 from pathlib import Path, PurePath
+import platform
 import shutil
+import signal
 import sys
 from subprocess import CalledProcessError
 import tempfile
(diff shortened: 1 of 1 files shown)


mbolivar-nordic:
Hmm, this isn't working the way I expect on Windows. DNM

mbolivar-nordic:
On a final note, I also verified that this doesn't affect control-C handling when you enter that within a gdb subprocess underneath `west debug` (on Linux). The existing correct behavior is maintained: the signal is delivered to gdb, which catches it and returns to the debugger REPL instead of exiting any processes.

marc-hb (commented):
The comments in the source code look very relevant and useful but let's pretend just for argument's sake that the person reading the code has no time or not enough background to dive deep into a 8 years old python bug or CPython source code. Any ressemblance with any real person is of course a pure coincidence.

The pointers seem extremely useful for a reader wondering whether this sort of black magic is the correct _solution_ or not. But long before that, it would be good to have a very basic introduction to the _problems_ solved by this code. This could be done by answering some very basic questions like "Why catch `KeyboardInterrupt` at all; why doesn't Python automagically does the right thing? Does it in a later/future Python? Is this mimicking newer Python versions to support older Python versions?" and "Why is some `signal` propagation required on Linux but not on Windows?"

You get the idea.

marc-hb:
> the person reading the code has no time or not enough background to dive deep into a 8 years old python bug or CPython source code.

Typically: someone trying to understand why things still don't work as expected on their system and _testing_ this code in various ways. Describing the problems tentatively solved by this exception handling code then guide the testing and the filing of a good bug with good reproduction steps.

mbolivar-nordic:
@marc-hb I have updated the commit message with the requested details; thanks for review.

marc-hb:
Thanks for the very detailed commit message, I'm learning a lot. However this is a lot more  than what I was asking and not _where_ I was asking it.

Looking at the newer commit message, now I know much better what I was asking. I'd like this new sentence in the _source code_, not just in the commit message:

> West catches KeyboardInterrupt because we do not want to dump stack 

And maybe this one too:

> On Unix, exiting with the expected status code is not enough, however.
If that's all we do, the calling shell will not detect that we were
interrupted.

If I had written this code myself I may have added a few more comments in the source code but you seem reluctant to add "too mundane" (?) comments in the source so I'm focusing on the essentials.

marc-hb (approved):
> basic assumed knowledge for people responsible for writing Python CLI apps

I can imagine it's second nature for these people but thanks for spending a one-line comment for the majority of other people: including very many people comfortable with Python yet who never created a new Python CLI app from scratch.

Without this intro, they wonder what is the point of all this seriously scary voodoo in the except block.

marc-hb:
> > basic assumed knowledge for people responsible for writing Python CLI apps

> seriously scary voodoo in the except block.

If it's basic knowledge, why does Python make it so hard and why didn't you get it right the first time? That does not make sense to me.

Never mind.

PR #640, opened 2023-03-06
West v1.0.1

mbolivar-nordic:
Backport https://github.com/zephyrproject-rtos/west/commit/9789617cd5becaf2b7d75aee996e9e3477cbc4a2 to the release branch. I don't think it's worth bothering with going through the entire testing procedure in an "alpha" 1.0.1a1 release as described in MAINTAINERS.rst since this is an isolated change that has been tested in its own right.

### DIFF
--- a/src/west/app/main.py
+++ b/src/west/app/main.py
@@ -18,7 +18,9 @@ from io import StringIO
 import logging
 import os
 from pathlib import Path, PurePath
+import platform
 import shutil
+import signal
 import sys
 from subprocess import CalledProcessError
 import tempfile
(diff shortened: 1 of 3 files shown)


marc-hb:
The fix is very well isolated but should we wait for at least the bug submitters to test the fix before backporting it? It is OS-dependent black magic...

mbolivar-nordic:
No harm in that and I've asked the user to confirm the fix, but I did test it myself as well :)

mbolivar-nordic:
Added DNM because I'd like to backport https://github.com/zephyrproject-rtos/west/pull/646 into this point release as well

Issue #642, opened 2023-03-10
west command to provide info about all repos which are not in the manifest revision


aescolar:
Similarly to how `repo status` works, it would be very nice if `west status` also provided information about all projects which are off the manifest revision; not just with uncommitted changes, but also in another sha/branch/tag.

This would facilitate working with several local changes to sub-projects projects.

aescolar:
(That other issue is all over the place, so let me be a bit clearer about what I would hope, which would cover my needs)

For each project:
1. If the project has been moved to a different commit (different that the one west last took it to(^1) in the last update), to print a line saying so (^2).
2. If the project has an unclean git status, print the output of git status
3. Otherwise no output whatsoever for that project (not even the project name)

(^1) I understad this is manifest-rev. So the condition would be if that project is not in manifest-rev right now.
(^2) Repo even tells you if you checked out a different branch but are still in the same commit. The benefit of this is marginal. So telling or not about this case is not so important.

You can be even nicer and provide a return code > 0 to the shell if any project hit 1. or 2.

The command can be called whatever.
The command should not change anything locally or in the .west or .git folders, or do anything smarter than that.

If the user moved manifest-rev on his own, that is not something this command should protect against.

marc-hb:
> (That other issue is all over the place, so let me be a bit clearer about what I would hope, which would cover my needs)

There was indeed a lot of alternatives discussed in #548. However I think the simplest one and the one I defended the most was exactly what you're asking here.

AFAIK it was closed by @mbolivar-nordic like this, verbatim:

> west status, as a wrapper around git status, reports status relative to HEAD, not manifest-rev

I hope I'm wrong but I'm not holding my breath.

marc-hb:
> With plain bash, something like this would do what I was hoping:

I highly recommend the excellent tool `apt/dnf install shellcheck` :-)

Off-topic sorry.

mbolivar-nordic:
@aescolar thanks for the clear and concise issue description.

Indeed I don't think it is a good idea to change the semantics of `west status`: the west commands that are just wrappers for git commands with the same names should be as simple and stupid as possible IMO, so `west status` should work like `git status`, which compares worktree+index with HEAD.

> The command can be called whatever.

I'm proposing `west compare` here: https://github.com/zephyrproject-rtos/west/pull/643

Please test it and provide feedback. If it matches your expectations, I will write test cases and try to get it merged.

Issue #548, opened 2021-10-08
[feature] allow west to report if any of the cloned repositories is out of sync with the manifest


xnlcasad:
In order to ensure that a Zephyr application is always built from the exact same sources it would be good if west could report if any of the west-handled repositories is out of sync with the manifest file. Much like a git superproject `git status` would report a dirty working tree if any of the submodules' HEADs point to a sha different from the one recorded in the superproject.

```
$ west forall -c 'git status'
=== running "git status" in manifest (nrf):
HEAD detached at v1.7.0
nothing to commit, working tree clean
=== running "git status" in zephyr (zephyr):
HEAD detached at v2.6.99-ncs1
$ cd -
~/work/zephyr-upstream/zephyr
$ git checkout HEAD^
$ cd -
~/work/zephyr-upstream/nrf
$ west forall -c 'git status'
=== running "git status" in manifest (nrf):
HEAD detached at v1.7.0
nothing to commit, working tree clean
=== running "git status" in zephyr (zephyr):
HEAD detached at 6bfc48ef0b
```
So basically I rolled back zephyr one commit, but west would not tell me that zephyr's HEAD is pointing to a different commit than whatever is on nrf/west.yml, which is:
```
- name: zephyr
        repo-path: sdk-zephyr
        revision: v2.6.99-ncs1
```

@carlescufi suggested that a new parameter to `west list` could be used. I don't have anything against that, I would only suggest the interface to be automation friendly, so that it could be invoked from a build script. If I take git as an analogy, `git diff --exit-code` would return 1 if anything has been changed from the working tree versus the HEAD revision.

@henrikbrixandersen @mbolivar

mbolivar-nordic:
`west status`, as a wrapper around `git status`, reports status relative to `HEAD`, not `manifest-rev`. So its output is not related to the manifest file, unless of course if you've just run `west update` and `HEAD` is the same commit as `manifest-rev`.

>  it would be good if west could report if any of the west-handled repositories is out of sync with the manifest file.

We get requests like this from time to time. The devil is in the details, as usual. For a manifest that does not use `import:` that only uses SHAs in its `revision:` values, like upstream zephyr/west.yml, the request seems clear. But it's less obvious what the request is when the manifest `import:` and/or non-SHA `revision:` values, especially when it contains both.

[...]

PR #643, opened 2023-03-10
West compare command

mbolivar-nordic:
Help text for the new command:

```
$ west help compare
usage: west compare [-h] [-a] [--exit-code] [--ignore-branches] [PROJECT ...]

Compare each project's working tree state against the results
of the most recent successful "west update" command.

This command prints "git status" output for a project if
(and only if) at least one of the following is true:

1. its checked out commit is different than the commit checked
   out by the most recent successful "west update"
2. its working tree is not clean (i.e. there are local uncommitted
   changes)
3. it has a local checked out branch

positional arguments:
  PROJECT            projects (by name or path) to operate on; defaults to active cloned projects

options:
  -h, --help         show this help message and exit
  -a, --all          include output for inactive projects
  --exit-code        exit with status code 1 if status output was printed for any project
  --ignore-branches  skip output for projects with checked out branches and clean
                     working trees if the branch is at the same commit as the
                     last "west update"

ACTIVE PROJECTS
---------------

Default output is limited to "active" projects as determined by the:

- "group-filter" manifest file section
- "manifest.group-filter" local configuration option in .west/config

To include inactive projects as well, use "--all" or give an explicit
list of projects (by name or path). See the west documentation for
more details on active projects.

Regardless of the above, output is limited to cloned projects.
```

Fixes: #642

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -670,15 +670,6 @@ class Status(_ProjectCommand):
         # to print. We manually use Popen in order to try to exit as
         # quickly as possible if 'git status' prints anything.
 
-        # This technique fails when tested on Python 3.6, which west
-        # still supports at time of writing. This seems likely to be
-        # due to https://bugs.python.org/issue35182.
-        #
-        # Users of old python versions will have to deal with the
-        # verbose output.
-        if sys.version_info < (3, 7):
-            return True
-
         popen = subprocess.Popen(['git', 'status', '--porcelain'],
                                  stdout=subprocess.PIPE,
                                  stderr=subprocess.PIPE,
@@ -693,15 +684,6 @@ class Status(_ProjectCommand):
                 stdout, stderr = popen.communicate(timeout=0.1)
             except subprocess.TimeoutExpired:
                 pass
-            except ValueError:
-                # In case this isn't issue35182, handle the exception
-                # anyway.
-                self.wrn(
-                    f'{project.name}: internal error; got ValueError '
-                    'from Popen.communicate() when checking status. '
-                    'Ignoring it, but please report this to the west '
-                    'developers.')
-                return True
             return stdout or stderr
 
         while True:
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -602,6 +602,134 @@ class ManifestCommand(_ProjectCommand):
         else:
             sys.stdout.write(to_dump)
 
+class Compare(_ProjectCommand):
+    def __init__(self):
+        super().__init__(
+            'compare',
+            "compare project status against the manifest",
+            textwrap.dedent('''\
+            Compare each project's working tree state against the results
+            of the most recent successful "west update" command.
+
+            This command prints output for a project if (and only if)
+            at least one of the following is true:
+
+            1. its checked out commit is different than the commit checked
+               out by the most recent successful "west update"
+            2. its working tree is not clean (i.e. there are local uncommitted
+               changes)
+            3. it has a local checked out branch (unless the configuration
+               option compare.ignore-branches is true or --ignore-branches
+               is given on the command line, either of which disable this)
+
+            The command also prints output for the manifest repository if it
+            has nonempty status.
+
+            The output is meant to be human-readable, and may change. It is
+            not a stable interface to write scripts against. This command
+            requires git 2.22 or later.''')
+        )
+
+    def do_add_parser(self, parser_adder):
+        parser = self._parser(parser_adder,
+                              epilog=ACTIVE_CLONED_PROJECTS_HELP)
+        parser.add_argument('projects', metavar='PROJECT', nargs='*',
+                            help='''projects (by name or path) to operate on;
+                            defaults to active cloned projects''')
+        parser.add_argument('-a', '--all', action='store_true',
+                            help='include output for inactive projects')
+        parser.add_argument('--exit-code', action='store_true',
+                            help='''exit with status code 1 if status output
+                            was printed for any project''')
+        parser.add_argument('--ignore-branches',
+                            default=None, action='store_true',
+                            help='''skip output for projects with checked out
+                            branches and clean working trees if the branch is
+                            at the same commit as the last "west update"''')
+        parser.add_argument('--no-ignore-branches', dest='ignore_branches',
+                            action='store_false',
+                            help='''overrides a previous --ignore-branches
+                            or any compare.ignore-branches configuration
+                            option''')
+        return parser
+
+    def do_run(self, args, ignored):
+        self.die_if_no_git()
+        if self.git_version_info < (2, 22):
+            # This is for git branch --show-current.
+            self.die('git version 2.22 or later is required')
+        self._setup_logging(args)
+
+        if args.ignore_branches is not None:
+            self.ignore_branches = args.ignore_branches
+        else:
+            self.ignore_branches = \
+                self.config.getboolean('compare.ignore-branches', False)
+
+        failed = []
+        printed_status = False
+        for project in self._cloned_projects(args, only_active=not args.all):
+            if isinstance(project, ManifestProject):
+                # West doesn't track the relationship between the manifest
+                # repository and any remote, but users are still interested
+                # in printing output for comparisons that makes sense.
+                if self._has_nonempty_status(project):
+                    try:
+                        self.banner(f'{project.name_and_path}:')
+                        self.small_banner('HEAD:')
+                        project.git('log --oneline -n 1 HEAD')
+                        self.small_banner('status:')
+                        project.git('status')
+                        printed_status = True
+                    except subprocess.CalledProcessError:
+                        failed.append(project)
+                continue
+
+            # 'git status' output for all projects is noisy when there
+            # are lots of projects.
+            #
+            # We avoid this problem in 2 steps:
+            #
+            #   1. Check if we need to print any output for the
+            #      project.
+            #
+            #   2. If so, run 'git status' on the project. Otherwise,
+            #      skip output for the project entirely.
+            #
+            # In verbose mode, we always print output.
+
+            def has_checked_out_branch(project):
+                if self.ignore_branches:
+                    return False
+
+                return bool(project.git('branch --show-current',
+                                        capture_stdout=True,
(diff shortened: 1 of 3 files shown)


mbolivar-nordic:
Putting a DNM on this until we get the requester to confirm it's what they want, and then I'll write some test cases.

marc-hb:
That's really nice but it seems to still suffer from git's "non-determistic" display behavior which I analyzed and documented a couple years ago in https://github.com/zephyrproject-rtos/west/issues/548#issuecomment-961624310

This git behavior makes it show `manifest-rev` sometimes and sometimes not - even when everything is on the `manifest-rev`. I was lucky enough to demonstrate this in just 5 minutes:

```
SOF$ west compare

=== status of hal_xtensa (modules/hal/xtensa):
HEAD detached at refs/heads/manifest-rev
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	foo

nothing added to commit but untracked files present (use "git add" to track)

=== status of mbedtls (modules/crypto/mbedtls):
HEAD detached at 6e1660500756
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	foo

nothing added to commit but untracked files present (use "git add" to track)
```
The output above really looks like `hal_xtensa` is on the manifest while `mbedtls` is not on the manifest.

WRONG! That's gits "non-deterministic" display deceiving you. In reality they're both on the manifest:

```
git -C modules/crypto/mbedtls/ log --oneline --decorate -n 1
6e1660500756 (HEAD, manifest-rev) Merge pull request for mbedtls-3.2.1
```


PS: `west compare` sounds like it could be extended in the future to "recursively" compare manifests, thoughts on that?

mbolivar-nordic:
> it seems to still suffer from git's "non-deterministic" display behavior

I am personally OK with this as a limitation unless there is a way to tell git status to try harder to avoid this.

> PS: west compare sounds like it could be extended in the future to "recursively" compare two manifests, thoughts on that?

The name `compare` was not arbitrary ;)

https://github.com/nrfconnect/sdk-nrf/blob/57fade75665b2d6e1af649ff47a72374dba31a0b/scripts/west_commands/ncs_commands.py#L400

marc-hb:
> unless there is a way to tell git status to try harder to avoid this.

I didn't find any; that's frustrating. `git log` has `--decorate` but nothing in "non-deterministic" `git status`

How about this one-liner? Works4me!
```diff
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -682,6 +682,9 @@ class Compare(_ProjectCommand):
                     continue
 
                 self.banner(f'status of {project.name_and_path}:')
+                # `git status` shows `manifest-rev` "sometimes", see #643
+                # If manifest-rev is missing then we already failed earlier anyway
+                project.git('log --oneline -n 1 manifest-rev')
                 project.git('status')
                 printed_status = True
             except subprocess.CalledProcessError:
```

> https://github.com/nrfconnect/sdk-nrf/blob/57fade75665b2d6e1af649ff47a72374dba31a0b/scripts/west_commands/ncs_commands.py#L400

Way ahead of "us" upstream! I didn't look at the code much but the name in the git history seems very familiar...

mbolivar-nordic:
> How about this one-liner? Works4me!

Something like this seems reasonable to me. Any issues @aescolar ?

> Way ahead of "us" upstream

We have a lot of tooling that I've been carrying around since my foundries days :slightly_smiling_face: . So far everything has proven to require some domain-specific knowledge, and I haven't bothered to try to untangle reusable from domain-specific. But it's all open source if people want to give it a go and play around.

aescolar:
> Putting a DNM on this until we get the requester to confirm it's what they want, and then I'll write some test cases.

Thank you very much @mbolivar-nordic , that was much faster than I expected.
Overall this solves my problem, so getting it already "as is" is more than welcome.
About skipping the manifest repo, my thinking was that it is still nice to show if it is dirty even if do not have a reference commit. It is conceptually a bit decoupled from the other projects checks, but nonetheless.

Otherwise I only see "extras" that could be added, but those can come later. (say things like providing more concise output than git status, when it is clean but we are in another commit (say a message like "xxxx checked out, manifest-rev was yyyy). Or having the option to get a short (-s) git status)

>> How about this one-liner? Works4me!
> Something like this seems reasonable to me. Any issues @aescolar ?

Ok.

marc-hb:
> About skipping the manifest repo, my thinking was that it is still nice to show if it is dirty even if do not have a reference commit. It is conceptually a bit decoupled from the other projects checks, but nonetheless.

I second this; I would like practically over purity here. It would also be consistent with `west status` which is already a bit "impure" that way.

Or how about this middle-ground: not by default but when some new `west compare --manifest-status` option is passed? Permanently settable in `~/.westconfig` somehow.

marc-hb:
So this additional line shows whether the module is on the `manifest-rev` or not:
```
project.git('log --oneline -n 1 manifest-rev')
```

As a bonus, it also shows whether it's on `upstream/main`  or not.

mbolivar-nordic:
@aescolar @marc-hb you were both totally right about including the manifest repository in the output. I even expected that to be the behavior when I sat down to write the tests tonight, and I was surprised that it wasn't.

Now updated, please give it a go and let me know if it meets your expectations. I will merge after that, and we'll get things rolling for west v1.1 with this feature.

src/west/app/project.py
```diff
@@ -602,6 +602,118 @@ def _dump(self, args, to_dump):
...
+        parser.add_argument('--ignore-branches', action='store_true',
+                            help='''skip output for projects with checked out
+                            branches and clean working trees if the branch is
+                            at the same commit as the last "west update"''')
+        return parser
+
+    def do_run(self, args, ignored):
+        self.die_if_no_git()
+        if self.git_version_info < (2, 22):
+            self.die('git version 2.22 or later is required')
```
tejlmand, line 650: remember to add to release notes that git >=2.22 is now required if using this command.

Also I believe it would be nice with a comment as to why git >=2.22 is required.
I assume it's to due to the `branch --show-current` which were introduced in 2.22.

But thanks for only having this as requirement for this command 👍 and not enforce it generally for `west`.

src/west/app/project.py
```diff
@@ -602,6 +602,118 @@ def _dump(self, args, to_dump):
...
+               out by the most recent successful "west update"
+            2. its working tree is not clean (i.e. there are local uncommitted
+               changes)
+            3. it has a local checked out branch
+
+            The command prints output for the manifest repository if it
+            has nonempty status.
+
+            The output is meant to be human-readable, and may change. It is
+            not a stable interface to write scripts against.''')
```
tejlmand, line 627: should we add a line that states that git >=2.22 is required for this command, so that it's also mentioned in the help text. After all, help text is a form of documentation for `west`.
I do know we also have docs here: https://docs.zephyrproject.org/latest/develop/west/index.html, but `west help <cmd>` is faster 😉

tejlmand (commented):
Looks good.

Couple of small proposals to be friendlier to users.

aescolar (approved):
It seems fine :)
(Maybe it would be a bit nicer without the minor banners, or if they did not jump to the next line, but that is quite minor)

mbolivar-nordic:
@tejlmand good points, thanks. I've implemented your suggestions.

tejlmand (approved):

marc-hb:
I just tested this and it looks pretty good, thanks! Two recommendations:

```diff
modified   src/west/app/project.py
@@ -666,6 +666,7 @@ class Compare(_ProjectCommand):
                         project.git('log --oneline -n 1 HEAD')
                         self.small_banner('status:')
                         project.git('status')
+                        print()
                         printed_status = True
                     except subprocess.CalledProcessError:
                         failed.append(project)
@@ -708,6 +709,7 @@ class Compare(_ProjectCommand):
                 project.git('log --oneline -n 1 manifest-rev')
                 self.small_banner('status:')
                 project.git('status')
+                print()
                 printed_status = True
             except subprocess.CalledProcessError:
                 failed.append(project)
```
... or something like it because `git status` adds this annoying blank line before its last line of output:

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	foo

nothing added to commit but untracked files present (use "git add" to track)
=== mbedtls (modules/crypto/mbedtls):
--- manifest-rev:
7fed49c9b9f9 (HEAD, manifest-rev) CC3XX: Hardcode entry points for the CC3XX driver
--- status:
```

marc-hb:
Second recommendation: can you please reverse the `--ignore-branches` logic and change it to `--include-branches` option instead? I really don't see the value of showing a personal branch that is 100% identical to the manifest in a repo that is in a 100% pristine state, this is just adding noise IMHO. The next `west update` command will totally ignore such a branch anyway, so I don't see why `west compare` should show it by default.

While this new `Compare` command looks incredibly useful, its output may get "crowded" and hard to parse pretty quickly (see comments about banners above), so the default should IMHO show only actual code differences that can make a difference and not include any "nice to have" git metadata like "BTW, did you remember you created a git branch there?" when said branch makes no difference whatsoever.

aescolar:
I don't mind the `Second recommendation: can you please reverse the --ignore-branches` request.
But @marc-hb one note about 

> The next west update command will totally ignore such a branch anyway, so I don't see why west compare should show it by default.

Not exactly, the next west update will get out of that branch and detach the head.
 
> While this new Compare command looks incredibly useful, its output may get "crowded" ..
 
Note as mentioned before, I'll be happy to get it already as is. Anyhow with regard to too much verbosity, I guess that is why `repo` has a very concise output: projects that are clean, but on a branch, only produce 1 line output (including the project name, with the branch that is checked out on the same line). And the status per project is a very compact custom notation of the status of each project. But that is minor nicety compared to having the functionality overall, which could be added later.

mbolivar-nordic:
I dithered over this decision for too long and then ended up just adding a chicken bit.

This now supports a compare.ignore-branches configuration option that people can use if they don't like the default behavior, which is still to print information about checked-out branches. (I think surfacing the behavior by default will satisfy people who want it, and people who don't want it will be motivated to figure out how to turn it off)

marc-hb:
> because git status adds this annoying blank line before its last line of output

"fix" submitted in
- #648

PR #648, opened 2023-04-13
compare: add a blank line after each project

marc-hb:
`git status` adds one annoying blank line just before its last line of output:

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	foo

nothing added to commit but untracked files present (use "git add" to track)
=== mbedtls (modules/crypto/mbedtls):
--- manifest-rev:
7fed49c9b9f9 (HEAD, manifest-rev) CC3XX: Hardcode entry points for the CC3XX driver
--- status:
```

This makes it harder to visualize the separation between projects.

Counters this by separating projects with one blank line:

```
Untracked files:
  (use "git add <file>..." to include in what will be committed)
	foo

nothing added to commit but untracked files present (use "git add" to track)

=== mbedtls (modules/crypto/mbedtls):
--- manifest-rev:
7fed49c9b9f9 (HEAD, manifest-rev) CC3XX: Hardcode entry points for the CC3XX driver
--- status:
```

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -680,6 +680,7 @@ class Compare(_ProjectCommand):
                         project.git('log --oneline -n 1 HEAD')
                         self.small_banner('status:')
                         project.git('status')
+                        print()
                         printed_status = True
                     except subprocess.CalledProcessError:
                         failed.append(project)
@@ -722,6 +723,7 @@ class Compare(_ProjectCommand):
                 project.git('log --oneline -n 1 manifest-rev')
                 self.small_banner('status:')
                 project.git('status')
+                print()
                 printed_status = True
             except subprocess.CalledProcessError:
                 failed.append(project)


aescolar (commented):
I'd rather we did not. The output already uses a lot of lines, and we have the green banners separating projects.

marc-hb:
> The output already uses a lot of lines,

That's why this one does not make much difference to the _total_ (while making a big readability difference in my experience)

> and we have the green banners separating projects.

It depends on the terminal/redirections, colors don't always work. For instance with `git` you can `git -p status` to get colors and a pager easily whereas `west compare | less` drops colors by default without some colors "acrobatics"

Even when they do, west's green separators don't stand out from the green and red already used by git. If west's separators were longer/bigger/bolder in some way then I agree this PR would not be needed.

mbolivar-nordic:
Is https://github.com/zephyrproject-rtos/west/pull/649 a valid compromise?

PR #649, opened 2023-04-13
compare: output improvements

mbolivar-nordic:
Fixes: #650

A few concerns have been voiced about the west compare output:

   1. its signal:noise ratio is too low (too many lines per
      project without enough useful information per line)
   2. the end of the status output from one project is hard to
      separate from the status output for the next
   3. it isn't obvious enough whether HEAD == manifest-rev or not

To try to resolve 2. without making 1. any worse, explicitly print
HEAD and manifest-rev if they differ, without adding any extra lines,
by consolidating some of the output into the relevant banner line.

To solve 3., indent the status output to the same level as the
'status:' text in the banner.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -614,8 +614,9 @@ class Compare(_ProjectCommand):
             This command prints output for a project if (and only if)
             at least one of the following is true:
 
-            1. its checked out commit is different than the commit checked
-               out by the most recent successful "west update"
+            1. its checked out commit (HEAD) is different than the commit
+               checked out by the most recent successful "west update"
+               (which the manifest-rev branch will point to)
             2. its working tree is not clean (i.e. there are local uncommitted
                changes)
             3. it has a local checked out branch (unless the configuration
@@ -667,7 +668,7 @@ class Compare(_ProjectCommand):
                 self.config.getboolean('compare.ignore-branches', False)
 
         failed = []
-        printed_status = False
+        printed_output = False
         for project in self._cloned_projects(args, only_active=not args.all):
             if isinstance(project, ManifestProject):
                 # West doesn't track the relationship between the manifest
@@ -675,12 +676,8 @@ class Compare(_ProjectCommand):
                 # in printing output for comparisons that makes sense.
                 if self._has_nonempty_status(project):
                     try:
-                        self.banner(f'{project.name_and_path}:')
-                        self.small_banner('HEAD:')
-                        project.git('log --oneline -n 1 HEAD')
-                        self.small_banner('status:')
-                        project.git('status')
-                        printed_status = True
+                        self.compare(project)
+                        printed_output = True
                     except subprocess.CalledProcessError:
                         failed.append(project)
                 continue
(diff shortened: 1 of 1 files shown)


mbolivar-nordic:
We could go even further with something like this:

image

LMK what y'all want me to do

marc-hb (approved):
I think there are probably ways to make it even better but this looks like good visual progress to me and the internal refactoring looks like a good starting point for further improvements like the suggestion I just made in
- #650

> We could go even further with something like this:

No because it would make the `manifest-rev:` prefix and SHA1 start in a variable column.

mbolivar-nordic:
> Was this change intended?

It was not; now fixed.

mbolivar-nordic:
image

mbolivar-nordic:
> * [Still difficult to tell for which repos HEAD == manifest-rev in `west compare` #650](https://github.com/zephyrproject-rtos/west/issues/650)

I'm starting to feel like the extra color and full `git log --oneline` output is noisier than it should be.

I could do something like this in this PR:

```
--- manifest-rev: cafef00d drivers: add all the bugs
            HEAD: deadbeef drivers: fix all the bugs
--- status:
    [...indented, colorized text goes here...]
```

Where the manifest-rev and HEAD lines are shown only if the two revisions are different SHAs, and no color is present in any case.

aescolar:
> It was not; now fixed.

This looks very fine to me

> Where the manifest-rev and HEAD lines are shown only if the two revisions are different SHAs

If very fine with this proposal.

mbolivar-nordic:
> Im very fine with this proposal.

OK, now done; looks like this:

image

Where the first banner is not printed at all if HEAD and manifest-rev are the same commit.

mbolivar-nordic:
@marc-hb since this has changed substantially since your first approval, I will give you some time to look again before merging

src/west/app/project.py
```diff
@@ -714,22 +711,57 @@ def has_checked_out_branch(project):
...
+        #
+        # For other repositories, print HEAD's and manifest-rev's
+        # SHAs and commit titles, but only if they are different.
+        #
+        # We force git not to print in color to reduce noise in the
+        # overall output.
+
+        # `git status` shows `manifest-rev` "sometimes", see
+        # #643. If manifest-rev is missing, then we already
+        # failed earlier anyway.
```
marc-hb, line 740: This comment was the rationale for running `git log` ONCE. Now that you are running `git log` twice it feels a bit out of place. It's still useful information though, can you move it below next to `git status`?

src/west/app/project.py
```diff
@@ -714,22 +711,57 @@ def has_checked_out_branch(project):
...
+        self.print_status(project)
+
+    def print_rev_info(self, project):
+        # For the manifest repository, unconditionally print
+        # HEAD's SHA and commit title.
+        #
+        # For other repositories, print HEAD's and manifest-rev's
+        # SHAs and commit titles, but only if they are different.
+        #
+        # We force git not to print in color to reduce noise in the
```
marc-hb, line 735:
```suggestion
        # We force git log not to print in color so west colored separators stand out
        # more.
```
git status is still colored by default.

src/west/app/project.py
```diff
@@ -714,22 +711,57 @@ def has_checked_out_branch(project):
...
+        # #643. If manifest-rev is missing, then we already
+        # failed earlier anyway.
+        def rev_info(rev):
+            return project.git(
+                ['log', '-1', '--color=never', '--pretty=%h %s', rev],
+                capture_stdout=True,
+                capture_stderr=True).stdout.decode().rstrip()
+        head_info = rev_info('HEAD')
+        if isinstance(project, ManifestProject):
+            self.small_banner(f'HEAD: {head_info}')
```
marc-hb, line 748: I would drop this because this is the `west compare` command and this is not comparing anything. Also, it can be misleading when scanning the output quickly for repos which are not on the manifest.

marc-hb (commented):
I don't think it helps with #648 much, especially not with uncolored `west compare | less` but it does not make it worse either.

On the other hand this really helps with #650, thanks!

Small changes suggested below.

mbolivar-nordic:
> I don't think it helps with #648 much, especially not with uncolored `west compare | less` but it does not make it worse either.

I am surprised to hear you say that. I think the indentation makes the separation between status output and the green lines quite easy to read, personally. Are you saying an extra newline reads significantly more clearly to you than an indented block?

mbolivar-nordic:
> and the green lines quite easy to read, personally.

... and indentation is of course preserved across pipelines that remove color

marc-hb:
> I think the indentation makes the separation between status output and the green lines quite easy to read, personally. Are you saying an extra newline reads significantly more clearly to you than an indented block?

I would take both but after taking a closer look at the before/after difference you're right: the indentation does help quite a lot. Apologies.

mbolivar-nordic:
To make it easier to review, here's a screenshot of how it looks as of now:

image

marc-hb (approved):
Still two confusing and/or misplaced comments and some strange color behavior (what's new?) but that shouldn't block this massive improvement, thanks!

src/west/app/project.py
```diff
@@ -714,22 +711,56 @@ def has_checked_out_branch(project):
...
+        if head_info != manifest_rev_info:
+            self.small_banner(f'manifest-rev: {manifest_rev_info}')
+            self.inf(f'            HEAD: {head_info}')
+
+    def print_status(self, project):
+        # `git status` shows `manifest-rev` "sometimes", see
+        # #643. If manifest-rev is missing, then we already
+        # failed earlier anyway.
+        if self.color_ui:
+            color = '-c status.color=always'
```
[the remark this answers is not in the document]
mbolivar-nordic: > Any idea why `git -c status.color=always status | less -R` shows colors but `west compare | less -R` does not with this PR?

west won't do color if stdout is not a tty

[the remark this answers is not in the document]
marc-hb: Answering myself after a chat with @mbolivar-nordic : `colorama` has no "always" concept and always strips colors when connected to a terminal. Commenting out `colorama.init()` "fixes" `west compare  | less -R`

PR #654, opened 2023-04-25
main: handle ManifestImportError from extensions better

mbolivar-nordic:
The default west manifest file in zephyrproject-rtos/zephyr recently added a project with an 'import'.

That means that 'west build' will fail with the confusing 'invalid choice: "build"' error if you pull zephyr without running 'west update'. This is a worse experience than what happens when you run e.g. 'west list': running a built-in command that needs the manifest to do its work will happily tell you you need to run 'west update' to resolve the failed import.

This is confusing people who pull and run 'west build', and we do have enough information to do better from WestArgumentParser.error(), which is the official argparse callback API invoked when parse_known_arguments() fails in this situation.

Extract the error message formatting for this situation into a helper function and use it from error() to print the same message when we get ManifestImportError in these situations.

Make the error message take up fewer lines while we are here.

### DIFF
--- a/src/west/app/main.py
+++ b/src/west/app/main.py
@@ -200,26 +200,7 @@ class WestApp:
                 if args.command == 'update':
                     return      # that's fine
 
-                p, imp = self.mle.project, self.mle.imp
-                ctxt = f'  Failed importing "{imp}"'
-                if not isinstance(p, ManifestProject):
-                    # Try to be more helpful by explaining exactly
-                    # what west.manifest needs to happen before we can
-                    # resolve the missing import.
-                    rev = p.revision
-
-                    ctxt += f' from revision "{rev}"\n'
-                    ctxt += '  Hint: for this to work:\n'
-                    ctxt += f'          - {p.name} must be cloned\n'
-                    ctxt += (f'          - its {MANIFEST_REV_BRANCH} ref '
-                             'must point to a commit with the import data\n')
-                    ctxt += '        To fix, run:\n'
-                    ctxt += '          west update'
-
-                self.queued_io.append(
-                    lambda cmd:
-                    cmd.die(f'failed manifest import in {p.name_and_path}\n' +
-                            ctxt))
+                self.queued_io.append(lambda cmd: cmd.die(mie_msg(self.mle)))
             elif isinst(FileNotFoundError):
                 # This should ordinarily only happen when the top
                 # level manifest is not found.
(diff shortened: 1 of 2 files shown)


mbolivar-nordic:
cc @nordicjm

mbolivar-nordic:
Tacked on a commit to improve the lives of API users.

aescolar (approved):

marc-hb:
I just tested (not reviewed) this and it does fix the `west build` error message; from the confusing `west: error: argument <command>: invalid choice: 'build` to the relevant `FATAL ERROR: failed manifest import in bsim (tools/bsim)`

- https://github.com/zephyrproject-rtos/zephyr/issues/57198

src/west/app/main.py
```diff
@@ -860,21 +841,61 @@ def add_argument(self, *args, **kwargs):
...
+                # import failed and thus we have no manifest to
+                # load extensions from.
+                #
+                # Just use the help command as a stand-in instead.
+                # We have to manually patch up its config
+                # attribute in order to do this outside of
+                # do_run().
+                cmd = app.builtins['help']
+                cmd.config = west.configuration.Configuration(
+                    topdir=app.topdir)
```
marc-hb, line 866: I don't understand why this `.config` "patch-up" is needed sorry.

src/west/app/main.py
```diff
@@ -860,21 +841,61 @@ def add_argument(self, *args, **kwargs):
...
-            self.west_app.cmd.die(mve_msg(self.west_app.mle))
+        if self.west_app and self.west_app.mle:
+            # If we have a known WestApp instance and the manifest
+            # failed to load, then try to specialize the error message
+            # to handle west-specific situations.
+
+            app = self.west_app
+            mle = self.west_app.mle
+            if app.cmd:
+                cmd = app.cmd
```
marc-hb, line 852: Maybe `failed_cmd = app.cmd`? I've been experiencing a fair amount of "shorthand" fatigue this review...

marc-hb (approved):
The force is strong with this one. But the "bus factor" is very low.

I hate to ask for these few cosmetic changes because the clarity of the code and comments in west is generally very high and this PR is no exception. However the complexity of this part of the code is... even higher? and I suspect the number of people familiar with it is very close to 1.

Non-blocking, if you're in the rush to get this merged then LGTM.

src/west/app/main.py
mbolivar-nordic: Because we aren't inside do_run(), and thus configuration isn't accessible in the usual way. This is part of the parameter contract for WestCommand.config:

```
        '''Property for the config which was passed to run().

        If `do_run` was given a *config* kwarg, it is returned.
        Otherwise, a fatal error occurs.
        '''
```

We're not running a command, so we're not in do_run(). So we have to patch this up manually.

src/west/app/main.py
mbolivar-nordic: > Maybe failed_cmd = app.cmd?

That's not accurate, though: we are just looking for a WestCommand we can call .die() on. That might be a command which has failed, but it also might be the case that we have no knowledge of what command to run, and are just grabbing some other WestCommand instance, somewhat at random, to use instead. This method is definitely a white-box sort of thing, relying on implementation details elsewhere.

mbolivar-nordic:
@marc-hb I addressed your first two comments and tried to provide explanations for the second two.

src/west/app/main.py
marc-hb: OK, it looks like I don't really know the meaning of the English verb "to patch up".

I thought "to patch up" meant modifying an existing `.config` but you're saying this code is making a _missing_ one up? That makes more sense to me now.

So I would have written "make up a missing .config attribute" but you're the native speaker and I'm not so never mind!

marc-hb (commented):
Thanks for the cosmetic changes. Still approved.

Issue #543, opened 2021-09-29
Enhancement: allowing 'groups:' + 'import:'


mbolivar-nordic:
As of west 0.11.x, you cannot combine `groups` with `import` in a single project definition. The documentation says:

> As a restriction, no project may use both import: and groups:. (This avoids some edge cases whose semantics are difficult to specify.)

If you try, you'll hit the error handling block that begins here:

https://github.com/zephyrproject-rtos/west/blob/38e656b05ea8f4c8d80b953f6d88b1ed604d11f8/src/west/manifest.py#L2027

This issue describes:

- a motivating use case for relaxing this restriction
- a rationale for why I introduced the restriction in the first place
- a proposal for how to remove the restriction and deal with the resulting difficulties
- a [chicken bit](https://en.wiktionary.org/wiki/chicken_bit) I would put in place to save us in case the whole idea is a great big mistake

# Use case

Consider a CI maintainer for a downstream Zephyr-based software distribution that provides additional platform support on top of the base OS.

As that CI maintainer, I (the hypothetical CI maintainer, not necessarily @mbolivar-nordic) would like to associate a set of test repositories with the manifest repository for my Zephyr distribution. I would like to do this by including some repositories containing test cases in my west manifest.

This will allow me to conveniently version the test repository revisions alongside the revisions of the code being tested. That allows me to maintain different branches of the same test case repositories to exercise different branches of the accompanying downstream distribution. Furthermore, putting my test repositories into west allows my platform developers to more easily reproduce results such as build errors in my test applications by running `west update` to get the test code.

I would like to adapt the [continuous integration overrides](https://docs.zephyrproject.org/latest/guides/west/manifest.html#west-manifest-ex2-3) pattern from the west documentation for my use case. This will also allow me to exercise changes that have to go into my test and platform repositories at the same time, which may occur for example if a Zephyr API used by both types of repository changes. I would like to adapt the documented pattern as follows.

First, I do not want developers to have to clone these test repositories by default into the workspace for several potential reasons:

- they are not relevant to day-to-day development
- they contain proprietary information
- they contain large amounts of binary test data or other artifacts

I therefore would like the test repositories to be inactive projects by default.

Second, I do not want the names or locations of the test repositories to be made public as they are also considered proprietary. (This is not incompatible with an open source platform, e.g. [sqlite](https://www.sqlite.org/testing.html) is another OSS project that has a proprietary test suite.)

I therefore would like to organize my manifests like this:

```yaml
[...]

PR #656, opened 2023-05-03
Initial import filtering implementation

mbolivar-nordic:
This feature provides a way to prevent west from performing an import
depending on a condition.

There are a variety of use cases for this:

- CI system maintainers want to be able to define an 'import' that
  contains repositories that contain test cases and which should be
  easy for developers to fetch with 'west update', but which are not
  distributed to users by default (by 'turning off' the import
  somehow)

- Similarly, but for projects which provide optional features that can
  be safely ignored unless the user takes an explicit action to enable
  the import. The main use case here isn't limiting visibility of
  the feature, but rather limiting the storage size and network
  bandwidth required to initialize a basic workspace

Provide this feature by extending the 'import: <map>' syntax with a
new key, 'if':

```yaml
  projects:
    - name: foo
      import:
        if:
          <condition>
```

Where the value of `<condition>` is itself a map, so that we can add
features to it over time. In this commit, we add the infrastructure
for the 'if' key and one simple way to write `<condition>`:

```yaml
  import:
    if:
      config-enabled: foo.bar
```

Here, the import will be resolved if and only if the west
configuration option 'foo.bar' is defined and set to a true boolean
value (see the python configparser docs for all the ways to write a
boolean option; "west config foo.bar true" is one way to set this up).

The configuration option can be set in any configuration file: local,
global, or system. This makes it easier for users who will want the
import performed but regularly set up workspaces in various places to
set it up permanently, without having to enable foo.bar every time
they set up a workspace, before the first "west update".

Design notes for the record:

- I was originally trying to do something like "import: if:
  group-enabled: bsim", but that runs into possibilities for
  self-contradiction, because a project that is imported later than
  you can disable a group, so you don't really know if a group is
  enabled or not until you are done parsing the manifest.

- It thus made sense to me to define a syntax for conditional
  imports that doesn't rely on groups at all. Configuration options
  are independent of groups and are set once and for all by the time
  we are loading the manifest, so it seemed like a good place to
  start.

- People are still rapidly ideating new ways to do import
  filtering for different use cases, so a one-size-fits-all solution
  seems unlikely. Making a flexible syntax that we can extend later
  therefore seemed more future-proof. In the future, we can do things
  like:

```yaml
    import:
      if:
        env: SOME_ENV_VAR=1
```

```yaml
    import:
      if:
        anyOf:
          config-enabled: foo.bar
          env: SOME_ENV_VAR=1
```

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -17,6 +17,7 @@ import subprocess
 import sys
 import textwrap
 from time import perf_counter
+from typing import List as ListType
 from urllib.parse import urlparse
 
 from west.configuration import Configuration
@@ -962,7 +963,7 @@ class Update(_ProjectCommand):
         else:
             self.update_some()
 
-    def init_state(self, args):
+    def init_state(self, args: argparse.Namespace):
         # Helper for initializing instance state in response to
         # command line args and configuration files.
 
@@ -977,7 +978,7 @@ class Update(_ProjectCommand):
         self.sync_submodules = config.getboolean('update.sync-submodules',
                                                  default=True)
 
-        self.group_filter: List[str] = []
+        self.group_filter: ListType[str] = []
 
         def handle(group_filter_item):
             item = group_filter_item.strip()
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -2191,10 +2191,11 @@ class Manifest:
             groups = []
 
         if imp and groups:
-            # Maybe there is a sensible way to combine the two of these.
-            # but it's not clear what it is. Let's avoid weird edge cases
-            # like "what do I do about a project whose group is disabled
-            # that I need to import data from?".
+            # There isn't a sensible way to handle this case. We can't
+            # decide whether or not to import because we don't know if
+            # the project is active or not until after all group filters
+            # are resolved, which won't happen until we're done resolving
+            # the manifest.
             self._malformed(
                 f'project {name}: "groups" cannot be combined with "import"')
 
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -224,6 +224,7 @@ class _import_map(NamedTuple):
     name_blocklist: List[str]
     path_blocklist: List[str]
     path_prefix: str
+    if_: Dict[str, Any]
 
 def _is_imap_list(value: Any) -> bool:
     # Return True if the value is a valid import map 'blocklist' or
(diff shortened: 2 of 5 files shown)


marc-hb:
Fixes https://github.com/zephyrproject-rtos/west/issues/543 or did I miss something?

marc-hb:
> Where the value of <condition> is itself a map, so that we can add features to it over time.

The flexibility looks great to avoid painting ourselves in a corner but I'm not sold on the first, `config`-based implementation.


> I was originally trying to do something like "import: if: group-enabled: bsim", but that runs into possibilities for self-contradiction, because a project that is imported later than you can disable a group, so you don't really know if a group is enabled or not until you are done parsing the manifest.

That contradiction and potential remedies were discussed in #543. One of the conclusions was: this contradiction stems from the fact that import filtering and group filtering are two very different things. So you seem to have made them very different indeed in this PR based on the description (I haven't read the code yet, I believe this is still at the "design" phase).

Aren't they _too_ different now?

Let's take for instance bsim issue https://github.com/zephyrproject-rtos/zephyr/issues/57198 which looks like it motivated this (57198 is one week old whereas #543 is 1-2 years old). Can you describe in detail the west configuration that would solve 57198 with this PR? (Ignoring of course other later attempts to solve 57198). More specifically: would enabling or disabling `bsim` require TWO settings with this PR? That seems very inconvenient.

A "combo switch" is the main suggestion I made in #543. Long story short: import filtering and group filtering are kept distinct, happening at distinct phases and they don't talk to each other. BUT they just happened to look at the same filter keywords for UI convenience. This gives import filtering some sort of "precedence" to resolve self-contradicting configurations when creative users try to break west. That precedence is only because it runs first; again they don't interact with each other.

aescolar:
> More specifically: would enabling or disabling bsim require TWO settings with this PR? That seems very inconvenient.

@marc-hb It is my understanding that we'd only use the import filter in that case and remove the "babblesim" group filter, as it would otherwise just be redundant.

nordicjm:
> Here, the import will be resolved if and only if the west
> configuration option 'foo.bar' is defined and set to a true boolean
> value (see the python configparser docs for all the ways to write a
> boolean option; "west config foo.bar true" is one way to set this up).
>
> The configuration option can be set in any configuration file: local,
> global, or system. This makes it easier for users who will want the
> import performed but regularly set up workspaces in various places to
> set it up permanently, without having to enable foo.bar every time
> they set up a workspace, before the first "west update".

If I read this rightly, it means that an additional west command is required to enable a group, but doesn't that then mean it can't be enabled by a custom west manifest? E.g. if someone wants to setup a manifest which imports zephyr and gets babblesim, they wouldn't be able to do that from their custom west.yml unless they ran a `west config` command to enable it? If that's the case, it does seem strange to have manifests with options that cannot be changed from manifests that include them and must be set in local west configuration files.

mbolivar-nordic:
Fair points @marc-hb and @nordicjm . Today @carlescufi @aescolar and I met and discussed an alternative approach I will prototype today.

mbolivar-nordic:
Closing this for now since we're going another way

Issue #652, opened 2023-04-24
Better support for projects with imports when users do not have them locally


aescolar:
When a manifest has a project with an import, west requires that project to be present locally so it can resolve the whole manifest before it can do quite a few of its functions.
If that is not the case, some west commands/API operations will fail in a bit mysterious ways.
Some users are either confused by the way west fails, and or / do not like this requirement, and/or consider it unnecessary.

I create this issue to try to list possible solutions and/or improvements, so maybe be arrive at a nice option.

1. west supports projects with `import` that are also part of a group https://github.com/zephyrproject-rtos/west/issues/543. This is not possible today, but would solve the problem when the project is not actually necessary (wrt to users being required to always have a project with an import)
2. west does not fail when a project with an import is missing, but instead prints a warning to stderr and continues like if the project did not have the import set: This would alleviate the problem for users who knowingly do not want that project. Additionally west could offer the option to wave that warning by configuration.
3. west provides a clearer error for west extension commands when this type of failure occurs (this would at least reduce the confusion and possible following frustration from users)
4. west provides clearer errors for all commands or API calls when it tries to collate the manifest but it fails due to a missing import.
5. `west update` having a new option to pull "whatever is the minimum needed" for west to work further. That is, all projects with imports, apart from the projects which may be explicitly listed. For ex. `west update --and_required project_x project_y`. 
6. (Variant of the previous). `west init` having a new option to pull "whatever is the minimum needed" for west to work further...

aescolar:
CC @mbolivar @carlescufi

mbolivar-nordic:
> 4\. west provides clearer errors for all commands or API calls when it tries to collate the manifest but it fails due to a missing import.

For commands: https://github.com/zephyrproject-rtos/west/pull/654

For API calls: unless the exceptions generated by the API are confusing, I think that's a problem for the user of the west API, not west itself.

mbolivar-nordic:
> 1. west supports projects with import that are also part of a group Enhancement: import filtering #543. This is not possible today, but would solve the problem when the project is not actually necessary (wrt to users being required to always have a project with an import)

To be clear, 'groups' and 'import' **cannot** be combined in a single project. This is not a question of priority, but semantics.

Import filtering is a separate enhancement that has been discussed to try to do filtering of imports in a way that doesn't use 'groups'.

[...]

Issue #653, opened 2023-04-25
Provide local project filter in west config


aescolar:
As an enhancement to west,
provide a new feature in which users can, thru local west workspace configuration, filter projects out of a manifest.
It could either be thru a deny-list or an allow-list (or both).

### Motivation/Use case:
A user wants to track a project main manifest (e.g. Zephyr), but only needs for his work a subset of the projects that project manifest has.
The users wants/needs to save disk space, "west update" time and/or bandwidth. Yet he wants to jump back and forth in branches of that project. The user would prefer to avoid needing to create another project or super-manifest that includes the main project of interest as an import with an allowlist, as that requires extra steps (as he would like to jump between several branches, which are changing over time, he would need to keep updating/branching that super-manifest).
The user does not have any means to affect the project of interest to require it to add group filters, or some of the projects cannot be added to group filters (for ex. due to imports)

The workflow would be something like:
A user does a `west init`, after which (probably before a west update) he configures the project to filter in/out what he does want/not want (`west config manifest.project-filter "-.*, +project_x, +project_y"`) . And then he does subsequent west updates, which just ignore the filtered out projects/all but the filtered-in projects.

---------

As an extension to this idea, so we would also cover the needs of https://github.com/zephyrproject-rtos/west/issues/652 . In manifests projects definitions, allow a new key "default-inactive".
When a project has this key, the project is filtered out unless the local configuration project-filter enables it.
All other projects (without 'default-inactive') would be enabled by default unless the local configuration project-filter disables them.
There is no "default-active" key (it would be redundant). 
Note: The project being filtered out means also a possible import done by that project is also ignored.
(Also, an imported manifest cannot define the same project and set "default-inactive : false", as the imported project with the same name is just ignored by west (this is the behavior already today)).

------

As a possible implementation (just to provide an understanding of a desired workflow and expected west behavior, not necessarily guide the solution):
The project-filter would be a list of projects names regexes, each preceded with "+" or "-".
When evaluating if a project is enabled or not, west would go thru this project-filter list from right to left. As soon as it finds the project name with a "-" it considers the project filtered out (and stops searching thru the list). As soon as it finds the project with a "+" , it considers it not filtered out (and stops searching thru the list).
The "default-inactive" key for <project_x> would be functionally equivalent to the list having as its left-most element -<project_x>

[...]

PR #660, opened 2023-05-09
'import-group-filters' top level manifest key

mbolivar-nordic:
**Abandoned**

@tejlmand has pointed out that this feature, if implemented, means that `west manifest --resolve` is no longer possible to implement. The semantics in this proposed feature would make the set of enabled/disabled groups context-specific, differing at different points in time when resolving imports. This in turn makes it impossible to write a single "equivalent" manifest file. We want to continue to support `west manifest --resolve` and believe the extra complexity and poorly specified semantics points to this idea being a nonstarter.

----------------------------------------------------------------------

This is the agreed upon approach for how we're going to handle the babblesim module in zephyr/west.yml (#656, #652, #653, #543).

The new 'import-group-filters' feature is a top level manifest key that takes a boolean. The default
value is 'true' and is west's current behavior. The behavior when this value is false can be shown by this example:

```yaml
  manifest:
    group-filter: [-foo]
    import-group-filters: false
    projects:
      - name: proj
        import: true
    self:
      import: submanifest.yml
```

Here, the final Manifest.group_filter attribute for the manifest whose data are shown above will always be [-foo], regardless of what the imported manifest data from 'proj' or 'submanifest.yml' are.

The purpose of this extension is to allow the combination of 'groups' and 'import':

- In previous versions of west, we could not combine the two, because
  in pathological edge cases, we could end up resolving an import for a
  group which was later disabled by an import that happened later on.

- With this new backwards-compatible enhancement to the manifest file
  format, we can combine 'import:' with 'groups:' in a single project
  whenever 'import-group-filters:' is 'false'. This is because we will
  know the final 'group_filter' attribute for the entire Manifest at
  the time we perform the import, and it cannot be changed later by
  any imports we might perform.

Since this new behavior would break backwards compatibility if made the default, we make it explicitly opt-in by requiring 'import-group-filters: false'.

Implementing this feature exposed a bug in the current group filter handling. The new implementation does not have this bug.

Fixes: #663
Fixes: #652

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -576,8 +576,7 @@ class ManifestCommand(_ProjectCommand):
         except _ManifestImportDepth:
             self.die("cannot resolve manifest -- is there a loop?")
         except ManifestImportFailed as mif:
-            self.die(f"manifest import failed\n  Project: {mif.project}\n  "
-                     f"File: {mif.filename}")
+            self.die(str(mif))
         except (MalformedManifest, ManifestVersionError) as e:
             self.die('\n  '.join(str(arg) for arg in e.args))
         dump_kwargs = {'default_flow_style': False,
--- a/src/west/manifest-schema.yml
+++ b/src/west/manifest-schema.yml
@@ -35,6 +35,16 @@ mapping:
     required: false
     type: text
 
+  # The "import-group-filters" key is an optional boolean that defaults
+  # to 'true'. If it's 'false', then the group filter for this manifest
+  # will not depend on the group filters of any manifests it imports.
+  # If it's 'true', it will represent the combination of its own group
+  # filter and the group filters of imported manifests. Setting this to
+  # 'false' is the only way to combine 'import:' with 'groups:' in a project
+  # element.
+  import-group-filters:
+    type: bool
+
   # The "defaults" key specifies some default values used in the
   # rest of the manifest.
   defaults:
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -1788,7 +1788,15 @@ class Manifest:
 
             current_relpath = manifest_path / manifest_file
             current_abspath = topdir_abspath / current_relpath
-            current_data = current_abspath.read_text(encoding='utf-8')
+            try:
+                current_data = current_abspath.read_text(encoding='utf-8')
+            except FileNotFoundError:
+                raise MalformedConfig(
+                    f'file not found: manifest file {current_abspath} '
+                    '(from configuration options '
+                    f'manifest.path="{manifest_path_option}", '
+                    f'manifest.file="{manifest_file}")')
+
             current_repo_abspath = topdir_abspath / manifest_path
 
             self.abspath = os.fspath(current_abspath)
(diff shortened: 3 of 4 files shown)


carlescufi:
After giving this a cursory review, I believe this to be the right solution for #543

carlescufi:
After giving this a cursory review, I believe this to be the right solution for #543

carlescufi:
After giving this a cursory review, I believe this to be the right solution for #543

mbolivar-nordic:
I'm polishing this up and will push when I've got something that doesn't introduce regressions

aescolar:
@mbolivar-nordic is this ready to give it a spin?

aescolar (approved):

aescolar:
@marc-hb , @mbolivar-nordic would better answer, but as he is in a different timezone I'll do my best:

> So in other words this proposition makes non-leaf projects able to join groups too
yes

> Please correct me again: what the new paradigm removes is the ability for every manifest except the top one to perform any sort of filtering (which was likely a design mistake in the first place). 

It prevents group filters from propagating "upwards" in the manifests hierarchy. A manifest can still set a group filter that affects its own projects or the project it imports, but a child manifest group filters does not affect the parent manifest group filters.

> So is the intention to switch the main, upstream zephyr/west.yml to this new import and filtering paradigm?
> After switching the main, upstream zephyr/west.yml to this new paradigm, then the (non-leaf) bsim project can be made part of the babblesim group which is disabled by default? And this will restore the ability to west update only what I need and fix https://github.com/zephyrproject-rtos/zephyr/issues/57198 ?

Yes to all, with 

```diff 
--- a/west.yml
+++ b/west.yml
@@ -23,6 +23,7 @@ manifest:
       url-base: https://github.com/zephyrproject-rtos
 
   group-filter: [-babblesim]
+  import-group-filters: false
 
   #
   # Please add items below based on alphabetical order
@@ -30,6 +31,8 @@ manifest:
     - name: bsim
       repo-path: babblesim-manifest
       revision: 908ffde6298a937c6549dbfa843a62caab26bfc5
+      groups:
+        - babblesim
       import:
         path-prefix: tools
     - name: canopennode
```

> BTW: when this new attribute is found anywhere but at the TOP, it will be silently ignored, correct?

@mbolivar-nordic better answer this.

aescolar:
@mbolivar-nordic 
It seems it is not there yet? :( 
```
west init -m git@github.com:aescolar/zephyr.git zp --mr bsim_new_west 
cd zp
west update
west config manifest.group-filter -- "+babblesim"
west update <- fail (everything fails now)
 File "west/src/west/manifest.py", line 2504, in _import_content_from_project
    content = self._ctx.project_importer(project, path)
  File "west/src/west/app/project.py", line 1051, in update_importer
    assert not project.groups
AssertionError
```
or 
```
west init -m git@github.com:aescolar/zephyr.git zp2 --mr bsim_new_west  
cd zp2
west config manifest.group-filter -- "+babblesim"
west update <- fail (everything fails now)
```

mbolivar-nordic:
> Why is #664 preferred now? Please don't close PRs or issues without a single word...

I updated the PR description when I closed it with rationale

PR #662, opened 2023-05-10
compare: add --decorate

marc-hb:
When a project is not on the `manifest-rev` it will often be on a branch or tag. Use the `%d` format to make `west compare` show that information.

As discussed in #643 and before, git status cannot be fully trusted to display branch information.

Sample new output:
```
west compare
=== zephyr (zephyr):
--- manifest-rev: e40859f78712 (some_tag) Revert "dma: dw: Do ...
            HEAD: e803b77463b4 (zephyrproject/main) samples: net: ...
--- status:
    HEAD detached at zephyrproject/main
```

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -736,10 +736,19 @@ class Compare(_ProjectCommand):
             return
 
         def rev_info(rev):
-            return project.git(
-                ['log', '-1', '--color=never', '--pretty=%h %s', rev],
+            title = project.git(
+                ['log', '-1', '--color=never', '--pretty=%h%d %s',
+                 '--decorate-refs-exclude=refs/heads/manifest-rev',
+                 rev],
                 capture_stdout=True,
                 capture_stderr=True).stdout.decode().rstrip()
+            # "HEAD" is special; '--decorate-refs-exclude=HEAD' doesn't work.
+            # Fortunately it's always first.
+            return (
+                title.replace('(HEAD) ', '').replace('(HEAD, ', '(')
+                .replace('(HEAD -> ', '(')
+            )
+
         head_info = rev_info('HEAD')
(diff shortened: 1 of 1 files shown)


src/west/app/project.py
```diff
@@ -736,10 +736,19 @@ def print_rev_info(self, project):
             return
 
         def rev_info(rev):
-            return project.git(
-                ['log', '-1', '--color=never', '--pretty=%h %s', rev],
+            title = project.git(
+                ['log', '-1', '--color=never', '--pretty=%h%d %s',
+                 '--decorate-refs-exclude=refs/heads/manifest-rev',
```
mbolivar-nordic, line 741: Did you check that this doesn't introduce a requirement for a more recent version of git?

mbolivar-nordic:
(LGTM otherwise)

src/west/app/project.py
marc-hb: No I hadn't, my bad. These options were introduced in git commit v2.16.0-rc0~64^2 https://github.com/git/git/commit/65516f586b69307f977

Git 2.16.0 was released in January 2018, 5+ years ago.

I couldn't find what is the minimum git version required by west, where is it?

mbolivar-nordic: That's fine then; we have a `        if self.git_version_info < (2, 22):` check up on line 659.

mbolivar-nordic (commented):
@marc-hb this LGTM now but offline you mentioned this being broken by west only fetching by URL -- can you confirm or deny that this is OK to merge?

marc-hb:
> but offline you mentioned this being broken by west only fetching by URL

It's not "broken", it may make it slightly less useful, showing less decorations in some cases.

mbolivar-nordic (approved):

PR #664, opened 2023-05-31
`manifest.project-filter`

### LINKED ISSUE #653: Provide local project filter in west config

mbolivar-nordic:
Take three at trying to provide an "import filtering" (#543) solution that will work to allow making the bsim project in zephyr/west.yml inactive, so you can `west build` etc without it cloned.

To test:

```
# set up a workspace without a bsim project any way you want. Example:
west init zephyrproject
cd zephyrproject

# exercise the new config option in a local workspce:
west list # error because bsim project has 'import' and is not cloned
west config manifest.project-filter -- -bsim
west list # works, and bsim project is not in the output (it's inactive)
west list bsim # works, prints the usual info for bsim
```

alternatively:

```
west config --global manifest.project-filter -- -bsim
west init zephyrproject
cd zephyrproject
west list # works
```

------

Add support for a new manifest.project-filter configuration option.

The option's value is a comma-separated list of regular expressions,
each prefixed with '+' or '-', like this:

```
    +re1,-re2,-re3
```

Project names are matched against each regular expression (re1, re2,
re3, ...) in the list, in order. If the entire project name matches
the regular expression, that element of the list either deactivates or
activates the project. The project is deactivated if the element
begins with '-'. The project is activated if the element begins with
'+'. (Project names cannot contain ',' if this option is used, so the
regular expressions do not need to contain a literal ',' character.)

If a project's name matches multiple regular expressions in the list,
the result from the last regular expression is used. For example,
if manifest.project-filter is:

```
    -hal_.*,+hal_foo
```

Then a project named 'hal_bar' is inactive, but a project named
'hal_foo' is active.

If a project is made inactive or active by a list element, the project
is active or not regardless of whether any or all of its groups are
disabled. (This is also now the only way to make a project that has no
groups inactive.)

Otherwise, i.e. if a project does not match any regular expressions in
the list, it is active or inactive according to the usual rules
related to its groups.

Within an element of a manifest.project-filter list, leading and
trailing whitespace are ignored. That means these example values
are equivalent:

```
    +foo,-bar
    +foo , -bar
```

The lists in the manifest.project-filter options in the system,
global, and local configuration files are concatenated together.
For example, on Linux, with:

```
    /etc/westconfig (system file):
    [manifest]
    project-filter = +foo

    ~/.westconfig (global file):
    [manifest]
    project-filter = -bar_.*

    <workspace>/.west/config (local file):
    [manifest]
    project-filter = +bar_local
```

Then the overall project filter when within the workspace <workspace>
is:

```
    +foo,-bar_.*,+bar_local
```

This lets you set defaults in the system or global files that can be
overridden within an individual workspace as needed, without having to
maintain the defaults across multiple workspaces.

----

Fixes #653

### DIFF
--- a/src/west/app/main.py
+++ b/src/west/app/main.py
@@ -318,8 +318,8 @@ class WestApp:
             if isinst(MalformedManifest, MalformedConfig):
                 self.queued_io.append(
                     lambda cmd:
-                    cmd.die('\n  '.join(["can't load west manifest"] +
-                                        list(self.mle.args))))
+                    cmd.die("can't load west manifest: " +
+                            "\n".join(list(self.mle.args))))
             elif isinst(_ManifestImportDepth):
                 self.queued_io.append(
                     lambda cmd:
--- a/src/west/app/main.py
+++ b/src/west/app/main.py
@@ -26,6 +26,8 @@ from subprocess import CalledProcessError
 import tempfile
 import textwrap
 import traceback
+from typing import NamedTuple, Optional
+from typing import List as ListType
 
 from west import log
 import west.configuration
@@ -39,6 +41,103 @@ from west.manifest import Manifest, MalformedConfig, MalformedManifest, \
     ManifestProject, MANIFEST_REV_BRANCH
 from west.util import quote_sh_list, west_topdir, WestNotFound
 from west.version import __version__
+
+class EarlyArgs(NamedTuple):
+    # Data type for storing "early" argument parsing results.
+    #
+    # We do some manual parsing of the command-line arguments before
+    # delegating the hard parts to argparse. This extra work figures
+    # out the command name, verbosity level, etc.
+    #
+    # This is necessary for:
+    #
+    # - nicer error-handling in situations where the command is an
+    #   extension but we're not in a workspace
+    #
+    # - setting up log levels from the verbosity level
+
+    # Expected arguments:
+    help: bool                  # True if -h was given
+    version: bool               # True if -V was given
+    zephyr_base: Optional[str]  # -z argument value
+    verbosity: int              # 0 if not given, otherwise counts
+    command_name: Optional[str]
+
+    # Other arguments are appended here.
+    unexpected_arguments: ListType[str]
+
+def parse_early_args(argv: ListType[str]) -> EarlyArgs:
+    # Hand-rolled argument parser for early arguments.
+
+    help = False
+    version = False
+    zephyr_base = None
+    verbosity = 0
+    command_name = None
+    unexpected_arguments = []
+
+    expecting_zephyr_base = False
+
+    def consume_more_args(rest):
+        # Handle the 'Vv' portion of 'west -hVv'.
+
+        nonlocal help, version, zephyr_base, verbosity, command_name
+        nonlocal unexpected_arguments
+        nonlocal expecting_zephyr_base
+
+        if not rest:
+            return
+
+        if rest.startswith('h'):
+            help = True
+            consume_more_args(rest[1:])
+        elif rest.startswith('V'):
+            version = True
+            consume_more_args(rest[1:])
+        elif rest.startswith('v'):
+            verbosity += 1
+            consume_more_args(rest[1:])
+        elif rest.startswith('z'):
+            if not rest[1:]:
+                expecting_zephyr_base = True
+            elif rest[1] == '=':
+                zephyr_base = rest[2:]
+            else:
+                zephyr_base = rest[1:]
+        else:
+            unexpected_arguments.append(rest)
+
+    for arg in argv:
+        if expecting_zephyr_base:
+            zephyr_base = arg
+        elif arg.startswith('-h'):
+            help = True
+            consume_more_args(arg[2:])
+        elif arg.startswith('-V'):
+            version = True
+            consume_more_args(arg[2:])
+        elif arg == '--version':
+            version = True
+        elif arg.startswith('-v'):
+            verbosity += 1
+            consume_more_args(arg[2:])
+        elif arg == '--verbose':
+            verbosity += 1
+        elif arg.startswith('-z'):
+            if arg == '-z':
+                expecting_zephyr_base = True
+            elif arg.startswith('-z='):
+                zephyr_base = arg[3:]
+            else:
+                zephyr_base = arg[2:]
+        elif arg.startswith('-'):
+            unexpected_arguments.append(arg)
+        else:
+            command_name = arg
+            break
+
+    return EarlyArgs(help, version, zephyr_base, verbosity,
+                     command_name, unexpected_arguments)
 
 class WestApp:
     # The west 'application' object.
@@ -328,6 +427,8 @@ class WestApp:
         # Remember to update zephyr's west-completion.bash if you add or
         # remove flags. This is currently the only place where shell
         # completion is available.
+        #
+        # If you update these, also update parse_early_args().
 
         parser.add_argument('-h', '--help', action=WestHelpAction, nargs=0,
                             help='get help for west or a command')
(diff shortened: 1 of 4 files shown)


aescolar:
Thanks @mbolivar-nordic, it is looking very nice so far.
A minor request: It would be nice if the manifest repo was not affected by the filter. Background: I expect users very often want to do a -*, or +* (for example to ignore global config, or have a clean slate), but there is no purpose of having west ignoring the manifest repo.

aescolar:
Regarding  the "default-disable" feature. Would it make sense to add it right away?
(I guess the logic would be something like adding just before
https://github.com/zephyrproject-rtos/west/pull/664/files#diff-6ef1afe17ca985687d1222e9ece942bbf4fdca50c2baa784b1dfb3504b078acbR1730
something like
```
​if  (ret == PFR.NONE) and (project.default_disabled)
   ret = PFR.INACTIVE
```
(apart from the extra changes to support that extra project key)

mbolivar-nordic:
>  Would it make sense to add it right away?

I don't think so. There is still a fair bit of work to do to get the current feature set in a merge-able state and I think such an addition is not a blocker.

mbolivar-nordic:
> A minor request: It would be nice if the manifest repo was not affected by the filter

That's a bug, thanks

mbolivar-nordic:
To be confirmed: `west manifest --resolve` desired behavior after this PR

mbolivar-nordic:
> For west commands that require the whole manifest to be resolved (e.g. `resolve`, `freeze`, etc), the proposed approach is to let west error out if the `project-filter` configuration option is set at all when invoking them.

Note that this is just a temporary solution. We can add full support for these commands at a later date when the enhancement is required for a concrete use case. This differs from #660, which would have prevented us from implementing these commands at all.

src/west/manifest.py
```diff
@@ -1758,7 +1835,27 @@ def get_option(option, default=None):
...
+                return config.get('manifest.project-filter',
+                                  configfile=configfile)
+
+            # Update our project filter based on the value in all the
+            # configuration files. This allows us to progressively
+            # build up a project filter with default settings in
+            # configuration files with wider scope, refining as
+            # needed. For example, you could do '-foo,-bar' in the
+            # global config file, and then '+bar' in the local config
+            # file, to have a final filter of '-foo'.
```
aescolar, line 1903: Nit: This comment is not fully correct. The correct comment would be
`.. in the local config file, to have a final filter of '-foo,-bar,+bar'.`

(The code is just appending more to the list.
`-foo,-bar,+bar != -foo` as on the left one, bar is active even if bar groups are disabled.)

aescolar (approved):
Just 3 minor nits, for which there is no need to hold this.

tejlmand:
> The lists in the manifest.project-filter options in the system, global, and local configuration files are concatenated together.

i'm not so sure on this.
`west` also supports `manifest.group-filter`, and one could with good reasons assume that both kind of filters would behave similar.

For example if I can have:
```
    /etc/westconfig (system file):
    [manifest]
    project-filter = +foo

    ~/.westconfig (global file):
    [manifest]
    project-filter = -bar

    <workspace>/.west/config (local file):
    [manifest]
    project-filter = +bar_local
```
why can I then not have similar for groups ?
```
    /etc/westconfig (system file):
    [manifest]
    group-filter = +foo

    ~/.westconfig (global file):
    [manifest]
    group-filter = -bar

    <workspace>/.west/config (local file):
    [manifest]
    group-filter = +bar_local
```

I know `project-filter` supports regex'es, but that's the only syntax difference between those filters, and not enough to justify such a behavioral difference.
I prefer consistency between those settings in this case.
(Feel free to propose `group-filter` to be concatenated in same fashion as `project-filter`)

carlescufi:
@aescolar and myself discussed [this comment](https://github.com/zephyrproject-rtos/west/pull/664#issuecomment-1571681797) with @tejlmand. The inconsistency in handling the aggregation of different config levels between `group-filter` and `project-filter` is noted, @mbolivar-nordic to decide if he wants to align `group-filter` to the current behavior of `project-filter` or the other way around.

carlescufi (approved):
Discussed together with @aescolar.

src/west/manifest.py
```diff
@@ -1650,23 +1660,53 @@ def is_active(self, project: Project,
...
+        - The elements of the manifest.project-filter options in each
+          of these files are checked against the project's name. If the
+          regular expression in the element matches the project's name,
+          then the project is active or inactive depending on if the
+          element begins with + or - respectively.
+
+        - If multiple elements have regular expressions matching the
+          project's name, the last element which has a match determines
+          the result.
+
```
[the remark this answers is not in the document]
mbolivar-nordic: I think this is redundant since the rules are documented explicitly in https://docs.zephyrproject.org/latest/develop/west/config.html

src/west/manifest.py
mbolivar-nordic: This paragraph is going away entirely since we're reversing course and just making the highest precedence one take effect

src/west/app/main.py
```diff
@@ -427,6 +561,42 @@ def run_command(self, argv):
         except WestNotFound as wnf:
             self.cmd.die(str(wnf))
 
+    def handle_early_arg_errors(self, early_args):
+        # If early_args indicates we should error out, handle it
+        # gracefully. This provides more user-friendly output than
+        # argparse can do on its own.
```
marc-hb, line 567: It's more user-friendly... unless the manifest fails to import. Then the original, relevant error message is lost, see issue #726 and tentative mitigation #727

PR #670, opened 2023-06-02
[RFC] west: allow list of manifest files

tejlmand:
Currently the `west config manifest.file` allows only a single file.

If multiple manifest files are required, then the user must manually create a custom manifest file and point manifest.file setting to the custom file, for example:
```
manifest:
  self:
    import:
    - fileA.yml
    - fileB.yml
```

This commit allows manifest.file to be a list so that the following `west config manifest.file fileA.yml,fileB.yml` will behave as the above manifest snippet.

`,` is used as separator, but is configurable using `west config manifest.file-separator` to support use-cases where manifest filename contains commas.

### DIFF
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -1869,22 +1869,31 @@ class Manifest:
                                 manifest_path_option)
 
             manifest_file = get_option('manifest.file', _WEST_YML)
+            manifest_file_separator = get_option('manifest.file-separator', ',')
 
-            current_relpath = manifest_path / manifest_file
-            current_abspath = topdir_abspath / current_relpath
-            try:
-                current_data = current_abspath.read_text(encoding='utf-8')
-            except FileNotFoundError:
-                raise MalformedConfig(
-                    f'file not found: manifest file {current_abspath} '
-                    '(from configuration options '
-                    f'manifest.path="{manifest_path_option}", '
-                    f'manifest.file="{manifest_file}")')
+            manifest_files = manifest_file.split(manifest_file_separator)
+            if len(manifest_files) > 1:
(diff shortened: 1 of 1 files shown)


mbolivar-nordic (commented):
Getting the syntax right so that we can express a list is a blocker for me.

Using commas as separators is a no-go in the current implementation since commas are valid characters in manifest file names.

mbolivar-nordic (changes requested):
Thought about this some more. Posting some comments I made internally here for the record. 

Making manifest.file a list is probably a nonstarter in my opinion:

- Potential impact on Manifest.from_file(), Manifest.from_topdir(), Manifest.from_data(), Manifest.__init__() needs to be investigated
- Impact on west manifest subcommands needs to be investigated (this breaks west manifest --path for sure -- others?)
- Getting the syntax right needs to be fixed (comma as separator does not work without other changes as I've commented on the draft PR from Rasmussen, Torsten)
- Precisely specifying the semantics needs to be done still (what if I "git pull" and one or more of the files goes away? any other edge cases?)
- Extending west init to support this needs to be investigated (do we just allow --mf to be given multiple times?)


The impact on the API is also pretty unclear to me and it looks like there are dealbreakers:

- For example, west.manifest.Manifest.abspath is an attribute that contains the absolute path to the manifest file. With multiple, files, this seems broken.
- Would west need to have to have multiple Manifest objects in flight at any given time in this idea? That would break the west.commands.WestCommand.manifest attribute, which both built-in and extension commands rely on to investigate the workspace.

TL;DR west has the assumption that there is one and only one manifest file pretty deeply baked into some core abstractions and I think it is likely that it is too late to change this.

PR #671, opened 2023-06-02
commands: Show less generic error when no manifest is loaded

57300:
Minor quirk that stuck out to me while testing the recent changes. Before https://github.com/zephyrproject-rtos/west/commit/23db6e1f3cc7fb50e0a3e0bda8235892db74c971, if I were to try `west manifest --resolve` in Zephyr without cloning `bsim`, I would see this:
```
FATAL ERROR: project bsim (tools/bsim): cannot import contents of west.yml; do you need to run "west update"?
```
which is helpful enough. Now, this generic message gets displayed:
```
FATAL ERROR: can't run west manifest; it requires the manifest, which was not available. Try "west manifest --validate" to debug.
```
Ironically, `west manifest --validate` produces the same error, which is not very helpful. So, I tried to fix that.

### DIFF
--- a/src/west/app/main.py
+++ b/src/west/app/main.py
@@ -37,8 +37,8 @@ from west.app.project import List, ManifestCommand, Compare, Diff, Status, \
     SelfUpdate, ForAll, Init, Update, Topdir
 from west.app.config import Config
 from west.manifest import Manifest, MalformedConfig, MalformedManifest, \
-    ManifestVersionError, ManifestImportFailed, _ManifestImportDepth, \
-    ManifestProject, MANIFEST_REV_BRANCH
+    _ManifestRequired, ManifestVersionError, ManifestImportFailed, \
+    _ManifestImportDepth, ManifestProject, MANIFEST_REV_BRANCH
 from west.util import quote_sh_list, west_topdir, WestNotFound
 from west.version import __version__
 
@@ -608,9 +608,18 @@ class WestApp:
             self.handle_builtin_manifest_load_err(args)
         for io_hook in self.queued_io:
             self.cmd.add_pre_run_hook(io_hook)
-        self.cmd.run(args, unknown, self.topdir,
-                     manifest=self.manifest,
-                     config=self.config)
+        try:
+            self.cmd.run(args, unknown, self.topdir,
+                         manifest=self.manifest,
+                         config=self.config)
+        except _ManifestRequired:
+            # We tried to run the command despite having failed to
+            # load_manifest(), but the manifest was needed after all.
+            # Show the deferred exception and exit.
+            if self.mle.args:
+                self.cmd.die('\n  '.join(str(arg) for arg in self.mle.args))
+            else:
+                self.cmd.die(str(self.mle))
 
     def run_extension(self, name, argv):
         # Check a program invariant. We should never get here
--- a/src/west/commands.py
+++ b/src/west/commands.py
@@ -25,7 +25,7 @@ import pykwalify
 import yaml
 
 from west.configuration import Configuration
-from west.manifest import Manifest, Project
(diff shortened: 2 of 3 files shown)


57300:
@mbolivar-nordic Hi, I can't add you as reviewer, but do you want to look over this?

mbolivar-nordic (changes requested):
I think this should probably be handled in the relevant built-in commands themselves by checking `WestCommand.has_manifest` and erroring out as necessary. Did you consider this? If so, why did you reject it?

57300:
> I think this should probably be handled in the relevant built-in commands themselves by checking `WestCommand.has_manifest` and erroring out as necessary. Did you consider this? If so, why did you reject it?

Yes, I had briefly considered it, but I figured it would probably involve passing `WestApp.mle` to each command, so that it could still report why `not self.has_manifest`. I decided against it, thinking it might be better to have common handling in one place. There's already some common handling in `handle_builtin_manifest_load_err()`, which seems to cover quite a few cases.

I'm fine with per-command handling, if you prefer, since it looks like only `west manifest` needs it right now anyway.

In fact, on second thought, the easiest fix for `west manifest` specifically might be to just remove it from this list:
https://github.com/zephyrproject-rtos/west/blob/9e8f5002f2ee86b7150c039cb69aa653ff454162/src/west/app/main.py#L277-L283

since it will always bail without a manifest:
https://github.com/zephyrproject-rtos/west/blob/9e8f5002f2ee86b7150c039cb69aa653ff454162/src/west/app/project.py#L547-L548

marc-hb:
`bsim` seems to have been made optional now by [Zephyr commit f30f08a3ce75](https://github.com/zephyrproject-rtos/zephyr/commit/f30f08a3ce75) and enabling babblesim in the `project-filter` now fails with `FATAL ERROR: "west manifest --resolve" is not (yet) supported when the manifest.project-filter option is set.` So I'm no sure it's possible to reproduce the original issue upstream.

However I tried with a totally different project and as of the current west version 0612a6ddaec3 the issue does not seem to have changed.

Considering Marti left the project, I have successfully tested a much simpler and zero-risk "fix" similar to 45762a01e30c

```diff
diff --git a/src/west/app/main.py b/src/west/app/main.py
index 2b4addb03e2d..b678893911f1 100755
--- a/src/west/app/main.py
+++ b/src/west/app/main.py
@@ -917,7 +917,7 @@ class WestArgumentParser(argparse.ArgumentParser):
                     # warn them again.
                     append('Cannot load extension commands; '
                            'help for them is not available.')
-                    append('(To debug, try: "west manifest --validate".)')
+                    append('(To debug, try: "west -vv manifest --validate".)')
                     append('')
             else:
                 # TODO we may want to be more aggressive about loading
diff --git a/src/west/commands.py b/src/west/commands.py
index 2e5a4b89b555..1abb0fcd17d2 100644
--- a/src/west/commands.py
+++ b/src/west/commands.py
@@ -260,7 +260,7 @@ class WestCommand(ABC):
         if self._manifest is None:
             self.die(f"can't run west {self.name};",
                      "it requires the manifest, which was not available.",
-                     'Try "west manifest --validate" to debug.')
+                     'Try "west -vv manifest --validate" to debug.')
         return self._manifest

     def _set_manifest(self, manifest: Optional[Manifest]):
```

marc-hb:
Minor `-vv` patch submitted in #734

pdgendt (approved):

marc-hb:
This change is located in a fairly complex area little or no test coverage: error handling.

I don't have the time to learn about it but I don't think the value is worth the risk. My 2 cents, sorry.

pdgendt:
@57300 any update on this?

57300:
I'm no longer invested in this PR, so I should probably close it. Thanks all for giving it a look though.

Issue #337, opened 2019-11-20
New "west grep" shortcut for west forall -c 'git grep <pattern>'


marc-hb:
`git` has a built-in `git grep` command that is convenient to perform a recursive search in sources. It's very convenient because
- it automatically excludes binary `.git/` database files and built artefacts,
- it works well "out of the box" requiring zero configuration or hard to remember option or parameter
- it supports `grep`'s standard and familiar user interface / syntax
- it's built-in requiring zero additional installation
- it's extremely short to type
- it's seems faster than the equivalent recursive grep . Maybe because it's multithreaded by default?

In a multi repo project, developers sometimes need to search across repos. Either because the project has (too) many repos and they need a reminder, or because some symbol is actually used across multiple repos, or because \<share other reasons below>.
`west forall -c 'git grep <pattern>'` is then the obvious solution, however it loses a lot of the convenience listed above and introduces additional and non-portable [quoting complexity](http://resources.mpi-inf.mpg.de/departments/rg1/teaching/unixffb-ss98/quoting-guide.html).

It's probably why Grepo already has the `repo grep` shortcut.

@mbolivar on Slack:

 **I would like people to start asking for features on GitHub, and voting with emoji.**

> I've recently learned that a feature I've been working on for a long time is not expected to be useful to the people I've been making it for, and there are a lot of directions we can go with west,
I want to start making some sort of priority queue based on user feedback
I want to be clear that I'm happy to discuss features in here [on Slack], but I'd like to start asking for more issues filed for new features before I at least start implementing. There are already quite a few enhancements open (edited)  And TBH I think it's rather easy to write a shell function to wrap your favorite grep-a-like, so without more people asking for it, I'm not inclined to make it a built-in command

In a different context:
>  just try to be conservative about adding [top-level built-in commands] unless they're really needed.

marc-hb:
Maybe `west grep` could be an _optional_ extension that users download and install from some random place on github ? For convenience and to avoid duplication of the simplest "user's choice". I don't know anything about west extensions so not sure I'm making sense.

mbolivar:
My position, from slack:

1. There is an easy workaround in place, namely defining a shell function that calls `west forall` with your favorite git-type program, so I don't think this is a blocker (we've spent way more time talking about this than it would have taken to write that shell  function)
2. I'm not happy with how our other existing commands that are basically `west forall` wrappers, namely `west diff` and `west status`, work, and I don't want to add more without solving some problematic pieces of their behavior
[...]

PR #678, opened 2023-08-25
commands: add grep

mbolivar-ampere:
This has been requested various times, but its lack never bothered me up until now, etc. Anyway, here it is.

Fixes: https://github.com/zephyrproject-rtos/west/issues/337

```
EXAMPLES
--------

To get "git grep foo" results from all cloned, active projects:

  west grep foo

To do the same with:

- git grep --untracked: west grep --untracked foo
- ripgrep:              west grep --tool ripgrep foo
- grep --recursive:     west grep --tool grep foo

To switch the default tool to:

- ripgrep:  west config grep.tool ripgrep
- grep:     west config grep.tool grep

GREP TOOLS
----------

This command runs a "grep tool" command in the top directory of each project.
Supported tools:

- git-grep (default)
- ripgrep
- grep

Set the "grep.tool" configuration option to change the default.

Use "--tool" to switch the tool from its default.

TOOL PATH
---------

Use --tool-path to override the path to the tool. For example:

  west grep --tool ripgrep --tool-path /my/special/ripgrep

Without --tool-path, the "grep.<TOOL>-path" configuration option
is checked next. For example, to set the default path to ripgrep:

  west config grep.ripgrep-path /my/special/ripgrep

If the option is not set, "west grep" searches for the tool as follows:

- git-grep: search for "git" (and run as "git grep")
- ripgrep: search for "rg", then "ripgrep"
- grep: search for "grep"

TOOL ARGUMENTS
--------------

The "grep.<TOOL>-args" configuration options, if set, contain arguments
that are always passed to the tool. The defaults for these are:

- git-grep: (none)
- ripgrep: (none)
- grep: "--recursive"

Command line options or arguments not recognized by "west grep" are
passed to the tool after that. This applies to --foo here, for example:

  west grep --foo --project=myproject

To force arguments to be given to the tool instead of west, put them
after a "--", like the --project and --tool-path options here:

  west grep -- --project=myproject --tool-path=mypath

Arguments before '--' that aren't recognized by west grep are still
passed to the grep tool.

To pass '--' to the grep tool, pass one for 'west grep' first.
For example, to search for '--foo' with grep, you can run:

  west grep --tool grep -- -- --foo

The first '--' separates west grep args from tool args. The second '--'
separates options from positional arguments in the 'grep' command line.

COLORS
------

By default, west will force the tool to print colored output as long
as the "color.ui" configuration option is true. If color.ui is false,
west forces the tool not to print colored output.

Since all supported tools have similar --color options, you can
override this behavior on the command line, for example with:

  west grep --color=never

To do this permanently, set "grep.color":

  west config grep.color never
```

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -1602,6 +1602,247 @@ class ForAll(_ProjectCommand):
                 failed.append(project)
         self._handle_failed(args, failed)
 
+GREP_EPILOG = '''
+EXAMPLES
+--------
+
+To get "git grep foo" results from all cloned, active projects:
+
+  west grep foo
+
+To do the same with:
+
+- git grep --untracked: west grep --untracked foo
+- ripgrep:              west grep --tool ripgrep foo
+- grep --recursive:     west grep --tool grep foo
+
+To switch the default tool to:
+
+- ripgrep:  west config grep.tool ripgrep
+- grep:     west config grep.tool grep
+
+GREP TOOLS
+----------
+
+This command runs a "grep tool" command in the top directory of each project.
+Supported tools:
+
+- git-grep (default)
+- ripgrep
+- grep
+
+Set the "grep.tool" configuration option to change the default.
+
+Use "--tool" to switch the tool from its default.
+
+TOOL PATH
+---------
+
+Use --tool-path to override the path to the tool. For example:
+
+  west grep --tool ripgrep --tool-path /my/special/ripgrep
+
+Without --tool-path, the "grep.<TOOL>-path" configuration option
+is checked next. For example, to set the default path to ripgrep:
+
+  west config grep.ripgrep-path /my/special/ripgrep
+
+If the option is not set, "west grep" searches for the tool as follows:
+
+- git-grep: search for "git" (and run as "git grep")
+- ripgrep: search for "rg", then "ripgrep"
+- grep: search for "grep"
+
+TOOL ARGUMENTS
+--------------
+
+The "grep.<TOOL>-args" configuration options, if set, contain arguments
+that are always passed to the tool. The defaults for these are:
+
+- git-grep: (none)
+- ripgrep: (none)
+- grep: "--recursive"
+
+Command line options or arguments not recognized by "west grep" are
+passed to the tool after that. This applies to --foo here, for example:
+
+  west grep --foo --project=myproject
+
+To force arguments to be given to the tool instead of west, put them
+after a "--", like the --project and --tool-path options here:
+
+  west grep -- --project=myproject --tool-path=mypath
+
+Arguments before '--' that aren't recognized by west grep are still
+passed to the grep tool.
+
+To pass '--' to the grep tool, pass one for 'west grep' first.
+For example, to search for '--foo' with grep, you can run:
+
+  west grep --tool grep -- -- --foo
+
+The first '--' separates west grep args from tool args. The second '--'
+separates options from positional arguments in the 'grep' command line.
+
+COLORS
+------
+
+By default, west will force the tool to print colored output as long
+as the "color.ui" configuration option is true. If color.ui is false,
+west forces the tool not to print colored output.
+
+Since all supported tools have similar --color options, you can
+override this behavior on the command line, for example with:
+
+  west grep --color=never
+
+To do this permanently, set "grep.color":
+
+  west config grep.color never
+'''
+# color.ui limitation:
+# https://github.com/zephyrproject-rtos/west/issues/651
+
+class Grep(_ProjectCommand):
+
+    # When adding a tool, make sure to check:
+    #
+    # - the 'failed' handling below for proper exit codes
+    # - color handling works as expected
+    TOOLS = ['git-grep', 'ripgrep', 'grep']
+
+    DEFAULT_TOOL = 'git-grep'
+
+    DEFAULT_TOOL_ARGS = {
+        'git-grep': [],
+        'ripgrep': [],
+        'grep': ['--recursive'],
+    }
+
+    DEFAULT_TOOL_EXECUTABLES = {
+        # git-grep is intentionally omitted: use self._git
+        'ripgrep': ['rg', 'ripgrep'],
+        'grep': ['grep'],
+    }
+
+    def __init__(self):
+        super().__init__(
+            'grep',
+            'run grep or a grep-like tool in one or more local projects',
+            'Run grep or a grep-like tool in one or more local projects.',
+            accepts_unknown_args=True)
+
+    def do_add_parser(self, parser_adder):
+        parser = self._parser(parser_adder, epilog=GREP_EPILOG)
+        parser.add_argument('--tool', choices=self.TOOLS, help='grep tool')
+        parser.add_argument('--tool-path', help='path to grep tool executable')
+        # Unlike other project commands, we don't take the project as
+        # a positional argument. This inconsistency makes the usual
+        # use case of "search the entire workspace" faster to type.
+        parser.add_argument('-p', '--project', metavar='PROJECT',
+                            dest='projects', default=[], action='append',
+                            help='''project to run grep tool in (may be given
+                            more than once); default is all cloned active
+                            projects''')
+        return parser
+
+    def do_run(self, args, tool_cmdline_args):
+        tool = self.tool(args)
+        command_list = ([self.tool_path(tool, args)] +
+                        self.tool_args(tool, tool_cmdline_args))
+        failed = []
+        for project in self._cloned_projects(args,
+                                             only_active=not args.projects):
+            completed_process = self.run_subprocess(
+                command_list,
+                capture_output=True,
+                text=True,
+                cwd=project.abspath)
+
+            # By default, supported tools generally exit 1 if
+            # nothing is found and 0 if something was found, with
+            # other return codes indicating an error.
+            #
+            # For git grep, this is experimentally true, even if
+            # the man page doesn't say so.
+            #
+            # Per grep's and ripgrep's man pages, this is true
+            # except that if an error occurred and --quiet was
+            # given and a match was found, then the exit status is
+            # still 0.
+            #
+            # Using --quiet can thus be used to print repositories
+            # with a match, while ignoring errors.
+            if completed_process.returncode == 1:
+                continue
+            self.banner(f'{project.name_and_path}:')
+            if completed_process.returncode != 0:
+                self.err(f'{util.quote_sh_list(command_list)}:\n'
+                         f'{completed_process.stderr}', end='')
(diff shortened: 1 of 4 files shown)


src/west/app/project.py
```diff
@@ -1602,6 +1602,204 @@ def do_run(self, args, user_args):
...
+that are always passed to the tool. The defaults for these are:
+
+- git-grep: (none)
+- ripgrep: (none)
+- grep: "--recursive"
+
+Additional command line arguments not recognized by "west grep" are
+passed to the tool after that. For example:
+
+  west grep --this-option-is-passed-to-the-tool
```
marc-hb, line 1669: So what about `west grep --this-option-is-passed-to-the-tool --project=A`? Who gets `--project`? In other words, is everything after the first unknown argument passed to the tool?

I could probably find the answer myself if I were less lazy but I'm not just asking for myself: I think the doc should clarify this.

src/west/app/project.py
```diff
@@ -1602,6 +1602,204 @@ def do_run(self, args, user_args):
...
+
+To switch the default tool:
+
+- to ripgrep:  west config grep.tool ripgrep
+- to grep:     west config grep.tool grep
+
+GREP TOOLS
+----------
+
+This command runs a "grep tool" to search inside projects.
```
marc-hb, line 1627:
```suggestion
This command runs a "grep tool" command in the top directory of each project.
```

src/west/app/project.py
```diff
@@ -1602,6 +1602,204 @@ def do_run(self, args, user_args):
...
+                                             only_active=not args.projects):
+            try:
+                output = self.check_output(command_list, cwd=project.abspath)
+            except subprocess.CalledProcessError as e:
+                # By default, supported tools exit 1 if nothing is
+                # found. Other return codes indicate error.
+                #
+                # This changes if you give the --quiet option, but
+                # that doesn't make much sense in this context, so
+                # assume it hasn't been passed.
```
marc-hb, line 1747: How about an assert or warning if `--quiet` has been passed? Assuming it takes only 1 or 2 lines of code.

src/west/app/project.py
```diff
@@ -1602,6 +1602,204 @@ def do_run(self, args, user_args):
...
+                            more than once); default is all cloned active
+                            projects''')
+        return parser
+
+    def do_run(self, args, tool_cmdline_args):
+        tool = self.tool(args)
+        command_list = ([self.tool_path(tool, args)] +
+                        self.tool_args(tool, tool_cmdline_args))
+        failed = []
+        for project in self._cloned_projects(args,
```
marc-hb, line 1737: This is literally begging for parallelisation. `concurrent.futures` should make that trivial, example: https://github.com/thesofproject/sof/commit/9b2453e2366da640760c4d0a0b0d2094c9622978

Hopefully colorama+CMD.EXE will not get in the way again? (cf. #533 and friends)

src/west/app/project.py
```diff
@@ -1602,6 +1602,204 @@ def do_run(self, args, user_args):
...
+                                             only_active=not args.projects):
+            try:
+                output = self.check_output(command_list, cwd=project.abspath)
+            except subprocess.CalledProcessError as e:
+                # By default, supported tools exit 1 if nothing is
+                # found. Other return codes indicate error.
```
marc-hb, line 1743: There should be short reminder of this above where `TOOLS` is defined; so this is not missed whenever someone adds a new tool.

marc-hb (commented):
Who is this new mbolivar-ampere guy? :-)

Could you bother adding just ONE, super basic, simple, minimal, incomplete and mostly hardcoded test? It would:
- already provide infinitely more test coverage than no test at all
- get the ball rolling and makes it vastly easier to expand later, especially for people not familiar with west's test suite (= everyone but you).


The perfect is the enemy of the good.

src/west/app/project.py
mbolivar-ampere: I think the doc is already clear so maybe you can suggest an edit?

The answer is that you have to use `--` to tell the command line parser "everything after this point is not an option, even if it looks like an option. This is per POSIX.

There is an example of how to pass `--project` to the underlying tool later in the help text.

src/west/app/project.py
mbolivar-ampere: It's fast enough for me so can I assume that we can leave this to future work, or, better yet, for you to implement?

Better is the enemy of done, they say.

mbolivar-ampere: Let's also keep in mind that:

- ripgrep is **already** parallelized, so running ripgrep in parallel is likely to over-commit your CPUs
- git grep has a `--threads` option and there is already a configuration option that people can use to use it by default

mbolivar-ampere:
@marc-hb for sure I'll be adding tests before merging. I'll update the --quiet and internal-reminder-when-adding-more-tools as well. Thanks for review -- other comments responded to.

src/west/app/project.py
marc-hb: > It's fast enough for me so can I assume that we can leave this to future work, or, better yet, for you to implement?

Yes, non-blocking comment of course. In fact I didn't use the "Request changes" option at all :-)

src/west/app/project.py
marc-hb: > I did my best to make the for-the-impatient help text clear,

As long as the header of the next section fits in the same screen, the impatient can easily skip to the next section :-) I'm all for conciseness but some completeness level matters to.

> West doesn't recognize --foo, but west grep is documented to pass unrecognized options along to the grep tool. It recognizes --project and therefore "consumes" it, not passing it along.

Thanks for the education, I had no idea that POSIX allowed interleaving consumed and passed-through arguments like this. In fact I didn't even know POSIX cared about this problem in general, I assumed each tool was free to do it the way it wants.

While this help should stay short and neither duplicate POSIX nor encourage this sort of interleaving; it should very briefly state that it is was happens. Not to encourage arguments interleaving (what a terrible idea), but to briefly state what happened when things go wrong. When things go wrong is actually the only moment where users read documentation, isn't it? Trial and error FTW!

I think part of what confused me a bit is the word "Additional" which I misinterpreted as "trailing". So just drop it? also, simply change the example so the unknown option is not at the end anymore:

```suggestion
  west grep --unknown-option-passed-to-the-grep-tool --project=A ...
```

And that's it; the documentation is not even longer!

src/west/app/project.py
```diff
@@ -1602,6 +1602,204 @@ def do_run(self, args, user_args):
...
+
+Additional command line arguments not recognized by "west grep" are
+passed to the tool after that. For example:
+
+  west grep --this-option-is-passed-to-the-tool
+
+To force an argument to be given to the tool instead of west, put
+it after a "--", like this:
+
+  west grep -- --project=val
```
marc-hb, line 1674: In theory it should be possible to use `--` twice, correct? The first one for `west`, the second `--` for `grep`. I think having two `--` in one test case would be great for test coverage.

src/west/app/project.py
mbolivar-ampere: Sorry -- it's not that POSIX is concerning itself with 'consumed' or 'passed-through'. POSIX concerns itself with recommendations for command line programs that delineate between positional arguments and options. A summary is available here: https://www.gnu.org/software/libc/manual/html_node/Argument-Syntax.html (longopts are kind of a separate GNU extension IIRC, but hopefully in a way that makes sense).

> I think part of what confused me a bit is the word "Additional" which I misinterpreted as "trailing". So just drop it? also, simply change the example so the unknown option is not at the end anymore:

OK, will do, thanks!

src/west/app/project.py
mbolivar-ampere: > How about an assert or warning if `--quiet` has been passed?

I gave this some more thought and revised the approach around --quiet a bit.

src/west/app/project.py
mbolivar-ampere: > There should be short reminder of this above where `TOOLS` is defined; so this is not missed whenever someone adds a new tool.

Added a checklist of new-tool-tasks.

tests/test_project.py
```diff
@@ -389,6 +389,27 @@ def test_forall(west_init_tmpdir):
...
+    # things we do.
+
+    actual_no_update = cmd('grep net-').strip()
+    actual_no_update_lines = actual_no_update.splitlines()
+    assert len(actual_no_update_lines) == 2
+    assert re.fullmatch(r'=== manifest \(zephyr\):', actual_no_update_lines[0])
+    assert re.search('net-tools', actual_no_update_lines[1])
+
+    assert not re.search('hello', cmd('grep hello'))
+    cmd('update')
```
marc-hb, line 404: Skip one blank line before `cmd('update')`

marc-hb (commented):
A few minor suggestions and questions left after a pretty thorough review, everything else LGTM

src/west/app/project.py
```diff
@@ -1602,6 +1602,234 @@ def do_run(self, args, user_args):
...
+                color = 'always'
+            else:
+                color = 'never'
+
+        ret.extend([f'--color={color}'])
+        ret.extend(self.config.get(f'grep.{tool}-args',
+                                   self.DEFAULT_TOOL_ARGS[tool]))
+
+        # The first '--' we see is "meant for" west grep. Take that
+        # out of the options we pass to the tool.
```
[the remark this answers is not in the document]
mbolivar-ampere: Yes, argparse respects the POSIX rules and stops parsing anything after `--` as an option, but it also includes the `--` in the arguments we see here.

src/west/app/project.py
```diff
@@ -1602,6 +1602,234 @@ def do_run(self, args, user_args):
...
+Command line options or arguments not recognized by "west grep" are
+passed to the tool after that. This applies to --foo here, for example:
+
+  west grep --foo --project=myproject
+
+To force arguments to be given to the tool instead of west, put them
+after a "--", like the --project and --tool-path options here:
+
+  west grep -- --project=myproject --tool-path=mypath
+
```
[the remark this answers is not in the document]
mbolivar-ampere:
```
+Arguments before '--' that aren't recognized by west grep are still
+passed to the grep tool.
+
+To pass '--' to the grep tool, pass one for 'west grep' first.
+For example, to search for '--foo' with grep, you can run:
+
+  west grep --tool grep -- -- --foo
+
+The first '--' separates west grep args from tool args. The second '--'
+separates options from positional arguments in the 'grep' command line.
```

mbolivar-ampere:
@marc-hb comments addressed and one additional bug in `grep.<TOOL>-args` handling found and fixed. Thanks for a great review!

marc-hb:
and... it hangs forever :-)

Hotfix submitted in #681

Looks like personal configs are still "polluting" the test suite...

PR #680, opened 2023-08-31
manifest: Add a new `description` project attribute

carlescufi:
This PR adds a new `description` project attribute that contains an optional text description of the project.


carlescufi:
@mbolivar-ampere before I start modifying/adding tests I'd like to discuss whether you would like the `Project()` constructor signature to look like. I currently have:
```
   def __init__(self, name: str, description: Optional[str], url: str,
                 revision: Optional[str] = None,
                 path: Optional[PathType] = None,
                 submodules: SubmodulesType = False,
                 clone_depth: Optional[int] = None,
                 west_commands: Optional[WestCommandsType] = None,
                 topdir: Optional[PathType] = None,
                 remote_name: Optional[str] = None,
                 groups: Optional[GroupsType] = None,
                 userdata: Optional[Any] = None):
```
but this will break existing apps. Should I instead place `description` as the last argument?

src/west/manifest-schema.yml
```diff
@@ -82,6 +82,10 @@ mapping:
           name:
             required: true
             type: str
+          # Project description.
```
marc-hb, line 85:
```suggestion
          # Project description. Has no effect.
```

marc-hb (commented):
Interesting! I'm aware the west documentation is unfortunately still in a different repo, so could elaborate in the commit message or source comments about where this new field will be displayed and how? Truncated in the default `west list` maybe?

carlescufi:
> Interesting! I'm aware the west documentation is unfortunately still in a different repo, so could elaborate in the commit message or source comments about where this new field will be displayed and how? Truncated in the default `west list` maybe?

Right now I've only added to the `repr()` but ultimately this is really to be displayed in the manifest itself. Today it is hard to identify projects in large manifests with functionality, and by adding this field I am hoping that this will be easier.

src/west/manifest.py
```diff
@@ -849,12 +853,17 @@ def posixpath(self) -> Optional[str]:
     def name_and_path(self) -> str:
         return f'{self.name} ({self.path})'
 
+    @property
+    def description(self) -> Optional[str]:
+        return self.description
```
mbolivar-ampere, line 858: Doesn't this cause infinite recursion? I think it is not necessary to have the property.

src/west/manifest.py
```diff
@@ -776,7 +778,7 @@ def __str__(self):
         path_repr = repr(self.abspath or self.path)
         return f'<Project {self.name} ({path_repr}) at {self.revision}>'
 
-    def __init__(self, name: str, url: str,
+    def __init__(self, name: str, description: Optional[str], url: str,
```
mbolivar-ampere, line 781: The commit message description is empty and I'm not sure why this is being added as an arg instead of as a kwarg. If we need to add it as an arg like this for a good reason, we can deal with the API change (it's not a big deal as I don't expect "anyone" is making their own Project objects). But if we can have this as `description: Optional[str] = None`, thus keeping compatibility, why not do that?

carlescufi: > If we need to add it as an arg like this for a good reason, we can deal with the API change

There isn't a good reason, this was just a first quick pass to gather feedback like the one I got from you now. I will change to become a kwarg. Thanks!

src/west/manifest.py
carlescufi: Yep, you are right. Leftover from a first pass where there was a `_description` field.

src/west/manifest.py
```diff
@@ -855,6 +859,8 @@ def as_dict(self) -> Dict:
         '''
         ret: Dict = {}
         ret['name'] = self.name
+        if self.description:
+            ret['description'] = self.description
```
mbolivar-ampere, line 863: If it's not too much work, could you add a description field to one of the projects in MANIFEST_TEMPLATE in conftest.py, and update test_manifest_freeze() accordingly? Thanks!

carlescufi: Of course! I have done this and, as part of it, added better handling of multi-line strings when dumping YAML. I have also described this in the commit message.

mbolivar-ampere (changes requested):
LGTM. Two minor requests then I am +1

carlescufi:
Thanks for the review @mbolivar-ampere! Comments addressed.

mbolivar-ampere (approved):

PR #681, opened 2023-08-31
grep: fix command hanging forever when grep.{tool}-args is empty

marc-hb:
Fixes commit 1d220bbd627b ("commands: add grep") / #678 

Fix misplaced closing parenthese in:

```python
shlex.split(self.config.get(f'grep.{tool}-args'), '')
```

The empty string is meant as a default value for config.get() but it was passed as a second argument to shlex.split() by mistake.

Funny enough this hangs forever:

```python
>>> shlex.split(None, comments='')
<stdin>:1: DeprecationWarning: Passing None for 's' to shlex.split() is deprecated.
 (hangs forever)
```

PS: debuggers rulez

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -1829,7 +1829,7 @@ class Grep(_ProjectCommand):
                 color = 'never'
 
         ret.extend([f'--color={color}'])
-        ret.extend(shlex.split(self.config.get(f'grep.{tool}-args'), '') or
+        ret.extend(shlex.split(self.config.get(f'grep.{tool}-args', '')) or
                    self.DEFAULT_TOOL_ARGS[tool])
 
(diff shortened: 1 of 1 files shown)


mbolivar-ampere (approved):

marc-hb:
Because I have no `grep.{tool}-args` configuration, `west grep` was ALWAYS hanging for me! So I really wondered how the test was passing.

Mystery solved: that's because the (complex) tox+pytest test environment closes the standard input.

In fact a workaround for this bug is to hit `Ctrl-D` repeatedly; once for each git repo.

I can see that the test environment solves a lot of problems. But it also creates some distance and divergence with the real thing. I just tried to document this gap here:
- #682

Issue #679, opened 2023-08-30
west cannot process a git branch name containing a single quote '


GarrettCarter-eaton:
If a repo that west clones and processes contains a branch name using a single quote, the following error is received:
```
Traceback (most recent call last):
  File "/usr/local/bin/west", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.8/dist-packages/west/app/main.py", line 793, in main
    app.run(argv or sys.argv[1:])
  File "/usr/local/lib/python3.8/dist-packages/west/app/main.py", line 109, in run
    self.run_command(argv)
  File "/usr/local/lib/python3.8/dist-packages/west/app/main.py", line 342, in run_command
    cmd.run(args, unknown, self.topdir, manifest=self.manifest,
  File "/usr/local/lib/python3.8/dist-packages/west/commands.py", line 135, in run
    self.do_run(args, unknown)
  File "/usr/local/lib/python3.8/dist-packages/west/app/project.py", line 806, in do_run
    self.update_all()
  File "/usr/local/lib/python3.8/dist-packages/west/app/project.py", line 871, in update_all
    self.update(project)
  File "/usr/local/lib/python3.8/dist-packages/west/app/project.py", line 1039, in update
    self.clean_refs_west(project, stats, take_stats)
  File "/usr/local/lib/python3.8/dist-packages/west/app/project.py", line 1296, in clean_refs_west
    _clean_west_refspace(project)
  File "/usr/local/lib/python3.8/dist-packages/west/app/project.py", line 1471, in _clean_west_refspace
    project.git(delete_ref_cmd)
  File "/usr/local/lib/python3.8/dist-packages/west/manifest.py", line 807, in git
    cmd_list = shlex.split(cmd)
  File "/usr/lib/python3.8/shlex.py", line 311, in split
    return list(lex)
  File "/usr/lib/python3.8/shlex.py", line 300, in __next__
    token = self.get_token()
  File "/usr/lib/python3.8/shlex.py", line 109, in get_token
    raw = self.read_token()
  File "/usr/lib/python3.8/shlex.py", line 191, in read_token
    raise ValueError("No closing quotation")
ValueError: No closing quotation
```

Example branch name:
`LTK-18338-No-link-for-privacy-policy-on-WebUI's-Login-page`

marc-hb:
Thanks for the report!

[...]

PR #683, opened 2023-09-01
Project.git(list/str): reduce reliance on shlex.split()

marc-hb:
For convenience, Project.git() supports passing either a list (good) or a string with whitespaces (bad). The latter is parsed with shlex.split()

This saves some typing but the caller has to be extremely careful to never use the shlex.split() convenience with unsanitized inputs.

Fixes commit 3ac600acaa11 ("git: clean west ref space after fetching") where the caller was not careful and concatenated `update-ref -d ` with unsanitized input, possibly containing special characters as found in bug #679. Fix this bug by converting the string to a list.

While at it, look for a few other, frequent and risky invocations and convert their string argument to a list too. The following test hack was used to semi-automate the search for these other locations:

```diff
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -897,6 +897,8 @@ class Project:
         :param cwd: directory to run git in (default: ``self.abspath``)
         '''
         if isinstance(cmd, str):
+            print(cmd)
+            breakpoint()
             cmd_list = shlex.split(cmd)
         else:
             cmd_list = list(cmd)
```

While at it, also convert to a list a couple non-risky but very frequent invocations. This speeds up the test hack above.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -19,6 +19,7 @@ envlist=py3
 ignore = E126,E261,E302,E305,W504
 # Don't lint setup.py, the .tox virtualenv directory, or the build directory
 exclude = setup.py,.tox,build
+max-line-length = 100
 
 [mypy]
 mypy_path=src
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -685,7 +685,7 @@ class Compare(_ProjectCommand):
                 if self.ignore_branches:
                     return False
 
-                return bool(project.git('branch --show-current',
+                return bool(project.git(['branch', '--show-current'],
                                         capture_stdout=True,
                                         capture_stderr=True).stdout.strip())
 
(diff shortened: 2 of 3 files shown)


tox.ini
```diff
@@ -36,5 +37,5 @@ setenv =
     TOXTEMPDIR={envtmpdir}
 commands =
   python -m pytest --cov=west {posargs:tests}
-  python -m flake8 --config={toxinidir}/tox.ini {toxinidir}
+  python -m flake8 --config={toxinidir}/tox.ini {toxinidir} # --color=never
```
mbolivar-ampere, line 40: ?

marc-hb: This lets you click on the error messages in your IDE (emacs, sorry for outing you).

This is learned the hard way because this can't go in a config file for some unknown reason.

mbolivar-ampere: Add this to your .emacs

```
(ignore-errors
  (require 'ansi-color)
  (defun my-colorize-compilation-buffer ()
    (when (eq major-mode 'compilation-mode)
      (ansi-color-apply-on-region compilation-filter-start (point-max))))
  (add-hook 'compilation-filter-hook 'my-colorize-compilation-buffer))
```

marc-hb: Thanks, removed.

mbolivar-ampere (approved):

PR #687, opened 2023-09-27
Add docker compose based test environment for Linux

mbolivar-ampere:
When preparing a west release, we like to make sure that it works on more Linux environments than whatever is currently tested in GitHub CI. Doing this by hand is a pain. Add some automation for making this more painless.


marc-hb:
Nice! No time to review at this moment but if you're looking for some Windows inspiration you can find some here:

https://github.com/thesofproject/sof/blob/main/.github/workflows/zephyr.yml

https://github.com/thesofproject/sof/actions/workflows/daily-tests.yml

docker-testing/in-container-test.sh
```diff
@@ -0,0 +1,29 @@
+#!/bin/bash
+
```
marc-hb, line 2:
```suggestion
set -e
```

It does not catch everything but it catches a lot.

docker-testing/in-container-test.sh
```diff
@@ -0,0 +1,29 @@
...
+# This is the test script that runs in the containers themselves.
+
+WEST=/west
+
+die() {
+    [ ! -z "$@" ] && echo "error: $@" >&2
```
marc-hb, line 8: I would prefer an empty `error: `. No one should `die` without a message.

I also prefer a `die` that is fully `printf` compatible: 

```shell
die()
{
        >&2 printf '%s ERROR: ' "$0"
        # We want die() to be usable exactly like printf
        # shellcheck disable=SC2059
        >&2 printf "$@"
        exit 1
}
```

marc-hb (commented):
I dunno much about Docker unfortunately...

docker-testing/in-container-test.sh
marc-hb, line 1: Forgot this: this script is more than big enough to deserver a `main()` function.

Very small change, all the benefits listed in:
- https://github.com/thesofproject/sof-test/issues/740

marc-hb:
> I dunno much about Docker unfortunately...

One thing we learned in SOF though;  the Dockerfiles should be in a separate repo:
- https://github.com/thesofproject/sof/issues/5073

Except for "unit" test code which is very dependent on the version of the production code it's testing, you usually want the convenience to combine any version of test code with any version of production code. Typically: git bisect production code with the latest version of test code. Or here: git bisect with the latest docker image.

Now don't get me wrong: anyone can take 2 random git repos mostly unrelated to each other, merge them into one and start interleaving their unrelated development histories. It is totally possible and even not that uncommon (especially when you don't know about `west` ;-) But it's really not convenient when the two projects have totally unrelated life cycles and release cycles.

Another instance: every Dockerfile change should trigger some Github Action that rebuilds the container and runs the test suite again. But any Dockerfile CI takes a very long time so you don't want that to happen for mosts west PRs which do NOT change any Dockerfile. Congratulations, now you need a path-based filter on every Github Action to separate two different CIs that should have never been together in the first place. Possible but pretty inconvenient.

mbolivar-ampere:
> One thing we learned in SOF though; the Dockerfiles should be in a separate repo:

This is meant for manual testing on various distros at release time; can we keep things simple for now?

docker-testing/in-container-test.sh
mbolivar-ampere: I added an error path when there's no argument. I'd rather not use printf syntax, though.

docker-testing/in-container-test.sh
marc-hb: `set -e` is not incompatible with manual error handling at all. It would be totally unusable otherwise. `set -e` catches many failures that the developer did not anticipate or was too lazy or too optimistic to handle manually - now and in the future.

The only problem with `set -e` is the false, 100% safety impression that it can give to some inexperienced people. For everything else it's fine.

https://github.com/thesofproject/sof-test/issues/312

docker-testing/in-container-test.sh
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
+
+# This is the test script that runs in the containers themselves.
+
+WEST=/west
+
+die() {
+    if [ ! -z "$@" ]; then
```
marc-hb, line 8: "$@" expands to multiple strings. That's not what you want here
http://mywiki.wooledge.org/BashPitfalls#for_arg_in_.24.2A
http://mywiki.wooledge.org/BashGuide/Parameters#Special_Parameters_and_Variables


```suggestion
    if [ -n "$*" ]; then
```

Or even just:
```suggestion
    if [ -n "$1" ]; then
```

Or `if [ $# -eq 0 ]`

Then for consistency:

`echo "error: $*" >&2`

marc-hb: BTW you don't HAVE to `printf 'error: %s' "$var"`, you can still do `printf "error: $var"` if you want. The well-tested `die()` function I recommended earlier supports either indifferently.

docker-testing/in-container-test.sh
```diff
@@ -0,0 +1,38 @@
...
+    # Verify the container environment set up meets this script's requirements.
+    [ ! -z "$WEST_TOX_OUT" ] || die "missing $WEST_TOX_OUT"
+    [ ! -z "$WEST_TOX_OUT_IN_HOST" ] || die "missing $WEST_TOX_OUT_IN_HOST"
+    [ -d "$WEST" ] || die "missing $WEST in the container"
+
+    TOX_LOG="$WEST_TOX_OUT/tox.log"
+    TOX_LOG_IN_HOST="$WEST_TOX_OUT_IN_HOST/tox.log"
+    WEST_TESTDIR="/tmp/west"
+
+    mkdir "$WEST_TOX_OUT" || die "failed to make $WEST_TOX_OUT in container ($WEST_TOX_OUT_IN_HOST in host)"
```
marc-hb, line 27: Nit
```suggestion
    mkdir "$WEST_TOX_OUT" ||
       die "failed to make $WEST_TOX_OUT in container ($WEST_TOX_OUT_IN_HOST in host)"
```

(No need for a backslash in this case)

marc-hb: How likely is this `mkdir` to fail really? Isn't `set -e` enough?

docker-testing/in-container-test.sh
```diff
@@ -0,0 +1,38 @@
...
+    else
+        echo "error: unknown error in $0"
+    fi
+    exit 1
+}
+
+main()
+{
+    # Verify the container environment set up meets this script's requirements.
+    [ ! -z "$WEST_TOX_OUT" ] || die "missing $WEST_TOX_OUT"
```
marc-hb, line 19:
```suggestion
    [ -n "$WEST_TOX_OUT" ] || die "missing $WEST_TOX_OUT"
```

The fewer negations the better and copy/paste of `!` can trigger history recall in interactive testing on the command line.

marc-hb (commented):
Still no time to look at the docker files yet sorry.

marc-hb:
Resurrected by @pdgendt in newer #733

Issue #622, opened 2023-01-14
Improvements for managing untracked projects


joerchan:
Whenever the manifest file is modified by removing entries in the manifest or by moving the manifest entry path the project folder may end up containing folders that are no longer needed or duplicated entries.

An example of this is the TF-M repository split, where the folder name was changed from `tfm` to `tf-m`.
Users doing `west update` after this change will end up with duplicated entries in their folder structure when the new project is imported by the old one still remains.

Usually what the user wants to do at this point is to remove the not-used project as it now has duplicated entries and this causes problems when searching in the projects for symbols.

Since the folders usually are git repositories cleaning this up should not be handled automatically by the tool as the user my still have branches in the old folder that they wish to clean up first.

Providing the user with the information about these old folders with by useful so that they can do the appropriate action to get rid of them.

Listing the untracked folders in a similar way as git does when doing `west status` would be useful.
West should skip folders that are inside existing projects, as this would otherwise be listed by git.

Example:

```
west status
=== status of manifest (zephyr):
On branch main
Your branch is up to date with 'upstream/main'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	build/

nothing added to commit but untracked files present (use "git add" to track)
=== 

Untracked folders:
	modules/tee/tfm

Add files to west.yml to track
```

This is very useful when doing bisect to find a bug, or checking out previous releases to reproduce reported bugs.
This could add older projects to the checkout that should be cleaned up afterwards.

joerchan:
> I would keep it simple and report only the top level directories (e.g.: modules/ here). Maybe top level files too for consistency with git?

But `modules/` and `modules/tee` contain projects that are tracked, i.e. `modules/tee/tf-m/trusted-firmware-m` for example in this case.

marc-hb:
Sorry I wrote too fast. I meant: if there is a `modules/foo/bar/` directory, then it should only report `modules/foo/` but not `modules/foo/bar` to keep the output short (assuming there is no west project anywhere below `modules/foo/`).

Different git clients report the above situation with different verbosity.

On the other hand, if there _is_ a west project `modules/foo/project/`, then it should report `modules/foo/bar/`.


So the pseudo-code should probably look like this:

```python
def west_find_untracked(parent):
   for child in parent:
        if child is a west module:
             continue # nothing to report
        if no west module below child:
             report child
             continue
        else:
             west_find_untracked(child)

west_find_untracked(west_top)
```

[...]

PR #702, opened 2024-01-14
west: Add untracked files and folders to west status output

joerchan:
Add untracked files and folder to west status output. This helps to keep the west checkout clean by alerting the user of folders that are no longer tracked by west.
This is useful for detecting when a module has been removed, or moved. and is no longer updated by west update.

Add the ability to ignore folders and files through a comma separated list in the west config file.
Example:
west config status.ignore .vscode/,optional/

Fixes: #622

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -844,6 +844,53 @@ class Status(_ProjectCommand):
                 failed.append(project)
         self._handle_failed(args, failed)
 
+        # List projects not tracked by manifest.
+        projects = [project.abspath for project in self._cloned_projects(args, only_active=True)]
+        topdir = Path(self.topdir)
+        untracked = []
+
+        self.do_find_untracked(topdir, projects, untracked)
+
+        ignored_config = self.config.get('status.ignore')
+        ignored = ignored_config.split(',') if ignored_config else []
+        # Ignore .west folder in untracked projects
+        ignored.append('.west/')
+        for ignore in ignored:
+            try:
+                untracked.remove(ignore)
+            except:
+                pass
+
+        if len(untracked) > 0:
+            self.banner('untracked:')
+            for project in untracked:
+                self.inf(textwrap.indent(f'{project}', ' ' * 4))
+
+    def is_manifest_subfolder(self, path, projects):
+        for project_path in projects:
+            try:
+                if Path(project_path).relative_to(Path(path)):
+                    return True
+            except ValueError:
+                pass
+        return False
+
+    def do_find_untracked(self, path, projects, untracked):
+        obj = os.scandir(path)
+
+        for entry in obj:
+            if entry.is_dir():
+                if entry.path in projects:
+                    continue
+                if self.is_manifest_subfolder(entry.path, projects):
+                    self.do_find_untracked(entry, projects, untracked)
+                else:
+                    untracked.append(os.path.relpath(entry.path, Path(self.topdir)) + '/')
+            if entry.is_file():
+                untracked.append(os.path.relpath(entry.path, Path(self.topdir)))
+
+        obj.close()
+
 class Update(_ProjectCommand):
 
     def __init__(self):


src/west/app/project.py
```diff
@@ -844,6 +844,53 @@ def do_run(self, args, user_args):
...
+        for entry in obj:
+            if entry.is_dir():
+                if entry.path in projects:
+                    continue
+                if self.is_manifest_subfolder(entry.path, projects):
+                    self.do_find_untracked(entry, projects, untracked)
+                else:
+                    untracked.append(os.path.relpath(entry.path, Path(self.topdir)) + '/')
+            if entry.is_file():
+                untracked.append(os.path.relpath(entry.path, Path(self.topdir)))
```
marc-hb, line 890: You're using a mix or `Path.relative_to()` and `os.path.relpath()`. If there's a reason why then add a short comment explaining why, otherwise stick to `Path`.

Generally speaking this what this code does is not always very obvious and there's zero zero comment right now.

src/west/app/project.py
```diff
@@ -844,6 +844,53 @@ def do_run(self, args, user_args):
...
+        for project_path in projects:
+            try:
+                if Path(project_path).relative_to(Path(path)):
+                    return True
+            except ValueError:
+                pass
+        return False
+
+    def do_find_untracked(self, path, projects, untracked):
+        obj = os.scandir(path)
```
marc-hb, line 879:
```suggestion
        entries = os.scandir(path)
```
This is a list, use plural. Also, stick to the standard Python terminology when possible.

src/west/app/project.py
```diff
@@ -844,6 +844,53 @@ def do_run(self, args, user_args):
...
+
+        for entry in obj:
+            if entry.is_dir():
+                if entry.path in projects:
+                    continue
+                if self.is_manifest_subfolder(entry.path, projects):
+                    self.do_find_untracked(entry, projects, untracked)
```
marc-hb, line 886: I think `os.walk()` does this recursion for you:
https://stackoverflow.com/questions/19859840/excluding-directories-in-os-walk

If not, add a comment explaining why you can't use os.walk() and have to perform the recursion yourself.

src/west/app/project.py
```diff
@@ -844,6 +844,53 @@ def do_run(self, args, user_args):
...
+                untracked.remove(ignore)
+            except:
+                pass
+
+        if len(untracked) > 0:
+            self.banner('untracked:')
+            for project in untracked:
+                self.inf(textwrap.indent(f'{project}', ' ' * 4))
+
+    def is_manifest_subfolder(self, path, projects):
```
marc-hb, line 869: Either I don't understand this function, or the name is very confusing. Do you mean something like `is_in_some_west_project()`?

The manifest is a list of git projects, it's not the projects themselves.

Anyway you should not need such a path comparison because the directory walk should prune `west` projects as soon as it finds one?

src/west/app/project.py
```diff
@@ -844,6 +844,53 @@ def do_run(self, args, user_args):
...
+        projects = [project.abspath for project in self._cloned_projects(args, only_active=True)]
+        topdir = Path(self.topdir)
+        untracked = []
+
+        self.do_find_untracked(topdir, projects, untracked)
+
+        ignored_config = self.config.get('status.ignore')
+        ignored = ignored_config.split(',') if ignored_config else []
+        # Ignore .west folder in untracked projects
+        ignored.append('.west/')
```
marc-hb, line 857: Why would anyone have a `.west/` folder in untracked projects? This looks like a problem most people would want to see (while others can explicitly ignore it)

src/west/app/project.py
```diff
@@ -844,6 +844,53 @@ def do_run(self, args, user_args):
...
+        for ignore in ignored:
+            try:
+                untracked.remove(ignore)
+            except:
+                pass
+
+        if len(untracked) > 0:
+            self.banner('untracked:')
```
marc-hb, line 865:
```suggestion
            self.banner('Unknown:')
```
Please do not re-use the same presentation as git to avoid misleading users into thinking this line comes from git.

src/west/app/project.py
```diff
@@ -844,6 +844,53 @@ def do_run(self, args, user_args):
...
+        self.do_find_untracked(topdir, projects, untracked)
+
+        ignored_config = self.config.get('status.ignore')
+        ignored = ignored_config.split(',') if ignored_config else []
+        # Ignore .west folder in untracked projects
+        ignored.append('.west/')
+        for ignore in ignored:
+            try:
+                untracked.remove(ignore)
+            except:
```
marc-hb, line 861: Does something this mundane really need a `try/except`? Dubious.

src/west/app/project.py
```diff
@@ -844,6 +844,53 @@ def do_run(self, args, user_args):
...
 
+        # List projects not tracked by manifest.
+        projects = [project.abspath for project in self._cloned_projects(args, only_active=True)]
+        topdir = Path(self.topdir)
+        untracked = []
+
+        self.do_find_untracked(topdir, projects, untracked)
+
+        ignored_config = self.config.get('status.ignore')
+        ignored = ignored_config.split(',') if ignored_config else []
```
marc-hb, line 855: We need a separate `.westignore` file for this for consistency with git and others. This will let users put each pattern on a separate line and let them ignore files with commas in their names.

marc-hb (changes requested):
Very quick review while this is "hot", apologies for any misunderstanding. 

Parts of the code are hard to read and there's zero comment. Please name symbols more carefully and  comment the least intuitive parts of the code.

New tests missing for this (non-trivial) feature.

As already discussed in #622. nested  `west` projects are probably the trickiest use cases for this. Unsurprisingly, `west` seems to have a "don't ask, don't tell" attitude with respect to nested projects.

At the very least you need to give detailed examples of how nested projects handled and some tests for these use cases.

BTW at least one test is failing right now with the current commit:
https://github.com/zephyrproject-rtos/west/actions/runs/7520440151/job/20470116532?pr=702. I haven't looked why.

Various other code issues, see inline.

Thanks for the PR but I'm a bit concerned about the typical "How hard could this be? Famous last words" effect, especially wrt. to nested projects. Could these be just left for git to show?

marc-hb:
Please re-open if still interested.

PR #703, opened 2024-01-18
project.py: request users to stop IDEs before running `west init`

marc-hb:
There have been many users reports in #558 that `west init` fails on the Windows NTFS filesystem when some IDEs are running concurrently.

There is no simple and reliable fix for this because:

- The lack of concurrency is a core limitation in the NTFS design that will not change anytime soon,
- IDEs have no reliable way to know when they should pause scanning the filesystem.
- Placing temporary files outside the workspace would invite other, complex and nasty issues with cross-filesystem moves (#558)

Short of a fix, explain why this happens very briefly in `west init -h` and request users to temporarily close their IDEs. It's the only simple and universal mitigation available.

Mention the issue number in case someone needs more details, some evidence or smarter workarounds specific to certain IDEs.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -153,6 +153,16 @@ manifest repository. If --mr is not given, the remote's default
 branch will be used, if it exists.
 
 With neither, -m {MANIFEST_URL_DEFAULT} is assumed.
+
+Warning: 'west init' renames and/or deletes temporary files inside the
+workspace being created. This fails on some filesystems when some
+development tool or any other program is trying to read/index these
+temporary files at the same time. For instance, it is required to stop
+Visual Studio Code before running 'west init` on the Windows NTFS
+filesystem. Find other, similar "Access is denied" examples in west
+issue #558.
+This is not required on inode-based, Linux filesystems that wait and
+finalize the deletion until there is no concurrent user left.
 ''',
             requires_workspace=False)
 


thewh1teagle:
The whole installation process is very inconvenient.
I don't know why you moved from MSYS2 which makes things much more simple.

marc-hb:
I don't remember `west` ever using MSYS2. You're probably confusing west with Zephyr.

BTW this particular NTFS design issue is not specific to west, it has been reproduced without west in #558.

marc-hb:
This is really off-topic but I heard that https://docs.platformio.org/en/latest/frameworks/zephyr.html is much more user-friendly than plain Zephyr. I never tried it myself.

src/west/app/project.py
```diff
@@ -153,6 +153,16 @@ def __init__(self):
...
 
 With neither, -m {MANIFEST_URL_DEFAULT} is assumed.
+
+Warning: 'west init' renames and/or deletes temporary files inside the
+workspace being created. This fails on some filesystems when some
+development tool or any other program is trying to read/index these
+temporary files at the same time. For instance, it is required to stop
+Visual Studio Code before running 'west init` on the Windows NTFS
+filesystem. Find other, similar "Access is denied" examples in west
+issue #558.
```
trond-snekvik, line 163:
```suggestion
filesystem. Find other, similar "Access is denied" examples in 
https://github.com/zephyrproject-rtos/west/issues/558.
```

Most terminals make links clickable these days, so this is worth spelling out, IMO

trond-snekvik (approved):
Bit of a drive-by review, but LGTM.  

For VS Code, I believe temporarily disabling the `git.autoRefresh` setting would be enough, actually, but it has a pretty extensive file watching system, so you never know.

src/west/app/project.py
marc-hb: I considered this but I tried to minimize the length as much as possible. In fact I was a bit reluctant to add this _at all_ because a help text should be really concise and not turn into a changelog or bug list. However this particular bug seems frequent, fatal and totally unfixable.

mbolivar-ampere (approved):

PR #715, opened 2024-06-29
project: Switch to pyproject

Ayush1325:
Starting with PEP 621, `pyproject.toml` is the standard way of specifying project metadata.

Also switch to using the version from `pyproject.toml` instead of having a west.version python module. Adjust documentation for the same.


pdgendt:
Could you merge `tox.ini` in `pyproject.toml` too?

Ayush1325:
> Could you merge `tox.ini` in `pyproject.toml` too?

Are you sure we want to add it as shown here: https://tox.wiki/en/latest/config.html#pyproject-toml ?
It does not seem like tox supportes `pyproject.toml` well.

pdgendt:
> It does not seem like tox supportes `pyproject.toml` well.

Didn't know this, I agree.

pdgendt (approved):

carlescufi (approved):

marc-hb (commented):
I wish I could review this but I really don't have the time to take a python packaging course right now. I hope the approvers know a bit more than me?

My lack of knowledge aside (!), I see nothing wrong here.

The relevant PEP is dated 2020 so I had a concern about Python 3.8 compatibility but CI is apparently testing that - great!

pdgendt (changes requested):
`twister` issue needs to be fixed.

pdgendt (approved):

Ayush1325:
I am testing the package from start to finish with upstream Zephyr once. Will update once that is done.

marc-hb (changes requested):
On line 91, MAINTAINERS.rst asks to run `python3 setup.py sdist bdist_wheel` to prepare a release package. This does not work anymore.

Ayush1325:
> I am testing the package from start to finish with upstream Zephyr once. Will update once that is done.
> 
> Update: Yup, everything including twister seems to be working.

I have update the instructions to build as described here: https://packaging.python.org/en/latest/tutorials/packaging-projects/#generating-distribution-archives

MAINTAINERS.rst
```diff
@@ -88,7 +88,8 @@ Building and uploading the release wheels
 You need the zephyr-project PyPI credentials for the 'twine upload' command. ::
 
   git clean -ffdx
-  python3 setup.py sdist bdist_wheel
+  python3 -m pip install --upgrade build twine
```
marc-hb, line 91: - Why `python3 -m pip install...` instead of the usual, shorter and clearer `pip3 install ...`?

- Rather than blindly upgrading "the latest", how about adding `build>=1.something` to `[dependencies]` in `pyproject.toml`?
`twine` is different because it is not needed for local installations.

So I think this script would be more consistent with the English text:

```
git clean -fdx
python3 -m build
pip3 install --upgrade twine
twine upload ...
```

marc-hb (commented):
> I have update the instructions to build as described here: https://packaging.python.org/en/latest/tutorials/packaging-projects/#generating-distribution-archives

This looks like a great page! Can you please add it somewhere in MAINTAINERS.rst so no one wastes time on some inferior stackoverflow or AI-generated duplicate?

Did you re-test MAINTAINERS.rst all the way now, including the twine upload to the test PyPI server?

MAINTAINERS.rst
Ayush1325: `pyproject-build` is also not needed for local installations. I can do `pip install .` without having `pyproject-build` installed.

pdgendt:
@marc-hb any open issues with this PR?

MAINTAINERS.rst
marc-hb: Indeed, thanks for correcting me. It's not "built-in" either: it just happened to be already installed on my system for some reason.

marc-hb (approved):

MAINTAINERS.rst
```diff
@@ -88,12 +88,15 @@ Building and uploading the release wheels
 You need the zephyr-project PyPI credentials for the 'twine upload' command. ::
 
   git clean -ffdx
-  python3 setup.py sdist bdist_wheel
+  pip3 install --upgrade build twine
```
marc-hb, line 91: I'm still not a fan of using the "random version of the day" but these instructions seem to work now and this file is only for MAINTAINERS who are supposed to know what they're doing.

PR #716, opened 2024-07-17
app: support alias commands

pdgendt:
This PR allows to add `alias` west commands to a configuration, similar to how git aliases work.

It can override existing commands to have default arguments, and iteratively replace alias commands.

Some examples:

```shell
# Add a global "west run" shortcut (will override pristine build)
$ west config --global alias.run "build --pristine=never -t run"

# Add a global "west menuconfig" shortcut
$ west config --global alias.menuconfig "build --pristine=never -t menuconfig"

# A shortcut to build my current project (prefix the path with $PWD/ to build anywhere from the workspace)
$ west config alias.my_proj "build -b native_sim samples/hello_world -- -DCONFIG_ASSERT=y"

# WARNING: overriding built-in or other commands is at your own risk
# build pristine by default locally
$ west config alias.build "build --pristine=always"
```

After applying you can run:

```shell
$ west run         #> west build --pristine=always --pristine=never -t run
$ west menuconfig  #> west build --pristine=always --pristine=never -t menuconfig
$ west my_proj     #> west build --pristine=always -b native_sim samples/hello_world -- -DCONFIG_ASSERT=y
```

### DIFF
--- a/src/west/app/main.py
+++ b/src/west/app/main.py
@@ -578,7 +578,7 @@ class WestApp:
         # gracefully. This provides more user-friendly output than
         # argparse can do on its own.
 
-        if (early_args.command_name and
+        if (early_args.command_name and not early_args.help and
             (early_args.command_name not in self.builtins and
              (not self.extensions or
               early_args.command_name not in self.extensions))):
--- a/src/west/app/main.py
+++ b/src/west/app/main.py
@@ -19,6 +19,7 @@ import logging
 import os
 from pathlib import Path, PurePath
 import platform
+import shlex
 import shutil
 import signal
 import sys
@@ -183,6 +184,7 @@ class WestApp:
         self.mle = None             # saved exception if load_manifest() fails
         self.builtins = {}          # command name -> WestCommand instance
         self.extensions = {}        # extension command name -> spec
+        self.aliases = {}           # alias -> WestCommand instance
         self.builtin_groups = OrderedDict()    # group name -> WestCommand list
         self.extension_groups = OrderedDict()  # project path -> ext spec list
         self.west_parser = None     # a WestArgumentParser
@@ -243,6 +245,7 @@ class WestApp:
         # Set self.manifest and self.extensions.
         self.load_manifest()
         self.load_extension_specs()
+        self.load_aliases()
 
         # Set up initial argument parsers. This requires knowing
         # self.extensions, so it can't happen before now.
@@ -402,6 +405,14 @@ class WestApp:
 
             self.extension_groups[path] = filtered
 
+    def load_aliases(self):
+        if not self.config:
+            return
+
+        self.aliases = {
+            k[6:]: Alias(k[6:], v) for k, v in self.config.items() if k.startswith('alias.')
+        }
+
     def handle_extension_command_error(self, ece):
         if self.cmd is not None:
             msg = f"extension command \"{self.cmd.name}\" couldn't be run"
@@ -424,9 +435,11 @@ class WestApp:
         # Set up and install command-line argument parsers.
 
         west_parser, subparser_gen = self.make_parsers()
+        real_command_names = set()
 
         # Add sub-parsers for the built-in commands.
-        for command in self.builtins.values():
+        for name, command in self.builtins.items():
+            real_command_names.add(name)
             command.add_parser(subparser_gen)
 
         # Add stub parsers for extensions.
@@ -436,9 +449,18 @@ class WestApp:
         # extension's code, which we won't do unless parse_known_args()
         # says to run that extension.
         if self.extensions:
-            for path, specs in self.extension_groups.items():
+            for specs in self.extension_groups.values():
                 for spec in specs:
+                    real_command_names.add(spec.name)
                     subparser_gen.add_parser(spec.name, add_help=False)
+
+        # Add aliases, but skip aliases that shadow other commands
+        # The help parser requires unique commands to be added
+        if self.aliases:
+            for name, alias in self.aliases.items():
+                # Advanced users shadowing real commands do not get "alias help"
+                if name not in real_command_names:
+                    alias.add_parser(subparser_gen)
 
         # Save the instance state.
         self.west_parser = west_parser
@@ -490,6 +512,28 @@ class WestApp:
         # Parse command line arguments and run the WestCommand.
         # If we're running an extension, instantiate it from its
         # spec and re-parse arguments before running.
+
+        if not early_args.help and early_args.command_name != "help":
+            # Recursively replace alias command(s) if set
+            aliases = self.aliases.copy()
+            while early_args.command_name in aliases:
+                # Make sure we don't end up in an infinite loop
+                alias = aliases.pop(early_args.command_name)
+
+                self.queued_io.append(lambda cmd, alias=alias: cmd.dbg(
+                    f'Replacing alias {alias.name} with {alias.args}'
+                ))
+
+                if len(alias.args) == 0:
+                    # This loses the cmd.dbg() above - too bad, don't use empty aliases
+                    self.print_usage_and_exit(f'west: empty alias "{alias.name}"')
+
+                # Find and replace the command name. Must skip any other early args like -v
+                for i, arg in enumerate(argv):
+                    if arg == early_args.command_name:
+                        argv = argv[:i] + alias.args + argv[i + 1:]
+                        break
+                early_args = early_args._replace(command_name=alias.args[0])
 
         self.handle_early_arg_errors(early_args)
         args, unknown = self.west_parser.parse_known_args(args=argv)
@@ -825,6 +869,8 @@ class Help(WestCommand):
             # exception handling block in app.run_command is in a
             # parent stack frame.
             app.run_extension(name, [name, '--help'])
+        elif app.aliases is not None and name in app.aliases:
+            app.aliases[name].parser.print_help()
         else:
             self.wrn(f'unknown command "{name}"')
             app.west_parser.print_help(top_level=True)
@@ -832,6 +878,25 @@ class Help(WestCommand):
                 self.wrn('your manifest could not be loaded, '
                          'which may be causing this issue.\n'
                          '  Try running "west update" or fixing the manifest.')
+
+class Alias(WestCommand):
+    # An alias command, it does not run itself
+
+    def __init__(self, cmd, args):
+        super().__init__(cmd, args or '<empty>', f'An alias that expands to: {args}')
+
+        self.args = shlex.split(args)
+
+    # Pseudo-parser that will never actually run except for ".print_help()"
+    def do_add_parser(self, parser_adder):
+        parser = parser_adder.add_parser(
+            self.name, help=self.help, description=self.description, add_help=False,
+            formatter_class=argparse.RawDescriptionHelpFormatter)
+
+        return parser
+
+    def do_run(self, args, ignored):
+        assert False
 
 class WestHelpAction(argparse.Action):
 
@@ -943,6 +1008,12 @@ class WestArgumentParser(argparse.ArgumentParser):
                     for spec in specs:
                         self.format_extension_spec(append, spec, width)
                     append('')
+
+            if self.west_app.aliases:
+                append('aliases:')
+                for alias in self.west_app.aliases.values():
+                    self.format_command(append, alias, width)
+                append('')
 
             if self.epilog:
                 append(self.epilog)
--- a/tests/test_alias.py
+++ b/tests/test_alias.py
@@ -1,0 +1,90 @@
+# Copyright (c) 2024, Basalte bv
+#
+# SPDX-License-Identifier: Apache-2.0
+
+import os
+import subprocess
+
+import pytest
+
+from conftest import cmd
+
+assert 'TOXTEMPDIR' in os.environ, "you must run these tests using tox"
+
+@pytest.fixture(autouse=True)
+def autouse_tmpdir(config_tmpdir, west_init_tmpdir):
+    # Since this module tests west's configuration file features,
+    # adding autouse=True to the config_tmpdir and west_init_tmpdir fixtures
+    # saves typing and is less error-prone than using it below in every test case.
+    pass
+
+def test_alias_commands():
+    cmd('config alias.test1 topdir')
+    cmd('config --global alias.test2 topdir')
+    cmd('config --system alias.test3 topdir')
+
+    topdir_out = cmd('topdir')
+
+    assert cmd('test1') == topdir_out
+    assert cmd('test2') == topdir_out
+    assert cmd('test3') == topdir_out
+
+def test_alias_help():
+    cmd('config alias.test topdir')
+
+    help_out = cmd('help test')
+
+    assert "An alias that expands to: topdir" in help_out
+    assert cmd('-h test') == help_out
+
+def test_alias_recursive_commands():
+    list_format = '{revision} TESTALIAS {name}'
+    cmd(['config', 'alias.test1', f'list -f "{list_format}"'])
+    cmd('config alias.test2 test1')
+
+    assert cmd('test2') == cmd(['list', '-f', list_format])
+
+def test_alias_infinite_recursion():
+    cmd('config alias.test1 test2')
+    cmd('config alias.test2 test3')
+    cmd('config alias.test3 test1')
+
+    with pytest.raises(subprocess.CalledProcessError) as excinfo:
+        cmd('test1', stderr=subprocess.STDOUT)
+
+    assert 'unknown command "test1";' in str(excinfo.value.stdout)
+
+def test_alias_empty():
+    cmd(['config', 'alias.empty', ''])
+
+    # help command shouldn't fail
+    cmd('help')
+
+    with pytest.raises(subprocess.CalledProcessError) as excinfo:
+        cmd('empty', stderr=subprocess.STDOUT)
+
+    assert 'empty alias "empty"' in str(excinfo.value.stdout)
+
+def test_alias_early_args():
+    cmd('config alias.test1 topdir')
+
+    # An alias with an early command argument shouldn't fail
+    assert "Replacing alias test1 with ['topdir']" in cmd('-v test1')
+
+def test_alias_command_with_arguments():
+    list_format = '{revision} TESTALIAS {name}'
+    cmd(['config', 'alias.revs', f'list -f "{list_format}"'])
+
+    assert cmd('revs') == cmd(['list', '-f', list_format])
+
+def test_alias_override():
+    before = cmd('list')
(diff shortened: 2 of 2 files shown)


src/west/app/main.py
```diff
@@ -479,6 +491,15 @@ def run_command(self, argv, early_args):
         # If we're running an extension, instantiate it from its
         # spec and re-parse arguments before running.
 
+        # Replace alias command(s) if set
+        while early_args.command_name in self.aliases:
+            new_args = shlex.split(self.aliases[early_args.command_name])
+            # Make sure we don't end up in an infinite loop
+            del self.aliases[early_args.command_name]
```
marc-hb, line 498: To be on the safe side, can you please work on a local clone of `self.aliases` here and leave `self.aliases` immutable? I don't think there is much concurrency in `west` yet but let's not leave "landmines" for anyone trying to add some in the future.

src/west/app/main.py
```diff
@@ -479,6 +491,15 @@ def run_command(self, argv, early_args):
         # If we're running an extension, instantiate it from its
         # spec and re-parse arguments before running.
 
+        # Replace alias command(s) if set
```
marc-hb, line 494:
```suggestion
        # Recursively replace alias command(s)
```

marc-hb (commented):
Nice! Amazing how little code this requires.

Can you please submit a documentation PR for https://docs.zephyrproject.org/latest/develop/west/config.html#built-in-configuration-options ?

It sucks that it is a in different repo but that's just the way it is now.


I would like the interface design to be reviewed before the implementation.  Please do mention that it's recursive. Also mention briefly that overriding/shadowing built-in commands is possible but please change the (commit messages) examples that do that: that looks like a sure support nightmare! I think it's fine if "advanced" users who know what they're doing override built-in commands but we really don't want beginners to do that and then have their friends, colleagues and support spend hours scratching their heads.

src/west/app/main.py
```diff
@@ -479,6 +491,16 @@ def run_command(self, argv, early_args):
         # If we're running an extension, instantiate it from its
         # spec and re-parse arguments before running.
 
+        # Recursively replace alias command(s) if set
```
marc-hb, line 494: Please guard this code with something like:
```python
   if not early_args.help and early_args.command_name != "help":
```
(UNTESTED)

This should stop `west help some_alias` and `west -h some_alias` from printing the help of a command resolved 3 or 4 levels down. In the future we want `west help some_alias` to print the alias just like `git help some_alias` does nicely. This guard will help get ready for that. Today it's better for `west help some_alias` to fail or print nothing rather than getting into the bad habit of `west help some_alias` showing the help of the resolved command (XKCD 1172)

tests/test_alias.py
```diff
@@ -0,0 +1,43 @@
...
+    cmd('config alias.test2 test1')
+
+    assert cmd('test2') == cmd('help')
+
+def test_alias_command_with_arguments():
+    cmd('config alias.th "topdir -h"')
+
+    assert cmd('th') == cmd('topdir -h')
+
+def test_alias_override():
```
marc-hb, line 40: thanks a lot for being thorough.

src/west/app/main.py
pdgendt: Hmn, currently all my tests use the `help` command, as it is built-in.

pdgendt:
Updated the PR to have `west help <alias>` print out info about the alias.

pdgendt:
Calling `west help` now also lists alias commands.

src/west/app/main.py
```diff
@@ -428,10 +442,17 @@ def setup_parsers(self):
         # extension's code, which we won't do unless parse_known_args()
         # says to run that extension.
         if self.extensions:
-            for path, specs in self.extension_groups.items():
+            for specs in self.extension_groups.values():
                 for spec in specs:
+                    command_names.add(spec.name)
                     subparser_gen.add_parser(spec.name, add_help=False)
 
+        # Add aliases, but skip aliases that shadow other commands
```
marc-hb, line 450: Please briefly allude to why these are skipped.

src/west/app/main.py
```diff
@@ -479,6 +500,20 @@ def run_command(self, argv, early_args):
...
 
+        if not early_args.help and early_args.command_name != "help":
+            # Recursively replace alias command(s) if set
+            aliases = self.aliases.copy()
+            while early_args.command_name in aliases:
+                new_args = shlex.split(aliases[early_args.command_name].args)
+                # Make sure we don't end up in an infinite loop
+                del aliases[early_args.command_name]
+                self.queued_io.append(lambda cmd: cmd.dbg(
+                    f'Replacing alias {early_args.command_name} with {new_args}'
```
marc-hb, line 511: While trying to see this log with `west -v mydiff` I'm afraid I found a command parsing issue

```shell
west config alias.mydiff "diff --stat"

west      diff  # ok
west -v   diff  # ok
west    mydiff  # ok
west -v mydiff  # fails

usage: west [-h] [-z ZEPHYR_BASE] [-v] [-V] <command> ...
west: unknown command "mydiff"; workspace ... does not define this extension command ...
```

tests/test_alias.py
```diff
@@ -0,0 +1,51 @@
...
+    assert cmd('test3') == topdir_out
+
+def test_alias_recursive_commands():
+    cmd('config alias.test1 topdir')
+    cmd('config alias.test2 test1')
+
+    assert cmd('test2') == cmd('topdir')
+
+def test_alias_command_with_arguments():
+    list_format = '{revision}'
```
marc-hb, line 36: Please add some whitespace here because:
1. It provides some important test coverage for quoting issues
2. Normal usage will typically involve whitespace

```suggestion
    list_format = '{revision} TESTALIAS {name}'
```

"TESTALIAS" in case someone wants to check the output in the future. Also a grep "anchor" for test logs/source/etc.

tests/test_alias.py
```diff
@@ -0,0 +1,51 @@
...
+    cmd(f'config alias.revs "list -f \"{list_format}\""')
```
marc-hb, line 37:
```suggestion
    cmd(f'config alias.revs "list -f '{list_format}'"')
```

src/west/app/main.py
```diff
@@ -815,6 +852,24 @@ def do_run(self, args, ignored):
                          'which may be causing this issue.\n'
                          '  Try running "west update" or fixing the manifest.')
 
+class Alias(WestCommand):
+    # An alias command, it does not run itself
+
+    def __init__(self, cmd, args):
+        super().__init__(cmd, f'"{args}"', f'An alias that expands to "{args}"')
```
marc-hb, line 859: Don't "naive" quote because the output maybe invalid, example:
```
aliases:
                     "list -f "{name} {revision}""
```
```suggestion
        super().__init__(cmd, f'"{args}"', f'An alias that expands to {args}')
```
Either quote correctly using `shlex`, or (much simpler and preferred) just don't quote here.

tests/test_alias.py
pdgendt: This isn't possible because of the outer `f'...'` formatted quotes.

pdgendt:
@marc-hb Thanks for the review!
I will look into the difference between `west help some_alias` and `west -h some_alias`

marc-hb:
> I will look into the difference between west help some_alias and west -h some_alias and the alias issue on windows.

I just did some interactive `west config` testing with both Powershell and CMD.EXE. You can nest single quotes inside double quotes in both but not the other way round. Don't ask me why. That's hopefully the only reason why https://github.com/zephyrproject-rtos/west/actions/runs/10585812818/job/29333273125?pr=716 failed.

If we really wanted to test quoting more then we should check some pre-canned config file in git and avoid command line peculiarities entirely. But for now I think restricting tests to single quotes nested in double quotes is portable and plenty good enough. We just want to make sure no one breaks the most basic whitespace usage and then trust `shlex` for the rest.

pdgendt:
@marc-hb I've handled all comments, I think.

tests/conftest.py
```diff
@@ -327,9 +327,13 @@ def cmd(cmd, cwd=None, stderr=None, env=None):
     # stdout from cmd is captured and returned. The command is run in
     # a python subprocess so that program-level setup and teardown
     # happen fresh.
-    cmd = 'west ' + cmd
-    if not WINDOWS:
-        cmd = shlex.split(cmd)
+    if isinstance(cmd, str):
+        cmd = 'west ' + cmd
+        if not WINDOWS:
+            cmd = shlex.split(cmd)
```
marc-hb, line 333: It's totally not your fault but using `shlex` here was a bad idea. Let me look it into it.

https://github.com/zephyrproject-rtos/west/pull/266#discussion_r284670076

marc-hb (dismissed):
Let me try getting rid of shlex in conftest.py first.

tests/conftest.py
marc-hb: I have some working, should submit a PR that removes shlex today.

marc-hb: Done, please take a look. It's long (and was tedious) but it's conceptually dead simple:
- #730

pdgendt: Rebased this PR on top of #730

marc-hb: #730 merged, please rebase

src/west/app/main.py
```diff
@@ -394,6 +397,15 @@ def load_extension_specs(self):
...
             self.extension_groups[path] = filtered
 
+    def load_aliases(self):
+        if not self.config:
+            return
+
+        for item in self.config.items():
+            ids = item[0].split('.', 1)
+            if ids[0] == 'alias':
+                self.aliases[ids[1]] = Alias(ids[1], item[1])
```
marc-hb, line 407: Codestyle nit: this looks like a good use case for list comprehension:

```python
self.aliases = [ Alias(key.split('.',1)[1], value) for
      key, value in self.config.items() if key.split('.')[0] == 'alias' ]
```

tests/test_alias.py
```diff
@@ -0,0 +1,62 @@
...
+    assert cmd('test3') == topdir_out
+
+def test_alias_help():
+    cmd('config alias.test topdir')
+
+    assert cmd('-h test') == cmd('help test')
+
+def test_alias_recursive_commands():
+    cmd('config alias.test1 topdir')
+    cmd('config alias.test2 test1')
```
marc-hb, line 36: Can you please add some whitespace and quoting in the recursive test case to make it more demanding? Maybe 3 levels too. Basically: something that breaks when you remove the `shlex` parsing that you just added.

It shouldn't require much more test code.

tests/test_alias.py
```diff
@@ -0,0 +1,62 @@
...
+    topdir_out = cmd('topdir')
+
+    assert cmd('test1') == topdir_out
+    assert cmd('test2') == topdir_out
+    assert cmd('test3') == topdir_out
+
+def test_alias_help():
+    cmd('config alias.test topdir')
+
+    assert cmd('-h test') == cmd('help test')
```
marc-hb, line 32: :-)

tests/test_alias.py
marc-hb: Also: some infinite loop as a "negative" test?

src/west/app/main.py
pdgendt: The aliases are a dictionary, so I changed it a bit, PTAL.

tests/test_alias.py
```diff
@@ -0,0 +1,72 @@
+# Copyright (c) 2024, Basalte bv
```
tejlmand, line 1: curious, why doesn't the files in `/tests` have a license tag :thinking:

pdgendt: Added it to this file.

src/west/app/main.py
```diff
@@ -491,6 +512,24 @@ def run_command(self, argv, early_args):
...
 
+        if not early_args.help and early_args.command_name != "help":
+            # Recursively replace alias command(s) if set
+            aliases = self.aliases.copy()
+            while early_args.command_name in aliases:
+                new_args = shlex.split(aliases[early_args.command_name].args)
+                # Make sure we don't end up in an infinite loop
+                del aliases[early_args.command_name]
+                self.queued_io.append(lambda cmd: cmd.dbg(
+                    f'Replacing alias {early_args.command_name} with {new_args}'
```
marc-hb, line 523: I'm testing the latest cead8d12a1d39f49ea88eaa733a2092ef9d992f2 and there is still a logging issue sorry:

```
west help

aliases:

  alpha:                beta
  beta:                 gamma
  gamma:                diff --stat

west -v alpha

Replacing alias diff with ['diff', '--stat']
Replacing alias diff with ['diff', '--stat']
Replacing alias diff with ['diff', '--stat']

```

pdgendt: Thanks for spotting, fixed.

marc-hb: Was this ONLY a logging bug? I could be wrong but the fix looks a big bigger than that...?

If and only if it was more than a logging bug, does this need more test coverage?

pdgendt: It was a deferred logging issue, where the logging lambda didn't capture copies. To keep it clean, I've captured the alias itself (using `pop` vs `del`)

One addition here is that `shlex.split` is called when the aliases are created (these never change anyway).

I should have been more clear about that, sorry.

marc-hb:
Just tried the latest https://github.com/zephyrproject-rtos/west/commit/93f9715c6410023986065da87709088cccea5b78 and infinite loops do not seem handled well. From reading the code I would expect this to print `west: unknown command alpha` but instead I get:

```
west help

aliases:
  alpha:                beta
  beta:                 alpha

west alpha

  File "~/.local/bin/west", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "src/west/app/main.py", line 1172, in main
    app.run(argv or sys.argv[1:])
  File "src/west/app/main.py", line 255, in run
  File "src/west/app/main.py", line 556, in run_command
    self.run_extension(args.command, argv)
  File "src/west/app/main.py", line 682, in run_extension
    self.cmd = self.extensions[name].factory()
               ~~~~~~~~~~~~~~~^^^^^^
KeyError: 'alpha'
```
More tests needed maybe?

I warned you recursion (and overriding built ins) was fraught with peril! :-)

pdgendt:
> I warned you recursion (and overriding built ins) was fraught with peril! :-)

Yes, the early argument handling does make it a bit more complicated, but I think I've managed to make it more robust now. 

> More tests needed maybe?

I added a test for the infinite recursion to contain "unknown command".

pdgendt:
Update: handle empty aliases and add a test

marc-hb:
> Update: handle empty aliases

`west help` still broken with empty aliases and latest commit https://github.com/zephyrproject-rtos/west/commit/f8c61d1e9a11f46f5f9977e957b60c883ec72c46

```
west config alias.bar ""

west bar
usage: west [-h] [-z ZEPHYR_BASE] [-v] [-q] [-V] <command> ...
west: empty alias "bar"

west help

  File "/var/home/.local/bin/west", line 8, in <module>
    sys.exit(main())
             ^^^^^^
  File "src/west/app/main.py", line 1175, in main
    app.run(argv or sys.argv[1:])
  File "src/west/app/main.py", line 255, in run
  File "src/west/app/main.py", line 557, in run_command
    self.run_builtin(args, unknown)
  File "src/west/app/main.py", line 673, in run_builtin
    self.cmd.run(args, unknown, self.topdir,
  File "src/west/commands.py", line 194, in run
    self.do_run(args, unknown)
  File "src/west/app/main.py", line 862, in do_run
    app.west_parser.print_help(top_level=True)
  File "src/west/app/main.py", line 927, in print_help
    print(self.format_help(top_level=top_level), end='',
          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "src/west/app/main.py", line 1014, in format_help
    self.format_command(append, alias, width)
  File "src/west/app/main.py", line 1039, in format_command
    self.format_thing_and_help(append, thing, command.help, width)
  File "src/west/app/main.py", line 1075, in format_thing_and_help
    help_lines[0] = thing + help_lines[0][thinglen:]
                            ~~~~~~~~~~^^^
IndexError: list index out of range
```

pdgendt:
> `west help` still broken with empty aliases and latest commit [f8c61d1](https://github.com/zephyrproject-rtos/west/commit/f8c61d1e9a11f46f5f9977e957b60c883ec72c46)

Ugh, sorry about that, and thanks for all the testing efforts! Empty help strings seem to get stripped. I now set an `<empty>` placeholder. Added it to the test.

pdgendt:
Fixed typo 🤦

src/west/app/main.py
```diff
@@ -833,6 +879,24 @@ def do_run(self, args, ignored):
...
 
+class Alias(WestCommand):
+    # An alias command, it does not run itself
+
+    def __init__(self, cmd, args):
+        super().__init__(cmd, args or '<empty>', f'An alias that expands to: {args}')
+
+        self.args = shlex.split(args)
+
+    def do_add_parser(self, parser_adder):
```
marc-hb, line 890: Do I get this right?
```suggestion
    # Pseudo-parser that will never actually run except for ".print_help()"
    def do_add_parser(self, parser_adder):
```

tests/test_alias.py
```diff
@@ -0,0 +1,87 @@
...
+    topdir_out = cmd('topdir')
+
+    assert cmd('test1') == topdir_out
+    assert cmd('test2') == topdir_out
+    assert cmd('test3') == topdir_out
+
+def test_alias_help():
+    cmd('config alias.test topdir')
+
+    assert cmd('-h test') == cmd('help test')
```
marc-hb, line 35: Can you make sure the output contains the words "alias" and "topdir"? Right now this test will not fail if both commands print the _same_ garbage.

marc-hb (commented):
A few questions and minor suggestions left.

src/west/app/main.py
```diff
@@ -578,10 +620,12 @@ def handle_early_arg_errors(self, early_args):
         # gracefully. This provides more user-friendly output than
         # argparse can do on its own.
 
-        if (early_args.command_name and
-            (early_args.command_name not in self.builtins and
-             (not self.extensions or
-              early_args.command_name not in self.extensions))):
+        if (
+            not early_args.help and
```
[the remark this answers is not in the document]
pdgendt: This fixes an issue with using `-h alias`, I thought about the side-effects of this too, could it break in other places?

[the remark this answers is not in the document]
marc-hb: > I thought about the side-effects of this too, could it break in other places?

In doubt, please make it a separate commit. This will make it easy to revert in the worst case. I just tried and a revert would only lose `west -h alias` which would really be minor.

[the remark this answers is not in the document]
pdgendt: Fixed

marc-hb (approved):

PR #718, opened 2024-07-19
app: Allow passing arguments to git status/diff

pdgendt:
`west` should pass unknown arguments to `git status` and `git diff` commands.

This allows things like:

```shell
$ west status --short --branch
$ west diff --stat
```

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -769,7 +769,10 @@ class Diff(_ProjectCommand):
         super().__init__(
             'diff',
             '"git diff" for one or more projects',
-            'Runs "git diff" on each of the specified projects.')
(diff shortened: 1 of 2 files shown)


aasinclair (approved):

src/west/app/project.py
```diff
@@ -769,7 +769,9 @@ def __init__(self):
         super().__init__(
             'diff',
             '"git diff" for one or more projects',
-            'Runs "git diff" on each of the specified projects.')
+            'Runs "git diff" on each of the specified projects.',
```
carlescufi, line 772: Should we document the fact that extra arguments are accepted?

PR #719, opened 2024-07-19
app: project: Allow to diff against manifest revision

pdgendt:
Add an argument to the west diff command that allows to set the manifest revision as the merge-base.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -779,6 +779,8 @@ class Diff(_ProjectCommand):
                             defaults to active cloned projects''')
         parser.add_argument('-a', '--all', action='store_true',
                             help='include output for inactive projects')
+        parser.add_argument('-m', '--manifest', action='store_true',
+                            help='show changes relative to the manifest revision')
         return parser
 
     def do_run(self, args, ignored):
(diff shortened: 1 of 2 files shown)


aasinclair (approved):

carlescufi (approved):

marc-hb:
While I'm  truly in awe of @pdgendt 's west productivity, I have concerns this one was a bit rushed... why `--merge-base`? `git diff` is useful both with and without `--merge-base` but:
- The --help text never mentions `--merge-base`.
- Why is the choice not given?
- While `merge-base` is simple in... simple cases, it can be very complicated in general: you can have 0 merge based, or more than one. What happens then? As `west` is a general purpose git tool, this should be at the very least documented. 


See for instance https://github.com/codecov/codecov-bash/issues/83

pdgendt:
Right, I've created a follow-up PR that simplifies it: #728

src/west/app/project.py
```diff
@@ -792,9 +794,10 @@ def do_run(self, args, ignored):
         for project in self._cloned_projects(args, only_active=not args.all):
             # Use paths that are relative to the base directory to make it
             # easier to see where the changes are
+            merge_base = ['--merge-base', project.revision] if args.manifest else []
```
marc-hb, line 797: `project.revision` is not OK: it can be a _remote_ branch name and then `git diff` will fail. Or something else and who knows what might happen.

I just submitted a replacement of `project.revision` by `manifest-rev` in #748

Longer story in #747

PR #722, opened 2024-07-21
commands: add the 'sdk' command to manage SDK.

soburi:
'west sdk' command can list and install SDK.

Run 'west sdk --install=latest' to simply install the latest SDK or
'west sdk' show installed SDKs informations.

This command is just a wrapper for SDK's setup command.
But this command makes the SDK install process simple.
It will be a good improvement for onboarding first-time users.

--

By introducing this command, you no longer need to check the SDK installation procedure to install the SDK,
so I think this will improve the experience for first-time users.

It depends on the SDK-NG installer and GitHub API,
but I think these are stable enough that I don't think it will be a big problem.


pdgendt:
I think this belongs in the `zephyr` project instead of `west`.

marc-hb (changes requested):
This definitely belongs to `zephyr/scripts/west_commands`. West is trying to become less and less Zephyr-specific and the Zephyr SDK is obviously Zephyr-specific.

src/west/app/project.py
```diff
@@ -1914,6 +1921,335 @@ def do_add_parser(self, parser_adder):
...
+    /opt
+    $ENV{HOME}
+    $ENV{HOME}/.local
+    $ENV{HOME}/.local/opt
+    $ENV{HOME}/bin)
+
+# Search for Zephyr SDK version 0.0.0 which does not exist, this is needed to
+# return a list of compatible versions and find the best suited version that
+# is available.
+find_package(Zephyr-sdk 0.0.0 EXACT QUIET CONFIG PATHS ${zephyr_sdk_search_paths})
```
marc-hb, line 2176: @stephanosio , @tejlmand , FYI.

soburi:
@pdgendt @marc-hb 

Thank you for your opinions

I've recreated this for the Zephyr repository.
https://github.com/zephyrproject-rtos/zephyr/pull/76283

PR #736, opened 2024-09-13
venv: Add support for virtual environment specs

yperess:
Adds handling for a venv key in the manifest. This key will then be used to provide virtual environment metadata to a west bootstrap command. The metadata includes:
- A virtual environment name
- A list of python requirement files, individual packages, constraints, and local directories.
- A mapping of binary requirements and which URLs can be used to find them. These include architecture and OS specific mappings.

### DIFF
--- a/src/west/manifest-schema.yml
+++ b/src/west/manifest-schema.yml
@@ -66,6 +66,78 @@ mapping:
             required: true
             type: str
 
+  # The venv key specifies requirements for the build environment for Zephyr
+  # West will handle merging any definitions into your binary dependency
+  # strings where allowed.
+  # Example:
+  #
+  #   urls:
+  #     "windows-*":
+  #       url: "https://my.download.server/win/${arch}.zip"
+  #
+  # The above will match any architecture for the windows OS and generate a url
+  # replacing ${arch} with the current machine's underlying architecture. West
+  # will automatically provide a few definitions:
+  # - os: The current OS (one of 'windows', 'linux', or 'mac')
+  # - arch: The current architecture. This value is standardized such that
+  #   x86_64 is represented as amd64 and aarch64 as arm64. All others match
+  #   Python's `platform.machine()`
+  # - platform: A concatenation of ${os}-${arch}
+  venv:
+    required: false
+    type: map
+    mapping:
+      definitions:
+        required: false
+        type: map
+        mapping:
+          regex;(.*):
+            type: str
+      name:
+        required: false
+        type: str
+      bin-requirements:
+        required: false
+        type: map
+        mapping:
+          regex;(.*):
+            type: map
+            mapping:
+              definitions:
+                required: false
+                type: map
+                mapping:
+                  regex;(.*):
+                    type: str
+              urls:
+                required: true
+                type: map
+                mapping:
+                  regex;(.*):
+                    type: map
+                    mapping:
+                      url:
+                        required: true
+                        type: str
+                      paths:
+                        required: false
+                        type: seq
+                        sequence:
+                          - type: str
+      py-requirements:
+        required: false
+        type: seq
+        sequence:
+          - type: map
+            mapping:
+              pattern:
+                required: true
+                type: str
+              type:
+                required: true
+                type: str
+                enum: ['directory', 'constraints', 'package', 'requirements']
+
   # The "projects" key specifies a sequence of "projects", each of which has a
   # remote, and may specify additional configuration.
   #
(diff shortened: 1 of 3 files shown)


yperess:
@nashif what's the process for getting this PR reviewed?

src/west/manifest.py
```diff
@@ -67,6 +68,64 @@
...
+    defs = {}
+    current_os = platform.system()
+    if current_os == "Windows":
+        defs["os"] = "windows"
+    elif current_os == "Darwin":
+        defs["os"] = "mac"
+    elif current_os == "Linux":
+        defs["os"] = "linux"
+    else:
+        raise RuntimeError("Unknown OS: " + current_os)
```
marc-hb, line 94: Does this code run when you don't use the new feature?

src/west/manifest.py
```diff
@@ -67,6 +68,64 @@
...
+    Returns:
+        A dictionary containing the keys 'os', 'arch', and 'platform' and their
+        current values.
+    """
+    defs = {}
+    current_os = platform.system()
+    if current_os == "Windows":
+        defs["os"] = "windows"
+    elif current_os == "Darwin":
+        defs["os"] = "mac"
```
marc-hb, line 90: Is this indirection really necessary?

src/west/manifest.py
```diff
@@ -67,6 +68,64 @@
 # Internal helpers
 #
 
+# Mapping of the architectures from platofrm.machine() to a common standard
```
marc-hb, line 71: Do you mean, to some "Zephyr" standard? I agree "Darwin" is a little bit cryptic but `aaarch64` is not.

src/west/manifest-schema.yml
```diff
@@ -66,6 +66,78 @@ mapping:
             required: true
             type: str
 
+  # The venv key specifies requirements for the build environment for Zephyr
```
marc-hb, line 69: The code in this repo is supposed to be not specific to Zephyr. Is this PR Zephyr-specific? I don't have time for a full review sorry.

If it is Zephyr-specific, then it belongs to https://github.com/zephyrproject-rtos/zephyr/tree/main/scripts/west_commands

Considering `west build` is Zephyr-specific, I suspect this PR is too.

Even if it's not, https://github.com/zephyrproject-rtos/zephyr/tree/main/scripts/west_commands could be a good "staging" area to gather feedback, real-use and... more reviewers.

marc-hb (commented):
This is IMHO missing some high-level descriptions of the problem statement and of the solution's design. I'm especially wondering what a "binary requirement" is?

I think the best way to fix that would be a corresponding documentation PR, for instance see how this unrelated feature had a separate documentation PR:
https://github.com/zephyrproject-rtos/zephyr/pull/76027

`west` documentation is unfortunately located in the Zephyr repo for various reasons:  fewer sites to maintain, single place for `west` built-ins and Zephyr extensions,...

The obvious drawback is: PRs like these must be split across two repos.

PS: there is a shortage of west maintainers right now.

src/west/manifest.py
yperess: It did, thanks for catching it. I've moved it to a lazily initialized class so it only gets run if you're using the feature.

yperess:
> This is IMHO missing some high-level descriptions of the problem statement and of the solution's design. I'm especially wondering what a "binary requirement" is?
> 
> I think the best way to fix that would be a corresponding documentation PR, for instance see how this unrelated feature had a separate documentation PR: [zephyrproject-rtos/zephyr#76027](https://github.com/zephyrproject-rtos/zephyr/pull/76027)
> 
> `west` documentation is unfortunately located in the Zephyr repo for various reasons: fewer sites to maintain, single place for `west` built-ins and Zephyr extensions,...
> 
> The obvious drawback is: PRs like these must be split across two repos.
> 
> PS: there is a shortage of west maintainers right now.

Thank you for the review. I've added a WIP CL that uses this and adds documentation at https://github.com/zephyrproject-rtos/zephyr/pull/78610. It's a WIP because I haven't had the chance to track down all the binary dependencies though I got a few of them (enough to see how this PR will be used).

tejlmand (changes requested):
this is a big no-go from my side.

The west manifest should not start to contain a bunch of keys and values not relevant or used by core west itself.

Nor should it contain a bunch of information which may not be of interest to several users of the manifest.

See also this comment for more details: https://github.com/zephyrproject-rtos/zephyr/pull/78610#pullrequestreview-2312682185

marc-hb:
Thanks for submitting https://github.com/zephyrproject-rtos/zephyr/pull/78610, I also replied there. Once the dust settles there we can revisit what belongs to Zephyr versus what belongs to west itself.

PR #737, opened 2024-09-18
.github: Pypi release workflows

pdgendt:
Add Github workflows to create Pypi packages and upload to Pypi when a release is published.

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,40 @@
+name: Release
+
+# This workflow follows Pypi guidelines for publishing using trusted publishers
+# See https://docs.pypi.org/trusted-publishers/using-a-publisher/ for more details.
+
+on:
+  release:
+    types: [published]
+
+jobs:
+  package:
+    name: Package
+    uses: ./.github/workflows/package.yml
+
+  release:
+    name: Release
+    needs: [package]
+    runs-on: ubuntu-latest
+
+    # Make the GH environment explicit
+    environment: release
+
+    # Mandatory for trusted publishing
+    permissions:
+      contents: write
+      id-token: write
+
+    steps:
+    - name: Download build artifacts
+      uses: actions/download-artifact@v4
+      with:
+        path: dist/
+
+    - name: Upload release assets
+      uses: softprops/action-gh-release@v2
+      with:
+        files: dist/*.whl
+
+    - name: Publish package to PyPI
+      uses: pypa/gh-action-pypi-publish@release/v1
--- a/MAINTAINERS.rst
+++ b/MAINTAINERS.rst
@@ -100,11 +100,16 @@ Pre-release test plan
 Building and uploading the release wheels
 -----------------------------------------
 
-You need the zephyr-project PyPI credentials for the 'twine upload' command. ::
+Creating Pypi releases is done automatically from Github. After publishing
+a release on Github a release build is packaged and uploaded with the
+version specified in pyproject.toml.
+
+To do these steps manually, you need the zephyr-project PyPI credentials
+for the 'twine upload' command. ::
 
   git clean -ffdx
   pip3 install --upgrade build twine
-  pyproject-build
+  python -m build
   twine upload -u zephyr-project dist/*
 
 The 'git clean' step is important. We've anecdotally observed broken wheels
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -1,0 +1,25 @@
+name: Python Package
+
+on: [push, pull_request, workflow_call]
+
+jobs:
+  package:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Set up Python
+        uses: actions/setup-python@v5
+        with:
+          python-version: "3.12"
+      - name: Install Python dependencies
+        run: |
+          pip3 install build
+      - name: Build a binary wheel and a source tarball
+        run: |
+          python -m build
+      - name: Store the distribution packages
+        uses: actions/upload-artifact@v4
+        with:
+          name: python-package-distributions
+          path: dist/
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: [3.8, 3.9, '3.10']
+        python-version: [3.8, 3.9, '3.10', '3.11', '3.12']
     steps:
       - uses: actions/checkout@v4
         # This is enough to find many quoting issues


.github/workflows/release.yml
```diff
@@ -0,0 +1,29 @@
...
+    steps:
+    - name: Download build artifacts
+      uses: actions/download-artifact@v4
+      with:
+        path: dist
+
+    - name: Publish package to PyPI
+      uses: pypa/gh-action-pypi-publish@release/v1
+      with:
+        password: ${{ secrets.PYPI_API_TOKEN }}
```
marc-hb, line 29: Is that publishing the final, official package? I'm not sure about this because this is adding one (big) entity to the supply chain: Github.

I'd rather to have reproducible packages instead, is that easy with Python?

pdgendt: I'm not sure how this could be done. The sources are packaged next to the release wheels and can be downloaded from the artifacts.
I think this automation adds traceability as there are logs from the GH actions and artifacts to download. The current (manual) actions are also done from "some" OS with "some" python version.

pdgendt: CC @carlescufi @stephanosio 
As this was discussed on discord to actually move this to a GH action rather than needing special permissions for users.

MAINTAINERS.rst
```diff
@@ -85,17 +85,9 @@ Pre-release test plan
...
-  pyproject-build
-  twine upload -u zephyr-project dist/*
-
-The 'git clean' step is important. We've anecdotally observed broken wheels
-being generated from dirty repositories.
-
-Check out [packaging.python.org](https://packaging.python.org/en/latest/tutorials/packaging-projects/#generating-distribution-archives) for more detailed instructions.
+Creating Pypi releases is done automatically from Github. After publishing
+a release on Github a release build is packaged and uploaded with the
+version specified in pyproject.toml.
```
stephanosio, line 90: might want to keep the documentation for manual package upload process as a backup, while also mentioning that it is normally automated and manual uploads are not necessary.

.github/workflows/release.yml
stephanosio: > Is that publishing the final, official package? I'm not sure about this because this is adding one (big) entity to the supply chain: Github.

I think there is not much to debate about having an automated process that builds and publishes the packages from a tag is a step forward compared to someone manually checking out the tag, building it on their local machine, and uploading the packages -- automated publishing is clearly a more streamlined, reproducible and secure process.

Below I suggested to keep the existing documentation for manually publishing a release as a backup in case GitHub-based automation becomes a no longer viable option for whatever reason.

stephanosio (dismissed):
Not seeing anything obviously wrong with it.

.github/workflows/release.yml
pdgendt: Updated

marc-hb: > automated publishing is clearly a more streamlined, reproducible and secure process.

Automation is great. It is more "streamlined".

Public build logs also make it much easier to _attempt_ to reproduce the build but that does not mean the build is actually reproducible. Did anyone tried?

More transparency is always good for security but 1) transparency is not the only security factor 2) the GitHub's infrastructure is completely opaque and has a very large attack surface. My main concern is this: GitHub is now trusted with the secret key or token that allows uploading to PyPI a release **that was not built by Github**.

This is the problem solved by reproducible builds.

For official releases, automating 99.9% of the work is enough. Because they happen only a few times per year.

> I think there is not much to debate... automated publishing is clearly ...

https://rationalwiki.org/wiki/Appeal_to_common_sense

marc-hb: I did a bit of testing and just shared good build reproducibility news in #738

marc-hb: > For official releases, automating 99.9% of the work is enough. Because they happen only a few times per year.

Another concern going from 99.9% automated release to 100% is the lack of "friction" itself: once each release is literally the push of a single button, then why would the release manager pause and look at release checklists and other test results? If something was missed or wrong then it's just the matter of pushing yet another "quick fix" and pushing the release button again; problem solved! Now the validation and quality difference between running releases versus the main branch is gone.

This is nothing wrong with such a "Continuous Delivery" channel as long as it does not pretend to be a quarterly release with a higher bar.

stephanosio: > > I think there is not much to debate... automated publishing is clearly ...
> 
> https://rationalwiki.org/wiki/Appeal_to_common_sense

Do I really need to explain why it is more reproducible and secure?

For reproducibility, GitHub runner environment is much more well defined than some developer's local machine -- it is less likely to produce different outputs due to execution environment-related factors. Of course, that should not happen in the first place; but, I am saying in case that happens ...

For security, I thought this really was a no-brainer, but since you insist ... Multiple entities having access to the secret token for publishing packages cannot possibly be safer than single entity having access to it, especially when the multiple "entities" involved are humans.

Automated single "cloud" secret storage is more likely to do a better job in keeping secrets than multiple maintainers keeping secrets in who knows what (did someone say `~/secrets.txt` in an unencrypted volume on their laptop?).

> the GitHub infrastructure is completely opaque and has a very large attack surface.

We already trust GitHub with so many things, including the source code of west itself. How do you feel safe building locally only to pull the source code from GitHub, which is a "very large attack surface"?

Heck, how do you trust PyPI even, especially when the pip package manager is completely void of the concept of version locking with package hashes?

> My main concern is this: GitHub is now trusted with the secret key or token that allows uploading to PyPI anything **that was not even built by Github**.

We already established above that GitHub is likely to do a better job in keeping secrets than multiple maintainers.

As for uploading to PyPI anything that was not even built by GitHub, that cannot happen unless GitHub is hacked, or malicious workflow changes are committed to the main branch.

stephanosio: > f something was missed or wrong then it's just the matter of pushing yet another "quick fix" and pushing the release button again

That is why we publish multiple "release candidates" before actually publishing a release?

marc-hb: > Do I really need to explain why it is more reproducible and secure?

Please stop being condescending; security is rarely ever "obvious". However, it should be "re-usable" as in: referencing prior art, existing threat models, analysis etc. So, yes: ideally you should not need to explain anything but you should point me at good references instead. Same as crypto: re-use, don't re-invent!


> For reproducibility, GitHub runner environment is much more well defined...

Yes, I never meant to question that, sorry for the confusion if any. I've been only questioning the 0.1%, ultimate step of automatically uploading the package to PyPI. That's the only step that requires trusting Github with a secret and with direct, fully automated write access to PyPI. This is more or less "merging" two security domains that used to be distinct.

> Multiple entities having access to the secret token for publishing packages cannot possibly be safer than single entity having access to it, especially when the multiple "entities" involved are humans.

Which "multiple" entities do you have in mind and how many? Why "multiple" = humans and "single" = Github? Which secrets are long-lived keys versus shorter-lived tokens? How often do you renew each? What redundancy is there when some secrets are lost or compromised?

If you found a way to make secret management "obvious" then you may want to start a new business because there are people, products and companies who make a living with what they think is a complex problem.


> We already trust GitHub with so many things, including the source code of west itself.
> How do you feel safe building locally only to pull the source code from GitHub, which is a "very large attack surface"?

That's a totally different security problem because the source code is not a _secret_. Also, git does not blindly trusts Github and offers some level of integrity.


> Heck, how do you trust PyPI even, especially when the pip package manager is completely void of the concept of version locking with package hashes?

I really don't think  security problems in one place is ever a good excuse to ignore security in another.  This is off-topic; this PR makes a significant security change with respect to how releases are published so it must carefully describe the security implications of that particular change (ideally: thanks to good references).

Again: I see nothing wrong with the _automation_ part. I'm only questioning the secret token and final, automated publishing part.


> As for uploading to PyPI anything that was not even built by GitHub, that cannot happen unless GitHub is hacked, ...

... and Github is invulnerable?

https://github.blog/news-insights/company-news/we-updated-our-rsa-ssh-host-key/


> That is why we publish multiple "release candidates" before actually publishing a release?

My concern was release managers potentially getting "sloppy". While Release Candidates are very useful, I don't see how they help with that specific concern. If anything they would rather make it worse, giving a potentially false impression that everything is great. No one knows how much actual test coverage release candidates get. Release Candidates only demonstrate "the presence of bugs, never their absence".

Anyway this isn't the main issue.

stephanosio: GitHub can be hacked. PyPI can be hacked. My computer can be hacked. Your computer can be hacked.

The deployment process implemented here is really nothing out of ordinary -- I do not think debating the very basis of modern CI/CD process here is going to yield any fruitful results.

If you are unhappy with this change, I would like to ask you to come up with an alternative solution that does not require the "release manager(s)" to have access to the publishing token, as that is the main thing we are trying to avoid (ironically ...) for security reasons.

And, yes, I am willing to bet that GitHub secret storage "system" is going to do a better job keeping the publishing token secure than the release manager "humans," who have consistently proven to be vulnerable to the "human factors" ...

marc-hb: > GitHub can be hacked. PyPI can be hacked. My computer can be hacked. Your computer can be hacked.

What a ridiculous threat model.

> The deployment process implemented here is really nothing out of ordinary

While it's slowly getting better and better, "everybody does it" has been historically bad from a security perspective.

> that does not require the "release manager(s)" to have access to the publishing token, as that is the main thing we are trying to avoid (ironically ...) for security reasons.

This sounds very desirable, until you realize that the release manager will now push a release tag or something which will automatically build and publish a release. This release tag will be pushed thanks to... a Github secret used daily and permanently stored on a development system with high exposure? unlike the PyPI release publishing secret which was used infrequently and was much better protected?

This is the sort of questions that this PR must answer.

> If you are unhappy with this change, I would like to ask you to come up with an alternative solution

We already have two "solutions": 1) before, and 2) after this PR. The security of these two must be compared before looking at some 3rd or 4th alternative. Any change needs a rationale whether security-related or not. The bar is just a bit higher when it's security-related.

This does not require writing a security novel: providing some _reputable_ references should be enough. Such references should be easy to find considering how many times you said this is was "obvious" or "common practice" yet I haven't seen any posted yet. So I spent 5-10 minutes searching myself and I found this official documentation which looks like a somewhat better process already:

https://packaging.python.org/en/latest/guides/publishing-package-distribution-releases-using-github-actions-ci-cd-workflows/

> And, yes, I am willing to bet that GitHub secret storage "system" is going to do a better job keeping the publishing token secure than the release manager "humans," who have consistently proven to be vulnerable to the "human factors" ...

I bet this is true statistically, however applying statistics to a single sample is pointless. Yes, a lot of humans have been statistically not careful enough with important secrets. That does imply west maintainers have to do the same. "Everybody does it" is neither automatically good nor automatically bad.

Also, Github storage system is a much higher value target and - guess what - it has to involve some human factors at some points too.

None of this is simple, if it were then documents like https://docs.pypi.org/trusted-publishers/ would not exist.

marc-hb: In parallel I asked advice from a colleague and security expert involved with these topics, here's some of what he replied:


> add constraints around branches where it can be applied, require specific users to approve the workflow jobs to run, etc.

> Also, Github released actions to implement SLSA build level 2 attestations for build artifacts. It's simple to add support for that in the standard build workflows. The attestations could be verified manually by a developer or automatically by other workflows to ensure they were generated by the expected workflows with the expected conditions--e.g., generated by the expected workflow in response to a push of a release tag with a specific format, etc.

> Ultimately what they really want pypi to support is uploading release based on workflow OIDC attestation, because that way the workflow itself would authenticate to pypi and get a very short-term upload token. No long-term tokens to store and protect any longer.

He confirmed that his last, OIDC  point is exactly what I already found and just shared: 
https://docs.pypi.org/trusted-publishers/

I don't understand his 2nd point yet :-)

stephanosio: > > GitHub can be hacked. PyPI can be hacked. My computer can be hacked. Your computer can be hacked.
>
> What a ridiculous threat model.

My point is that, based on your reasoning, PyPI itself is a "development system with high exposure" and we should not be publishing packages to the interweb at all because it can be hacked, as with GitHub.

It is quite contradictory to claim that GitHub is dangerous because it is a high exposure target, when there exists PyPI that is just as exposed.

> > The deployment process implemented here is really nothing out of ordinary
> 
> While it's slowly getting better and better, "everybody does it" has been historically bad from a security perspective.

My point is that we need to be practical. Setting a high bar is good, until it starts hindering progress in the name of ideality.

> Yes, a lot of humans have been statistically not careful enough with important secrets. That does not imply west maintainers have to do the same. "Everybody does it" is neither automatically good nor automatically bad.

Maybe you and Pieter are good at this -- well, are you really sure your local development machine is clean and void of any backdoors and malicious code that can inject wonderful things into the package? or are you disciplined enough to use a fresh machine just for building and publishing packages?

What about the next person that comes along? Am I supposed to just manually hand out the publishing token to that person assuming they will be just as good?

Unfortunately, I have very little spare time to engage in these philosophical discussions ... I leave it up to the west maintainers to work out.

_Now back to the technical side (thanks for finally providing something we can work with)._

> So I spent 5-10 minutes searching myself and I found this official documentation which looks like a somewhat better process already:

"Trusted publisher" is indeed a new improved mechanism; though, it suffers from essentially the same problem as `${{ secrets.PYPI_API_TOKEN }}`.

> [Any workflow defined in a repository can request an OIDC token, with any audience, so long as it has the id-token: write permission.](https://docs.pypi.org/trusted-publishers/security-model/)

As with `${{ secrets.PYPI_API_TOKEN }}`, if a malicious workflow code is committed to the upstream repository, the publishing process can be ... well, you know what I mean. I will not go into too much details in a public thread for obvious reasons (ha, I said "obvious" again!) -- please feel free to ping me on Discord if you want more details.

As for the excessive "streamlining" of the publishing process, we can look into creating a [dedicated release "environment" in GitHub with deployment protection rules](https://docs.github.com/en/actions/managing-workflow-runs-and-deployments/managing-deployments/managing-environments-for-deployment#deployment-protection-rules) that can require additional approvals from the required reviewers before publishing happens.

marc-hb: > It is quite contradictory to claim that GitHub is dangerous because it is a high exposure target, when there exists PyPI that is just as exposed.

I disagree: two high exposures is more risk than one high and one low. Also, PyPI is more "downstream" which means less work to recover from a breach.


> Maybe you and Pieter are good at this -- well, are you really sure your local development machine is clean and void of any backdoors and malicious code that can inject wonderful things into the package?

I'm doing my best but I'm not perfect and obviously at the mercy of my OS vendor. But more relevant here: I never held any PyPI secret and never performed any official `west` release.

> My point is that we need to be practical. Setting a high bar is good, until it starts hindering progress in the name of ideality.

I'm not sure where this "high bar" comes from: so far I only tried to request a _description_ of this change justifying how it is more secure and why. 

> (thanks for finally providing something we can work with).

Sorry to repeat myself again: I've been mostly _questioning_ so far because I am not the one submitting a significant change to the security of the release process. I'm not sure what Marti used to do but in any case, any change (security-related or not) must have a bit of a rationale and a description that demonstrates _progress_. Progress "obvious" to you but not to me; sorry. Not until we have a basic description of the security of the release process.

> or are you disciplined enough to use a fresh machine just for building and publishing packages?

Some variants of this are actually part of the formal release process in some projects. It could also be here; why not!

This is exactly where build reproducibility shines BTW: automate the build 100% on Github (as done by this PR!), have all maintainers build the same release locally and compare the contents of the `.whl` ZIP files with diffoscope[*]. Less need for a "fresh machine" because it's very unlikely an attacker compromised all these systems at the same time. Then publish to PyPI... somehow :-)

[*] Don't Python wheels provide a content checksum to save the diffoscope step? They really should.

Speaking of which, some projects also have a process where each formal, signed release requires authenticated approval from _more than one_ maintainer. Can Github and/or PyPI do that?

I would also expect formal releases to request the second authentication factor - as always with sensitive actions. No second factor when giving Github automation a key straight to the PyPI kingdom?


> What about the next person that comes along? Am I supposed to just manually hand out the publishing token to that person assuming they will be just as good?

Ignoring the non-technical, "pure trust" part of that question, you could/should require that this person understands and follows some documented release policies similar to the ones being... "brainstormed" right now.

Again, these release security questions seem relatively "classic" so there must be good examples and references about this. You just reminded me of that bit: https://www.debian.org/devel/join/nm-step2

> if a malicious workflow code is committed to the upstream repository,

I'm no expert but that looks like an extra step for an attacker = extra risk of getting caught. Great! No?

> I will not go into too much details in a public thread for obvious reasons 

I understand not every single detail of the security process for making releases should be documented but there's a lot of room between that extreme and this current PR which has only one-line long commit messages...

Well, technically there this discussion now but so far it's been messy and will always be hard to find. Something better needs to be defined and written down, at least partially.

pdgendt: Follow-up update:

- A publisher has been added to the Pypi project
image

- A `release` environment has been created together with a [team to review releases](https://github.com/orgs/zephyrproject-rtos/teams/west)
image

Thanks @carlescufi for doing this!
Thanks @stephanosio @marc-hb for the effort and patience

stephanosio (approved):

pdgendt:
@marc-hb any objections or open comments here?

This is currently the last PR required to start the v1.3 release plan.

.github/workflows/release.yml
```diff
@@ -0,0 +1,34 @@
...
+    name: Release
+    needs: [package]
+    runs-on: ubuntu-latest
+
+    # Make the GH environment explicit
+    environment: release
+
+    # Mandatory for trusted publishing
+    permissions:
+      id-token: write
```
stephanosio, line 25: Today I migrated the windows-curses repo to use the "trusted publisher" mechanism and found that [`contents: write` is also required when triggered on "release published" event](https://github.com/zephyrproject-rtos/windows-curses/blob/61281da18178d49d505be2ec285feee9a776ccad/.github/workflows/release.yml#L19).

stephanosio (changes requested):

.github/workflows/release.yml
pdgendt: Updated

pdgendt: I think this was because of the release assets being uploaded using the `softprops/action-gh-release` action.

But you make a good point, I've added this step.

stephanosio (approved):

.github/workflows/release.yml
```diff
@@ -0,0 +1,40 @@
...
+      uses: actions/download-artifact@v4
+      with:
+        path: dist/
+
+    - name: Upload release assets
+      uses: softprops/action-gh-release@v2
+      with:
+        files: dist/*.whl
+
+    - name: Publish package to PyPI
```
marc-hb, line 39: This last step depends only on the "Download" step, correct? Not on the gh-release step. It wasn't clear to me until I checked the documentation so maybe state it in a comment?

Actually, wait: shouldn't `gh-release` be last, _after_ `pypi-publish`? What is its purpose, just a "backup" of the PyPI release, isn't it? In case of a failure/glitch, I think it's better to have the official PyPI release published without the backup rather than the opposite?

pdgendt: It's the same file but the gh-release step attaches it to the GitHub release, instead of uploading it manually.

marc-hb: I think you missed my point. I'll submit a follow-up PR.

stephanosio: >  In case of a failure/glitch, I think it's better to have the official PyPI release published without the backup rather than the opposite?

In case of a failure/glitch, you can retrigger the workflow to ensure that the full release workflow completes.

marc-hb:
I unfortunately don't have time to dive in but these seem relevant:

- https://lwn.net/Articles/1001215/  "Abusing Git branch names to compromise a PyPI package"

- https://blog.pypi.org/posts/2024-11-14-pypi-now-supports-digital-attestations/

pdgendt:
> I unfortunately don't have time to dive in but these seem relevant:
> 
> * [lwn.net/Articles/1001215](https://lwn.net/Articles/1001215/)  "Abusing Git branch names to compromise a PyPI package"
> * [blog.pypi.org/posts/2024-11-14-pypi-now-supports-digital-attestations](https://blog.pypi.org/posts/2024-11-14-pypi-now-supports-digital-attestations/)

I think the current workflow prevents anyone from abusing, or overriding whatever is on pypi?

PR #738, opened 2024-09-19
README.rst: Update installing from source section

pdgendt:
This section was outdated as we have moved to pyproject.toml.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -117,16 +117,11 @@ Installing from Source
 ~~~~~~~~~~~~~~~~~~~~~~
 
 You can create and install a wheel package to install west as well.
-The `wheel`_ Python package is required to do this. See "Installing Wheel"
-below if you need to do this.
 
 To build the west wheel file::
 
-  # macOS, Linux
-  python3 setup.py bdist_wheel
-
-  # Windows
-  py -3 setup.py bdist_wheel
+  pip3 install --upgrade build
+  python -m build
 
 This will create a file named ``dist/west-x.y.z-py3-none-any.whl``,
 where ``x.y.z`` is the current version in setup.py.
@@ -137,17 +132,3 @@ To install the wheel::
 
 You can ``pip3 uninstall west`` to remove this wheel before re-installing the
 version from PyPI, etc.
-
-Installing Wheel
-~~~~~~~~~~~~~~~~
-
-On macOS and Windows, you can install wheel with::
-
-  pip3 install wheel
-
-That also works on Linux, but you may want to install wheel from your
-system package manager instead -- e.g. if you installed pip from your
-system package manager. The wheel package is likely named something
-like ``python3-wheel`` in that case.
-
-.. _wheel: https://wheel.readthedocs.io/en/latest/


README.rst
```diff
@@ -117,16 +117,11 @@ Installing from Source
...
-below if you need to do this.
 
 To build the west wheel file::
 
-  # macOS, Linux
-  python3 setup.py bdist_wheel
-
-  # Windows
-  py -3 setup.py bdist_wheel
+  pip3 install --upgrade build
```
marc-hb, line 123: As I said before in https://github.com/zephyrproject-rtos/west/pull/715#discussion_r1757613300, I'm not a fan of using the random version of the day. This is not good for reproducibility.

marc-hb:
According to c399c012570c51ae806fdb5bb940d7ab1bd1c3d2, the new way still creates a wheel, but it does not require the "wheel" package anymore, correct? Packaging noob here sorry.

The same commit refers to a `pyproject-build` command but this commit uses `python -m build` instead, can you explain?

Despite the recent efforts, it feels like we still have a bit of copy/paste/diverge between the README.rst and MAINTAINERS.rst files... maybe it's by design but I'd like to understand why.

README.rst
pdgendt: This is done as described by python itself.

marc-hb: I think you mean https://packaging.python.org/en/latest/tutorials/packaging-projects/#generating-distribution-archives?

Any author of any package in any language will always tell you to run "the latest" because they're not interested in reports of bugs that they have already fixed. They're only interested in reports of new bugs. That bias is not necessarily aligned with packaging and release goals - notably build reproducibility.

Somewhat related and on-going discussion in #737

marc-hb: I built and inspected `west-1.2.99-py3-none-any.whl` on very different Operating Systems: Ubuntu 22, Arch Linux and macOS and I have very good build reproducibility news. Ubuntu 22 shipped with old `python-build` version 0.7.0 while the other two had the latest 1.2.2 version. Yet the content of the wheel packages were identical (except for the expected .zip metadata differences). So `python-build` does not seem to inject any timestamp of version-dependent metadata in the package, which means its version does not seem to affect build reproducibility.

pdgendt:
> According to [c399c01](https://github.com/zephyrproject-rtos/west/commit/c399c012570c51ae806fdb5bb940d7ab1bd1c3d2), the new way still creates a wheel, but it does not require the "wheel" package anymore, correct? Packaging noob here sorry.

It does create a wheel indeed. Not an expert, just trying to follow [documentation](https://packaging.python.org/en/latest/tutorials/packaging-projects/#generating-distribution-archives).

> The same commit refers to a `pyproject-build` command but this commit uses `python -m build` instead, can you explain?

Again, trying to update steps as per the python documentation, and `python -m build` seems to be preferred.

> Despite the recent efforts, it feels like we still have a bit of copy/paste/diverge between the README.rst and MAINTAINERS.rst files... maybe it's by design but I'd like to understand why.

I [updated the maintainers file](https://github.com/zephyrproject-rtos/west/pull/737/files#diff-37ee1f8dae662f435f88ca5896f6d3ac7605332b0004c262070761e72c55b84bR97) in the release update to also use the same build command. Do you feel that other places are diverged?

marc-hb (approved):

PR #740, opened 2024-09-20
manifest: Add submodule representation in as_dict

### LINKED ISSUE #739: west manifest freeze/resolve will lose submodules information
the project in raw west.yml
```yaml
  - name: test_repo
    path: manifest_repo/test_repo
    revision: main
    submodules:
    - path: modules/test_module
```
And run `west manifest --resolve -o freeze.ym`
freezed west.yml
```yaml
  - name: test_repo
    path: manifest_repo/test_repo
    revision: main
```
The submodules key gone.
Not sure whether this field was ignored in project's as_dict api: https://github.com/zephyrproject-rtos/west/blob/49f0a101df45ea735437bd388606a1219d9e76f5/src/west/manifest.py#L866

pdgendt:
Submodules were missing when calling --freeze or --resolve

Fixes #739

### DIFF
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -882,6 +882,15 @@ class Project:
                 _west_commands_maybe_delist(self.west_commands)
         if self.groups:
             ret['groups'] = self.groups
+        if isinstance(self.submodules, bool) and self.submodules:
+            ret['submodules'] = True
+        elif isinstance(self.submodules, list):
+            ret['submodules'] = []
+            for s in self.submodules:
+                obj: Dict = {'path': s.path}
+                if s.name:
+                    obj['name'] = s.name
+                ret['submodules'].append(obj)
         if self.userdata:
             ret['userdata'] = self.userdata
 
@@ -1143,7 +1152,7 @@ class ManifestProject(Project):
         # Pretending that this is a Project, even though it's not (#327)
         self.description: Optional[str] = None
         self.url: str = ''
-        self.submodules = False
+        self.submodules: SubmodulesType = False
         self.revision: str = 'HEAD'
         self.remote_name: str = ''
         self.clone_depth: Optional[int] = None
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -81,6 +81,11 @@ def _list_f(format):
     return ['list', '-f', format]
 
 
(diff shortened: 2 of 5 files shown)


tests/test_project.py
```diff
@@ -806,6 +807,28 @@ def test_update_submodules_list(repos_tmpdir):
...
+                    f'^    url: {re.escape(str(net_tools))}$',
+                    '^    revision: [a-f0-9]{40}$',
+                    '^    submodules:$',
+                    f'^    - path: {re.escape(str(kconfiglib_submodule))}$',
+                    '^      name: Kconfiglib$',
+                    '^  self:$',
+                    '^    path: mp$']
+
+    for eline_re, aline in zip(expected_res, actual):
+        assert re.match(eline_re, aline) is not None, (aline, eline_re)
```
marc-hb, line 830: At first, this (wrongly) felt overkill. But now I think this is actually really nice and pretty generic. Can you "promote" this duplicated `for` loop to a named function? Something like `match_multiline_regex()` or something.

The only potential "indirection" issue I foresee: make sure it's still possible to quickly tell which test failed when one does. If needed then maybe let the `match_multiline_regex()` pass a (sub)test name as a string. Hopefully not even needed.

Oh, and you did check that `pytest` does not have this already, right?

src/west/manifest.py
```diff
@@ -882,6 +882,15 @@ def as_dict(self) -> Dict:
                 _west_commands_maybe_delist(self.west_commands)
         if self.groups:
             ret['groups'] = self.groups
+        if isinstance(self.submodules, bool) and self.submodules:
+            ret['submodules'] = self.submodules
```
marc-hb, line 886:
```suggestion
            ret['submodules'] = True
```

marc-hb (commented):
Minor suggestions below, everything else looks good, thanks!

src/west/manifest.py
pdgendt: Updated

tests/test_project.py
pdgendt: I created a helper method for this, didn't find any `pytest` out-of-the-box helper.

marc-hb (approved):

PR #756, opened 2024-10-19
Add some NTFS checks

marc-hb:
NTFS: the gift that keeps on giving...

See October 2024 discussion in #558

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -16,6 +16,7 @@ import shlex
 import subprocess
 import sys
 import textwrap
+import time
 from time import perf_counter
 from urllib.parse import urlparse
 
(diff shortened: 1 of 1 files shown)


src/west/app/project.py
```diff
@@ -201,6 +206,12 @@ def do_add_parser(self, parser_adder):
                             MANIFEST_URL; .west is created next to "directory"
                             in this case, and manifest.path points at
                             "directory"''')
+        parser.add_argument('--rename-delay-hack', type=int,
```
pdgendt, line 209: While I agree this is a hack, does the parameter name need to be so... screamy?
```suggestion
        parser.add_argument('--rename-delay', type=int,
```

pdgendt (approved):

PR #760, opened 2024-10-23
tests: Wrap or remove deprecated functions

pdgendt:
Make deprecated calls explicit during testing or remove them if unnecessary.

This allows us to promote warnings to errors with `pytest`.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -5,11 +5,13 @@
 import configparser
 import os
 import pathlib
+from typing import Any, Optional
 import subprocess
 
 import pytest
 
 from west import configuration as config
+from west.util import PathType
 
 from conftest import cmd
 
@@ -30,8 +32,22 @@ def autouse_config_tmpdir(config_tmpdir):
 def cfg(f=ALL, topdir=None):
     # Load a fresh configuration object at the given level, and return it.
     cp = configparser.ConfigParser(allow_no_value=True)
-    config.read_config(configfile=f, config=cp, topdir=topdir)
+    # TODO: convert this mechanism without the global deprecated read_config
+    with pytest.deprecated_call():
+        config.read_config(configfile=f, config=cp, topdir=topdir)
     return cp
+
+def update_testcfg(section: str, key: str, value: Any,
+                   configfile: config.ConfigFile = LOCAL,
+                   topdir: Optional[PathType] = None) -> None:
+    c = config.Configuration(topdir)
+    c.set(option=f'{section}.{key}', value=value, configfile=configfile)
+
+def delete_testcfg(section: str, key: str,
+                   configfile: Optional[config.ConfigFile] = None,
+                   topdir: Optional[PathType] = None) -> None:
+    c = config.Configuration(topdir)
+    c.delete(option=f'{section}.{key}', configfile=configfile)
 
 def test_config_global():
     # Set a global config option via the command interface. Make sure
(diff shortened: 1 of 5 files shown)


tests/test_config.py
```diff
@@ -29,10 +28,22 @@ def autouse_config_tmpdir(config_tmpdir):
...
 def cfg(f=ALL, topdir=None):
     # Load a fresh configuration object at the given level, and return it.
-    cp = configparser.ConfigParser(allow_no_value=True)
-    config.read_config(configfile=f, config=cp, topdir=topdir)
+    cp = config.configparser.ConfigParser(allow_no_value=True)
+    cp.read(config._gather_configs(f, topdir), encoding='utf-8')
     return cp
 
+def update_config(*args, **kwargs):
+    # For testing purposes it's easiest to use the deprecated functions
```
marc-hb: Nit
```suggestion
    # For testing purposes it's easier to use the deprecated functions
```

tests/test_config.py
```diff
@@ -29,10 +28,22 @@ def autouse_config_tmpdir(config_tmpdir):
 
 def cfg(f=ALL, topdir=None):
     # Load a fresh configuration object at the given level, and return it.
-    cp = configparser.ConfigParser(allow_no_value=True)
-    config.read_config(configfile=f, config=cp, topdir=topdir)
+    cp = config.configparser.ConfigParser(allow_no_value=True)
```
marc-hb, line 33: Mostly not your fault but this commit is incredibly hard to review... first, this import rename is horrible because it hides both the origin and nature of `config`:
```python
from west import configuration as config
```
(don't change it now because it would have a massive ripple effect)

Then, I can't find `west.configuration.configparser` in the documentation. Is this is some "private" object (ab)used for "white box" testing? If yes please add a comment stating that. If not, sorry I missed what it is.

The icing on the cake is the fact that this commit replaces Python's built-in `configparser` with `west.configuration.configparser` which has the very same name while being a bit of a mystery...

I hate to ask but... could you slow down a little bit? Maybe move this to a separate commit. Either way, provide a bit more comments, commit message, etc. Thanks in advance!

tests/test_config.py
```diff
@@ -29,10 +28,22 @@ def autouse_config_tmpdir(config_tmpdir):
...
+    cp.read(config._gather_configs(f, topdir), encoding='utf-8')
     return cp
 
+def update_config(*args, **kwargs):
+    # For testing purposes it's easiest to use the deprecated functions
+    # but make these calls explicit.
+    with pytest.deprecated_call():
+        return config.update_config(*args, **kwargs)
```
marc-hb: Nice.

tests/test_project.py
```diff
@@ -1631,7 +1630,6 @@ def test_init_local_with_manifest_filename(repos_tmpdir):
     # success
     cmd(['init', '--mf', 'project.yml', '-l', zephyr_install_dir])
     workspace.chdir()
-    config.read_config()
```
marc-hb, line 1634: Should this be a separate commit?

tox.ini
```diff
@@ -37,6 +37,6 @@ setenv =
     # For instance: ./.tox/py3/tmp/
     TOXTEMPDIR={envtmpdir}
 commands =
-  python -m pytest --cov-report=html --cov=west {posargs:tests} --basetemp='{envtmpdir}/pytest with space/'
+  python -m pytest -W error --cov-report=html --cov=west {posargs:tests} --basetemp='{envtmpdir}/pytest with space/'
```
marc-hb, line 40: Nice, but could there be a convenient way/setting to temporarily turn this off? I mean without `git diff` and `git status` pollution.

When hacking, there is often a intermediate stage where you really want the tests to pass on work in progress that still has warnings.

tests/test_config.py
```diff
@@ -29,10 +28,22 @@ def autouse_config_tmpdir(config_tmpdir):
...
+
+def delete_config(*args, **kwargs):
```
marc-hb: Warning: this review comment will be hard to find because that line  is changed in a later commit in the PR and Github doesn't like the multi-commit, force-push workflow.

There is already an API named `delete_config()`, so please use a different name for internal symbols. Example:
```suggestion
def delete_testcfg(*args, **kwargs):
```

Same for `update_config` above.

Sincere condolences for the ripple effect on consecutive commits :-(

Generally speaking, please choose less "API-looking" names for internal symbols. Pretty much everything here is called `something_config` and it's incredibly hard to keep track of what is: 1) Builtin Python (e.g. `configparser`) 2) deprecated west API 3) current west API 4) private west objects abused for "white box" testing 5) local/private test symbols,....

It's never going to be perfect and please do not cause massive git churn but brand new symbols is where progress can and should be made.

This is a cleanup PR so it should bring more clarity.

marc-hb (changes requested):

tests/test_config.py
pdgendt: I was contemplating the same things, but tried to touch it "minimally" I'll try to clean this up so that it makes sense what is builtin and what not.

tests/test_config.py
pdgendt: Yeah, I tried to do it in some reasonable steps, but I think it can make it harder to follow.
I'll come back to this PR, but there's no urgency here, as I didn't intend on getting into 1.3.

tests/test_config.py
pdgendt: I've reverted to the original `configparser`.

tests/test_project.py
pdgendt: Moved to a separate commit.

tests/test_config.py
pdgendt: Renamed as suggested, which is clearer.

tox.ini
pdgendt: I've removed it from the `tox.ini` file and added it to the CI run.

tests/test_config.py
```diff
@@ -30,9 +32,21 @@ def autouse_config_tmpdir(config_tmpdir):
 def cfg(f=ALL, topdir=None):
     # Load a fresh configuration object at the given level, and return it.
     cp = configparser.ConfigParser(allow_no_value=True)
-    config.read_config(configfile=f, config=cp, topdir=topdir)
+    cp.read(config._gather_configs(f, topdir), encoding='utf-8')
```
marc-hb, line 35: After searching the code, I found that `_gather_configs()` is just an internal implementation detail of `read_config()` and used only by that deprecated function. This means `_gather_configs()` is effectively deprecated too. So I don't see what this particular line change brings except some confusion, sorry.

I think you should either "upgrade" this function to the newer API if possible, or not change it at all yet if not possible and/or too much work.

marc-hb (commented):
Looks great except for one confusing, one-line change.

tests/test_config.py
pdgendt: I've wrapped the call in `with pytest.deprecated_call()` as it requires a lot more work in this case. Also added a `TODO` comment.

marc-hb (approved):

PR #763, opened 2024-11-02
project: Fix and check pyupgrade issues

pdgendt:
Depending on the minimum supported python version (3.9 in our case), `pyupgrade` will report deprecated python features.
Extend `ruff` with `pyupgrade` checks.

Depends on #762

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -44,3 +44,11 @@ include-package-data = true
 [tool.setuptools.packages.find]
 where = ["src"]
 namespaces = false
+
+[tool.ruff.lint]
+extend-select = [
+  "UP",  # pyupgrade
+]
+ignore = [
+  "UP027", # deprecated pyupgrade rule
+]
--- a/src/west/app/main.py
+++ b/src/west/app/main.py
@@ -28,7 +28,6 @@ import tempfile
 import textwrap
 import traceback
 from typing import NamedTuple, Optional
-from typing import List as ListType
 
 from west import log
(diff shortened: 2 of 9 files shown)


tests/conftest.py
```diff
@@ -200,9 +200,9 @@ def repos_tmpdir(tmpdir, _session_repos):
         path: zephyr
 
     '''
-    kconfiglib, tagged_repo, net_tools, zephyr = [
+    kconfiglib, tagged_repo, net_tools, zephyr = (
         os.path.join(_session_repos, x) for x in
-        ['Kconfiglib', 'tagged_repo', 'net-tools', 'zephyr']]
+        ['Kconfiglib', 'tagged_repo', 'net-tools', 'zephyr'])
```
marc-hb, line 205: I'm curious what was the warning here...

marc-hb (commented):
The cleanup commit is a bit too big for my taste hence a bit painful to check... could you split it in say 2 or 3? Maybe one commit with all the type changes and another commit for the rest...

tests/conftest.py
pdgendt: It was actually a [deprecated rule](https://docs.astral.sh/ruff/rules/unpacked-list-comprehension/) itself, which was still "automagically" fixed with `ruff check --fix`.

I've ignored this rule.

carlescufi (approved):
LGTM, thanks!

marc-hb (approved):

PR #764, opened 2024-11-04
project: Extend ruff check with pycodestyle, flake8-bugbear, pyflakes

pdgendt:
Enable pycodestyle, flake8-bugbear, pyflakes rules when running ruff check.

### DIFF
--- a/src/west/app/main.py
+++ b/src/west/app/main.py
@@ -1019,7 +1019,7 @@ class WestArgumentParser(argparse.ArgumentParser):
                 #
                 # This has its own wrinkle: we can't let a failed
                 # import break the built-in commands.
-                for path, specs in self.west_app.extension_groups.items():
+                for _path, specs in self.west_app.extension_groups.items():
                     # This may occur in case a project defines commands already
                     # defined, in which case it has been filtered out.
                     if not specs:
--- a/src/west/app/main.py
+++ b/src/west/app/main.py
@@ -920,7 +920,7 @@ class Alias(WestCommand):
         return parser
 
     def do_run(self, args, ignored):
-        assert False
+        raise AssertionError("Alias command can't run directly")
 
 class WestHelpAction(argparse.Action):
 
--- a/src/west/app/main.py
+++ b/src/west/app/main.py
@@ -409,17 +409,17 @@ class WestApp:
             for spec in specs:
                 if spec.name in self.builtins:
                     self.queued_io.append(
-                        lambda cmd: cmd.wrn(
-                            f'ignoring project {spec.project.name} '
-                            f'extension command "{spec.name}"; '
+                        lambda cmd, spec_const=spec: cmd.wrn(
+                            f'ignoring project {spec_const.project.name} '
+                            f'extension command "{spec_const.name}"; '
                             'this is a built in command'))
                     continue
                 if spec.name in extension_names:
                     self.queued_io.append(
-                        lambda cmd: cmd.wrn(
-                            f'ignoring project {spec.project.name} '
-                            f'extension command "{spec.name}"; '
-                            f'command "{spec.name}" is '
+                        lambda cmd, spec_const=spec: cmd.wrn(
+                            f'ignoring project {spec_const.project.name} '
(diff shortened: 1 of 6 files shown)


src/west/app/main.py
```diff
@@ -384,14 +384,14 @@ def load_extension_specs(self):
...
-                        lambda cmd: cmd.wrn(
+                        lambda cmd, spec=spec: cmd.wrn(
                             f'ignoring project {spec.project.name} '
                             f'extension command "{spec.name}"; '
                             'this is a built in command'))
                     continue
                 if spec.name in extension_names:
                     self.queued_io.append(
-                        lambda cmd: cmd.wrn(
+                        lambda cmd, spec=spec: cmd.wrn(
```
marc-hb, line 394: Wait, this one is not just silencing a warning: this looks like an actual bug! So it needs a different commit message.

Unsurprisingly, no one ever complained because this is error handling code which is rarely used and never tested.

So I think the bug should go like this: when west is "ignoring" a project, the error message will always refer to the last `specs[len(spec)-1]` no matter which `spec` was actually involved.

How would one reproduce this wrong error message? I mean interactively at least. If that's complicated then feel free to drop this commit from this cleanup PR so it can be merged quickly.

marc-hb (changes requested):
Looks good except for the late binding bug fix.

src/west/app/main.py
marc-hb: I trust the `spec=spec` fix suggested by https://docs.astral.sh/ruff/rules/function-uses-loop-variable/ is functionally correct, however it's pretty horrible style because it completely obscures the very subtle issue. The two should be different, something like this:

```python
            for spec in specs:
                if spec.name in self.builtins:
                    self.queued_io.append(
                        lambda cmd, spec_const=spec: cmd.wrn(
                            f'ignoring project {spec_const.project.name} '
                            f'extension command "{spec_const.name}"; '
                            'this is a built in command'))
```

pdgendt: The issue won't occur once the check has been enabled, I wanted to make as little changes as possible to also preserve git history.

But I can make the changes.

pdgendt:
Updated the commit that fixed the actual bug with a more descriptive message.

marc-hb (commented):
I'm still disappointed by the commit message of the bug fix, sorry.

I'm aware no one ever cares about error handling and error messages (at least not until it hurts them personally) but I think someone experiencing this bug should be able to look at the git log of a newer west version and at least think "Mmmm... could that commit be a fix for this problem I observe?" I don't think that would happen right now.

Did you try to reproduce this bug? If that takes too long, set aside that commit for now so we can merge all the rest?


https://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git/commit/?id=a2fe16b9d878a101b6767
> "A patch subject line should describe the change, not the tool that found it"

pdgendt:
> I'm still disappointed by the commit message of the bug fix, sorry.
> 
> I'm aware no one ever cares about error handling and error messages (at least not until it hurts them personally) but I think someone experiencing this bug should be able to look at the git log of a newer west version and at least think "Mmmm... could this be a fix for this problem I observe?" I don't think that would happen right now.

I'll try to describe it better or even add a testcase.

> Did you try to reproduce this bug? If that takes too long, set aside that commit for now so we can merge all the rest?

Setting the commit aside, would mean we can't enable `flake8-bugbear` checks.

marc-hb:
> I'll try to describe it better or even add a testcase.

Thanks! An _automated_ testcase could be overkill, I'm certainly not asking for that.

> Setting the commit aside, would mean we can't enable flake8-bugbear checks.

There's quite a lot more in this PR.

There might also be a way to tell `flake8-bugbear` to temporarily ignore this line or type of warning.

pdgendt:
I've updated the commit message again. I'm sorry to put the ball back in your camp, but what would be an acceptable message? 

I'm not a fluent writer for these things, and I want to do it right, but it should be easy to make progress here.

marc-hb (approved):
Perfect thanks. I just wanted the effect of the bug to be described from a user perspective.

PR #765, opened 2024-11-04
project: Format and sort imports

pdgendt:
Apply `isort --profile black` on files in `src/` and `tests/`

### DIFF
--- a/src/west/app/main.py
+++ b/src/west/app/main.py
@@ -12,35 +12,60 @@ Nothing in here is public API.
 '''
 
 import argparse
-from collections import OrderedDict
-import colorama
-from io import StringIO
 import logging
 import os
-from pathlib import Path, PurePath
 import platform
 import shlex
 import shutil
 import signal
 import sys
-from subprocess import CalledProcessError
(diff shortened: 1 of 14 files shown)


src/west/app/main.py
```diff
@@ -12,36 +12,61 @@
...
+    Diff,
+    ForAll,
+    Grep,
+    Init,
+    List,
+    ManifestCommand,
+    SelfUpdate,
+    Status,
+    Topdir,
+    Update,
```
marc-hb, line 47: 1 per line is a significant waste of screen estate but no big deal...

marc-hb (approved):
As discussed on Discord, I'm generally not a fan of these mass reformats because they break `git cherry-pick/rebase/merge/revert/blame/...` and they tend to do so _repeatedly_ when the formatting tools gets upgraded or the `west/ruff.config` file gets "optimized". Moreover, this does not seem to add a lot of value.

HOWEVER this one looks small and harmful enough; a git conflict in `imports` should never  be hard to fix.

PR #766, opened 2024-11-04
ci: Add format check workflow

pdgendt:
Add a CI workflow to report formatting issues on changed files. This is to gradually update files in the repository to be conform with PEP8 formatting.


.github/workflows/format.yml
```diff
@@ -0,0 +1,17 @@
...
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v4
+
+    - uses: astral-sh/ruff-action@v1
+      continue-on-error: true
+      with:
+        args: "format --check"
+        changed-files: "true"
```
marc-hb, line 17: Nice.

I checked the source code and this relies on https://github.com/tj-actions/changed-files

Years ago, I could not find anything like it _and_ simple and reliable so hopefully this one does the job.

I wrote this `git` script at the time: https://github.com/thesofproject/sof-test/commit/336094b869979bb6fab47158787a6802f4bddad6

marc-hb (commented):
Can this check be easily overriden when needed? For instance we don't want one-line fix PRs to be drowned in whitespace changes.

pdgendt:
Went a step further, new files are always fully checked, changed chunks now also need to formatted.

Experimenting shown here: https://github.com/pdgendt/west/actions/workflows/format.yml

marc-hb:
Checking only modified chunks sounds great in theory but I'm concerned the code of the Github workflow is getting too complicated and hard to maintain. It's not like `west` has had an oversupply of maintainers :-/

As long as this check does not block merges, I think checking only new and modified files (the entire file each time) is a good enough trade-off. For sure that trade-off has been working well for years with `pylint` and `shellcheck` in https://github.com/thesofproject/sof-test/commit/336094b869979bb6fab47158787a6802f4bddad6

pdgendt:
> Checking only modified **chunks** sounds great in theory but I'm concerned that the code of the Github workflow is getting too complicated and hard to maintain. It's not like `west` has had an oversupply of maintainers :-/
> 
> As long as this check does not block merges, I think checking only new and modified files (the entire file each time) is a good enough trade-off. For sure that trade-off has been working well for years with `pylint` and `shellcheck` in [thesofproject/sof-test@336094b](https://github.com/thesofproject/sof-test/commit/336094b869979bb6fab47158787a6802f4bddad6)

Updated the PR accordingly:

- Ignore deleted python files
- Run `ruff format --check --diff` on entire files
- Add `continue-on-error: true` to make the workflow pass if the format job fails (making it non-blocking)
- Add a warning annotation instead of an error

.github/workflows/format.yml
```diff
@@ -0,0 +1,56 @@
...
+
+on:
+  pull_request:
+    branches:
+      - main
+    paths:
+      - '**.py'
+
+jobs:
+  generate:
```
marc-hb, line 11: I think this name is part of the "user interface" so can it be a bit longer and more specific? `find-changed-files` or better...

.github/workflows/format.yml
```diff
@@ -0,0 +1,56 @@
...
+      files: ${{ steps.git-diff-filter.outputs.files }}
+    name: Detect added and changed files
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      - uses: GrantBirki/git-diff-action@v2
+        id: git-diff-action
```
marc-hb, line 23: Can you make this name (slightly) different from the action name? Right now it's confusing to tell which object is being referred to.

.github/workflows/format.yml
```diff
@@ -0,0 +1,56 @@
...
+        with:
+          base_branch: origin/main
+          search_path: '**.py'
+          # Ignore deleted files
+          git_options: '--no-color --diff-filter=d'
+
+      - name: Filter json diff
```
marc-hb, line 30: Filter what? It should be possible to have a clue without reading the code.

marc-hb: Doesn't `file_output_only` do this? It's "highly recommended"

.github/workflows/format.yml
```diff
@@ -0,0 +1,56 @@
...
+        id: git-diff-filter
+        env:
+          JSON_DIFF: ${{ steps.git-diff-action.outputs.json-diff }}
+        run: |
+          files=$(echo $JSON_DIFF | jq -c -r '[.files[] | {path: .path}]')
+          echo "files=$files" >> $GITHUB_OUTPUT
+
+  check-files:
```
marc-hb, line 38:
```suggestion
  ruff-format:
```

.github/workflows/format.yml
```diff
@@ -0,0 +1,56 @@
...
+    needs: generate
+    if: ${{ needs.generate.outputs.files != '[]' }}
+    runs-on: ubuntu-latest
+    # Allow the workflow run to pass when this job fails
+    continue-on-error: true
+    strategy:
+      matrix:
+        files: ${{ fromJSON(needs.generate.outputs.files) }}
```
marc-hb, line 46: Wow!

.github/workflows/format.yml
```diff
@@ -0,0 +1,56 @@
...
+        run: |
+          files=$(echo $JSON_DIFF | jq -c -r '[.files[] | {path: .path}]')
+          echo "files=$files" >> $GITHUB_OUTPUT
+
+  check-files:
+    needs: generate
+    if: ${{ needs.generate.outputs.files != '[]' }}
+    runs-on: ubuntu-latest
+    # Allow the workflow run to pass when this job fails
+    continue-on-error: true
```
marc-hb, line 47: Does this just change red to green? If yes then I don't like it, sorry for the confusion. 

- https://github.com/orgs/community/discussions/15452

If formatting fails then the check should be red to draw attention, then maintainers should be able to merge anyway.

.github/workflows/format.yml
```diff
@@ -0,0 +1,56 @@
...
+        with:
+          base_branch: origin/main
+          search_path: '**.py'
+          # Ignore deleted files
+          git_options: '--no-color --diff-filter=d'
+
+      - name: Filter json diff
+        id: git-diff-filter
+        env:
+          JSON_DIFF: ${{ steps.git-diff-action.outputs.json-diff }}
```
marc-hb, line 33: Using `json_diff_file_output` is "highly recommended".

.github/workflows/format.yml
```diff
@@ -0,0 +1,56 @@
...
+        with:
+          fetch-depth: 0
+
+      - uses: GrantBirki/git-diff-action@v2
+        id: git-diff-action
+        with:
+          base_branch: origin/main
+          search_path: '**.py'
+          # Ignore deleted files
+          git_options: '--no-color --diff-filter=d'
```
marc-hb, line 28: Are you really sure `--no-color` is required? It's very surprising not get this by default in such a context...

marc-hb (dismissed):

.github/workflows/format.yml
pdgendt: Under the hood `git-diff-action` has this as a default. It doesn't harm IMO.

.github/workflows/format.yml
pdgendt: No, it's still red, but won't block merging the PR.

.github/workflows/format.yml
pdgendt: Updated

.github/workflows/format.yml
pdgendt: Updated

.github/workflows/format.yml
pdgendt: Used the recommended json file output.

.github/workflows/format.yml
pdgendt: Updated.

.github/workflows/format.yml
pdgendt: Updated

.github/workflows/format.yml
marc-hb: I'm still not sure `continue-on-error` is what you need here. Unfortunately, it's very confusing and barely documented - just one cryptic line https://github.com/actions/runner/issues/2347

Are you looking for `fail-fast: false` instead by any chance? Also: much better documented:

https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast

marc-hb: Whatever happens with `continue-on-error`, from past experience with `fail-fast: false` I'm 99% sure you need it here.

pdgendt: `continue-on-error` during my testing showed that jobs failed and showed up as red. But the workflow was still successful.
I've added the `fail-fast: false`.

marc-hb: I'm still not sure about `continue-on-error` but this can easily be changed later. Also, I know first hand it's incredibly hard to test Github Actions before they are merged.

marc-hb (approved):

.github/workflows/format.yml
pdgendt: Tested this on PR #768, seems to be working.

PR #768, opened 2024-11-14
app: config: Add support for appending to the config string

carlescufi:
In some cases, and specifically in the manifest.group-filter and manifest.project-filter options, it is sometimes useful to be able to append to a value instead of replacing it completely.

For example, assuming one wants to add to an existing group filter, without this patch the user needs to do:

(assuming the group filter is currently +unstable,-optional, and the
user wants to add +extras).
$ west config manifest.group-filter
$ west config manifest.group-filter +unstable,-optional,+extras

With this patch instead:

$ west config -a manifest.group-filter ,+extras

### DIFF
--- a/src/west/app/config.py
+++ b/src/west/app/config.py
@@ -88,12 +88,17 @@ class Config(WestCommand):
             description=self.description,
             epilog=CONFIG_EPILOG)
 
-        parser.add_argument('-l', '--list', action='store_true',
-                            help='list all options and their values')
-        parser.add_argument('-d', '--delete', action='store_true',
-                            help='delete an option in one config file')
-        parser.add_argument('-D', '--delete-all', action='store_true',
-                            help="delete an option everywhere it's set")
+        group = parser.add_argument_group(
+            "action to perform (give at most one)"
+        ).add_mutually_exclusive_group()
+
+
+        group.add_argument('-l', '--list', action='store_true',
+                           help='list all options and their values')
+        group.add_argument('-d', '--delete', action='store_true',
+                           help='delete an option in one config file')
+        group.add_argument('-D', '--delete-all', action='store_true',
+                           help="delete an option everywhere it's set")
 
         group = parser.add_argument_group(
             "configuration file to use (give at most one)"
@@ -121,13 +126,9 @@ class Config(WestCommand):
         if args.list:
             if args.name:
                 self.parser.error('-l cannot be combined with name argument')
-            elif delete:
-                self.parser.error('-l cannot be combined with -d or -D')
         elif not args.name:
             self.parser.error('missing argument name '
                               '(to list all options and values, use -l)')
-        elif args.delete and args.delete_all:
-            self.parser.error('-d cannot be combined with -D')
 
         if args.list:
             self.list(args)
--- a/src/west/app/config.py
+++ b/src/west/app/config.py
@@ -48,6 +48,14 @@ To get a value for <name>, type:
 To set a value for <name>, type:
     west config <name> <value>
 
+To append to a value for <name>, type:
+    west config -a <name> <value>
+A value must exist in the selected configuration file in order to be able
+to append to it. The existing value can be empty.
+Examples:
+    west config -a build.cmake-args -- " -DEXTRA_CFLAGS='-Wextra -g0' -DFOO=BAR"
+    west config -a manifest.group-filter ,+optional
+
 To list all options and their values:
     west config -l
 
@@ -64,7 +72,7 @@ To delete <name> everywhere it's set, including the system file:
 
 CONFIG_EPILOG = '''\
 If the configuration file to use is not set, reads use all three in
-precedence order, and writes use the local file.'''
+precedence order, and writes (including appends) use the local file.'''
 
 ALL = ConfigFile.ALL
 SYSTEM = ConfigFile.SYSTEM
@@ -92,13 +100,14 @@ class Config(WestCommand):
             "action to perform (give at most one)"
         ).add_mutually_exclusive_group()
 
-
         group.add_argument('-l', '--list', action='store_true',
                            help='list all options and their values')
         group.add_argument('-d', '--delete', action='store_true',
                            help='delete an option in one config file')
         group.add_argument('-D', '--delete-all', action='store_true',
                            help="delete an option everywhere it's set")
+        group.add_argument('-a', '--append', action='store_true',
+                           help='append to an existing value')
 
         group = parser.add_argument_group(
             "configuration file to use (give at most one)"
@@ -129,6 +138,9 @@ class Config(WestCommand):
         elif not args.name:
             self.parser.error('missing argument name '
                               '(to list all options and values, use -l)')
+        elif args.append:
+            if args.value is None:
+                self.parser.error('-a requires both name and value')
 
         if args.list:
             self.list(args)
@@ -136,6 +148,8 @@ class Config(WestCommand):
             self.delete(args)
         elif args.value is None:
             self.read(args)
+        elif args.append:
+            self.append(args)
         else:
             self.write(args)
 
@@ -179,6 +193,16 @@ class Config(WestCommand):
         else:
             self.dbg(f'{args.name} is unset')
             raise CommandError(returncode=1)
+
+    def append(self, args):
+        self.check_config(args.name)
+        where = args.configfile or LOCAL
+        value = self.config.get(args.name, configfile=where)
+        if value is None:
+            self.die(f'option {args.name} not found in the {where.name.lower()} '
+                     'configuration file')
+        args.value = value + args.value
+        self.write(args)
 
     def write(self, args):
         self.check_config(args.name)
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -229,6 +229,45 @@ def test_local_creation_with_topdir():
     assert 'pytest' not in cfg(f=GLOBAL)
     assert cfg(f=LOCAL, topdir=str(topdir))['pytest']['key'] == 'val'
 
+def test_append():
+    update_testcfg('pytest', 'key', 'system', configfile=SYSTEM)
+    update_testcfg('pytest', 'key', 'global', configfile=GLOBAL)
+    update_testcfg('pytest', 'key', 'local', configfile=LOCAL)
+    # Appending with no configfile specified should modify the local one
+    cmd('config -a pytest.key ,bar')
+
+    # Only the local one will be modified
+    assert cfg(f=SYSTEM)['pytest']['key'] == 'system'
+    assert cfg(f=GLOBAL)['pytest']['key'] == 'global'
+    assert cfg(f=LOCAL)['pytest']['key'] == 'local,bar'
+
+    # Test a more complex one, and at a particular configfile level
+    update_testcfg('build', 'cmake-args', '-DCONF_FILE=foo.conf', configfile=GLOBAL)
+    assert cfg(f=GLOBAL)['build']['cmake-args'] == '-DCONF_FILE=foo.conf'
+
+    # Use a list instead of a string to avoid one level of nested quoting
+    cmd(['config', '--global', '-a', 'build.cmake-args', '--',
+         ' -DEXTRA_CFLAGS=\'-Wextra -g0\' -DFOO=BAR'])
+
+    assert cfg(f=GLOBAL)['build']['cmake-args'] == \
+        '-DCONF_FILE=foo.conf -DEXTRA_CFLAGS=\'-Wextra -g0\' -DFOO=BAR'
+
+def test_append_novalue():
+    with pytest.raises(subprocess.CalledProcessError) as exc_info:
+        cmd('config -a pytest.foo', stderr=subprocess.STDOUT)
+    # Get the output into a variable to simplify pytest error messages
+    err_msg = exc_info.value.output.decode("utf-8")
+    assert '-a requires both name and value' in err_msg
+
+def test_append_notfound():
+    update_testcfg('pytest', 'key', 'val', configfile=LOCAL)
+    with pytest.raises(subprocess.CalledProcessError) as exc_info:
+        cmd('config -a pytest.foo bar', stderr=subprocess.STDOUT)
+    # Get the output into a variable to simplify pytest error messages
+    err_msg = exc_info.value.output.decode("utf-8")
+    assert 'option pytest.foo not found in the local configuration file' in err_msg
+
+
 def test_delete_basic():
     # Basic deletion test: write local, verify global and system deletions
     # don't work, then delete local does work.


src/west/app/config.py
```diff
@@ -194,8 +206,14 @@ def read(self, args):
     def write(self, args):
         self.check_config(args.name)
         what = args.configfile or LOCAL
+        if args.append:
+            value = self.config.get(args.name,
+                                    configfile=args.configfile or ALL) or ''
```
pdgendt, line 211: Does the `ALL` make sense? If the value exists at global or system level, you'd create a new local copy with the appended data.

carlescufi: I thought the same thing, but my reasoning was along these lines: if the user reads the value without any additional parameters, it'll get the global or system value if there's no local one. Then, if the user intends to append, perhaps they expect to append to whatever the currently active value is when reading? I agree it's not obvious.

pdgendt:
I do have a concern about the appending if the value doesn't exist yet.

Something like:
```shell
$ west config -a manifest.group-filter ,+extras
```

Could result in an error because of the leading comma.

carlescufi:
> I do have a concern about the appending if the value doesn't exist yet.

I also thought about that, but did not want to overcomplicate the code. But yes, let me do that as well.

src/west/app/config.py
pdgendt: Maybe appending should require an explicit target then? This hidden logic would confuse users too much IMO.

src/west/app/config.py
```diff
@@ -182,20 +194,27 @@ def check_config(self, option):
...
-            self.dbg(f'{args.name} is unset')
-            raise CommandError(returncode=1)
+        value = self._read(args.name, args.configfile or ALL)
+        self.inf(value)
 
     def write(self, args):
         self.check_config(args.name)
         what = args.configfile or LOCAL
+        if args.append:
+            value = self._read(args.name, args.configfile or ALL) or ''
```
pdgendt, line 212: This would raise an exception if not found, is this desired?

src/west/app/config.py
```diff
@@ -182,20 +194,27 @@ def check_config(self, option):
...
-        else:
-            self.dbg(f'{args.name} is unset')
-            raise CommandError(returncode=1)
+        value = self._read(args.name, args.configfile or ALL)
+        self.inf(value)
 
     def write(self, args):
         self.check_config(args.name)
         what = args.configfile or LOCAL
+        if args.append:
```
pdgendt, line 211: I'd prefer to create a separate `append` function instead of putting all the logic in `write`.

src/west/app/config.py
carlescufi: Yes, it is caught and you get an exit code (though no error message)

src/west/app/config.py
carlescufi: Done

pdgendt: You need to call the function 😉

src/west/app/config.py
carlescufi: Requiring an explicit target beats the point of this option, IMO, which is to simplify the usage of the write operation. Instead I'd rather follow the logic of write() for append() as well.

carlescufi: @pdgendt I changed it now so it defaults to `LOCAL`, just like `write`.

src/west/app/config.py
```diff
@@ -182,14 +196,23 @@ def check_config(self, option):
...
-            self.dbg(f'{args.name} is unset')
-            raise CommandError(returncode=1)
+        value = self._read(args.name, args.configfile or ALL)
+        self.inf(value)
+
+    def append(self, args):
+        self.check_config(args.name)
+        value = self._read(args.name, args.configfile or LOCAL) or ''
+        args.value = value + args.value
+        self.write(args)
```
pdgendt, line 215: The tricky bit here, is that if the name doesn't exist, the value isn't created and nothing is printed. The user needs to check the return value

```shell
$ west config -a non.existing whatever
$ echo $?
1
```

Together with the scenario you describe, where a separator is needed for the particular option, maybe this needs something more.
Perhaps the `-a` argument could be passed a separator value?

carlescufi: Sure, I get your point. But aren't those 2 separate issues?

1. Report the "unable to find value to append to" error to the user in a more visible way (I can add that very easily)
2. Appending without a separator may confuse users

src/west/app/config.py
```diff
@@ -135,18 +140,27 @@ def do_run(self, args, user_args):
...
                               '(to list all options and values, use -l)')
-        elif args.delete and args.delete_all:
-            self.parser.error('-d cannot be combined with -D')
+        elif args.delete:
+            if args.delete_all:
+                self.parser.error('-d cannot be combined with -D')
+            elif args.append:
+                self.parser.error('-d cannot be combined with -a')
+        elif args.value is None and args.append:
+            self.parser.error('-a requires a value')
```
pdgendt, line 154: These checks are becoming a mess, can we use a [mutually exclusive group](https://docs.python.org/3/library/argparse.html#argparse.ArgumentParser.add_mutually_exclusive_group) instead?

pdgendt (commented):
LGTM, can you add tests?

src/west/app/config.py
```diff
@@ -48,6 +48,11 @@
 To set a value for <name>, type:
     west config <name> <value>
 
+To append to a value for <name>, type:
+    west config -a <name> <value>
+A value must exist in the selected configuration file in order to be able
+to append to it.
```
marc-hb, line 54: Can you please add a short example? One with whitespace, quoting and `shlex.split()` to warn the user about the usual whitespace and quoting perils. Something like `west config -a build.cmake-args -- "-DCMAKE_EXPORT_COMPILE_COMMANDS=ON -DCMAKE_VERBOSE_MAKEFILE=ON"` (UNTESTED)

See https://docs.zephyrproject.org/latest/develop/west/build-flash-debug.html#permanent-cmake-arguments and 0186eade513a

marc-hb (commented):
> > LGTM, can you add tests?

> yep, on it!

I think these won't involve `shlex.split()` because Zephyr will not be involved but please use some whitespace for more "stress".

src/west/app/config.py
```diff
@@ -121,20 +132,20 @@ def do_run(self, args, user_args):
...
                 self.parser.error('-l cannot be combined with name argument')
-            elif delete:
-                self.parser.error('-l cannot be combined with -d or -D')
         elif not args.name:
             self.parser.error('missing argument name '
                               '(to list all options and values, use -l)')
-        elif args.delete and args.delete_all:
-            self.parser.error('-d cannot be combined with -D')
+        elif args.value is None and args.append:
+            self.parser.error('-a requires a value')
```
marc-hb, line 139:
```suggestion
            self.parser.error('-a requires both name and value')
```

src/west/app/config.py
```diff
@@ -179,6 +190,16 @@ def read(self, args):
             self.dbg(f'{args.name} is unset')
             raise CommandError(returncode=1)
 
+    def append(self, args):
+        self.check_config(args.name)
+        what = args.configfile or LOCAL
```
marc-hb, line 195:
```suggestion
        where = args.configfile or LOCAL
```

src/west/app/config.py
marc-hb, line 54:
```suggestion
to append to it. The existing value can be empty.
```

src/west/app/config.py
```diff
@@ -64,7 +69,7 @@
 
 CONFIG_EPILOG = '''\
 If the configuration file to use is not set, reads use all three in
-precedence order, and writes use the local file.'''
+precedence order, and writes (including appends) use the local file.'''
```
marc-hb, line 72: I'm afraid you just discovered a documentation bug (or worse). The `west config` works just fine outside of any workspace - when there is no "local file". What happens then?

src/west/app/config.py
```diff
@@ -88,12 +93,18 @@ def do_add_parser(self, parser_adder):
...
+        ).add_mutually_exclusive_group()
+
+        group.add_argument('-l', '--list', action='store_true',
+                           help='list all options and their values')
+        group.add_argument('-d', '--delete', action='store_true',
+                           help='delete an option in one config file')
+        group.add_argument('-D', '--delete-all', action='store_true',
+                           help="delete an option everywhere it's set")
+        group.add_argument('-a', '--append', action='store_true',
+                           help='append to an existing value')
```
marc-hb, line 107: Nice simplification thanks to `add_mutually_exclusive_group()` but I don't find it trivial enough to be combined with the addition of a feature that is not really trivial either. Please use (at least) 2 separate commits.

src/west/app/config.py
carlescufi: You get this:
`ValueError: ConfigFile.LOCAL: file not found; retry in a workspace or set WEST_CONFIG_LOCAL`
So, because you can actually set `WEST_CONFIG_LOCAL` I don't think this is really a bug?

src/west/app/config.py
carlescufi: I don't mind adding the example, but:

1. there are no examples for the other actions
2. this example is particularly convoluted, perhaps not representative?

I have added two just in case in the end :)

tests/test_config.py
```diff
@@ -229,6 +229,9 @@ def test_local_creation_with_topdir():
     assert 'pytest' not in cfg(f=GLOBAL)
     assert cfg(f=LOCAL, topdir=str(topdir))['pytest']['key'] == 'val'
 
+def test_append_basic():
```
carlescufi, line 232: tests are WIP!

src/west/app/config.py
marc-hb: Thanks, I should have tried. It's a small documentation... gap, not worth filling because the error message is good enough and this help message is already long enough.

src/west/app/config.py
```diff
@@ -48,6 +48,14 @@
 To set a value for <name>, type:
     west config <name> <value>
 
+To append to a value for <name>, type:
+    west config -a <name> <value>
+A value must exist in the selected configuration file in order to be able
+to append to it. The existing value can be empty.
+Examples:
+    west config -a build.cmake-args -- " -DEXTRA_CFLAGS='-Wextra -g0' -DFOO=BAR"
```
carlescufi, line 56: @marc-hb is this now what you were looking for?

pdgendt (approved):
TEST APPROVAL, DO NOT MERGE

src/west/app/config.py
marc-hb: LGTM, thanks.

marc-hb (commented):
@carlescufi please try to follow the formatting suggestions from `ruff` for NEW or CHANGED code but please avoid git churn in unchanged code, thanks!

carlescufi:
> @carlescufi please try to follow the formatting suggestions from `ruff` for NEW or CHANGED code but please avoid git churn in unchanged code, thanks!

Honestly I have to say I disagree with this approach, and I already told @pdgendt. I think we either reformat the codebase to match `ruff` or we don't add the (IMHO misleading) check that is there today in CI.

kartben:
> > @carlescufi please try to follow the formatting suggestions from `ruff` for NEW or CHANGED code but please avoid git churn in unchanged code, thanks!
> 
> Honestly I have to say I disagree with this approach, and I already told @pdgendt. I think we either reformat the codebase to match `ruff` or we don't add the (IMHO misleading) check that is there today in CI.

@pdgendt my guess is you just forgot to add a ruff-excludes.toml file, did you not? The way it's done today it's preventing folks to contribute until they fix the entire file they're touching, which sounds broken (and unintentional). FWIW this bit me here https://github.com/zephyrproject-rtos/west/pull/713 too.

pdgendt:
> @pdgendt my guess is you just forgot to add a ruff-excludes.toml file, did you not? The way it's done today it's preventing folks to contribute until they fix the entire file they're touching, which sounds broken (and unintentional). FWIW this bit me here #713 too.

Not really, this was the outcome after some offline discussions with @marc-hb, which were also reasonable, but not the same as the approach that was taken for Zephyr itself.

kartben:
> > @pdgendt my guess is you just forgot to add a ruff-excludes.toml file, did you not? The way it's done today it's preventing folks to contribute until they fix the entire file they're touching, which sounds broken (and unintentional). FWIW this bit me here #713 too.
> 
> Not really, this was the outcome after some offline discussions with @marc-hb, which were also reasonable, but not the same as the approach that was taken for Zephyr itself.

So you mean you will be just ignoring the CI failures for the time being if they are for lines that were not changed by the contributor? Sorry if I am missing something obvious, but I don't understand why it was done differently than for Zephyr itself (plus, of course, I have no idea about the content and outcome of the offline discussions you're referring to ;))

pdgendt:
> So you mean you will be just ignoring the CI failures for the time being if they are for lines that were not changed by the contributor? Sorry if I am missing something obvious, but I don't understand why it was done differently than for Zephyr itself (plus, of course, I have no idea about the content and outcome of the offline discussions you're referring to ;))

Yeah, I understand the confusion, a GH job can only be green or red, nothing in between. The current idea was; contributors **can** be asked to update the format, but if only a few lines are touched, it can get merged without it.

There aren't as many contributors to `west` as there are for Zephyr luckily...

pdgendt:
A large part of the discussion is online: https://github.com/zephyrproject-rtos/west/pull/766

marc-hb:
> I think we (1) either reformat the codebase to match ruff or we don't add the (IMHO misleading) check that is there today in CI.

(1) means the git history is practically destroyed. Especially problematic considering the main author left.
(2) means no formatting progress.

> So you mean you will be just ignoring the CI failures for the time being if they are for lines that were not changed by the contributor?

Yes, formatting issues are obviously not blocking. As mentioned by Pieter, Github has unfortunately no yellow color. BTW this is not just a problem with formatting; every warning in every linter has the same CI problem.

> Sorry if I am missing something obvious, 

Agree it could/should be more obvious. Pieter could you maybe rename "Format check" to "Format _suggestion"? Basically: try to express "yellow" in words.

> but I don't understand why it was done differently than for Zephyr itself (plus, of course, I have no idea about the content and outcome of the offline discussions you're referring to ;))

We're even then because I don't know how it was done for Zephyr itself :-)

In #766 I gave a real-world example where this approach worked well. It was not formatting but `pylint` and `shellcheck`; from a CI perspective the process is the same.

> my guess is you just forgot to add a ruff-excludes.toml file

An exclude file is interesting, is that how Zephyr solves this problem in general? Not just with formatting but with every linter?

Unlike (1) and (2), this allows some form of incremental progress but there are still smaller "big bangs" where the entire git history gets broken on a file-by-file basis. `west` has unfortunately a small number of giant files.

An exclude file does not solve the problem of changes in the code or the configuration of the formatting tool itself.

marc-hb:
To be clear: I do not share Pieter's (and others') "enthusiasm" for formatting. Unlike C/C++, Python already enforces indentation = a little bit of formatting and in practice that seems enough to me, I mean I rarely ever scratch my head trying to understand badly formatted Python code. So, `ruff` only looks like a nice-to-have but nothing more.

On the other hand, I use `git blame/rebase/cherry-pick/merge` all the time and degrading those commands affects productivity greatly. Even worse when there's a reformatting every year because of some `ruff` upgrade or configuration change.

pdgendt:
> To be clear: I do not share Pieter's (and others') "enthusiasm" for formatting. Unlike C/C++, Python already enforces indentation = a little bit of formatting and in practice that seems enough to me, I mean I rarely ever scratch my head trying to understand badly formatted Python code. So, `ruff` only looks like a nice-to-have but nothing more.

I would say the opposite, I "hate" formatting, as in, I don't want to spend time on it. Working on a project where code is in a certain style, I need to "learn" how formatting is done, and carefully curate whatever modifications or additions I do.
I want to focus on the content rather than the looks, and in the best case scenario, tools to do formatting for me.

For python projects I have worked with black in the past, and the ruff formatter is like black, but faster. They have a [commitment](https://black.readthedocs.io/en/latest/the_black_code_style/index.html#stability-policy) towards compatibility too, so have some trust in that the formatting doesn't change drastically.

> On the other hand, I use `git blame/rebase/cherry-pick/merge` all the time and degrading those commands affects productivity greatly. Even worse when there's a reformatting every year because of some `ruff` upgrade or configuration change.

And I agree, that's why I'm not a fan of blindly applying it to all files, but gradually. Either by exclusion or non-forcing checks. Both have downsides, but I'm happy to adjust.

Currently tooling in and around Zephyr is the wild-wild-west. Maintainers get to call the shots, and for good reason, but this does have the downside of different approaches in each. I introduced ruff in west and Zephyr hoping to get a more uniform tooling codebase. And formatting in an uncompromising way.

marc-hb:
> I would say the opposite, I "hate" formatting, as in, I don't want to spend time on it. 
> [...]
> Currently tooling in and around Zephyr is the wild-wild-west. Maintainers get to call the shots, and for good reason, but this does have the downside of different approaches in each.

I don't know about Zephyr more broadly, but in `west`  there had never been a formatter and there was no "time spent" on formatting either. There could have been some suggestions to try this or that formatting tool to help with "extreme" cases (e.g. very long lines) but if there were, it was so rare that I don't remember any such event. So I don't think formatting has ever caused anyone to "spend any time" in west. I don't remember that for _Python_ in Zephyr either.

As I said, C/C++ offers a lot more editing "creativity" than Python so maybe formatting has been a problem there but for `west` specifically I'm afraid `ruff` is a solution in search of a problem.

What would be perfectly non-intrusive is some sort of `ruff_format_patch` tool that touches only new or changed code like checkpatch.pl does. BTW that's what allows checkpatch not to be worried about being "subjective", or having contradictions or inconsistencies across checkpatch versions: because it looks **only** at changed lines. It makes a massive... difference! (pun intended)

https://www.kernel.org/doc/html/latest/dev-tools/checkpatch.html
> Checkpatch is not always right. Your judgement takes precedence over checkpatch messages. If your code looks better with the violations, then its probably best left alone.

pdgendt:
> I don't know about Zephyr more broadly, but in `west` there had never been a formatter and there was no "time spent" on formatting either. There could have been some suggestions to try this or that formatting tool to help with "extreme" cases (e.g. very long lines) but if there were, it was so rare that I don't remember any such event. So I don't think formatting has ever caused anyone to "spend any time" in west. I don't remember that for _Python_ in Zephyr either.

By "time spent", I meant what I do while developing, not commenting on contributions. I have developed in many different languages, mostly out of curiosity and learning new things. And languages like go, rust, zig, ... (yes all the new kids) come with formatting guidelines and **tooling** out of the box.
There is no nit-picking, or preference of appearance as the language has set clear defaults.

Python has the PEP8 guideline which allows much more than what black or ruff currently output, but we currently don't comply either.

> As I said, C/C++ offers a lot more editing "creativity" than Python so maybe formatting has been a problem there but for `west` specifically I'm afraid `ruff` is a solution in search of a problem.

That was my earlier point on trying have uniform code across projects under the Zephyr umbrella, and without tooling to enforce it, it's always a never-ending discussion.

> What would be perfectly non-intrusive is some sort of `ruff_format_patch` tool that touches only new or changed code like checkpatch.pl does. BTW that's what allows checkpatch not to be worried about being "subjective", or having contradictions or inconsistencies across checkpatch versions: because it looks **only** at changed lines. It makes a massive... difference! (pun intended)
> 
> [kernel.org/doc/html/latest/dev-tools/checkpatch.html](https://www.kernel.org/doc/html/latest/dev-tools/checkpatch.html)
> 
> > Checkpatch is not always right. Your judgement takes precedence over checkpatch messages. If your code looks better with the violations, then its probably best left alone.

You can turn off formatting just as easy for code blocks, but this shouldn't be the default IMO.

marc-hb:
> ... engage in that discussion in the first place. That has never been the case with west until now. [...] Indentation likely helped.

I just remembered that `west` had some codestyle checks in CI since forever. That likely helped with readability and reduced formatting needs even more. Formatting is just a small, less important part of code style and readability.

carlescufi:
@marc-hb @pdgendt getting back to this now after a while. I will finish the actual work first, and once we are happy with that, we can discuss with the two of you which `ruff` complaints you'd like me to address.

carlescufi:
@marc-hb @pdgendt tests are ready, I think they add good coverage of the feature.

tests/test_config.py
```diff
@@ -229,6 +229,45 @@ def test_local_creation_with_topdir():
...
+    assert cfg(f=GLOBAL)['pytest']['key'] == 'global'
+    assert cfg(f=LOCAL)['pytest']['key'] == 'local,bar'
+
+    # Test a more complex one, and at a particular configfile level
+    update_testcfg('build', 'cmake-args', '-DCONF_FILE=foo.conf', configfile=GLOBAL)
+    assert cfg(f=GLOBAL)['build']['cmake-args'] == '-DCONF_FILE=foo.conf'
+
+    # Use a list instead of a string to get quoting right
+    cmd(['config', '--global', '-a', 'build.cmake-args', '--',
+         ' -DEXTRA_CFLAGS=\'-Wextra -g0\' -DFOO=BAR'])
```
pdgendt, line 250: Nit, use outer `"` quotes so you don't need to escape the inner `'` ones?

```suggestion
         " -DEXTRA_CFLAGS='-Wextra -g0' -DFOO=BAR"])
```

pdgendt (approved):

tests/test_config.py
carlescufi: I had it like that before, but I personally actually prefer it this way, because it is consistent with the rest of the elements in the list.

tests/test_config.py
```diff
@@ -229,6 +229,45 @@ def test_local_creation_with_topdir():
...
+
+    assert cfg(f=GLOBAL)['build']['cmake-args'] == \
+        '-DCONF_FILE=foo.conf -DEXTRA_CFLAGS=\'-Wextra -g0\' -DFOO=BAR'
+
+def test_append_novalue():
+    with pytest.raises(subprocess.CalledProcessError) as exc_info:
+        cmd('config -a pytest.foo', stderr=subprocess.STDOUT)
+    # Get the output into a variable to simplify pytest error messages
```
marc-hb, line 258: I don't think this comment adds a lot of value. No big deal.

marc-hb (commented):
A few style suggestions, otherwise LGTM!

> once we are happy with that, we can discuss with the two of you which ruff complaints you'd like me to address.

As "usual", I think new and modified code should be "ruff-clean" while unchanged code should be... unchanged. Clicking on the link added by Pieter should help with that.

https://wiki.openstack.org/wiki/GitCommitMessages#Things_to_avoid_when_creating_commits

src/west/app/config.py
```diff
@@ -180,6 +193,16 @@ def read(self, args):
...
 
+    def append(self, args):
+        self.check_config(args.name)
+        where = args.configfile or LOCAL
+        value = self.config.get(args.name, configfile=where)
+        if value is None:
+            self.die(f'option {args.name} not found in the {where.name.lower()} '
+                     'configuration file')
+        args.value = value + args.value
+        self.write(args)
```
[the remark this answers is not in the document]
carlescufi: The doc says:
```
If the configuration file to use is not set, reads use all three in
precedence order, and writes (including appends) use the local file.'''

```
we provide `args` to `self.write()`, so `append` behaves exactly like `write` as documented.

tests/test_config.py
carlescufi: I think it does, if you don't do this the error message thrown by pypi is unreadable.

src/west/app/config.py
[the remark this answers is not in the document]
marc-hb: I should have looked at the `write()` code sorry.

marc-hb (approved):

PR #770, opened 2024-12-02
project+ci: Succeed ruff format check but keep a notice annotation

pdgendt:
Prevent ruff issues being reported by updates of the tool itself.
UP027 has been removed in version 0.8.0, remove it from the ignore list.

For CI move the `continue-on-error` to the format step instead of the job. This should set the job status to green, we only print a notice annotation.

See #771 how the result looks.

### DIFF
--- a/.github/workflows/format.yml
+++ b/.github/workflows/format.yml
@@ -43,8 +43,6 @@ jobs:
     needs: find-changed-files
     if: ${{ needs.find-changed-files.outputs.files != '[]' }}
     runs-on: ubuntu-latest
-    # Allow the workflow run to pass when this job fails
-    continue-on-error: true
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


.github/workflows/format.yml
```diff
@@ -55,12 +53,16 @@ jobs:
...
           args: "format --check --diff"
           src: "${{ matrix.files.path }}"
+          version: 0.8.1
 
       - name: Annotate unformatted file
-        if: ${{ failure() }}
+        if: ${{ steps.format-check.outcome }} == 'failure'
         run: |
-          echo "::warning file=${{ matrix.files.path }},title=File format check failed::Run 'ruff format ${{ matrix.files.path }}'"
+          echo "::notice file=${{ matrix.files.path }},title=Unformatted file::Consider running 'ruff format ${{ matrix.files.path }}'"
```
marc-hb, line 68: Please tell the user that the formatted code can be found in the check logs.

(technically: the diff but no need to mention that here)

marc-hb (approved):

PR #775, opened 2024-12-18
app: more versatile initialization for path in config

### LINKED ISSUE #774: `west init` gives wrong config `path` for a manifest file in nested directories

```
.
├── .west
├── bootloader
│   └── mcuboot
├── modules
│   ├── ...
├── orb
│   ├── private/west.yml
│   └── public
└── zephyr
```

```
[manifest]
path = orb
file = private/west.yml
```

fouge:
in case a manifest file is contained into some sub directories, the command to init a workspace locally should be
```
west init -l sub1/sub2 --mf west.yml
```

but in that case, the west top dir is going to be into `sub1` which is an issue.

one workaround is to use:
```
west init -l sub1 --mf sub2/west.yml
```

but in that case, the config is initialized incorrectly: `manifest.path` contains only `sub1` and `manifest.file` is `sub2/west.yml`.

`manifest.path` should contain a path, even if `--mf` is a relative path. `manifest.file` should only contain the filename.

this change allows any usage so that the workspace top dir is created next to `sub1` and the manifest repo can be located into nested directories

fixes #774

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -204,13 +204,13 @@ below.
                             help='''manifest repository branch or tag name
                             to check out first; cannot be combined with -l''')
         parser.add_argument('--mf', '--manifest-file', dest='manifest_file',
-                            help='manifest file name to use')
+                            help='manifest file name to use, relative to "directory"')
         parser.add_argument('-l', '--local', action='store_true',
-                            help='''use "directory" as an existing local
-                            manifest repository instead of cloning one from
-                            MANIFEST_URL; .west is created next to "directory"
-                            in this case, and manifest.path points at
-                            "directory"''')
+                            help='''use "directory" to create the workspace
+                            from a local manifest directory, instead of cloning
+                            one from MANIFEST_URL; .west is created next to
+                            "directory" in this case, and manifest.path points at
+                            the manifest file location.''')
         parser.add_argument('--rename-delay', type=int,
                             help='''Number of seconds to wait before renaming
                             some temporary directories. Some filesystems like NTFS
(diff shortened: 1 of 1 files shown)


fouge:
IMO to go even further, I think `directory` should point to the location of the west top dir
not: 
> .west is created next to  "directory"

but
> `.west is created in "directory"`

the command would become 
```
west init --mf sub1/sub2/west.yml .
```

thoughts?

fouge:
@pdgendt tox fails with errors even on `main` on my machine :/

src/west/app/project.py
```diff
@@ -287,7 +287,7 @@ def local(self, args) -> Path:
         os.chdir(topdir)
         self.config = Configuration(topdir=topdir)
         self.config.set('manifest.path', os.fspath(rel_manifest))
-        self.config.set('manifest.file', manifest_filename)
+        self.config.set('manifest.file', manifest_file.name)
```
pdgendt, line 290: This doesn't allow a nested manifest file, I think? It should be relative to the manifest path.

marc-hb (changes requested):
Thanks for catching this issue and submitting a PR. I haven't time to look at it in detail yet but I already know we _definitely_ need additional test coverage for this. Either adding some new test(s) or tweaking existing tests or a combination of both.

marc-hb:
Again, thanks for  the excellent report.

> this is a breaking change though, thoughts?

Interestingly, even your very first `west init -l development/west_project1/ --mf west.yml` suggestion at the top would already be a backwards incompatible change.

I'm afraid we need to clearly and formally define the user interface before even looking at the code.

> @pdgendt tox fails with errors even on main on my machine 😞

Please file a separate bug for this. Don't forget the OS and OS version.

marc-hb:
@fouge please test and review newer 
- #854

marc-hb:
Superseded by #854

EDIT: #854 superseded by #953

PR #777, opened 2025-01-04
tests: config: Fix assertions checks with pytest.raises

### LINKED ISSUE #776: Multiple tests in `tests/test_config.py` are broken
The following construct used pervasively in `tests/test_config.py`:

```
    with pytest.raises(subprocess.CalledProcessError) as e:
        cmd('...')
        assert 'error string' in str(e)
```

but this doesn't work for two reasons:

1. The `assert` statement must be outside the context manager (i.e. dedented one level) for it to execute at all if the exception is raised
2. The error string is very often not part of the exception object, so one needs to check the captured stderr instead

Something like this works instead:

```
    with pytest.raises(subprocess.CalledProcessError) as exc_info:
        cmd('...', stderr=subprocess.STDOUT)
    err_msg = exc_info.value.output.decode("utf-8")
    assert 'str' in err_msg
```

pdgendt:
Dedent assertion checks as they were currently skipped, and make sure the correct error message is validated.

Fixes #776

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -351,6 +351,15 @@ def cmd(cmd, cwd=None, stderr=None, env=None):
         print('cmd: west:', shutil.which('west'), file=sys.stderr)
         raise
 
+
+def cmd_raises(cmd_str_or_list, expected_exception_type, cwd=None, env=None):
+    # Similar to 'cmd' but an expected exception is caught.
+    # Returns the output together with stderr data
+    with pytest.raises(expected_exception_type) as exc_info:
+        cmd(cmd_str_or_list, stderr=subprocess.STDOUT, cwd=cwd, env=env)
+    return exc_info.value.output.decode("utf-8")
+
+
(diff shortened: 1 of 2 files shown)


marc-hb (approved):

pdgendt:
@carlescufi #768 might change the message with the grouped argument change, this PR can go before or after yours but might need changes accordingly.

pdgendt:
Updated after merging #768 with updated error messages.

marc-hb (commented):
I took a longer look and realized now that this does much more than moving the asserts : this also fixes the test code which of course did not work because it never ran!

That's great, but now I see a lot of repetition. Have you considered a new, generic function, something like:

```python
def run_bad_cmd(cmd_str_or_list, expected_exception_type):
   with pytest.raises(expected_exception_type) as e:
       ...

   return err_msg


emsg = run_bad_cmd("config pytest")
assert 'invalid ..." in emsg
```

carlescufi (approved):
Thanks!

PR #780, opened 2025-01-21
tests: Extend test_extension_command_duplicate

57300:
This test is meant to cover the handling of duplicate command names, including the warnings to be emitted. Previously, it only tested the case where two extension commands had the same name. The other case is duplicating a built-in command name, which prompts a different warning message, so update the test to include it.

Additionally, ensure that each duplicate command in the test comes from a different project, so that each warning message is expected to contain a different project name. This serves to verify the bug fix from commit edd1cee2ebb6191cb6fa4ec874d0113c1fd2b9a3.

### DIFF
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -1941,8 +1941,10 @@ def test_extension_command_multiproject(repos_tmpdir):
 
 
 def test_extension_command_duplicate(repos_tmpdir):
-    # Test to ensure that in case to subprojects introduces same command, it
-    # will print a warning.
+    # West should disregard an extension command if its name is already taken,
+    # either by a built-in command or another extension command added earlier.
+    # A warning should also appear for each such case.
+
     rr = repos_tmpdir.join('repos')
     remote_kconfiglib = str(rr.join('Kconfiglib'))
     remote_zephyr = str(rr.join('zephyr'))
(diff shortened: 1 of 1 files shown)


pdgendt (approved):

tests/test_project.py
```diff
@@ -2002,13 +2019,18 @@ def do_run(self, args, ignored):
     west_tmpdir.chdir()
     cmd('update')
 
-    actual = cmd('test-extension', stderr=subprocess.STDOUT).splitlines()
-    expected = [
+    expected_warns = [
+        'WARNING: ignoring project Kconfiglib extension command "list"; this is a built in command',  # noqa: E501
```
marc-hb, line 2023: Nit: these `noqa` do not seem necessary:
```suggestion
        'WARNING: ignoring project Kconfiglib extension command "list"; '
        'this is a built in command',
```
Python concatenates strings similar to what C does.

tests/test_project.py
```diff
@@ -1941,8 +1941,10 @@ def do_run(self, args, ignored):
 
 
 def test_extension_command_duplicate(repos_tmpdir):
-    # Test to ensure that in case to subprojects introduces same command, it
-    # will print a warning.
+    # West should disregard an extension command if its name is already taken,
+    # either by a built-in command or another extension command added earlier.
+    # A warning should also appear for each such case.
```
marc-hb, line 1946: pydoc syntax instead?

marc-hb (commented):
Nice, thanks. I'm curious how you noticed this? Were you using a `west` version older than the fix?

57300:
> Nice, thanks. I'm curious how you noticed this? Were you using a `west` version older than the fix?

Yes, I was on an older version. To be honest, I noticed the bug while looking for a crude way to override `west flash` (before out-of-tree runner support was added to Zephyr).

tests/test_project.py
57300: Do you mean: move this comment above the function, or make it a docstring?
So far, I just kept it consistent with other comments in this particular file.

mbolivar: >  I can't see a good reason why.

I tried to write pydoc for APIs and comments for things that weren't API, as a pattern.

marc-hb: Thanks @mbolivar , now I see https://docs.zephyrproject.org/latest/develop/west/west-apis.html is extracted from the source. Is that automated somewhere or manual?

> I tried to write pydoc for APIs and comments for things that weren't API, as a pattern.

Makes sense but the `tests/` are located in an entirely different, top-level directory so they should not interfere.

Either way this should really be written down somewhere.

Apologies @57300 for the off-topic discussion, feel free to ignore.

tests/test_project.py
57300: Updated

Issue #519, opened 2021-07-28
New west update --if-present option (better name wanted)


marc-hb:
From a digression in #518.

To please everyone and provide a good "out of the box"  experience, west manifests can have many repos that are unnecessary to most users. As a random example, out of the 45-ish projects currently in https://github.com/zephyrproject-rtos/zephyr/blob/main/west.yml I need only 2 right now. So to save disk, network and time I never ever ran `west update` and never downloaded more than these 2 repos. Instead I always `west update proj1 proj2`. This also makes `west status`  much shorter and readable (similar to #518, which solves a different problem).

`west update proj1 proj2` works but it is inconvenient. First because you have to remember to never ever run `west update`, then because the real command is actually `west update dir1/dir2/what_was_that_name_again?` So it would be nice to have to run `west update dir1/dir2/what_was_that_name_again?` only the first time and then run just `west update --if-present`. Even better: have `if-present` in west config and then just `west update`.

marc-hb:
From @mbolivar-nordic  in #518
> I think that since west status already ignores inactive projects by default, and the upstream zephyr manifest now has groups, you may have the ingredients for most of what you want already by just cloning a subset of active projects.

I will look into that but I suspect a `if-present` in west config  would beat groups from a convenience perspective. If/when that setting exists then I would for sure permanently add it once to my global user config and never touch it again (which means some `--no-if-present` CLI option would be useful to override the config on rare occasions)

mbolivar-nordic:
Thanks for filing a feature. I'd like to start to clarify some behavior around imports.

Here are a few simple cases.

1. I have project A in my main manifest file, and I have `import: true` for project A. A is cloned in my workspace. A/west.yml defines projects B and C, which are not cloned.
2. I have A and B cloned, but not C.
3. I have B cloned, but not A or C.

What does `west update --if-present` do in each case? I think these would be the "least surprising" outcomes:

1. A is updated
2. A and B are updated
3. Nothing is updated (maybe a little surprising actually!)

mbolivar-nordic:
OK, thanks. Sketching out an implementation then, as prep work we would need:

- a new `ImportFlag.FORCE` that asks the Manifest constructor to call `importer` on every single `import` statement
- support for it in the Manifest import machinery

Then `west update --if-present` would need to:

- pass ImportFlag.FORCE to the Manifest constructor
- pass an `importer` callback which returns `None` if the project is not cloned

Plus any other unanticipated changes I can't think of yet.

[...]

PR #781, opened 2025-02-06
app: project: Allow updating only cloned projects

pdgendt:
Add an update command argument and configuration option to allow updating the currently cloned projects.

This can help in the scenario where a user updated an initial set of projects manually and would like to only update those in the future.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -952,6 +952,9 @@ class Update(_ProjectCommand):
         parser.add_argument('--stats', action='store_true',
                             help='''print performance statistics for
                             update operations''')
+        parser.add_argument('--cloned', action='store_true',
+                            help='''update the projects that are already cloned
+                            in the current workspace''')
 
         group = parser.add_argument_group(
             title='local project clone caches',
@@ -1048,6 +1051,7 @@ class Update(_ProjectCommand):
         self.name_cache = args.name_cache or config.get('update.name-cache')
         self.sync_submodules = config.getboolean('update.sync-submodules',
                                                  default=True)
+        self.cloned = args.cloned or config.getboolean('update.cloned')
 
         self.group_filter: List[str] = []
 
@@ -1092,6 +1096,9 @@ class Update(_ProjectCommand):
                     project.name in self.updated):
                 continue
             try:
+                if self.cloned and not project.is_cloned():
+                    self.dbg(f'{project.name}: skipping missing project')
+                    continue
                 if not self.project_is_active(project):
                     self.dbg(f'{project.name}: skipping inactive project')
                     continue
@@ -1163,6 +1170,9 @@ class Update(_ProjectCommand):
             if isinstance(project, ManifestProject):
                 continue
             try:
+                if self.cloned and not project.is_cloned():
+                    self.dbg(f'{project.name}: skipping missing project')
+                    continue
                 self.update(project)
             except subprocess.CalledProcessError:
                 failed.append(project)
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -1521,6 +1521,28 @@ def test_update_narrow_depth1(tmpdir):
     assert len(refs) == 1
 
 
+def test_update_cloned(west_init_tmpdir):
+    # Manually update a single project
+    cmd('update Kconfiglib')
+
+    cmd('update --cloned')
+    assert (west_init_tmpdir / 'subdir' / 'Kconfiglib').check(dir=1)
+    assert (west_init_tmpdir / 'tagged_repo').check(dir=0)
+    assert (west_init_tmpdir / 'net-tools').check(dir=0)
+
+    cmd('config update.cloned true')
+    cmd('update')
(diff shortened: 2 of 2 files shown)


src/west/app/project.py
```diff
@@ -952,6 +952,9 @@ def do_add_parser(self, parser_adder):
         parser.add_argument('--stats', action='store_true',
                             help='''print performance statistics for
                             update operations''')
+        parser.add_argument('--cloned', action='store_true',
+                            help='''update the projects that are already cloned
+                            in the current workspace''')
```
mbolivar, line 957: Since `update_some()` is being changed to match, I think the meaning of this option is closer to "only update a project if it's already cloned locally" or so, right?

I'm thinking about the behavior of `west update --cloned foo` for an uncloned project `foo` and feeling the current help doesn't make clear how it should work in this case.

mbolivar (commented):
The biggest source of doubt I have about this comes from how it should work when you have a project that is:

1. not cloned
2. contains an `import:` in the manifest

The behavior here can be pretty counter-intuitive and confusing -- did you consider this in detail? What's the plan here?

For a real world example, consider this top level west.yml fragment:

```
manifest:
  projects:
    - name: p1
      import: true
    - name: p2
      import: true
```

then `p1/west.yml`  says:

```
manifest:
  projects:
    - name: foo
      url: url1
```

while `p2/west.yml` says:

```
manifest:
  projects:
    - name: foo
      url: url2
```

If you do `west update --cloned` with `p1` and `p2` both cloned, you get `foo` from `url1`. But if you do `west update --cloned` with just `p2` cloned, you get `foo` from `url2`.

I find this sort of behavior pretty confusing and I'm not sure what the "right" thing is in terms of what the user would reasonably expect:

1. should we just fail if any project with an `import` is not cloned?
2. should we ignore it and therefore allow us to get imported projects from surprising places (`url2` instead of `url1`)?
3. something else?

pdgendt:
> The biggest source of doubt I have about this comes from how it should work when you have a project that is:
> 
> 1. not cloned
> 2. contains an `import:` in the manifest
> 
> The behavior here can be pretty counter-intuitive and confusing -- did you consider this in detail? What's the plan here?
> 
> If you do `west update --cloned` with `p1` and `p2` both cloned, you get `foo` from `url1`. But if you do `west update --cloned` with just `p2` cloned, you get `foo` from `url2`.

I see, but how is this different from calling `west update p2` without having cloned `p1`? The idea was to give users the option of working in/with the upstream Zephyr repository, and having only a subset of HALs/tools/... cloned, without the need to remember which were already updated.

An "end-goal" I envisioned is to have a West extension in Zephyr, for example `west getting-started`, that would present a cli wizard to the user to install a bare minimum setup (HAL + toolchain + ..) instead of cloning/installing everything. This option could be useful in such a setup IMO. 
 
> I find this sort of behavior pretty confusing and I'm not sure what the "right" thing is in terms of what the user would reasonably expect:
> 
> 1. should we just fail if any project with an `import` is not cloned?
> 2. should we ignore it and therefore allow us to get imported projects from surprising places (`url2` instead of `url1`)?
> 3. something else?

I tried to keep it as simple as possible, but if it would cause more confusion than it could be useful, I'm happy to drop it.

pdgendt:
> > The biggest source of doubt I have about this comes from how it should work when you have a project that is:
> > 
> > 1. not cloned
> > 2. contains an `import:` in the manifest
> > 
> > The behavior here can be pretty counter-intuitive and confusing -- did you consider this in detail? What's the plan here?
> > If you do `west update --cloned` with `p1` and `p2` both cloned, you get `foo` from `url1`. But if you do `west update --cloned` with just `p2` cloned, you get `foo` from `url2`.
> 
> I see, but how is this different from calling `west update p2` without having cloned `p1`? The idea was to give users the option of working in/with the upstream Zephyr repository, and having only a subset of HALs/tools/... cloned, without the need to remember which were already updated.

Hmn, it does differ as only `p2` would be updated and `foo` wouldn't. The devil _is_ in the details.

> An "end-goal" I envisioned is to have a West extension in Zephyr, for example `west getting-started`, that would present a cli wizard to the user to install a bare minimum setup (HAL + toolchain + ..) instead of cloning/installing everything. This option could be useful in such a setup IMO.

Maybe this would be more suited for the `example-application` repository or even a dedicated one.

> > I find this sort of behavior pretty confusing and I'm not sure what the "right" thing is in terms of what the user would reasonably expect:
> > 
> > 1. should we just fail if any project with an `import` is not cloned?
> > 2. should we ignore it and therefore allow us to get imported projects from surprising places (`url2` instead of `url1`)?
> > 3. something else?
> 
> I tried to keep it as simple as possible, but if it would cause more confusion than it could be useful, I'm happy to drop it.

I'll close this for now :)

marc-hb:
Just for the record, this looked like an attempt to implement:
- #519

> I find this sort of behavior pretty confusing and I'm not sure what the "right" thing is in terms of what the user would reasonably expect:

Sounds like we just got yet another PR/issue with the infamous label https://github.com/zephyrproject-rtos/west/labels/Partial%20imports

PR #783, opened 2025-02-26
project: launch subprocesses without shell

jhol:
In the Project method `set_new_manifest_rev`, the `_update_manifest_rev` helper function tries to run the following git command:

```
git update-ref -m .. refs/heads/ref <commit-id>^{commit}
```

However, on Msys2's MinGW build of python handles the escaping of `...^{commit}` incorrectly, converting it into `...commit`.

This issue highlights the pitfalls of using the Python `subprocess` module's `shell=True` option, which can apply string conversions to commands that are unexpected by the calling code.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -1731,8 +1731,7 @@ class ForAll(_ProjectCommand):
 
             self.banner(
                 f'running "{args.subcommand}" in {project.name_and_path}:')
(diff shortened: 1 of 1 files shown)


marc-hb:
I agree `shell=True` is evil. As I'm guessing you discovered, some tests depend on it. Maybe some features even?

If you don't have time for a complete fix, could you please turn  this into a new bug? I could be wrong but I don't remember this being already filed.

PR #786, opened 2025-03-19
app: manifest: Add the ability to display untracked files and directories

carlescufi:
This new command-line switch implements something similar to `git
    status`'s untracked files display.
    
    The code examines the workspace and prints out all files and folders
    that are not inside of a west project.

Fixes: #622
Supersedes #702

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -584,6 +584,14 @@ class ManifestCommand(_ProjectCommand):
               If this file uses imports, it will not contain all the
               manifest data.
 
+            - --untracked: print all files and directories inside the workspace that
+              are not tracked or managed by west. This effectively means any
+              file or directory that is outside all of the projects' directories on
+              disk (regardless of whether those projects are active or inactive).
+              This is similar to `git status` for untracked files. The
+              output format is relative to the current working directory and is
+              stable and suitable as input for scripting.
+
             If the manifest file does not use imports, and all project
             revisions are SHAs, the --freeze and --resolve output will
             be identical after a "west update".
@@ -604,6 +612,9 @@ class ManifestCommand(_ProjectCommand):
                            exiting with an error if there are issues''')
         group.add_argument('--path', action='store_true',
                            help="print the top level manifest file's path")
+        group.add_argument('--untracked', action='store_true',
+                           help='''print all files and directories not managed or
+                           tracked by west''')
 
         group = parser.add_argument_group('options for --resolve and --freeze')
         group.add_argument('-o', '--out',
@@ -624,6 +635,8 @@ class ManifestCommand(_ProjectCommand):
         elif args.freeze:
             self._die_if_manifest_project_filter('freeze')
             self._dump(args, manifest.as_frozen_yaml(**dump_kwargs))
+        elif args.untracked:
+            self._untracked()
         elif args.path:
             self.inf(manifest.path)
         else:
@@ -639,6 +652,70 @@ class ManifestCommand(_ProjectCommand):
                      'west developers if you have a use case for resolving '
                      'the manifest while projects are made inactive by the '
                      'project filter.')
+
+    def _untracked(self):
+        ''' "Performs a top-down search of the west topdir,
+        ignoring every directory that corresponds to a west project.
+        '''
+        ppaths = []
+        untracked = []
+        for project in self._projects(None):
+            # We do not check for self.manifest.is_active(project) because
+            # inactive projects are still considered "tracked directories".
+            ppaths.append(Path(project.abspath))
+
+        # Since west tolerates nested projects (i.e. a project inside the directory
+        # of another project) we must sort the project paths to ensure that we
+        # hit the "enclosing" project first when iterating.
+        ppaths.sort()
+
+        def _find_untracked(directory):
+            '''There are three cases for each element in a directory:
+            - It's a project -> Do nothing, ignore the directory.
+            - There are no projects inside -> add to untracked list.
+            - It's not a project directory but there are some projects inside it -> recurse.
+            The directory argument cannot be inside a project, otherwise all bets are off.
+            '''
+            self.dbg(f'looking for untracked files/directories in: {directory}')
+            for e in [e.absolute() for e in directory.iterdir()]:
+                if not e.is_dir() or e.is_symlink():
+                    untracked.append(e)
+                    continue
+                self.dbg(f'processing directory: {e}')
+                for ppath in ppaths:
+                    # We cannot use samefile() because it requires the file
+                    # to exist (not always the case with inactive or even
+                    # uncloned projects).
+                    if ppath == e:
+                        # We hit a project root directory, skip it.
+                        break
+                    elif e in ppath.parents:
+                        self.dbg(f'recursing into: {e}')
+                        _find_untracked(e)
+                        break
+                else:
+                    # This is not a project and there is no project inside.
+                    # Add to untracked elements.
+                    untracked.append(e)
+                    continue
+
+        # Avoid using Path.walk() since that returns all files and directories under
+        # a particular directory, which is overkill in our case. Instead, recurse
+        # only when required.
+        _find_untracked(Path(self.topdir))
+
+        # Exclude the .west directory, which is maintained by west
+        try:
+            untracked.remove((Path(self.topdir) / Path(WEST_DIR)).resolve())
+        except ValueError:
+            self.die(f'Directory {WEST_DIR} not found in workspace')
+
+        # Sort the results for displaying to the user.
+        untracked.sort()
+        for u in untracked:
+            # We cannot use Path.relative_to(p, walk_up=True) because the
+            # walk_up parameter was only added in 3.12
+            self.inf(os.path.relpath(u, Path.cwd()))
 
     def _dump(self, args, to_dump):
         if args.out:
@@ -881,7 +958,10 @@ class Status(_ProjectCommand):
             'status',
             '"git status" for one or more projects',
             '''Runs "git status" for each of the specified projects.
-            Unknown arguments are passed to "git status".''',
+            Unknown arguments are passed to "git status".
+
+            Note: If you are looking to find untracked files and directories
+            in the workspace use "west manifest --untracked".''',
             accepts_unknown_args=True,
         )
 
@@ -1747,7 +1827,8 @@ To get "git grep foo" results from all cloned, active projects:
 
 To do the same with:
 
-- git grep --untracked: west grep --untracked foo
+- git grep --untracked: west grep --untracked foo (see also west manifest --untracked
+  for finding untracked files outside the project directories)
 - ripgrep:              west grep --tool ripgrep foo
 - grep --recursive:     west grep --tool grep foo
 
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -5,6 +5,7 @@ import os
 import re
 import shutil
 import subprocess
+import sys
 import textwrap
 from pathlib import Path, PurePath
 
@@ -238,6 +239,122 @@ def test_list_sha(west_update_tmpdir):
 
     assert cmd('list -f {sha}').startswith("N/A")
 
+
+def test_manifest_untracked(west_update_tmpdir):
+
+    def check(expected, cwd=None):
+        out_lines = cmd("manifest --untracked", cwd=cwd).splitlines()
+        assert out_lines == expected
+
+    # Ensure that all projects are active
+    projs = cmd('list -f {name}').splitlines()
+    assert projs == ['manifest', 'Kconfiglib', 'tagged_repo', 'net-tools']
+
+    # Disable Kconfiglib
+    cmd('config manifest.group-filter -- -Kconfiglib-group')
+
+    # Ensure that Kconfiglib is inactive
+    projs = cmd('list -f {name}').splitlines()
+    assert projs == ['manifest', 'tagged_repo', 'net-tools']
+    projs = cmd('list --all -f {name}').splitlines()
+    assert projs == ['manifest', 'Kconfiglib', 'tagged_repo', 'net-tools']
+
+    topdir = Path(west_update_tmpdir)
+
+    # No untracked files yet
+    check(list())
+
+    (topdir / "dir").mkdir()
+    # Untracked dir
+    check(['dir'])
+
+    (topdir / "unt").mkdir()
+    (topdir / "unt" / "file.yml").touch()
+    # Ensure we show the dir as untracked, not the file
+    check(['dir', 'unt'])
+
+    # Add a file to see it's displayed correctly as untracked
+    (topdir / "file.txt").touch()
+    check(['dir', 'file.txt', 'unt'])
+    (topdir / 'subdir' / "z.py").touch()
+    check(['dir', 'file.txt', str(Path('subdir/z.py')), 'unt'])
+
+    (topdir / "subdir" / "new").mkdir()
+    (topdir / "subdir" / "new" / "afile.txt").touch()
+    check(['dir', 'file.txt', str(Path('subdir/new')),
+           str(Path('subdir/z.py')), 'unt'])
+
+    # Check relative paths
+    check([str(Path('../dir')), str(Path('../file.txt')),
+           str(Path('../subdir/new')), str(Path('../subdir/z.py')), '.'],
+          cwd=str(Path("unt/")))
+
+    # Add a file to an existing project, ignored by --untracked
+    (topdir / "net-tools" / "test_manifest_untracked.file").touch()
+    check(['dir', 'file.txt', str(Path('subdir/new')), str(Path('subdir/z.py')),
+           'unt'])
+
+    kconfiglib = Path(topdir / "subdir" / "Kconfiglib")
+    # Same but with an inactive project
+    (kconfiglib / "test_manifest_untracked.file").touch()
+    check(['dir', 'file.txt', str(Path('subdir/new')), str(Path('subdir/z.py')),
+           'unt'])
+
+    # Copy (via clone) a full Git repo so we verify that those are also
+    # displayed as untracked.
+    clone(topdir / "net-tools", Path('subdir/acopy'))
+    clone(topdir / "net-tools", Path('tmpcopy'))
+
+    check(['dir', 'file.txt', str(Path('subdir/acopy')), str(Path('subdir/new')),
+           str(Path('subdir/z.py')), 'tmpcopy', 'unt'])
+
+    # Empty a project so it's not a Git repo anymore
+    (topdir / "net-tools" / ".git").rename(topdir / "net-tools" / "former-git")
+    # Should make no difference
+    check(['dir', 'file.txt', str(Path('subdir/acopy')), str(Path('subdir/new')),
+           str(Path('subdir/z.py')), 'tmpcopy', 'unt'])
+
+    # Same with an inactive project
+    (kconfiglib / ".git").rename(kconfiglib / "former-git")
+    # Should make no difference
+    check(['dir', 'file.txt', str(Path('subdir/acopy')), str(Path('subdir/new')),
+           str(Path('subdir/z.py')), 'tmpcopy', 'unt'])
+
+    # Even if we make the whole inactive project disappear it should make no
+    # difference at all, except that the renamed dir will show up.
+    (kconfiglib).rename(topdir / "subdir" / "other")
+    check(['dir', 'file.txt', str(Path('subdir/acopy')), str(Path('subdir/new')),
+           str(Path('subdir/other')), str(Path('subdir/z.py')), 'tmpcopy', 'unt'])
+
+@pytest.mark.skipif(sys.platform.startswith("win"), reason="symbolic links not tested on Windows")
+def test_manifest_untracked_with_symlinks(west_update_tmpdir):
+
+    def check(expected, cwd=None):
+        out_lines = cmd("manifest --untracked", cwd=cwd).splitlines()
+        assert out_lines == expected
+
+    # Disable Kconfiglib to have an inactive project
+    cmd('config manifest.group-filter -- -Kconfiglib-group')
+
+    # Ensure that Kconfiglib is inactive
+    projs = cmd('list -f {name}').splitlines()
+    assert projs == ['manifest', 'tagged_repo', 'net-tools']
+
+    # Create a folder symlink
+    Path('asl').symlink_to(Path('subdir/Kconfiglib'))
+    # Create another one
+    Path('anothersl').symlink_to(Path('../'))
+    # Yet another one
+    Path('subdir/yetanothersl').symlink_to(Path('tagged_repo'))
+    # check that symlinks are displayed like any other regular untracked file or
+    # directory
+    check(['anothersl', 'asl', str(Path('subdir/yetanothersl'))])
+
+    # File symlink tests, should all be displayed as untracked as well
+    Path('filesl.yml').symlink_to(Path('zephyr/west.yml'))
+    Path('subdir/afsl.py').symlink_to(Path('net-tools/scripts/test.py'))
+    check(['anothersl', 'asl', 'filesl.yml', str(Path('subdir/afsl.py')),
+           str(Path('subdir/yetanothersl'))])
 
 def test_manifest_freeze(west_update_tmpdir):
     # We should be able to freeze manifests.


src/west/app/project.py
```diff
@@ -640,6 +644,48 @@ def _die_if_manifest_project_filter(self, action):
...
+    def _status(self):
+        topdir = Path(self.topdir)
+        ppaths = []
+        untracked = []
+        for project in self._projects(None):
+            # Skip inactive projects unless the user said
+            # --all or named some projects explicitly.
+            if not self.manifest.is_active(project):
+                self.dbg(f'{project.name}: skipping inactive project')
+                continue
```
pdgendt, line 656: Is this desired? Now inactive projects will be printed as untracked, AFAICS?

carlescufi: right, the problem is that with the APIs I am using the folder needs to be present on disk for it to work. I will think about it and provide something better.

src/west/app/project.py
```diff
@@ -640,6 +644,46 @@ def _die_if_manifest_project_filter(self, action):
...
+                        untracked.append(e)
+                        continue
+                else:
+                    # Untracked file
+                    untracked.append(e)
+
+        # Avoid using Path.walk() since that returns all files and folders under
+        # a particular folder, which is overkill in our case. Instead, recurse
+        # only when required.
+        _folder_status(Path.cwd())
```
pdgendt, line 682: Why not `topdir`? I think it would be confusing to have different output depending on where you are in the workspace, also because I would hope the paths are relative to `topdir` in the output?

carlescufi: Because this is how `git status` behaves. This is another thing I wanted to discuss

src/west/app/project.py
```diff
@@ -640,6 +645,45 @@ def _die_if_manifest_project_filter(self, action):
...
+                    # Untracked file
+                    untracked.append(e)
+
+        # Avoid using Path.walk() since that returns all files and folders under
+        # a particular folder, which is overkill in our case. Instead, recurse
+        # only when required.
+        _find_untracked(Path(self.topdir))
+        self.inf('Untracked files and folders:')
+        for u in untracked:
+            self.inf(u)
```
pdgendt, line 685: I'd prefer to print out the paths relative to `topdir` here.

carlescufi: Relative to `topdir`, or relative to `cwd` like `git status` does?

src/west/app/project.py
```diff
@@ -640,6 +645,45 @@ def _die_if_manifest_project_filter(self, action):
...
+                            # We hit a project root folder, skip it
+                            break
+                        elif e in ppath.parents:
+                            self.dbg(f'recursing into: {e}')
+                            _find_untracked(e)
+                            break
+                    else:
+                        # No match, untracked folder
+                        untracked.append(e)
+                        continue
```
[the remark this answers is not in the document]
carlescufi: No, I am using Python's [for/else](https://book.pythontips.com/en/latest/for_-_else.html)

src/west/app/project.py
pdgendt: Relative to cwd is perhaps even better.

src/west/app/project.py
```diff
@@ -640,6 +645,45 @@ def _die_if_manifest_project_filter(self, action):
                      'the manifest while projects are made inactive by the '
                      'project filter.')
 
+    def _untracked(self):
+        ppaths = []
+        untracked = []
+        for project in self._projects(None):
+            # We do not check for self.manifest.is_active(project) because
+            # inactive projects are still considered "tracked folders"
+            ppaths.append(Path(project.abspath).resolve(strict=False))
```
pdgendt, line 654: I think these `ppaths` needs to be sorted so parent directories are checked first, otherwise in `_find_untracked` the `elif e in ppath.parents:` is triggered before `if ppath == e:`

```suggestion
        for project in self._projects(None):
            # We do not check for self.manifest.is_active(project) because
            # inactive projects are still considered "tracked folders"
            ppaths.append(Path(project.abspath).resolve(strict=False))

        ppaths.sort() # make sure hierarchy is sorted
```

src/west/app/project.py
```diff
@@ -640,6 +645,45 @@ def _die_if_manifest_project_filter(self, action):
...
+                        continue
+                else:
+                    # Untracked file
+                    untracked.append(e)
+
+        # Avoid using Path.walk() since that returns all files and folders under
+        # a particular folder, which is overkill in our case. Instead, recurse
+        # only when required.
+        _find_untracked(Path(self.topdir))
+        self.inf('Untracked files and folders:')
```
pdgendt, line 683: Does this need to be printed? This would prevent us from using this command in some scriptsl

src/west/app/project.py
carlescufi: You're very right. Fixed.

src/west/app/project.py
carlescufi: Removed

src/west/app/project.py
```diff
@@ -640,6 +650,59 @@ def _die_if_manifest_project_filter(self, action):
...
+                        # uncloned projects).
+                        if ppath == e:
+                            # We hit a project root folder, skip it.
+                            break
+                        elif e in ppath.parents:
+                            self.dbg(f'recursing into: {e}')
+                            _find_untracked(e)
+                            break
+                    else:
+                        # No match, untracked folder.
```
marc-hb, line 678: Comment hidden by force-push:
- https://github.com/zephyrproject-rtos/west/pull/786#discussion_r2004029315

src/west/app/project.py
```diff
@@ -640,6 +650,59 @@ def _die_if_manifest_project_filter(self, action):
...
+                        untracked.append(e)
+                        continue
+                else:
+                    # Untracked file.
+                    untracked.append(e)
+
+        # Since west supports nested projects (i.e. a project inside the folder
+        # of another project) we must sort the project paths to ensure that we
+        # hit the "enclosing" project first when iterating.
+        ppaths.sort()
```
marc-hb, line 688: Please move this up, immediately after the first `ppaths` initialization step. It's confusing to have 25 lines between these two steps.

carlescufi:
Thanks @marc-hb for the review! Will address those comments soon!

src/west/app/project.py
```diff
@@ -640,6 +650,59 @@ def _die_if_manifest_project_filter(self, action):
                      'the manifest while projects are made inactive by the '
                      'project filter.')
 
+    def _untracked(self):
```
marc-hb, line 653:
```suggestion
    def _untracked(self):
    """Performs a top-down search of the west topdir, pruned every time we hit a west project.
```

src/west/app/project.py
marc-hb: Are we committing to a certain, scriptable format? I don't have a strong opinion but if not then that should be clear in the --help.

src/west/app/project.py
carlescufi: The discussion was whether to find untracked files from `cwd()` and below or from `topdir` and below (i.e. the whole workspace) we ended up with the latter because that's how `git status` works for untracked files.

src/west/app/project.py
carlescufi: Good point. I will add it.

src/west/app/project.py
```diff
@@ -640,6 +650,59 @@ def _die_if_manifest_project_filter(self, action):
...
+        ppaths = []
+        untracked = []
+        for project in self._projects(None):
+            # We do not check for self.manifest.is_active(project) because
+            # inactive projects are still considered "tracked folders".
+            ppaths.append(Path(project.abspath).resolve(strict=False))
+
+        def _find_untracked(folder):
```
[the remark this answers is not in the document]
carlescufi: I am OK with the addition of pydoc, but not with the renaming of the functions. I find the names I used clearer. Ultimately this is a matter of taste, but I will not change them for now.

pdgendt (approved):

marc-hb (commented):
I tend to have a lot of untracked stuff in one workspace so I gave this a very quick try. Seems to work but I found one annoying issue: symbolic links are resolved in the display :-( Could you avoid that?

```shell
west manifest --untracked

../other_workspace/real_repo        # resolved symbolic link

crosstool-ng      # plain directory, expected
crosstool-ng/builds/xtensa-apl-elf  # more resolved symlinks
crosstool-ng/builds/xtensa-byt-elf
crosstool-ng/builds/xtensa-cnl-elf
```

The logging with `west -vv manifest --untracked` is unexpected: It starts with prefixes but the new statements miss them?!

```
west.manifest: DEBUG: resolving self import submanifests
west.manifest: DEBUG: found submanifest directory: submanifests
west.manifest: DEBUG: resolved self import

_find_untracked in: /home/me/WS
processing folder: /home/me/WS/.west
processing folder: /home/me/WS/acpi-scripts
```

Code LGTM! I hoped the pydocs could be in a better English but they're OK ;-D

carlescufi:
> I tend to have a lot of untracked stuff in one workspace so I gave this a very quick try. Seems to work but I found one annoying issue: symbolic links are resolved in the display :-( Could you avoid that?

@marc-hb this was intentional, is it not what `git status` does?

>The logging with west -vv manifest --untracked is unexpected: It starts with prefixes but the new statements miss them?!

Indeed, will fix

carlescufi:
> Code LGTM! I hoped the pydocs could be in a better English but they're OK ;-D

I tried to improve things a bit!

src/west/app/project.py
```diff
@@ -640,6 +652,68 @@ def _die_if_manifest_project_filter(self, action):
...
+            for e in [e.absolute() for e in folder.iterdir()]:
+                if not e.is_dir():
+                    untracked.append(e)
+                    continue
+                self.dbg(f'processing folder: {e}')
+                for ppath in ppaths:
+                    # We cannot use samefile() because it requires the file
+                    # to exist (not always the case with inactive or even
+                    # uncloned projects).
+                    if ppath == e:
```
marc-hb, line 688: I wonder whether symbolic links could break this, let me try.
If yes then maybe you should first check whether ppath exists, then compare.

Or maybe even simpler, something like `ppaths = [ p for p in ppaths if p.exists() ]`

Symbolic links can be a nightmare: https://lwn.net/Articles/899543/

But symlinks are also incredibly popular on Linux. I don't think `west` should bend over backwards to support symbolic links BUT it should work reasonably well in simple cases like `/home/me -> /disk5/home/me`.

- #313
- https://github.com/zephyrproject-rtos/zephyr/issues/50284 EVERYTHING is a symlink in NixOS!!

carlescufi: @marc-hb  could you explain what you mean by "break this"? i.e. could you provide a way for me to reproduce the potential issue locally for me to understand it better before proposing a solution?

marc-hb:
> The logging with west -vv manifest --untracked is unexpected: It starts with prefixes but the new statements miss them?!

> Indeed, will fix

Still happening with 665c1ec132170e23c

src/west/app/project.py
marc-hb: > > I wonder whether symbolic links could break this, let me try.

>  i.e. could you provide a way for me to reproduce the potential issue locally for me to understand it better before proposing a solution?

Wait, I don't know whether I can "break" it yet! Please give me a moment.

marc-hb:
> I don't see function prefixes anymore?

I meant these prefixes: `west.manifest: DEBUG: `

carlescufi:
> > I don't see function prefixes anymore?
> 
> I meant these prefixes: `west.manifest: DEBUG: `

ah but that has nothing to do with this PR. That is (I think) something to do with the use of `_logger.debug` in `west.manifest` (the library code) vs `self.dbg` in app commands. Perhaps @pdgendt has more background on this.

src/west/app/project.py
```diff
@@ -640,6 +645,51 @@ def _die_if_manifest_project_filter(self, action):
...
+                            break
+                        elif e in ppath.parents:
+                            self.dbg(f'recursing into: {e}')
+                            _find_untracked(e)
+                            break
+                    else:
+                        # No match, untracked folder
+                        untracked.append(e)
+                        continue
+                else:
```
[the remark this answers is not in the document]
marc-hb: You're not a fan of "Return Early" in general? Just curious.

src/west/app/project.py
```diff
@@ -640,6 +652,68 @@ def _die_if_manifest_project_filter(self, action):
...
+                        break
+                else:
+                    # There is not a project and there is no project inside.
+                    # Add to untracked elements.
+                    untracked.append(e)
+                    continue
+
+        # Avoid using Path.walk() since that returns all files and folders under
+        # a particular folder, which is overkill in our case. Instead, recurse
+        # only when required.
```
marc-hb, line 703: I think this comment can be dropped because it should be obvious how smart and efficient is the recursion above. Some directories have a LOT of files and directories, especially some build directories! That would make this command orders of magnitude slower.

marc-hb (commented):
I didn't find a way to "break" this with symbolic links but I don't have infinite time and I only tried simple usage. However I found a simple case where things seem to work... by chance! So, see small `is_symlink()` suggestion below.

On the other hand, I noticed this:

```
.abspath: absolute and normalized path to the project
.posixpath: like abspath, but in posix style, that is, with '/'
```

This also reminded me this nasty 3.13 issue: 85fd5979f2e

I rarely ever use Windows so I'm not in a position to test or even review the code with respect to slash and drive letter pitfalls. Could you:
- Check that this pitfalls don't affect the code here?
- If needed, add comments wherever applicable to make sure Linux people don't inadvertently break it in the future?

Again, this is because we have no test for this feature.

marc-hb:
>  That is (I think) something to do with the use of _logger.debug in west.manifest (the library code) vs self.dbg in app commands. Perhaps @pdgendt has more background on this.

Maybe related to https://github.com/zephyrproject-rtos/west/issues/149#issuecomment-1447000986 ? Wild guess sorry, I need to timebox this for today.

> ah but that has nothing to do with this PR.

Hopefully :-) As long as this is not using the "wrong" logging method and adding technical debt, I don't mind the inconsistent debug logs much.

src/west/app/project.py
carlescufi: I cannot use this @marc-hb. When projects are inactive (or even uncloned!) the folder is not present on disk, so `p.exists()` would not include those.

src/west/app/project.py
```diff
@@ -640,6 +652,68 @@ def _die_if_manifest_project_filter(self, action):
...
+
+        def _find_untracked(folder):
+            '''There are three cases for each element in a folder:
+            - It's a project -> Do nothing, ignore the folder.
+            - There are no projects inside -> add to untracked list.
+            - It's not a project folder but there are some projects inside it -> recurse.
+            The folder argument cannot be inside a project, otherwise all bets are off.
+            '''
+            self.dbg(f'looking for untracked files/folders in: {folder}')
+            for e in [e.absolute() for e in folder.iterdir()]:
```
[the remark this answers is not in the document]
carlescufi: It is, but just like you said, better safe than sorry!

src/west/app/project.py
```diff
@@ -640,6 +652,68 @@ def _die_if_manifest_project_filter(self, action):
...
+                if not e.is_dir():
```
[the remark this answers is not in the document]
carlescufi: I have no problem adding this for safety.

src/west/app/project.py
carlescufi: Here I must disagree. I think it's good for future readers to explain why we did not take the easy way out with `Path.walk()`, I do not agree with the statement that it should be obvious.

carlescufi:
> I rarely ever use Windows so I'm not in a position to test or even review the code with respect to slash and drive letter pitfalls. Could you:
> 
> * Check that this pitfalls don't affect the code here?

I can check with Windows, but I reviewed the code and I am quite convinced there are no pitfalls in this case, because we just compare absolute paths to each other.

> * If needed, add comments wherever applicable to make sure Linux people don't inadvertently break it in the future?

Sure, but I have found no need for them so far.

src/west/app/project.py
marc-hb: @carlescufi found a great test case where my suggestion changes the output:

- project `foo/bar/baz` is not cloned and ppath.exists() is false
- User creates  `foo/bar/untracked1`, `foo/bar/untracked2`, ...

Then the output differs.

I think this demonstrates that tests are the "real" specification and the only way to formalize ambiguities.

pdgendt:
> @marc-hb Regarding moving this to `west status --untracked`, I think we can have it there, but it is a bit misleading because:
> 
> 1. `west status` operates on projects, not the workspace
> 2. `status` implies general status, not just untracked
> 
> But, if you really prefer it, I won't object.

I initially suggested to add it to `west manifest` so that would still be my first choice, but no strong objections to moving it either.

> However, regarding the tracking of inactive and uncloned projects I must respectfully disagree. While I understand the analogy with the git branch, for me this is not comparable (and that's also why I prefer `manifest` to `status`). For me, `west manifest --untracked` shows you the folders that are not managed, and will never be managed, by west. This is easy to understand for users and is conceptually simple and solid. If a user has their own folder in a location where a currently inactive or uncloned project exists, `west update` will fail (as it should). The fact that the user does not see that when using `--untracked`, in my opinion is not only not a real issue, given the unlikelihood of the use case and the fact that `west update` will give a clear error message, but distorts the meaning of the `--untracked` concept. That said, I understand that both approaches are to an extent reasonable, and so this becomes a choice to be made. I am happy to let other users chime in to see what they'd prefer.

I agree with @carlescufi, the current proposal is simpler and more intuitive IMHO. I sometimes re-use a west workspace by changing the `manifest.path` and `manifest.file` in `.west/config`

src/west/app/project.py
```diff
@@ -640,6 +652,68 @@ def _die_if_manifest_project_filter(self, action):
...
+        # Remove the .west folder, which is maintained by west
+        try:
+            untracked.remove((Path(self.topdir) / Path(WEST_DIR)).resolve())
+        except ValueError:
+            self.die(f'Folder {WEST_DIR} not found in workspace')
+
+        # Sort the results for displaying to the user.
+        untracked.sort()
+        for u in untracked:
+            self.inf(u.relative_to(Path.cwd(), walk_up=True))
```
pdgendt, line 715: `walk_up` was added in Python 3.12 and will not work on older versions.

carlescufi: Good catch! I now have to find an alternative.

pdgendt:
> EDIT: One more use case that I think becomes really ugly with @marc-hb's propose approach. Let's say you are switching projects from active to inactive using the group filter, and that is part of your workflow. With your proposal, actual (inactive) west projects cloned by west would show as untracked when you disable a group filter after having it enabled! I think that is very ugly

Agreed on the addition here, besides that I sometimes edit the `manifest.path` or `manifest.file` in my `.west/config` file, and the current proposal would work better in seeing what I left behind from switching over.

src/west/app/project.py
carlescufi: Fixed, thanks @pdgendt

marc-hb:
> Let's say you are switching projects from active to inactive using the group filter, and that is part of your workflow. With your proposal, actual (inactive) west projects cloned by west would show as untracked when you disable a group filter after having it enabled! I think that is very ugly

No, because `.is_cloned()` is true in that case. It would only show "mistakes" that break `west update`.

marc-hb:
> walk_up was added in Python 3.12 and will not work on older versions.

The most basic test would have found that.

> Agreed on the addition here, besides that I sometimes edit the manifest.path or manifest.file in my .west/config file, and the current proposal would work better in seeing what I left behind from switching over.

I don't see why sorry.

Maybe we don't need a test for this example specifically, but the number of discussions about corner cases and specifying their behavior changed my mind and convinced me this feature is complex enough to require at least one test to get it started. It's really not that simple of a feature.

carlescufi:
> > Let's say you are switching projects from active to inactive using the group filter, and that is part of your workflow. With your proposal, actual (inactive) west projects cloned by west would show as untracked when you disable a group filter after having it enabled! I think that is very ugly
> 
> No, because `.is_cloned()` is true in that case. It would only show "mistakes" that break `west update`.

OK, I was referring to your earlier proposal using `.exists()` instead of `is_cloned()`.
But I will reiterate my statement before `status`, `diff`, `compare` do not operate on the workspace at all, they operate on the individual Git projects. That is also why I prefer this in `west manifest`, which operates on the manifest and the workspace. I also dislike that the output depends on whether the projects are cloned or not, because, as I've mentioned before, I think this command should restrict itself to analyzing the parts of the workspace that are not covered by the manifest.

But I have to agree with you that, while I do not share your view, I can understand why others would prefer your approach, that's why I've asked or additional opinions.

>Maybe we don't need a test for this example specifically, but the number of discussions about corner cases and specifying their behavior changed my mind and convinced me this feature is complex enough to require at least one test to get it started. It's really not that simple of a feature.

Completely agree here. I am working on a basic test.

marc-hb:
> But I will reiterate my statement before status, diff, compare do not operate on the workspace at all, they operate on the individual Git projects.

They operate on the list of `.is_cloned()` projects and ignore everything else. So a command considering projects-that-could-be would be new and an (admittedly minor) inconsistency.

> That is also why I prefer this in west manifest, which operates on the manifest and the workspace.

Fair.

> I also dislike that the output depends on whether the projects are cloned or not, because, as I've mentioned before, I think this command should restrict itself to analyzing the parts of the workspace that are not covered by the manifest.

This sounds like the manifest is a very static thing and this just reminded that because of `import` it is not really static. This causes plenty of tricky corner cases and "dynamic" manifest dilemma 

- https://github.com/zephyrproject-rtos/west/labels/Partial%20imports
- #652 
- etc.

Curious how this new feature deals with partial and changing imports.

> But I have to agree with you that, while I do not share your view, I can understand why others would prefer your approach, that's why I've asked or additional opinions.

If it is really just a one-line change (not 100% sure of that yet), then maybe we could have both based on some additional flag? I don't mind which one we get first and what the default would be.

BTW, did you ever consider and discuss a brand `west untracked` command?

> Completely agree here. I am working on a basic test.

Thank you! That has become my biggest concern now.

carlescufi:
@marc-hb why I don't like operating on `is_cloned()` is that that function does not check whether the "right repo" is cloned. It only [checks if there's a Git repo](https://github.com/zephyrproject-rtos/west/blob/6202160b797435e754c8daef46aca5f3c9c934b5/src/west/manifest.py#L1037-L1055) at all. So for me then this is confusing, because if I have another Git repo in it, sure, `west update` will not fail because it adds a remote and just fetches, and that will work with any Git repo, but it will leave a mix of repos in the folder.
That's why I would rather not complicate it and use the simplest approach.

Alternatively, the only viable option to this is `west untracked` and then add some sort of `--all` or `--uncloned` to enable the behaviour you'd like. But I think this is overkill.

carlescufi:
@marc-hb @pdgendt added a test for the feature.

src/west/app/project.py
```diff
@@ -584,6 +584,13 @@ def __init__(self):
               If this file uses imports, it will not contain all the
               manifest data.
 
+            - --untracked print all files and folders inside the workspace that
+              are not tracked or managed by west. This effectively means any
+              file or folder that lays outside all of the projects' folders on
+              disk. This is similar to `git status` for untracked files. The
+              output format is relative to the current working directory and is
+              stable and suitable as input for scripting.
```
pdgendt, line 592: "Folder" is a Windows thing 🤢, and not used anywhere else in the project.

```suggestion
            - --untracked: print all files and folders inside the workspace that
              are not tracked or managed by west. This effectively means any
              file or directory that is outside all of the projects' directories on
              disk. This is similar to `git status` for untracked files. The
              output format is relative to the current working directory and is
              stable and suitable as input for scripting.
```

src/west/app/project.py
```diff
@@ -604,6 +611,9 @@ def do_add_parser(self, parser_adder):
                            exiting with an error if there are issues''')
         group.add_argument('--path', action='store_true',
                            help="print the top level manifest file's path")
+        group.add_argument('--untracked', action='store_true',
+                           help='''print any files and folders not managed or
+                           tracked by west''')
```
pdgendt, line 616:
```suggestion
        group.add_argument('--untracked', action='store_true',
                           help='''print all files and directories not managed or
                           tracked by west''')
```

src/west/app/project.py
```diff
@@ -640,6 +652,70 @@ def _die_if_manifest_project_filter(self, action):
                      'the manifest while projects are made inactive by the '
                      'project filter.')
 
+    def _untracked(self):
+        ''' "Performs a top-down search of the west topdir,
+        ignoring every folder that corresponds to a west project.
+        '''
```
pdgendt, line 658:
```suggestion
        ''' "Performs a top-down search of the west topdir,
        ignoring every directory that corresponds to a west project.
        '''
```

src/west/app/project.py
```diff
@@ -640,6 +652,70 @@ def _die_if_manifest_project_filter(self, action):
...
+        ppaths = []
+        untracked = []
+        for project in self._projects(None):
+            # We do not check for self.manifest.is_active(project) because
+            # inactive projects are still considered "tracked folders".
+            ppaths.append(Path(project.abspath))
+
+        # Since west tolerates nested projects (i.e. a project inside the folder
+        # of another project) we must sort the project paths to ensure that we
+        # hit the "enclosing" project first when iterating.
```
pdgendt, line 668:
```suggestion
        # Since west tolerates nested projects (i.e. a project inside the directory
        # of another project) we must sort the project paths to ensure that we
        # hit the "enclosing" project first when iterating.
```

src/west/app/project.py
```diff
@@ -640,6 +652,70 @@ def _die_if_manifest_project_filter(self, action):
...
+                        break
+                    elif e in ppath.parents:
+                        self.dbg(f'recursing into: {e}')
+                        _find_untracked(e)
+                        break
+                else:
+                    # This is not a project and there is no project inside.
+                    # Add to untracked elements.
+                    untracked.append(e)
+                    continue
```
pdgendt, line 699: Rename the `folder` variable name and text here.

src/west/app/project.py
```diff
@@ -640,6 +652,70 @@ def _die_if_manifest_project_filter(self, action):
...
+
+        # Avoid using Path.walk() since that returns all files and folders under
+        # a particular folder, which is overkill in our case. Instead, recurse
+        # only when required.
```
pdgendt, line 703:
```suggestion
        # Avoid using Path.walk() since that returns all files and directories under
        # a particular directory, which is overkill in our case. Instead, recurse
        # only when required.
```

src/west/app/project.py
```diff
@@ -640,6 +652,70 @@ def _die_if_manifest_project_filter(self, action):
...
+        _find_untracked(Path(self.topdir))
+
+        # Remove the .west folder, which is maintained by west
```
pdgendt, line 706:
```suggestion
        # Exclude the .west directory, which is maintained by west
```

src/west/app/project.py
```diff
@@ -881,7 +957,10 @@ def __init__(self):
             'status',
             '"git status" for one or more projects',
             '''Runs "git status" for each of the specified projects.
-            Unknown arguments are passed to "git status".''',
+            Unknown arguments are passed to "git status".
+
+            Note: If you are looking to find untracked files and folders
+            in the workspace use "west manifest --untracked".''',
```
pdgendt, line 963:
```suggestion
            Note: If you are looking to find untracked files and directories
            in the workspace use "west manifest --untracked".''',
```

pdgendt (commented):
`s/folders/directories/` and `s/folder/directory/` 😅

pdgendt (approved):
Re-approved

mbolivar:
@carlescufi I know you've been looking for this feature for a long time. Thanks for making the effort to work on it.

Given that a lot of the discussion centered around the behavior related to symlinks, I think it would make sense to have at least some basic coverage of the behavior of this command when symbolic links are encountered.

Since that is not always possible on Windows, I would recommend a separate test function like this:

```py
@pytest.mark.skipif(sys.platform.startswith("win"), reason="symbolic links not tested on Windows")
def test_manifest_untracked_with_symlinks():
    # basic test coverage for desired behavior identified by yourself and marc-hb goes here
```

(see https://docs.pytest.org/en/7.1.x/how-to/skipping.html)

I'm not going to -1 over this, but I think it would give us all some peace of mind to bite the bullet and get those tests in.

mbolivar (approved):

carlescufi:
Last push:

- Added a bit more documentation on the help of --untracked
- Added a few more checks in the test

PR #787, opened 2025-03-20
.github: Harden action workflows

pdgendt:
Some improvements to the Github workflows:

- Pin action version hashes
- Set explicit default permission
- Add a dependabot configuration to check for updates

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,13 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"
+    commit-message:
+      prefix: "ci: github: "
+    labels: []
+    groups:
+      actions-deps:
+        patterns:
+          - "*"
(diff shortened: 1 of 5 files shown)


kartben (approved):

pdgendt:
> Is there a bit more context for this?
> 
> I imagine the idea is to review action changes before upgrading them which sounds more secure in theory but... who is going to actually perform the audit in practice? Or maybe the idea is to just delay the upgrade, hoping someone _else_'s supply chain gets compromised first and finds out first?

So there has been some [fuss](https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised) around a compromised action, also used in the zephyr repository.
These changes are recommended by the [Open Source Security Foundation](https://openssf.org/) and similar efforts are being done for zephyr:
https://github.com/zephyrproject-rtos/zephyr/pull/87311
https://github.com/zephyrproject-rtos/zephyr/pull/87254
https://github.com/zephyrproject-rtos/zephyr/pull/87153
https://github.com/zephyrproject-rtos/zephyr/pull/87309

This is a first step, more will probably follow.

PR #794, opened 2025-03-26
app: list: additional support for inactive projects

carlescufi:
See commit messages.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -428,6 +428,8 @@ class List(_ProjectCommand):
             epilog=f'''\
 {ACTIVE_PROJECTS_HELP}
 
+Note: To list only inactive projects you can use --inactive.
+
 FORMAT STRINGS
 --------------
 
@@ -456,8 +458,11 @@ The following arguments are available:
 - clone_depth: project clone depth if specified, "None" otherwise
 - groups: project groups, as a comma-separated list
 ''')
-        parser.add_argument('-a', '--all', action='store_true',
-                            help='include inactive projects'),
+        group = parser.add_mutually_exclusive_group(required=False)
+        group.add_argument('-a', '--all', action='store_true',
+                           help='include inactive projects'),
+        group.add_argument('-i', '--inactive', action='store_true',
+                           help='list only inactive projects'),
         parser.add_argument('--manifest-path-from-yaml', action='store_true',
                             help='''print the manifest repository's path
                             according to the manifest file YAML, which may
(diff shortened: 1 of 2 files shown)


src/west/app/project.py
```diff
@@ -428,6 +428,9 @@ def do_add_parser(self, parser_adder):
             epilog=f'''\
 {ACTIVE_PROJECTS_HELP}
 
+Note: To list only inactive projects you can use --inactive. Giving an explicit
```
carlescufi, line 431: The only question in my mind about this feature is whether `--inactive` should filter out projects when they are explicitly provided by the user. No particular preference from my side.

So, should:
`west list -i my-foo-active my-bar-inactive` print both foo and bar or only bar?

pdgendt: My first thought would be to have only print bar, so `-i` would invert the printed projects.

carlescufi: But remember that `west list foo bar` prints **both** `foo` and `bar`!

src/west/app/project.py
```diff
@@ -491,15 +498,23 @@ def cloned_thunk(project):
...
+
         def delay(func, project):
             return DelayFormat(partial(func, project))
 
         for project in self._projects(args.projects):
-            # Skip inactive projects unless the user said
+            # Include the project based on the inactive flag. If the flag is
+            # set, include only inactive projects. Otherwise, include only
+            # active ones.
+            include = self.manifest.is_active(project) != args.inactive
```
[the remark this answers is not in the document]
carlescufi: Ooops good point. Surprised that the tests pass.

src/west/app/project.py
carlescufi: Yes, I think it's the "safe" choice here.

joerchan: Make sense to disallow that combination, asking which of list of projects is inactive does not sound like a very useful thing to do to me.

carlescufi: Implemented.

joerchan (approved):
+1 for the feature, but have not reviewed the implementation.

pdgendt (approved):

src/west/app/project.py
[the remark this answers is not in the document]
pdgendt: Apparently `store_true` arguments default to `False`, sorry for the noise.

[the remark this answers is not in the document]
carlescufi: Oh right. So this was superfluous, but not a big deal, will leave it as is.

PR #802, opened 2025-04-09
project: Pin Python dependencies

pdgendt:
Generate pinned requirement files from pyproject.toml for tox.

Additionally create pinned requirement files for the Github workflows.


requirements-tox.txt
```diff
@@ -0,0 +1,91 @@
...
+#    uv pip compile --universal --python-version 3.9 --generate-hashes - -o requirements-tox.txt
+cachetools==5.5.2 \
+    --hash=sha256:1a661caa9175d26759571b2e19580f9d6393969e5dfca11fdb1f947a23e640d4 \
+    --hash=sha256:d26a22bcc62eb95c3beabd9f1ee5e820d3d2704fe2967cbe350e20c8ffcd3f0a
+    # via tox
+chardet==5.2.0 \
+    --hash=sha256:1b3b6ff479a8c414bc3fa2c0852995695c4a026dcd6d0633b2dd092ca39c1cf7 \
+    --hash=sha256:e1cf59446890a00105fe7b7912492ea04b6e6f06d4b742b2c788469e34c82970
+    # via tox
+colorama==0.4.6 \
```
marc-hb, line 11: Can `colorama` really affect release output? That sounds like serious security design  attack surface issues if it can.

.github/workflows/package.yml
```diff
@@ -15,12 +15,17 @@ jobs:
         uses: actions/setup-python@8d9ed9ac5c53483de85588cdf95a591a75ab9f55 # v5.5.0
         with:
           python-version: "3.13"
+          cache: pip
+          cache-dependency-path: requirements-build.txt
```
marc-hb, line 19: Could you maybe add a one-line comment clue for people not familiar with this `cache` feature? Is it part of the security or just an optimization?

requirements-build.txt
```diff
@@ -0,0 +1,62 @@
...
+    --hash=sha256:c954d2250168d28797dd4e3ac5cf812a406cd5a92674ee4c8f123c889786aa8e \
+    --hash=sha256:cb55c73c5f4408779d0cf3eef9f762b9c9f147a77de7b258bef0a5628adc85cc \
+    --hash=sha256:cd45e1dc79c835ce60f7404ec8119f2eb06d38b1deba146f07ced3bbc44505ff \
+    --hash=sha256:d3f5614314d758649ab2ab3a62d4f2004c825922f9e370b29416484086b264ec \
+    --hash=sha256:d920f33822747519673ee656a4b6ac33e382eca9d331c87770faa3eef562aeb2 \
+    --hash=sha256:db2b95f9de79181805df90bedc5a5ab4c165e6ec3fe99f970d0e302f384ad222 \
+    --hash=sha256:e59e304978767a54663af13c07b3d1af22ddee3bb2fb0618ca1593e4f593a106 \
+    --hash=sha256:e85e99945e688e32d5a35c1ff38ed0b3f41f43fad8df0bdf79f72b2ba7bc5272 \
+    --hash=sha256:ece47d672db52ac607a3d9599a9d48dcb2f2f735c6c2d1f34130085bb12b112a \
+    --hash=sha256:f4039b9cbc3048b2416cc57ab3bda989a6fcf9b36cf8937f01a6e731b64f80d7
```
marc-hb, line 57: Any idea why some dependencies need a crazy number of hashes?

requirements-ci.txt
```diff
@@ -0,0 +1,346 @@
+# This file was autogenerated by uv via the following command:
+#    uv pip compile --universal --python-version 3.9 --generate-hashes --group ci -o requirements-ci.txt pyproject.toml
```
marc-hb, line 2: Why do we need the same hashes 4 times? Is there no way to include and re-use?

marc-hb (commented):
first impressions

requirements-ci.txt
pdgendt: Also installing all the different requirement files would simply skip already installed packages. The hashes are identical.

marc-hb: Thanks for your patience, I'm slowly making sense of all this.

I don't see why we need to pin tox dependencies. I hope tox is not involved in the release process, is it? I mean, I hope there is no way a compromised `tox` could affect the release output in any way, correct?

pdgendt: It's one of the medium security issues checked by the OpenSSF tool: https://github.com/zephyrproject-rtos/west/security/code-scanning/26

kartben: I think you're focusing on "released bits" too much and missing the bigger picture.

The recent supply chain attack that triggered all this hardening work highlighted that a lot of bad things can happen in CI too, if we just randomly pull untrusted packages.
Arguably west CI probably has a lot less things that can leak compared to Zephyr CI but imagine someone able to publish a new colorama package that does funny business, printing env variables and what not on stdout (more or less what was happening with the aforementioned supply chain attack, except it was JS code in a GH action - https://www.cisa.gov/news-events/alerts/2025/03/18/supply-chain-compromise-third-party-tj-actionschanged-files-cve-2025-30066-and-reviewdogaction). We run West test suite in CI... Env variables get printed for everyone looking at workflow runs to see... Plus all the other similar scenarios we're not even thinking of. In the context of Zephyr CI this could mean people able to get malicious JS added to the docs website that CI regularly publishes, getting access to our AWS tokens, ...

requirements-build.txt
pdgendt: Every distribution of a package (specific python version, platform, output format like wheels vs tarball) needs to be added to the list of hashes.

requirements-ci.txt
marc-hb: > I think you're focusing on "released bits" too much and missing the bigger picture.

On the contrary, I'm trying to think in terms of threat models, attack surface and security principles instead of blindly trying to [fix warnings](https://github.com/zephyrproject-rtos/west/security/code-scanning/26) with a "every dependency looks like a nail" approach. That does not mean I'm right and thanks for trying to correct me but it means commit messages and reviews should have a little bit of security rationale and discussions, not just "let's apply this universal recipe because someone said so".

> imagine someone able to publish a new colorama package that does funny business,...

If you read my comments above, you can see it is indeed what I've been trying to imagine.

> ... printing env variables and what not on stdout.

The  "least privilege" is one of the most important security principles. So the critical questions I'm asking here are: 1. Does GitHub have any concept of a "read-only", less privileged action 2. Are workflows for test tools (tox) and scanners (CodeQL) in such a low privilege class? If it exists, they definitely should be. 3. What is the extend of "funny business" such a "read-only" class do? It should certainly not be able to print secrets when it needs access to zero secret to do its job.

If the answer is "No, GitHub does not know least privilege" then OK, let's sob, write it down and pin everything. But someone must really check and be able to clearly answer these basic questions. Not just for west, very far from it.

The problem with security "theater" is the false impression of security it gives. Example: "Dependencies are pinned so we're safe, no need to care about what they have access to".

Another critical issue is usability: when security becomes tedious, people stop paying attention. For instance, imagine you have a _routine_ where you have to update a ton of checksums: some for dependencies which are critical, others not. In theory, you're supposed to audit the corresponding updates but who has time to do that for so many of them?  So instead you just approve and merge... a recent backdoor. This is not fictional: we barely started pinning and rubber stamping already started to happen in #800 - on the very first routine update.

pdgendt: > 1. Does GitHub have any concept of a "read-only", less privileged action 2. Are workflows for test tools (tox) and scanners (CodeQL) in such a low privilege class? If it exists, they definitely should be. 3. What is the extend of "funny business" such a "read-only" class do? It should certainly not be able to print secrets when it needs access to zero secret to do its job.

Even in the "read-only" case we want to ensure that our simplest CI actions are running the tools we expect them to run. Be it `setuptools`, `build`, `colorama`, `tox`, ... really any tool or package. As @kartben mentioned, access to CI already opens that "read-only" view.
So if that's what you're referring to as "least privilege" then I guess we can't trust Github for that matter.

> The problem with security "theater" is the false impression of security it gives. Example: "Dependencies are pinned so we're safe, no need to care about what they have access to".
>
> Another critical issue is usability: when security becomes tedious, people stop paying attention. For instance, imagine you have a routine where you have to update a ton of checksums: some for dependencies which are critical, others not. In theory, you're supposed to audit the corresponding updates but who has time to do that for so many of them? So instead you just approve and merge... a recent backdoor. This is not fictional: we barely started pinning and rubber stamping already started to happen in https://github.com/zephyrproject-rtos/west/pull/800 - on the very first routine update.

I'm putting a lot of effort in trying to get things right, and learning a lot in the process. But, it isn't an easy matter, and following best practices as recommended by the OpenSSF would be my first choice of moving forward. The current implementation is my interpretation of those best practices, and I'm happy to go with better options if need be.
I've created another PR #803 that would hopefully help alongside the dependency update mechanism. Yes, _yet another tool_ that can help reduce the burden of trying to track the world and its vulnerabilities.

Lastly, I want to note that security isn't the only goal here, the pinning helps in traceability and reproducibility, both for releases and testing.

marc-hb: > I'm putting a lot of effort in trying to get things right, and learning a lot in the process. 

Indeed - and maintainer burnout is a very real thing. So, security theater overhead is the last thing we want. Time-consuming things must be done for clearly stated reasons and benefits, not just because "everyone does this" or because "everything looks like a nail to pin".

> Lastly, I want to note that security isn't the only goal here, the pinning helps in traceability and reproducibility, both for releases and testing.

I know a little bit of about build reproducibility (https://github.com/zephyrproject-rtos/zephyr/pull/14593, https://github.com/zephyrproject-rtos/zephyr/pull/51954#issuecomment-2555373460, ...) and again, there is a huge difference between what affects the build output versus not. Pure test dependencies do obviously not affect the build. Now, reproducing test _results_ matters too but 1) that's a very different thing, and 2) _recording_ versions is enough for that, not need to pin there.

requirements-tox.txt
marc-hb, line 1: I'm confused by the file names: `requirements-tox.txt` is used in `.github/` CI while `requirements-ci.txt` is used in `tox.ini`... is this intended?

Automation aside, after this PR how can I simply test the latest version of all dependencies and in my workspace and detect any new incompatibility or regression in them?

pdgendt: Yes, the file names maybe aren't clear, currently:
`requirements-tox.txt` -> the dependency chain to install `tox` itself
`requirements-ci.txt` -> the dependencies to run `tox`
`requirements.txt` -> the dependencies of `west`
`requirements-build.txt` -> the dependencies to create a package

> Automation aside, after this PR how can I simply test the latest version of all dependencies and in my workspace and detect any new incompatibility or regression in them?

I was thinking about making this easier too, I can add a script that would create/update the requirements files (without dependabot).

requirements-ci.txt
marc-hb: > Even in the "read-only" case we want to ensure that our simplest CI actions are running the tools we expect them to run. Be it setuptools, build, colorama, tox, ... really any tool or package.

That's just a statement without a rationale.

> As @kartben mentioned, access to CI already opens that "read-only" view. So if that's what you're referring to as "least privilege" then I guess we can't trust Github for that matter.

It's not about "trusting GitHub.com" in general, I agree there is not a lot we can do there. It's about trusting project-specific, automated workflows and controlling their respective access levels. 

So are you saying that all GitHub workflows must have the same, full access level no matter what side effects they need? In other words, running `pylint` or creating a new, official release requires the same level of complete access? Not having something that simple and that effective would be very sad, even more sad when compared to the incredibly fine-grained https://docs.github.com/en/organizations/managing-peoples-access-to-your-organization-with-roles/roles-in-an-organization that no one ever fully understands and to all the incoming CodeQL, pinning and Dependabot complexity.

requirements-build.txt
marc-hb: Isn't there a way to "checksum the list of checksums" and reduce the number?

Checksums are meant to be compared, that's their only purpose. Often enough, they need to be compared by humans but no one is going to compare **dozens** of checksums visually.

pdgendt: I don't think so, not at least where other tooling (dependabot) can still parse them.

marc-hb: > not at least where other tooling (dependabot) can still parse them.

It does not have to be mutually exclusive. We could have both the list and its sum (if there were such a sum)

pdgendt: Not sure what to do here, this is generated using tools I frequently use.

requirements-ci.txt
pdgendt: > > Even in the "read-only" case we want to ensure that our simplest CI actions are running the tools we expect them to run. Be it setuptools, build, colorama, tox, ... really any tool or package.
> 
> That's just a statement without a rationale.

Apologies, this was not intentional, I guess I am unable to answer your initial question.
 
> > As @kartben mentioned, access to CI already opens that "read-only" view. So if that's what you're referring to as "least privilege" then I guess we can't trust Github for that matter.
> 
> It's not about "trusting GitHub.com" in general, I agree there is not a lot we can do there. It's about trusting project-specific, automated workflows and controlling their respective access levels.
> 
> So are you saying that all GitHub workflows must have the same, full access level no matter what side effects they need? In other words, running `pylint` or creating a new, official release requires the same level of complete access? Not having something that simple and that effective would be very sad, even more sad when compared to the incredibly fine-grained [docs.github.com/en/organizations/managing-peoples-access-to-your-organization-with-roles/roles-in-an-organization](https://docs.github.com/en/organizations/managing-peoples-access-to-your-organization-with-roles/roles-in-an-organization) that no one ever fully understands and to all the incoming CodeQL, pinning and Dependabot complexity.

@marc-hb I appreciate your efforts and value your opinion, but in the context of the PR, I have no clue what to do with this comment. Is this a showstopper? Can you give any suggestions on what would be a better workflow/solution?
I'd like to move forward with this, as I think it is an improvement, both for testing and packaging. It follows best practices. It relies on tooling, yes, but alongside automation that should help maintainers.

marc-hb: > > So are you saying that all GitHub workflows must have the same, full access level no matter what side effects they need? In other words, running pylint or creating a new, official release requires the same level of complete access? 

Do you know the answer?

henrikbrixandersen: > > > So are you saying that all GitHub workflows must have the same, full access level no matter what side effects they need? In other words, running pylint or creating a new, official release requires the same level of complete access? 
> 
> 
> 
> Do you know the answer?

The closest, I can find, is this: https://docs.github.com/en/actions/writing-workflows/workflow-syntax-for-github-actions#jobsjob_idpermissions

requirements-tox.txt
pdgendt: Renamed the files, now:

`requirements.txt` -> the dependencies of `west`
`requirements-tox.txt` -> the dependencies to run `tox`
`requirements-install-build.txt` -> the dependencies to create a package
`requirements-install-tox.txt` -> the dependencies to install `tox`

requirements-ci.txt
pdgendt: Yes, this is already applied to all current workflows with:
```yaml
permissions:
  contents: read
```

Any other permission is default to `none`:
> If you specify the access for any of these permissions, all of those that are not specified are set to none.

marc-hb:
> I have no clue what to do with this comment. Is this a showstopper? Can you give any suggestions on what would be a better workflow/solution?

Let me try to clarify.

- Least privilege: this is the most important, much more than pinning. All dependencies must have the minimum possible level of access to do their job. Because of configuration complexity, this does not come for free but it is much more important than pinning dependencies and it is also much better security "value" as it reduces attacks surface completely and permanently: no routine reviews of changes needed. Looks like you just checked that?

- "If it's not tested then it does not work": because GitHub permission levels seem complex, after configuring workflow permissions "someone" should attempt to perform side effects or expose secrets and make sure access is actually denied. This should be done again every time GitHub makes some change to how permissions work (hopefully: rarely).

- Privileged dependencies that do have a need to perform any side-effect ("write" access) or that have "read" access to any secret token or any other non-public data should absolutely be pinned like this PR does. I never questioned that. I'm afraid I might simplify GitHub access levels but you should get the idea.

- Usage of privileged dependencies should be reduced to an absolute minimum in the first place.

- Other, de-privileged dependencies maybe be pinned to some major version for functional reasons but they should IMHO **not** be micro-managed with checksums etc. because this unnecessarily and considerably increases the workload of routine updates and steals precious time from other, actually effective security tasks - like detailed and very time-consuming reviews of changes in privileged dependencies.

"Security theater" sounds harmless and even fun maybe. But it is actually a dangerous distraction because it steals scarce attention and brain cycles and because routine tasks are mind-numbing.

Speaking of scarce attention and distractions, OS images of runners offer a much larger attack surface. Are those all pinned too and who's going to review updates of entire images like Ubuntu and others? I noticed that GitHub performs "continuous" updates of its default runners. 
https://github.com/actions/runner-images/issues/741#issuecomment-1731799182
Not sure these can even be pinned?

marc-hb:
> Other, de-privileged dependencies maybe be pinned to some major version for functional reasons but they should IMHO not be micro-managed with checksums etc. because ...

Even if you disagree with this, you'll agree the risk level is much lower. So these could be done later - after learning from experience with privileged dependencies. "Smaller PRs".

carlescufi:
@marc-hb @henrikbrixandersen @kartben is anything here not ready for merging? It'd be great to move this forward.

kartben (approved):

requirements-tox.txt
marc-hb: If `colorama` is a dependency of `tox` only, then it does not need to be pinned to a specific SHA because `tox` should never have permission to affect source or releases.

Same for this entire requirements-tox.txt file.

marc-hb (commented):
> @marc-hb in particular, I am not sure by reading your comments whether you are OK with this change or not.

I'm OK with only "half" of it. Pinning `requirements.txt` makes sense, pinning `requirements-tox.txt` does not.

I'm very concerned about the perception that pinning is the latest security silver bullet. For years we were told to "update, update and update" to stay secure. Now it's the opposite. Which is it? Neither, it's complicated. And it can never be that "cheap".

marc-hb:
> I have to disagree here, pinning requirements-tox.txt is no different than pinning requirements.txt IMO.

I tried to phrase the difference in 2 or 3 different ways. Maybe none were well written, but your reply does not even try to mention any of them. I give up!

pdgendt:
> I tried to phrase the difference in 2 or 3 different ways. Maybe none were well written, but your reply does not even try to mention any of them. I give up!

I'm sorry, I may have difficulties explaining this, but requirements.txt is only relevant during CI, as is requirements-tox.txt. During installation of west itself, it isn't used.

marc-hb:
> I'm sorry, I may have difficulties explaining this, but requirements.txt is only relevant during CI, as is requirements-tox.txt. During installation of west itself, it isn't used.

I think that is still besides my main point.

> I would argue we still do "update, update, update", but I prefer to offload that to tooling, and in this case dependabot.

dependabot is nice but it does absolutely nothing to help with actual security audits. Except giving a bit of a false sense maybe.

marc-hb:
> It [homebrew] is super automated and pins EVERYTHING. 

Off-topic but since this thread is doomed anyway...

I've used homebrew on my mac since forever. Once, I tried to make a one-line change in one of the packages. That's the "essential" litmus test. I tried for hours and gave up. Not impressed.

carlescufi:
@marc-hb this has been now open for a while. My impression when reading your comments is that you disagree with pinning everything as a rule, but are not strictly opposed to using pinning as tool.

So here's my question: if @pdgendt only pinned `requirements.txt` but not `requirements-tox.txt` you'd be OK with this proposed change? If that's the case then I suggest @pdgendt makes a call on whether to pin only one or both. Also, if you are strictly opposed to pinning both then please NACK the PR. Thanks!

pdgendt:
I'm not in favour of doing only one, but not the other. I'll close this and not include it in v1.4, maybe try to come up with something for v2.0.
Not a hill I want to die on.

PR #803, opened 2025-04-10
.github: Add dependency review workflow

pdgendt:
Add the dependency review action to run against PRs and scan for vulnerabilities.
This will also list the licenses introduced.

### DIFF
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -1,0 +1,24 @@
+# Dependency Review Action
+#
+# This Action will scan dependency manifest files that change as part of a Pull Request,
+# surfacing known-vulnerable versions of the packages declared or updated in the PR.
+# Once installed, if the workflow run is marked as required,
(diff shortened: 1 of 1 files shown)


.github/workflows/dependency-review.yml
```diff
@@ -0,0 +1,24 @@
...
+
+jobs:
+  dependency-review:
+    runs-on: ubuntu-latest
+    steps:
+      - name: 'Checkout Repository'
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+
+      - name: 'Dependency Review'
+        uses: actions/dependency-review-action@67d4f4bd7a9b17a0db54d2a7519187c65e339de8 # v4
```
marc-hb, line 24: More security noise again?
- https://github.com/zephyrproject-rtos/west/pull/802#issuecomment-2798893236

marc-hb (approved):
Hey it even reviews itself: I like it.

PR #804, opened 2025-04-11
project: Move docker testing to a Github workflow

pdgendt:
Convert manual docker testing using docker-compose to a Github workflow.
This workflow should be manually started to reduce CI overhead.

We increase visibility and traceability with having action runs to refer to, instead of someone running these locally and dumping the log output somewhere.

~~Depends on #802 only the last commit is added on top.~~


marc-hb (commented):
It's great that GitHub provides this "for free" and it might provide better integration and test results may look better but how to debug now when a Fedora or Arch-specific regression is found? Long after it was introduced since this is now tested only at release time... I think the Docker configuration files should be kept for local usage.

Also, this should be tested weekly or monthly without waiting for release time.

pdgendt:
> It's great that GitHub provides this "for free" and it might provide better integration and test results may look better but how to debug now when a Fedora or Arch-specific regression is found? Long after it was introduced since this is now tested only at release time... I think the Docker configuration files should be kept for local usage.

You mean to keep bother docker-testing directory and the Github action?
I think the test is trivial enough to run on these distros? Why keep maintaining both?

> Also, this should be tested weekly or monthly without waiting for release time.

Sure, but this currently isn't done either. And wouldn't we prefer to test released versions instead then?

marc-hb:
> I think the test is trivial enough to run on these distros?

I see of lot lines deleted, were they too verbose and/or not useful? Maybe I'm missing something.

If it's trivial to reproduce distro-specific issues locally with docker and without all these config files, then maybe you can replace them with a short README-docker.md or something with manual steps?


> Why keep maintaining both?

Good question: why switch to GitHub if this only loses the ability to run locally?

.github/workflows/test-distros.yml
```diff
@@ -0,0 +1,60 @@
...
+      fail-fast: false
+      matrix:
+        container:
+          - "archlinux:latest"
+          - "debian:stable"
+          - "debian:testing"
+          - "fedora:latest"
+          - "fedora:rawhide"
+          - "opensuse/tumbleweed:latest"
+          - "ubuntu:devel"
```
marc-hb, line 31: Pardon my ignorance but where do all these images come from and are they "pinned" somehow? Also, what do they have access to? The volume of code potential attack surface of an entire distribution is orders of magnitude bigger than for a simple Github Action (#802)

(not like I remember where they came from in the previous method either, apologies)

marc-hb (commented):
> Sure, but this currently isn't done either.

Running existing workflows daily is surprisingly short and easy, see for example https://github.com/thesofproject/sof/actions/workflows/daily-tests.yml


> And wouldn't we prefer to test released versions instead then?

Both, why not. Same code.

pdgendt:
> > I think the test is trivial enough to run on these distros?
> 
> I see of lot lines deleted, were they too verbose and/or not useful? Maybe I'm missing something.
> 
> If it's trivial to reproduce distro-specific issues locally with docker and without all these config files, then maybe you can replace them with a short README-docker.md or something with manual steps?

For every distro it boils down to: install git+python+pip, install tox, checkout repo, run tox. Is it that complex that we need to keep both options?
The action workflow can be taken as the reference on how to produce it locally. 

> > Why keep maintaining both?
> 
> Good question: why switch to GitHub if this only loses the ability to run locally?

To gain visibility and traceability, it's is described in the release process to test on major linux distros, which can now be done with pointing to the action having a successful run instead of pasting the output.

.github/workflows/test-distros.yml
pdgendt: The pull logic is the same as running the images locally, they are fetched from [docker hub](https://hub.docker.com/).

The reason these aren't pinned is to actually validate the running heads, hence the reason it should be considered to be run manually when needed.

marc-hb: Thanks, so maybe we could have weekly runs with pinned images and manual runs without pinning?

But... is pinning less secure here? I mean, isn't pinning missing security updates? Aren't we supposed to upgrade operating systems all the time?

Funny how every dependency in #802 looks like a nail to pin but not here. Maybe there are differences and good reasons not to pin here but the corresponding rationales and policies should be summarized and referenced somewhere and I haven't seen this anywhere so far.

marc-hb (commented):
> For every distro it boils down to: install git+python+pip, install tox, checkout repo, run tox. Is it that complex that we need to keep both options?
The action workflow can be taken as the reference on how to produce it locally.

Yes - that's not the part what worries me. It's the preliminary Docker bit that does and that bit is "magically" performed by Github. So maybe a 2-3 lines comment that provides the right `docker ...` incantation to download any distro from dockerhub is enough.

> which can now be done with pointing to the action having a successful run instead of pasting the output.

Fair!

.github/workflows/test-distros.yml
carlescufi: This, IMHO, is outside the scope of this PR. I suggest we take that separately since this PR contains a good change that is worth getting in ASAP.

carlescufi (approved):
Looks good to me, thanks!

kartben (approved):

marc-hb:
> So maybe a 2-3 lines comment that provides the right docker incantation to download any distro from dockerhub is enough.

@pdgendt , ping?

Not everyone is familiar with Docker. The code this PR removes was complex and it's good to get rid of it - but it provided local, "push-button" reproduction of any distro-specific test failure. Now that's gone. I'm only asking for 2-3 sample commands that someone new to Docker can copy/paste locally.

PS: last I checked, file sharing between host and container was still a royal PITA. I digress a bit.

pdgendt:
> > So maybe a 2-3 lines comment that provides the right docker incantation to download any distro from dockerhub is enough.
> 
> @pdgendt , ping?
> 
> Not everyone is familiar with Docker. The code this PR removes was complex and it's good to get rid of it - but it provided local, "push-button" reproduction of any distro-specific test failure. Now that's gone. I'm only asking for 2-3 sample commands that someone new to Docker can copy/paste locally.
> 
> PS: last I checked, file sharing between host and container was still a royal PITA. I digress a bit.

RIght, I forgot about that comment, I'll add that.

marc-hb (approved):
> RIght, I forgot about that comment, I'll add that.

Thanks! I have not tested them but this is what I wanted and it looks simple enough.

PR #805, opened 2025-04-14
project: Add CodeCov reporting

pdgendt:
Add a Github workflow and configuration file for CodeCov. Create more visibility about the coverage in PRs.


nashif (approved):

.github/workflows/test.yml
```diff
@@ -23,9 +24,49 @@ jobs:
...
+      - name: Create coverage report
+        run: |
+          coverage combine coverage-*/.coverage
+          coverage xml
+
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@0565863a31f2c772f9f0395002a31e3f06189574 # v5.4.0
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          slug: zephyrproject-rtos/west
```
marc-hb, line 72:
```suggestion
          slug: ${{github.repository_owner }}/west
```

.github/workflows/test.yml
```diff
@@ -23,9 +24,49 @@ jobs:
...
+      - name: Install coverage
+        run: pip3 install coverage
+
+      - name: Create coverage report
+        run: |
+          coverage combine coverage-*/.coverage
+          coverage xml
+
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@0565863a31f2c772f9f0395002a31e3f06189574 # v5.4.0
```
marc-hb, line 69: More security noise again? :-(
- https://github.com/zephyrproject-rtos/west/pull/802#issuecomment-2798893236

marc-hb (commented):
This new tool looks great, thanks!

Small change recommended below.

.github/workflows/test.yml
pdgendt: Why? This is pinning the GitHub action which we already established in all the workflows.

marc-hb:
> This new tool looks great, thanks!

Forgot to say: these reports really nail the common knowledge that error handling is never tested, it's impressive!

While it would be ideal to have coverage for every `raise` statement, it would simply require way too much effort that would be much better spent elsewhere. So, I wonder whether CodeCov has any option to filter the lack of coverage of single-line `raise` statements? And maybe some other, `return ERRXX` too statements too. This would considerably increase the signal/noise ratio of their reports.

marc-hb (approved):
> It is possible, 

Great news! I had low CodeCov expectations there.

> but I'd like to create a baseline first, follow-up PRs for increasing coverage by either adding tests,

Yes of course, I didn't mean otherwise, sorry for not being clear.

pdgendt:
> > It is possible,
> 
> Great news! I had low CodeCov expectations there.

It's not done by CodeCov, but the Python coverage tooling itself.

> > but I'd like to create a baseline first, follow-up PRs for increasing coverage by either adding tests,
> 
> Yes of course, I didn't mean otherwise, sorry for not being clear.

Thanks!

Issue #667, opened 2023-06-01
Support `west manifest --resolve` and `--freeze` with `manifest.project-filter`


mbolivar-nordic:
#664 disabled `west manifest --resolve` and `west manifest --freeze` whenever `manifest.project-filter` is set. This is intended as a temporary limitation. This issue tracks an enhancement to west where the two are made compatible again.

The proposed semantics are:

- `inactive-by-default: true` (if implemented, see #666 ) and `manifest.project-filter` will NOT affect the output of `west manifest --resolve`
- corollary: if bsim includes `inactive-by-default: true` in the upstream zephyr manifest, users will need to activate bsim manually in order to be able to use west resolve, because e.g. `west manifest --resolve` will need to resolve the import in the bsim project definition. If the project is missing, the command should produce an error saying that `west update` is needed (to get bsim itself and all of the projects it imports).

mbolivar-nordic:
Expected behavior in an example:

```
manifest.project-filter == `-bsim` # forbidden to combine with west manifest --resolve today
west update bsim                   # updates bsim, but does not do its import; bsim remains inactive
west manifest --resolve            # works, since bsim is present
```

marc-hb:
Fix submitted in https://github.com/zephyrproject-rtos/west/pull/808 by  @pdgendt, please review.

Issue #491, opened 2021-03-20
`west manifest --freeze` fails since it doesn't ignore excluded groups


JelmerT:
When a group is disabled in a manifest, those project(s) don't get cloned, so no folder(s) is created.

When doing `west manifest --freeze` the command fails with 

`RuntimeError: cannot freeze; project XXXX is uncloned 
`

In #314 , where a repo has no commits, this would be expected behavior. But in this case where projects are specifically excluded, would this still be expected behavior?

If yes -> more robust handling would be nice instead of a stack trace.

marc-hb:
> But in this case where projects are specifically excluded, would this still be expected behavior?

I'm guessing not.

Speaking of stack traces: they're very useful... in bug reports :-)

JelmerT:
Now, by popular demand! The stack trace:
```
$ west manifest --freeze
Traceback (most recent call last):
  File "/usr/local/bin/west", line 8, in <module>
    sys.exit(main())
  File "/usr/local/lib/python3.9/site-packages/west/app/main.py", line 779, in main
    app.run(argv or sys.argv[1:])
  File "/usr/local/lib/python3.9/site-packages/west/app/main.py", line 106, in run
    self.run_command(argv)
  File "/usr/local/lib/python3.9/site-packages/west/app/main.py", line 336, in run_command
    cmd.run(args, unknown, self.topdir, manifest=self.manifest)
  File "/usr/local/lib/python3.9/site-packages/west/commands.py", line 129, in run
    self.do_run(args, unknown)
  File "/usr/local/lib/python3.9/site-packages/west/app/project.py", line 621, in do_run
    self._dump(args, manifest.as_frozen_yaml(**dump_kwargs))
  File "/usr/local/lib/python3.9/site-packages/west/manifest.py", line 1498, in as_frozen_yaml
    return yaml.safe_dump(self.as_frozen_dict(), **kwargs)
  File "/usr/local/lib/python3.9/site-packages/west/manifest.py", line 1475, in as_frozen_dict
    return self._as_dict_helper(pdict=pdict)
  File "/usr/local/lib/python3.9/site-packages/west/manifest.py", line 1430, in _as_dict_helper
    project_dicts = [pdict(p) for p in projects]
  File "/usr/local/lib/python3.9/site-packages/west/manifest.py", line 1430, in <listcomp>
    project_dicts = [pdict(p) for p in projects]
[...]

PR #808, opened 2025-04-17
manifest: Support --active-only argument for --resolve/--freeze

pdgendt:
EDIT: The discussions in #491 and #667 made it clear that `--freeze` and `--resolve` should not have their behavior changed.

Updated the PR to introduce `--active-only` for `--resolve/--freeze`.

~~Running `west manifest --freeze` expects projects to be cloned, but the purpose of inactive projects is to not clone these explicitly.~~

~~Currently it's not possible to call `west manifest --freeze` on a clean zephyr installment because of the `optional` projects.~~

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -638,6 +638,8 @@ class ManifestCommand(_ProjectCommand):
         group = parser.add_argument_group('options for --resolve and --freeze')
         group.add_argument('-o', '--out',
                            help='output file, default is standard output')
+        group.add_argument('--active-only', action='store_true',
+                           help='only resolve active projects')
 
         return parser
 
@@ -649,11 +651,13 @@ class ManifestCommand(_ProjectCommand):
         if args.validate:
             pass              # nothing more to do
         elif args.resolve:
-            self._die_if_manifest_project_filter('resolve')
-            self._dump(args, manifest.as_yaml(**dump_kwargs))
+            if not args.active_only:
+                self._die_if_manifest_project_filter('resolve')
+            self._dump(args, manifest.as_yaml(active_only=args.active_only, **dump_kwargs))
         elif args.freeze:
-            self._die_if_manifest_project_filter('freeze')
-            self._dump(args, manifest.as_frozen_yaml(**dump_kwargs))
+            if not args.active_only:
+                self._die_if_manifest_project_filter('freeze')
+            self._dump(args, manifest.as_frozen_yaml(active_only=args.active_only, **dump_kwargs))
         elif args.untracked:
             self._untracked()
         elif args.path:
@@ -666,7 +670,9 @@ class ManifestCommand(_ProjectCommand):
         if self.config.get('manifest.project-filter') is not None:
             self.die(f'"west manifest --{action}" is not (yet) supported '
                      'when the manifest.project-filter option is set. '
-                     'Please clear the project-filter configuration '
+                     f'Add --active-only to {action} only the projects '
+                     'currently active in the workspace. Alternatively, '
+                     'please clear the project-filter configuration '
                      'option and re-run this command, or contact the '
                      'west developers if you have a use case for resolving '
                      'the manifest while projects are made inactive by the '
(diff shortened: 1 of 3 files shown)


marc-hb (commented):
> LGTM

I spent more than 2s looking at the code and I take that back, sorry.

`_as_dict_helper()` already has an optional `pdict` argument which is already used to perform filtering. So, this PR scatters the project filtering in two different places; one "contextual" one and another, new, more "hardcoded" place. Does not look great. Would something like this make more sense? (UNTESTED)

```diff
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -1634,7 +1634,7 @@ class Manifest:
         Raises ``RuntimeError`` if a project SHA can't be resolved.
         '''
         def pdict(p):
-            if not p.is_cloned():
+            if not p.is_active():
                 raise RuntimeError(f'cannot freeze; project {p.name} '
                                    'is uncloned')
             try:

```

marc-hb:
> I don't think that would work? The pdict isn't filtering? It would throw.

Yes my bad, ignore. Sorry for the noise.

marc-hb:
This was already discussed in #491, #667 and probably others. At the time, a new command was suggested instead of changing the semantic of the existing one. Please review past discussions before proceeding with this.

> Discussion in
https://discord.com/channels/720317445772017664/733037524498382881/1362257185597554808

Discord is useful but its discussions are not public, please avoid use as a reference (try opening this links in private/incognito mode)

marc-hb:
> west manifest --freeze would throw an exception on the first uncloned repo 

So this never worked in a default or almost default configuration? Whatever happens, it sounds like more test coverage is needed! CodeCov & friends only know so much.,,

EricNRS:
Here is the original bug report from Discord that prompted this change.

### Bug report from the Discord discussion

Any tips on how to get west manifest --freeze to work?  When I run it, it complains about the projects in zephyr/submanifests/optional.yaml not being cloned starting with the first in the list of canopennode.  Since it is optional software that I am not using, it is not cloned, so the entire freeze operation fails.

I am using the "forest" model of west and have a zephyrproject/applications folder which imports the zephyr repo:
```
manifest:
  remotes:
    - name: zephyrproject-rtos

  projects:
    - name: zephyr
      revision: v4.0-mainline-dev
      import: true
    # private application repo configurations go here

$ west manifest --freeze
. . .
RuntimeError: cannot freeze; project canopennode is uncloned
```
I was able to update west to skip uncloned projects to get a release out.  Would still like to know if this is a west issue or a usage/configuration issue on my part.

pdgendt:
> This was already discussed in #491, #667 and probably others. At the time, a new command was suggested instead of changing the semantic of the existing one[*]. Please review past discussions before proceeding with this.

Thanks for pointing to those discussions, will update this PR to have `--freeze-active` and `--resolve-active` instead.

carlescufi:
@pdgendt Overall this looks great, but just a question from myself. Why not have something like `west manifest --resolve --active` and `west manifest --freeze --active`?

pdgendt:
> @pdgendt Overall this looks great, but just a question from myself. Why not have something like `west manifest --resolve --active` and `west manifest --freeze --active`?

No particular reason, both options are probably valid, I took the comment from [here](https://github.com/zephyrproject-rtos/west/issues/491#issuecomment-2557547958).

carlescufi:
> > @pdgendt Overall this looks great, but just a question from myself. Why not have something like `west manifest --resolve --active` and `west manifest --freeze --active`?
> 
> No particular reason, both options are probably valid, I took the comment from [here](https://github.com/zephyrproject-rtos/west/issues/491#issuecomment-2557547958).

I see, I personally would rather see a new `--active-only` flag that can be combined with `--resolve` and `--freeze`, as well as help text pointing to this new `--active-only` when the resolving or freezing fails. But I won't block the PR for this. CC @mbolivar @marc-hb

carlescufi (approved):

carlescufi (approved):
Looks great!

EricNRS:
Thank you @pdgendt for pushing this through and taking care of several old issues!

PR #811, opened 2025-04-17
ci: Cancel ongoing builds on new changes

pdgendt:
Reduce the amount of CI cycles by canceling ongoing builds if a PR is updated.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,6 +5,11 @@ on: [push, pull_request]
 permissions:
   contents: read
 
+# Cancel ongoing builds on new changes
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.head_ref || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   build:
     runs-on: ${{ matrix.os }}


marc-hb (approved):
As long as this not used as an excuse for rapid-fire force pushes :-)

.github/workflows/test.yml
```diff
@@ -5,6 +5,11 @@ on: [push, pull_request]
 permissions:
   contents: read
 
+# Cancel ongoing builds on new changes
+concurrency:
+  group: ${{ github.workflow }}-${{ github.event_name }}-${{ github.head_ref || github.ref }}
```
marc-hb, line 10: I'm surprised GitHub does not provide a shorter, more user friendly way to express this.

pdgendt: It is [documented](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs#example-using-concurrency-and-the-default-behavior) like that, but I copied this from the Zephyr project workflows. Adding more specific IDs to the `group` will make sure we only cancel our PR's runs.

Issue #820, opened 2025-07-02
Document the default revision


yashi:
Even though https://github.com/zephyrproject-rtos/west/blob/670024b9a18285ef8f927984a7cc0288b063ee62/src/west/manifest.py#L197 sets `_DEFAULT_REV` to `master`, https://github.com/zephyrproject-rtos/west/blob/main/src/west/manifest-schema.yml doesn't say the default value.

Document the default value in the schema like:

```diff
diff --git a/src/west/manifest-schema.yml b/src/west/manifest-schema.yml
index 6982300..ebf82b5 100644
--- a/src/west/manifest-schema.yml
+++ b/src/west/manifest-schema.yml
@@ -103,6 +103,7 @@ mapping:
             required: false
             type: str
           # Revision to check out. May be a branch, tag, or SHA.
+          # Default: "master"
           revision:
             required: false
             type: text        # SHAs could be only numbers
```

kartben:
FWIW it should really be `main`, not `master`

marc-hb:
Or, much better:
- #821

PR #822, opened 2025-07-02
schema: Document default revision value

### LINKED ISSUE #821: Determine the default branch of a remote repo
Should we keep using `master` as the default? Or, we could determine the default branch of a remote repo using:

```
git ls-remote --symref $REPO_RUL HEAD
```

Such as

GitHub
```
❯ git ls-remote --symref https://github.com/zephyrproject-rtos/zephyr HEAD
ref: refs/heads/main	HEAD
4bd1d39c67dfb4860fbd4f9ccc17e084aa32d763	HEAD
```

GitLab
```
❯ git ls-remote --symref https://gitlab.gnome.org/GNOME/glib.git HEAD
ref: refs/heads/main	HEAD
804553b12606a728e9d6fab7b12d1d8f4277f03e	HEAD
```

yashi:
Document the default behavior of the `revision` key. The `revision` key is optional; if unspecified, west defaults to the `master` branch instead of the repository’s default branch. As a result, `west update` will fail when the repo’s default branch isn’t `master` and no revision is provided.

This fixes #821.

### DIFF
--- a/src/west/manifest-schema.yml
+++ b/src/west/manifest-schema.yml
@@ -103,6 +103,7 @@ mapping:
             required: false
             type: str
           # Revision to check out. May be a branch, tag, or SHA.
+          # Default: "master"
           revision:
             required: false
             type: text        # SHAs could be only numbers


pdgendt:
We should actually favour using main instead of master.

marc-hb:
Unfortunately, this fixes ~#822~ #820, not #821

yashi:
Oops. Thanks. Fixed. (The commit log doesn't have the reference to the github issue, just comment here)

marc-hb:
> Oops. Thanks. Fixed. 

OMG, I got it wrong too. This was #820. I fixed the description again.

Next time please don't file any issue for a one-line PR, the PR itself is enough. This is not zephyr.git, it's less formal :-)

Issue #824, opened 2025-07-03
Keep consistent list orders on `west manifest --resolve` and `--freeze`


matheusmbar:
Commands `west manifest --freeze` and `--resolve` don't enforce the order of entries on list `group-filter:`. Items list shuffle on successive executions. 

While that does not modify the content of the manifest, it generates unnecessary noise on repos that hold tag history for a project.

`projects` list seem to hold the order at every execution, at least as long as I have tested.

**Steps to reproduce:**
```bash
$ west init -m https://github.com/zephyrproject-rtos/example-application
$ west update

# execute any of these a few times, compare order at group-filter
$ west manifest --freeze
$ west manifest --resolve
```

**Expected behavior:** (every time)
```yaml
manifest:
  group-filter:
  - -optional
  - -babblesim
  projects:
  - name: zephyr
    ...
```

**Actual behavior:**

Order of `-optional` and `-babblesim` shuffles on each execution, but it does not change every time.

```yaml
manifest:
  group-filter:
  - -optional
  - -babblesim
  projects:
  - name: zephyr
    ...
```

```yaml
manifest:
  group-filter:
  - -babblesim
  - -optional
  projects:
  - name: zephyr
    ...
```

marc-hb:
Thanks for the detailed report; I can reproduce this:

```
west -vv manifest --resolve
...
west.manifest: DEBUG: group-filter: ['-babblesim', '-optional']
...
west.manifest: DEBUG: group filters in precedence order (later is higher): deque([['-babblesim', '-optional']])
west.manifest: DEBUG: final top level group-filter: ['-optional', '-babblesim']
```
```
west -vv manifest --resolve
...
west.manifest: DEBUG: group-filter: ['-babblesim', '-optional']
...
west.manifest: DEBUG: group filters in precedence order (later is higher): deque([['-babblesim', '-optional']])
west.manifest: DEBUG: final top level group-filter: ['-babblesim', '-optional']
```

marc-hb:
Fix submitted and test coverage added in:
- #827

matheusmbar:
Thanks for taking some time and putting the effort to improve this @marc-hb 
I'll put it to good use as soon as a new west release is available.

marc-hb:
@matheusmbar testing the west main branch is pretty easy.

1. uninstall your current west release; depends on how you installed it.
2. git clone this repo
3. https://github.com/zephyrproject-rtos/west?tab=readme-ov-file#editable-install

[...]

PR #826, opened 2025-07-11
tests: manifest: add test for `manifest --resolve` output

marc-hb:
Try to catch non-deterministic output issues like #824


marc-hb:
Superseded by #827.

This was useful to test commits separately.

PR #827, opened 2025-07-11
make manifest.group_filter list order deterministic + add test coverage

### LINKED ISSUE #824: Keep consistent list orders on `west manifest --resolve` and `--freeze`

marc-hb:
3 detailed commit messages

Fixes #824

### DIFF
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -2766,7 +2766,7 @@ class Manifest:
                 self._ctx.group_filter_q)
             for group_filter in self._ctx.group_filter_q:
                 _update_disabled_groups(self._disabled_groups, group_filter)
-            ret = [f'-{g}' for g in self._disabled_groups]
+            ret = [f'-{g}' for g in sorted(self._disabled_groups)]
             _logger.debug('final top level group-filter: %s', ret)
             return ret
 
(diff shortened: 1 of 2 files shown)


marc-hb:
For the record, I think this regression was introduced in the (huge) 28c76a52067d part of the (bigger) #482

Issue #828, opened 2025-07-14
Cache for initial "west update" in CI


thorsten-klein:
### Summary

Hello,

we are currently setting up our first cloud-based CI build for our zephyr project.

In our initial runs, we’ve observed the following timings:

Note: It is a freshly started VM in CI.

- Cloning our repository: **1 second**
- Setting up the initial pip3 environment (`pip3 install -r requirements.txt`): **15 seconds**
- `west update`: **90 seconds**
- `west packages pip --install`: **15 seconds**
- `west build`: **30 seconds**

As you can see, the `west update` step is by far the most time-consuming and generates significant network traffic (because of the `git clone` commands which are run). We would like to optimize this by introducing caching.

Do you already have a recommended solution for caching `west update`?

### Describe the solution you'd like

One idea we’re considering is to cache the Git repositories that west clones. For example:

- Keep a centralized mirror of the modules in a local directory (e.g., `~/.west/git`)
- Configure west to clone from these local mirrors instead of remote URLs
  - This could be done using `git worktree`, or
  - By cloning from a local `--mirror` repository using file paths

This would allow the `west update` step to pull from the local system, significantly reducing clone time and external traffic.

### Alternatives

Are there already alternative solutions existing?

pdgendt:
There already is caching support, isn't this sufficient?

```shell
$ west help update
...
local project clone caches:
  Projects are usually initialized by fetching from their URLs, but
  they can also be cloned from caches on the local file system.

  --name-cache NAME_CACHE
                        cached repositories are in subdirectories matching the names of projects to update
  --path-cache PATH_CACHE
                        cached repositories are in the same relative paths as the workspace being updated
```

thorsten-klein:
Thank you for your prompt response.

I have tested the process, and I expected that setting `--name-cache NAME_CACHE` would automatically clone the repositories to the specified location on the first run. Unfortunately, this does not happen.

It appears that I need to manually clone the repositories to that location beforehand. This is quite challenging for me, especially since most of the repositories are transitive dependencies.

marc-hb:
Have you tried `--fetch-opt=--filter=tree:0` ? In my CI experience this is the "best value" by far.

See #638 and https://github.com/zephyrproject-rtos/west/labels/performance , including the closed tickets.

marc-hb:
> It appears that I need to manually clone the repositories to that location beforehand.

[...]

Issue #799, opened 2025-04-02
Make a proper dedicated top-level documentation site


yashi:
https://discord.com/channels/720317445772017664/733037524498382881/1356918374231445524

Make a proper West document and host it by itself. We can based on the current West sections in the Zephyr Doc.

Once we get it up running, we can start trimming the West section in the Zephyr doc or replace it with links to West doc.

marc-hb:
That would be great but please edit the description and add a summary of the Discord discussion which is NOT available unless you have a Discord account (reproduce with incognito mode). Being a Discord project member may be required too.

yashi:
Hope you like it

marc-hb:
Work in progress in #860

PR #829, opened 2025-07-16
Add new flag `update --auto-cache` 

thorsten-klein:
Proposal for Solution for Issue #828

Introducing a new flag, `west update --auto-cache`, which automatically caches all repositories at the specified `--name-cache` or `--path-cache`. The repositories are cloned with the `--bare` option to the designated cache location, allowing them to be subsequently cloned from this local cache. Note: The bare repositories are automatically synchronized with the remote repository when running `west update`.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -15,7 +15,7 @@ import sys
 import textwrap
 import time
 from functools import partial
-from os.path import abspath, relpath
+from os.path import abspath, dirname, relpath
 from pathlib import Path, PurePath
 from time import perf_counter
 from urllib.parse import urlparse
@@ -1069,6 +1069,9 @@ class Update(_ProjectCommand):
         group.add_argument('--path-cache',
                            help='''cached repositories are in the same relative
                            paths as the workspace being updated''')
+        group.add_argument('--auto-cache', action='store_true',
+                           help='''automatically setup local cache repositories
+                           (clone on first run and sync on subsequent runs)''')
 
         group = parser.add_argument_group(
             title='fetching behavior',
@@ -1151,6 +1154,7 @@ class Update(_ProjectCommand):
         self.narrow = args.narrow or config.getboolean('update.narrow')
         self.path_cache = args.path_cache or config.get('update.path-cache')
         self.name_cache = args.name_cache or config.get('update.name-cache')
+        self.auto_cache = args.auto_cache or config.get('update.auto-cache')
         self.sync_submodules = config.getboolean('update.sync-submodules',
                                                  default=True)
 
@@ -1501,11 +1505,38 @@ class Update(_ProjectCommand):
             if take_stats:
                 stats['init'] = perf_counter() - start
 
+    def handle_auto_cache(self, project):
+        # update() helper. Initialize the specified cache directory if it has
+        # not been cloned yet. If the cache directory is already existing, it
+        # will be synced with remote.
+        if self.name_cache is not None:
+            cache_dir_expected = Path(self.name_cache) / project.name
+        elif self.path_cache is not None:
+            cache_dir_expected = Path(self.path_cache) / project.path
+        else:
+            # no cache dir could be determined
+            return
+
+        cache_dir = self.project_cache(project)
+        if cache_dir is None:
+            cache_dir_parent = dirname(cache_dir_expected)
+            # cache has not been created yet at the expected location.
+            # Clone it as bare repository now.
+            os.makedirs(cache_dir_parent, exist_ok=True)
+            project.git(['clone', '--bare', project.url, os.fspath(cache_dir_expected)],
+                        cwd=cache_dir_parent)
+        else:
+            # cache is already existing. Sync it with remote now.
+            project.git(['--bare', 'fetch'], cwd=cache_dir)
+
     def init_project(self, project):
         # update() helper. Initialize an uncloned project repository.
         # If there's a local clone available, it uses that. Otherwise,
         # it just creates the local repository and sets up the
         # convenience remote without fetching anything from the network.
+
+        if self.auto_cache:
+            self.handle_auto_cache(project)
 
         cache_dir = self.project_cache(project)
 
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -1580,6 +1580,57 @@ def setup_cache_workspace(workspace, remote, foo_head, bar_head):
         ''')
 
 
+def test_update_name_cache_auto(tmpdir):
+    # Test that 'west update --name-cache' works and doesn't hit the
+    # network if it doesn't have to.
+
+    remote = tmpdir / 'remotes' / 'remote'
+
+    create_repo(remote)
+    create_repo(tmpdir / 'remotes' / 'foo')
+    create_repo(tmpdir / 'remotes' / 'bar')
+
+    foo_head = rev_parse(tmpdir / 'remotes' / 'foo', 'HEAD')
+    bar_head = rev_parse(tmpdir / 'remotes' / 'bar', 'HEAD')
+
+    name_cache_dir = tmpdir / 'name_cache'
+
+    workspace = tmpdir / 'workspace'
+    setup_cache_workspace(workspace, remote, foo_head, bar_head)
+    workspace.chdir()
+    foo = workspace / 'subdir' / 'foo'
+    bar = workspace / 'bar'
+
+    # Test the command line option.
+    cmd(['update', '--name-cache', name_cache_dir, '--auto-cache'])
+    assert foo.check(dir=1)
+    assert bar.check(dir=1)
+    assert (name_cache_dir / "foo").check(dir=1)
+    assert (name_cache_dir / "bar").check(dir=1)
+    assert rev_parse(foo, 'HEAD') == foo_head
+    assert rev_parse(bar, 'HEAD') == bar_head
+
+    # Move the repositories out of the way and test the configuration option.
+    # (We can't use shutil.rmtree here because Windows.)
+    shutil.move(os.fspath(foo), os.fspath(foo) + ".moved")
+    shutil.move(os.fspath(bar), os.fspath(bar) + ".moved")
+    shutil.move(os.fspath(name_cache_dir / "foo"),
+                os.fspath(name_cache_dir / "foo") + ".moved")
+    shutil.move(os.fspath(name_cache_dir / "bar"),
+                os.fspath(name_cache_dir / "bar") + ".moved")
+    cmd(['config', 'update.name-cache', name_cache_dir])
+    cmd(['config', 'update.auto-cache', 'true'])
+    cmd('update')
+    assert foo.check(dir=1)
+    assert bar.check(dir=1)
+    assert (name_cache_dir / "foo").check(dir=1)
+    assert (name_cache_dir / "bar").check(dir=1)
+    assert rev_parse(foo, 'HEAD') == foo_head
+    assert rev_parse(bar, 'HEAD') == bar_head
+
+    # Run update again
+    cmd('update')
+
 def test_update_name_cache(tmpdir):
     # Test that 'west update --name-cache' works and doesn't hit the
     # network if it doesn't have to.
@@ -1616,6 +1667,55 @@ def test_update_name_cache(tmpdir):
     assert rev_parse(foo, 'HEAD') == foo_head
     assert rev_parse(bar, 'HEAD') == bar_head
 
+
+def test_update_path_cache_auto(tmpdir):
+    # Test that 'west update --path-cache' works and doesn't hit the
+    # network if it doesn't have to.
+
+    remote = tmpdir / 'remotes' / 'remote'
+    create_repo(remote)
+    create_repo(tmpdir / 'remotes' / 'foo')
+    create_repo(tmpdir / 'remotes' / 'bar')
+    foo_head = rev_parse(tmpdir / 'remotes' / 'foo', 'HEAD')
+    bar_head = rev_parse(tmpdir / 'remotes' / 'bar', 'HEAD')
+
+    path_cache_dir = tmpdir / 'path_cache_dir'
+
+    workspace = tmpdir / 'workspace'
+    setup_cache_workspace(workspace, remote, foo_head, bar_head)
+    workspace.chdir()
+    foo = workspace / 'subdir' / 'foo'
+    bar = workspace / 'bar'
+
+    # Test the command line option.
+    cmd(['update', '--path-cache', path_cache_dir, '--auto-cache'])
+    assert foo.check(dir=1)
+    assert bar.check(dir=1)
+    assert (path_cache_dir / "subdir" / "foo").check(dir=1)
+    assert (path_cache_dir / "bar").check(dir=1)
+    assert rev_parse(foo, 'HEAD') == foo_head
+    assert rev_parse(bar, 'HEAD') == bar_head
+
+    # Move the repositories out of the way and test the configuration option.
+    # (We can't use shutil.rmtree here because Windows.)
+    shutil.move(os.fspath(foo), os.fspath(foo) + ".moved")
+    shutil.move(os.fspath(bar), os.fspath(bar) + ".moved")
+    shutil.move(os.fspath(path_cache_dir / "subdir" / "foo"),
+                os.fspath(path_cache_dir / "subdir" / "foo") + ".moved")
+    shutil.move(os.fspath(path_cache_dir / "bar"),
+                os.fspath(path_cache_dir / "bar") + ".moved")
+    cmd(['config', 'update.path-cache', path_cache_dir])
+    cmd(['config', 'update.auto-cache', 'true'])
+    cmd('update')
+    assert foo.check(dir=1)
+    assert bar.check(dir=1)
+    assert (path_cache_dir / "subdir" / "foo").check(dir=1)
+    assert (path_cache_dir / "bar").check(dir=1)
+    assert rev_parse(foo, 'HEAD') == foo_head
+    assert rev_parse(bar, 'HEAD') == bar_head
+
+    # Run update again
+    cmd('update')
 
 def test_update_path_cache(tmpdir):
     # Test that 'west update --path-cache' works and doesn't hit the


pdgendt:
@thorsten-klein thanks for creating this PR, could you look into the CI issues?

thorsten-klein:
Hello,

I’m still working on improving the proposed solution and would appreciate your input. Which of the following concepts would you prefer?

### Option 1: Use `--auto-cache` alongside `--name-cache` or `--path-cache`

This approach would automatically set up the caches based on the specified options.

- **Pros:** Nice improvement for existing behavior.
- **Cons:** If the remote URL changes, the local cache will not reflect that change and might become outdated. Users / CI has to remove the local cache so that it can be correctly re-created.

---

### Option 2: Use `--auto-cache <DIRECTORY>`

With this approach, a directory is specified via `--auto-cache`, and the cache is automatically created under that directory (analog to `--path-cache`). The last directory is derived from the hashed remote URL (and possibly further factors if needed).

- **Example path:**
  ```
  /path/to/auto/cache/foo/81b90dfaec466d20bd5fb334aa80b008e6f5940f
  ```

- **Pros:** Allows automatic cache organization by URL, avoiding stale entries.
- **Cons:** additional cache concept is added.

---

I am going with Option 2 now.

Let me know which direction you'd prefer, or if you have another idea in mind. Thanks!

thorsten-klein:
> thanks for creating this PR, could you look into the CI issues?

All CI builds should pass now 👍

thorsten-klein:
Now also the code coverage should be close to 100% 👍

pdgendt:
I'm liking the implementation of option 2, will test it out locally and do a deep dive review later on.

src/west/app/project.py
```diff
@@ -1556,6 +1589,16 @@ def init_project(self, project):
...
 
+    def project_auto_cache(self, project):
+        project_auto_cache = None
+        if self.auto_cache:
+            # get the location of the project within the auto-cache.
+            # Note: The url is hashed and used as a subfolder to accomodate
+            # changes in the manifest
+            url_hash = hashlib.sha1(project.url.encode('utf-8')).hexdigest()
+            project_auto_cache = os.fspath(Path(self.auto_cache) / project.path / url_hash)
+        return project_auto_cache
```
pdgendt: Not entirely sure about the hash, it makes sense, but for someone who doesn't know the internals it could be a bit confusing.

Nit, prefer early return:

```python
    def project_auto_cache(self, project):
        if self.auto_cache is None:
            return None

        # get the location of the project within the auto-cache.
        # Note: The url is hashed and used as a subfolder to accomodate
        # changes in the manifest
        url_hash = hashlib.sha1(project.url.encode('utf-8')).hexdigest()
        return os.fspath(Path(self.auto_cache) / project.path / url_hash)
```

src/west/app/project.py
```diff
@@ -1501,12 +1506,40 @@ def ensure_cloned(self, project, stats, take_stats):
...
+            return
+
+        if not Path(cache_dir).exists():
+            # The cache has not been created yet at the expected location.
+            # Ensure that its parent directory exist before cloning anything.
+            # Then bare-clone the repository into the local cache.
+            cache_dir_parent = dirname(cache_dir)
+            os.makedirs(cache_dir_parent, exist_ok=True)
+            project.git(['clone', '--bare', project.url, os.fspath(cache_dir)],
+                        cwd=cache_dir_parent)
```
kartben: Not sure how it's done elsewhere but a manifest containing malicious url could cause trouble here. Add an explicit `--` between actual options and the repo url & destination.

src/west/app/project.py
```diff
@@ -1556,6 +1589,16 @@ def init_project(self, project):
                 # f'refs/remotes/{project.remote_name}/{branch}'.
                 project.git(['update-ref', '-d', branch])
 
+    def project_auto_cache(self, project):
+        project_auto_cache = None
+        if self.auto_cache:
+            # get the location of the project within the auto-cache.
+            # Note: The url is hashed and used as a subfolder to accomodate
+            # changes in the manifest
+            url_hash = hashlib.sha1(project.url.encode('utf-8')).hexdigest()
```
kartben: Don't see any reason not to use SHA-256 over SHA-1 here

thorsten-klein: I'm using SHA-1 for now, as it's faster and produces shorter hashes, and this isn't a security-sensitive use case. Shall I change to SHA-256?

Would it make sense to truncate the hash for folder names—say, limit it to the first 10 characters?

kartben: Yeah the length, especially on windows, is definitely a concern. And I agree it's not sensitive, but on the other hand performance shouldn't really be an issue either, here, on such a small amount of data, right?

src/west/app/project.py
thorsten-klein: I will check in detail tomorrow. But at other places `project.url` is used in a similar way, e.g. `project.git(['remote', 'set-url', project.remote_name, project.url])`. I will check if the clone command works with `--` as well.

src/west/app/project.py
pdgendt: Or use `md5`

src/west/app/project.py
marc-hb: @thorsten-klein there is no need for consistency in such a case. The `--` separator exists to make things more robust, so using it is here a no-brainer. Other places should also use it - but off-topic here. It's optional because it saves typing but saving 3 characters does not matter in source code.

src/west/app/project.py
marc-hb: Which hash is IPFS using? https://docs.ipfs.tech/how-to/host-git-repo/

Could there be some "convergence" there?

> Would it make sense to truncate the hash for folder names—say, limit it to the first 10 characters?

Is this referring to Windows MAX_PATH? In that case, isn't it enough to tell the user not to use an already long argument?

Vaguely related: maybe you can use the first two characters for a directory name, same as `ls .git/objects/`

marc-hb (commented):
No time to really review the code right now but it looks much simpler than what I was afraid of, nice! Very promising.

> /path/to/auto/cache/subdir/foo/81b90dfaec466d20bd5fb334aa80b008e6f5940f

Could you also create some short, informal `/path/to/auto/cache/subdir/foo/README.txt` there with basic, static info like: "this was created by west to cache XXX".

I like the "everything is content-addressable" idea in general but it sucks when someone wants to delete old files and has no idea who put them there and why.

src/west/app/project.py
```diff
@@ -1069,6 +1070,9 @@ def do_add_parser(self, parser_adder):
         group.add_argument('--path-cache',
                            help='''cached repositories are in the same relative
                            paths as the workspace being updated''')
+        group.add_argument('--auto-cache',
+                           help='''automatically setup local cache repositories
+                           (clone on first run and sync on subsequent runs)''')
```
marc-hb: Isn't this mutually exclusive with "manual" caching command line options?

thorsten-klein: In this PR a new caching command has been introduced, which has a lower priority compared to the existing ones. It fits to existing caching mechanisms.

Currently, you can specify both `--path-cache X` and `--name-cache Y`, with the name cache having higher priority.

The lowest priority cache is now `--auto-cache`. This cache will only be used if no other cache is specified or if the repository cannot be found in the specified caches.

I will update the help documentation to highlight these cache priorities. 👍

pdgendt: I was going to suggest to replace it with `group = parser.add_mutually_exclusive_group(required=False)`, but for that some tests probably need to be changed too.

thorsten-klein:
> Could you also create some short, informal `/path/to/auto/cache/foo/README.txt` there with basic, static info like: "this was created by west to cache XXX".

Since there can be multiple folders under `/path/to/auto/cache/foo` it is hard to create only one `README.txt` there.
Instead `/path/to/auto/cache/foo/<hash>.info` is now created alongside which contains the according basic static info.

> Would it make sense to truncate the hash for folder names—say, limit it to the first 10 characters?

I prefer shorter directory names (8 bytes) over longer ones (e.g. 40 bytes). Even with a 8 bytes hash, west can support up to `(26+10)^8` different remotes at the same time (which is still more than enough)

> Vaguely related: maybe you can use the first two characters for a directory name, same as ls .git/objects/

The hash folder is already placed under the project name (e.g. `/path/to/auto/cache/foo`), so I guess there is no need to use the first two characters of the project name. Since most URLs will start with `http[s]://...` it also does not make sense to use the first two characters of the URL.

src/west/app/project.py
thorsten-klein: Currently I have made my changes so that the caches can be combined (see according test which I have added).
IMO This behavior makes sense, since users might have some local caches which they want to manually maintain (e.g. their current `--name-cache` or `--path-cache`), but all further repositories they want to be automatically cached with `--auto-cache`.

pdgendt: Hmn, while I agree that this should be the intended behavior, it does change the behaviour slightly.

src/west/app/project.py
```diff
@@ -1501,12 +1512,56 @@ def ensure_cloned(self, project, stats, take_stats):
...
+        # not been cloned yet. If the cache directory is already existing, it
+        # will be synced with remote.
+
+        cache_dir = self.project_cache(project)
+        auto_cache_dir = self.project_auto_cache(project)
+
+        if not auto_cache_dir or (cache_dir != auto_cache_dir):
+            # nothing to do: auto-cache directory was either not specified
+            # or another cache was specified with higher priority.
+            return
```
pdgendt: We can do a slight improvement here:

```python
        auto_cache_dir = self.project_auto_cache(project)

        if auto_cache_dir is None:
            # nothing to do: auto-cache directory was not specified
            return
            
        cache_dir = self.project_cache(project)
        if cache_dir != auto_cache_dir:
            # another cache was specified with higher priority.
            return
```

We can then (optionally) replace
```python
        if self.auto_cache:
            self.handle_auto_cache(project)
```
with simply (as it will be a no-op)
```python
        self.handle_auto_cache(project)
```

src/west/app/project.py
```diff
@@ -1501,12 +1512,58 @@ def ensure_cloned(self, project, stats, take_stats):
             if take_stats:
                 stats['init'] = perf_counter() - start
 
+    def create_auto_cache_info(self, project, cache_dir):
+        # auto-cache helper: Create some short informal file with basic info
+        # about the local cache folder.
+        content = textwrap.dedent(f"""
+        The following local cache directory  was automatically created by west:
```
pdgendt: Nit

```suggestion
        The following local cache directory was automatically created by west:
```

tests/test_project.py
```diff
@@ -1617,20 +1619,181 @@ def test_update_name_cache(tmpdir):
...
+    assert (auto_cache_dir / "subdir" / "foo").check(dir=1)
+    assert (auto_cache_dir / "bar").check(dir=1)
+    assert rev_parse(foo, 'HEAD') == foo_head
+    assert rev_parse(bar, 'HEAD') == bar_head
+
+    # Check that some info file was created with basic info
+    # e.g. /path/to/auto/cache/subdir/foo/<hash>.info
+    foo_hash = sorted(os.listdir(auto_cache_dir / "subdir" / 'foo'))[0]
+    expected_info = textwrap.dedent(f"""
+        The following local cache directory  was automatically created by west:
```
pdgendt: Same

```suggestion
        The following local cache directory was automatically created by west:
```

src/west/app/project.py
```diff
@@ -1556,6 +1613,16 @@ def init_project(self, project):
...
 
+    def project_auto_cache(self, project):
+        if self.auto_cache is None:
+            return None
+        # get the location of the project within the auto-cache.
+        # Note: The url is hashed and used as a subfolder to accomodate
+        # changes in the manifest. To avoid long paths, only the first 8
+        # characters  of the hash are used.
+        url_hash = hashlib.md5(project.url.encode('utf-8')).hexdigest()
+        return os.fspath(Path(self.auto_cache) / project.path / url_hash[:8])
```
pdgendt: Playing with this, I think with the added `.info` file, we shouldn't add the `project.path` here. The cache could even be used if the project's path differ in multiple workspaces. I'd also keep the full `md5` hash to increase collision resistance.

```suggestion
        return os.fspath(Path(self.auto_cache) / url_hash)
```

src/west/app/project.py
```diff
@@ -1501,12 +1512,58 @@ def ensure_cloned(self, project, stats, take_stats):
...
+    def create_auto_cache_info(self, project, cache_dir):
+        # auto-cache helper: Create some short informal file with basic info
+        # about the local cache folder.
+        content = textwrap.dedent(f"""
+        The following local cache directory  was automatically created by west:
+        - Local Cache:  {basename(cache_dir)}
+
+        It caches the following project:
+        - Project Name: {project.name}
+        - Project Path: {project.path}
```
pdgendt: If the cache could be used by different workspaces as commented [here](https://github.com/zephyrproject-rtos/west/pull/829/files#r2212903818), we shouldn't add the name or path in this file.

thorsten-klein: The project path is a relative path, e.g. "subdir/foo". This will be same for all workspaces.

For example the file looks like following:
```
The following local cache directory  was automatically created by west:
- Local Cache:  63dce786

It caches the following project:
- Project Name: foo
- Project Path: subdir/foo
- Project Url:  file:///home/user/work/GIT/west/.tox/py3/tmp/pytest with space/test_update_auto_cache0/remotes/foo

```

pdgendt: > The project path is a relative path, e.g. "subdir/foo". 

I know 🙂 

> This will be same for all workspaces.

Not necessarily, it's up to the manifest to set the project's path. Different workspaces can have different manifests. The cache should only depend on the URL.

thorsten-klein: > Not necessarily, it's up to the manifest to set the project's path. Different workspaces can have different manifests. The cache should only depend on the URL.

In this case it will be placed at another path in the auto-cache folder (since we use `project.path` to set up the auto cache).

But it is actually a nice idea. We can use `project.name` for setting up the auto cache and omit the `project.path` info in the info file since the path might be different. With this the chances of the auto-cache being reused are higher 👍

pdgendt: _Currently, yes. And this might be an over-optimization, but I don't think it is._

I can have two manifest projects:

```yaml
manifest:
  projects:
    - name: awesome-stuff
      path: modules/lib/awesome-stuff
      url: https://github.com/zephyrproject-rtos/awesome-stuff
      revision: main
```
and
```yaml
manifest:
  projects:
    - name: util
      path: util
      url: https://github.com/zephyrproject-rtos/awesome-stuff
      revision: v1.0.0
```

As the URL is the same, the cache can be shared.

thorsten-klein: Ah ok, so you would even go one step further and put the hash directories directly into auto-cache?
With this the cache becomes very efficient (because it is shared as much as possible), but it is not so user-friendly anymore (since you have to check the info files to get readable infos).

I am open for any solution.
Personally I like the "name/hash" solution most.

pdgendt: `name/hash` might be an acceptable compromise, the chances of having different names is rather small probably.

thorsten-klein: Or we could take `basename(project.url)`?

pdgendt: Yeah, I was just going to comment, we should be careful with characters in names that aren't allowed in paths.

tests/test_project.py
pdgendt: Not updated yet.

src/west/app/project.py
```diff
@@ -1501,12 +1512,54 @@ def ensure_cloned(self, project, stats, take_stats):
...
+            # Ensure that its parent directory exist before cloning anything.
+            # Then bare-clone the repository into the local cache.
+            cache_dir_parent = Path(cache_dir).parent
+            cache_dir_parent.mkdir(parents=True, exist_ok=True)
+            project.git(['clone', '--bare', '--', project.url, os.fspath(cache_dir)],
+                        cwd=cache_dir_parent)
+            self.create_auto_cache_info(project, cache_dir)
+        else:
+            # The local cache is already existing. Sync it with remote.
+            project.git(['--bare', 'fetch'], cwd=cache_dir)
```
marc-hb: What is `git --bare fetch`? It's not documented in `git help fetch`.

I tried this command inside a git checkout and I get the very strange error "this is not a git repository". The command line parser accepts it, super weird.

marc-hb (commented):
> Even with a 8 bytes hash, west can support up to (26+10)^8 different remotes at the same time (which is still more than enough) [...]  The hash folder is already placed under the project path (e.g. /path/to/auto/cache/subdir/foo), so I guess there is no need to use the first two characters of the project name.

Could you please use the Linux `tree` command or similar and add some ASCII "art" of (simplified) examples that shows the structure of the different caches? It's very time-consuming to infer from the code and review comments. This should have been already done for existing cache options but better late than never.

These tree structures should probably be documented in some new section at https://github.com/zephyrproject-rtos/zephyr/tree/main/doc/develop/west but this is inconveniently located in a different repo (#799) and  more overhead generally speaking. So, in the mean time, you could just put these structures as a long comment in `west.git/tests/` near the corresponding test code. Good enough for now and good to have it very close to the corresponding test code. Also, there is a fine line between "user interface" and "implementation detail" for this; it can be a matter of perspective. Either way, this is useful information that should be documented somewhere.

Speaking of test code, `tests/test_project.py` is becoming a bit bloated with tests for everything and anything. Could you start a new `tests/test_caching.py` file and move `test_update_path_cache()` and `test_update_name_cache()` etc. there as an initial commit that merely moves stuff without any code change? I know many developers don't like spending time rearranging multi-commit patch series but this is west here so this should be trivial for you ;-)

Documenting the existing structures could ideally be a separate commit too but no big deal.

> > Currently I have made my changes so that the caches can be combined (see according test which I have added).

> Hmn, while I agree that this should be the intended behavior, it does change the behaviour slightly.

I think it's great to support multiple levels of caching at the same time: caching is 100% about performance so it should try everything it can find, all at once.

On the other hand, changing the behavior of existing options should be a separate commit with its own commit message. Maybe its own PR even to give more visibility? I mean merged before this one.

src/west/app/project.py
marc-hb: > With this the cache becomes very efficient (because it is shared as much as possible), but it is not so user-friendly anymore (since you have to check the info files to get readable infos).

I think it's important to be able to tell easily and quickly WHAT some subdirectory is caching. On the other hand, the project or reason why it was created is not that important, sorry for any misleading suggestion earlier. If possible, sharing caches across multiple projects saves disk space and sounds great! Just keep in mind  different projects use different git branches and tags though...

My concern was really just users finding big directories using a lot disk space and having no idea what they are for. There are many different ways to address that concern.

src/west/app/project.py
thorsten-klein: This command only works in a local cache repository, which is a repository that was cloned via `git clone --bare`.

### Example

```bash
# Setup a local cache
git clone --bare https://github.com/zephyrproject-rtos/zephyr zephyr.git

# Clone from the local cache
git clone ./zephyr.git zephyr

# Fetch works in the bare repository
git -C zephyr.git --bare fetch # works

# Fetch does not work in the non-bare repository
git -C zephyr --bare fetch # does not work

# Fetch succeeds in both repositories without the --bare flag
git -C zephyr.git fetch # works
git -C zephyr fetch # works
```

If I remember correctly, yesterday I encountered an issue where a normal `git fetch` succeeded but did not fetch any new changes.

### Next Steps

I plan to create some tests, most likely next week, to investigate this issue further. In the test, a change will be committed on the remote repository, and the automatic sync should occur with the next `west update`. After the sync, the new revision should be usable in the workspace manifest.

src/west/app/project.py
```diff
@@ -1556,6 +1609,16 @@ def init_project(self, project):
...
                 project.git(['update-ref', '-d', branch])
 
+    def project_auto_cache(self, project):
+        if self.auto_cache is None:
+            return None
+        # get the location of the project within the auto-cache.
+        # Note: The url is hashed and used as a subfolder to accomodate
+        # changes in the manifest. To avoid long paths, only the first 8
+        # characters of the hash are used.
+        subdir_hash = hashlib.md5(project.url.encode('utf-8')).hexdigest()
```
marc-hb, line 1683: If you truncate it anyway, I don't see the point of using an obsolete checksum like md5.

src/west/app/project.py
marc-hb: From your example, it looks like `git --bare fetch` is never needed. In that case, don't use it because it's not documented (and weird looking IMHO).

If you do have to use some undocumented git command for some reason, then definitely add a comment explaining why, maybe with a link to a relevant github issue (in west or elsewhere).

pdgendt: It is documented: https://git-scm.com/docs/git#Documentation/git.txt---bare

marc-hb: Thx @pdgendt, I should have looked there. It's good that this syntax is not an "accident". This is still very cryptic (single line description), not saying anything about what happens when you ask git to treat a non-bare repo as a bare repo and why anyone would want to do that. Even more confusing when just fetching.

In any case, git is perfectly capable of distinguishing bare and non-bare repos on its own, so there is no need to try to "force" anything and I still think this `--bare` can be dropped.

src/west/app/project.py
```diff
@@ -1645,6 +1714,9 @@ def fetch(self, project, stats, take_stats):
         tags = (['--tags'] if not self.narrow else [])
         clone_depth = (['--depth', str(project.clone_depth)] if
                        project.clone_depth else [])
+        # sync the auto_cache before fetching the project
+        self.handle_auto_cache(project)
```
thorsten-klein, line 1783: I think this line does not make sense here since the remote URL was changed to point to the original repository (not to the local auto-cache). Therefore it is enought to do the auto-syncing only on the first init of the repository (right before cloning from the local auto-cache).

I will investigate next week if it works as expected with submodules as well.

src/west/app/project.py
thorsten-klein: Comment was outdated

src/west/app/project.py
thorsten-klein: It seems that it also works with `git fetch`. So no need for `git --bare fetch` 😊

thorsten-klein:
I have added another test now which is using submodules. I tried to not break any current existing behavior.

PS: I have created separate commit for my newest changes. If you want me to combine my commits into one single commit (to keep the history clean), please ping me before merging the PR 👍🏻

marc-hb:
> If you want me to combine my commits into one single commit (to keep the history clean),

Most GitHub projects seem to prefer a "dirty" git history like this: https://fossil-scm.org/home/doc/trunk/www/rebaseharm.md

But Zephyr, like Linux and many other open-source projects, takes the opposite, "clean history" stance: the final, merged git history should be bisectable, one logical change per commit and each commit should be as "perfect" as possible.

- https://docs.zephyrproject.org/latest/contribute/contributor_expectations.html
- https://github.com/zephyrproject-rtos/zephyr/issues/39194

marc-hb:
I received this (worrying) comment by email notification but I cannot find it online:
- https://github.com/zephyrproject-rtos/west/pull/829#issuecomment-3096695448

> I have investigated the flaky CI test on Windows. For any reason sometimes this command is failing: [...]
> west.manifest: INFO: running 'git rev-parse
> [...]
> unknown revision or path not in the working tree.\nUse '--' to separate paths from revisions, like this:\n'git <command> [<revision>...] -- [<file>...]'\n"
> ERROR: update failed for project bar
> It seems, that the git --bare fetch / git fetch both sporadically fail.

Either this is an artifact of not using GitHub as intended (see https://github.com/zephyrproject-rtos/zephyr/issues/39194), or you deleted this comment? Please don't delete comments, it's very confusing. Minor editions and striking are fine.

I'm not too worried about flaky runners and Windows. I mean it could be a serious problem but it would not be a west problem. I'm more worried about `west` not catching `git fetch` failures and failing much later: that would be very confusing. If you know which `git fetch` command was failing and where, could you artificially insert a typo in the git command at that location in the code and manually make sure west "fails fast" when that git fetch command fails; not much later in the code? Making sure the exception is not discarded, etc. That does not require Windows.

thorsten-klein:
Sorry for the confusion — I deleted the comment just seconds after posting it (when I identified the cause). 

The issue was a bit misleading, as it only occurred on Windows.

It turned out to be a copy-paste mistake in my test: I was checking for a commit ID from `foo` in the `bar` repository. This happened to work about 50% of the time because if the commits were made at the same moment, the IDs were identical for both `foo` and `bar`. But if the second commit in my test happened a second later, the IDs differed, and naturally, the commit couldn’t be found in the other repository.

So all good — Git is behaving as expected, and so is west 👍

src/west/app/project.py
```diff
@@ -1373,12 +1374,40 @@ def update_submodules(self, project):
             ref = []
             if (cache_dir):
                 submodule_ref = Path(cache_dir, submodule.path)
-                if any(os.scandir(submodule_ref)):
+                # do not break existing behavior: search for the submodule
```
marc-hb: "existing" sounds like this belongs to a commit message. git tracks points in time whereas comments in source code should ideally be "timeless".

There are exceptions but they need context and I miss the context here.

src/west/app/project.py
marc-hb, line 1: "handle" auto-caching of submodules sounds a bit ambiguous, it could be misinterpreted as a bug fix. I would just say "Add auto-caching of submodules" or "Implement auto-caching of submodules.

src/west/app/project.py
```diff
@@ -1645,6 +1712,7 @@ def fetch(self, project, stats, take_stats):
         tags = (['--tags'] if not self.narrow else [])
         clone_depth = (['--depth', str(project.clone_depth)] if
                        project.clone_depth else [])
+
```
marc-hb, line 1781: Spurious whitespace

src/west/app/project.py
marc-hb, line 1: > Avoid syncing the auto-cache on each update

Why not? Is fetching from the (refreshed) cache too slow for small updates?

This really needs more documentation details. Such documentation will also provide some design rationales and choices which is needed to review the implementation. Just like those ASCII art trees I asked before.

thorsten-klein: After cloning a repository into the workspace (using the local cache), the remote URL is updated to point to the actual remote. This means that when running `west update`, the repository in the workspace is fetched and updated from the original remote source.

The local cache is **not** updated during this process because it's no longer used after the initial clone. It only plays a role when the workspace repository is first created from the cache.

We could think about keeping the remote URL pointed to the local cache. But IMO it makes sense to set the remote URL in the workspace to the original remote (as it is now). This way, users can run commands like `git fetch` directly in the workspace without confusion.

Instead, the following logic could be implemented during `west update` for the auto cache:

1. Temporarily change the remote URL of the workspace repository to point to the local cache.
2. `git fetch` the local cache (from original remote).
3. `git fetch` the workspace repository (from the local cache).
4. Restore the remote URL in the workspace repository to the original remote.

I am afraid that this becomes tricky, especially when the `git fetch` command fails. We need to ensure that in any case the remote URL is pointing to remote url again.

src/west/app/project.py
thorsten-klein: > On the other hand, changing the behavior of existing options should be a separate commit with its own commit message. Maybe its own PR even to give more visibility? I mean merged before this one.

I will create a separate Pull Request where I change the existing behavior slightly, so that west supports multiple caches at the same time 👍

src/west/app/project.py
marc-hb: > Temporarily change the remote URL of the workspace repository to point to the local cache.

I don't have any strong opinion on when to refresh cache etc., or at least not yet. Just that the behavior should be clearly stated somewhere, I mean no one should have to infer it from reading the source code. Then we could review the behavior first ("top-down") and finally review the code to make sure it does what's expected.

On the other hand, I would hope for a way not to temporarily change remote URLs because it implies restoring them in case of any failure anywhere and it's really hard not to miss a spot. Plus, that sort of recovery is rarely ever tested and certainly not in every case.

Instead, it could be possible to perform "literal" git fetches from unnamed URL (e.g.: `git fetch https://....`) or maybe some ephemeral `insteadOf` configuration that vanishes when `west` exits.

src/west/app/project.py
```diff
@@ -1363,13 +1373,40 @@ def update_submodules(self, project):
...
+                                       f'submodule.{submodule.path}.url'],
+                                      capture_stdout=True)
+                    if not res.stdout or res.returncode:
+                        self.die("Submodule url cannot be determined for project:"
+                                 f"{project.name} ({project.path}).")
+                    submodule_url = res.stdout.decode('utf-8').strip()
+
+                    submodule_project = Project(name=submodule.name,
+                                                path=submodule.path,
+                                                url=submodule_url)
```
pdgendt: Not a huge fan of this artificially created project to have submodules cached next to actual projects.
Instead, can the project bare clones keep submodules in sync?

thorsten-klein: > Can the project bare clones keep submodules in sync?

With bare clones, this isn't feasible. However, it's worth noting that Git no longer strictly requires `--bare` clones in many contexts. For instance, both the current `--name-cache` and `--path-cache` implementations operate without `--bare` clones. So to maintain consistency with these approaches, I will avpod `--bare` clones in the auto-cache as well. Instead, I’ll rely on standard clones of the remote repositories now.

> Not a huge fan of this artificially created project to have submodules cached next to actual projects

Personally, I actually like the idea of treating submodules the same way as full projects. This reduces overall complexity and helps ensure that each submodule is checked out only once in the auto-cache, at a single central location per remote URL, rather than scattered across multiple places.

Nevertheless, let me try implementing it your way.

thorsten-klein:
Update: I face some issues when trying to cache submodules recursively. 
Would it be okay if this is some limitation for the auto-cache, so that it can only handle first layer of submodules?

tests/test_project_caching.py
```diff
@@ -160,12 +181,228 @@ def test_update_path_cache(tmpdir):
...
+    # cache correctly.
+
+    # The tree of the auto cache looks like following:
+    # (path cache)
+    # ├── bar
+    # │   └── <hash>
+    # │   └── <hash>.info
+    # └── foo
+    #     └── <hash>
+    # │   └── <hash>.info
```
marc-hb, line 196: Thanks for adding this and the ones in the previous PR but I feel like I'm still missing the connection with the workspace tree structure (`west update --help` says "in the same relative paths as the workspace being updated")

So could you expand all those a bit? Imagine this is an example in documentation for end users. In the longer term I think they should be in some user-visible documentation but here is good enough to get started (and avoids painful reviews across multiple repos #799)

marc-hb (commented):
> Would it be okay if this is some limitation for the auto-cache, 

100% fine by me.

- submodules are "competing" with West and not recommended. Their support in west is only a convenience for projects stuck with them. They should still be 100% functional but...
- ... caching is only an optimization, it should never affect functionality. So, "best effort" is good enough.
- If anyone cares about caching more submodules and/or better, they can always submit followup patches (I bet no one will ever)

However and as usual: the caching expectations should be clear. So, this limitation (like everything else) must be clearly written down somewhere. Maybe start a new "caching"
 section at the bottom of  `update --help` ? Similar to the bottom of `west list -h`

marc-hb:
Just remembered that submodules path can be relative or absolute:
- #545 

This means the test code that caches submodules should check for both options.

Not sure how this interacts with the cache tree structure.

Or, you could just not cache submodules at all in this first PR and postpone that work for later: the perfect is the enemy of the good. People using west _and_ submodules are a minority, meaning the perfect for a minority really is the enemy of the good for the majority.

src/west/app/project.py
```diff
@@ -1503,12 +1518,55 @@ def ensure_cloned(self, project, stats, take_stats):
...
+            return
+
+        if not Path(cache_dir).exists():
+            # The cache has not been created yet at the expected location.
+            # Ensure that its parent directory exist before cloning anything.
+            # Then clone the repository into the local cache.
+            cache_dir_parent = Path(cache_dir).parent
+            cache_dir_parent.mkdir(parents=True, exist_ok=True)
+            project.git(['clone', '--', project.url, os.fspath(cache_dir)],
+                        cwd=cache_dir_parent)
```
pdgendt: If the project specified that it has submodules, we could already clone recursive maybe? Not sure if that would conflict with the `fetch` logic in the `else` case..

tests/test_project_caching.py
thorsten-klein: > If anyone cares about caching more submodules and/or better, they can always submit followup patches

Very good idea. I will remove submodules support completely now in this PR.

 > I'm still missing the connection with the workspace tree structure

I will add the workspace tree above each cache tree in the according test cases.

> Maybe start a new "caching" section at the bottom of update --help

I will add some `epilog` there.

src/west/app/project.py
thorsten-klein: I will remove the handling of submodules within this PR completely (as aligned with @marc-hb  here: https://github.com/zephyrproject-rtos/west/pull/829#pullrequestreview-3055691609) 

If there is some follow up request for submodule caching, this should be considered there

src/west/app/project.py
```diff
@@ -1071,6 +1072,67 @@ def do_add_parser(self, parser_adder):
                            help='''cached repositories are in the same relative
                            paths as the workspace being updated. This cache has
                            lower priority (Prio 1).''')
+        group.add_argument('--auto-cache',
+                           help='''automatically setup local cache repositories
+                           in a flat folder hierarchy, but with an additional
+                           subfolder (hashed name) for different remote urls.
```
```suggestion
                           subfolder (hashed name) for different remote URLs.
```

src/west/app/project.py
```diff
@@ -1071,6 +1072,67 @@ def do_add_parser(self, parser_adder):
...
+                           Each local cache repository is automatically cloned
+                           on first usage and synced on subsequent clones.
+                           This cache has lowest priority (Prio 2).''')
```
```suggestion
                           This cache has the lowest priority (Prio 2).''')
```

src/west/app/project.py
```diff
@@ -1071,6 +1072,67 @@ def do_add_parser(self, parser_adder):
...
+          > Path Cache
+          Repositories should be present in a folder hierarchy that matches the
+          workspace's path layout.
+            (path cache directory)
+            ├── bar
+            └── subdir
+                └── foo
+
+          > Auto Cache
+          Use this when you prefer not to manage local caches manually.The complete
```
```suggestion
          Use this when you prefer not to manage local caches manually. The complete
```

src/west/app/project.py
```diff
@@ -1647,12 +1774,23 @@ def fetch(self, project, stats, take_stats):
...
+
+        # automatically fetch the auto-cache so that it is up-to-date
+        self.handle_auto_cache(project)
+
+        # fetch workspace repository from freshly synced auto-cache if
+        # auto-cache is used. Otherwise fetch directly from remote url.
+        if self.auto_cache:
+            fetch_url = self.project_auto_cache(project)
+        else:
+            fetch_url = project.url
```
pdgendt: Nit:

```suggestion
        fetch_url = self.project_auto_cache(project) or project.url
```

pdgendt (approved):
Looking good!

alainsanguinetti:
This looks great! Looking forward to seeing it merged and released!

pdgendt:
> This looks great! Looking forward to seeing it merged and released!

Same here, once merged I'm going to start working on a 1.5 release, as there don't appear to be any other ongoing feature/bugfix PRs.
Just needs some more eyes and approvals here.

marc-hb:
I'll try to find some  time this week, sorry I've been very busy lately

> This looks great! Looking forward to seeing it merged and released!

@alainsanguinetti, others: you do know that you can test this PR without waiting for any of that, right? There are multiple ways listed in the README.txt, pick the one you like.

"I've used this in anger and it surpassed my wildest expectations" puts many times more pressure than "this looks cool, looking forward to try it". Who knows, you could even find some issues and/or things to make better.

pdgendt:
> you do know that you can test this PR without waiting for any of that, right?

Yes, I do, but I also think a feature like this looks great to use in CI environments with caching options. You'd probably don't want to install unreleased/experimental versions in those cases.
 
> "I've used this in anger and it surpassed my wildest expectations" puts many times more pressure than "this looks cool, looking forward to try it". Who knows, you could even find some issues and/or things to make better.

We can only hope, but the few people following PRs probably won't compare to when a release would be published on the mailing list, triggering more users to take a look at the new goodies.

thorsten-klein:
While using the auto cache in our CI today I see following error:

```
=== updating zephyr (zephyr-rtos):
HEAD is now at c14a756cc03 dts: arm: st: c0: stm32c091 set exti lines to 32
 ! [rejected]                dependabot/github_actions/actions-deps-b0008bc5e3 -> dependabot/github_actions/actions-deps-b0008bc5e3  (non-fast-forward)
   b7b6d39a452..5c7e12b53ac  main       -> main
FATAL ERROR: command exited with status 1: fetch -- https://github.com/zephyrproject-rtos/zephyr 'refs/heads/*:refs/heads/*'
```

I have been able to solve it by removing the zephyr folder from auto cache. 
Has someone seen such an error before?

Is it maybe better to do a `git --bare clone` within the auto cache to avoid this kind of error?

pdgendt:
> While using the auto cache in our CI today I see following error:
> 
> ```
> === updating zephyr (zephyr-rtos):
> HEAD is now at c14a756cc03 dts: arm: st: c0: stm32c091 set exti lines to 32
> From https://github.com/zephyrproject-rtos/zephyr
>  ! [rejected]                dependabot/github_actions/actions-deps-b0008bc5e3 -> dependabot/github_actions/actions-deps-b0008bc5e3  (non-fast-forward)
>    b7b6d39a452..5c7e12b53ac  main       -> main
> FATAL ERROR: command exited with status 1: fetch -- https://github.com/zephyrproject-rtos/zephyr 'refs/heads/*:refs/heads/*'
> ```

I'm guessing the branch was updated remotely? Maybe even forcefully.

> I have been able to solve it by removing the zephyr folder from auto cache. Has someone seen such an error before?
> 
> Is it maybe better to do a `git --bare clone` within the auto cache to avoid this kind of error?

Are you still able to have branch names in the manifest revision this way?

pdgendt:
Would it be better to do a `--mirror` clone? From the man pages:

> Set up a mirror of the source repository. This implies `--bare`. Compared to `--bare`, `--mirror` not only maps local branches of the source to local branches of the target, it maps all refs (including remote-tracking branches, notes etc.) and sets up a refspec configuration such that all these refs are overwritten by a `git remote update` in the target repository.

thorsten-klein:
> I'm guessing the branch was updated remotely? Maybe even forcefully.

I am using now `git fetch --force` which has worked in my situation as well.

> Would it be better to do a --mirror clone?

If you prefer I can adapt the PR to use `git clone --mirror`. I have used normal git clone since also all other existing caches work this way.

pdgendt:
> If you prefer I can adapt the PR to use `git clone --mirror`. I have used normal git clone since also all other existing caches work this way.

I'm thinking this is closer to what we want to achieve, but done by `git`. Instead of making sure we pass in the correct arguments to `git fetch` ourselves.

Would it be possible to simplify to `git clone --mirror` for the case where the cache misses, and `git remote update --prune` for fetching?

EDIT: I toyed a bit with some local repositories, and this combination appears to work very nicely, even for forced updates.

thorsten-klein:
> Would it be possible to simplify to git clone --mirror for the case where the cache misses, and git remote update --prune for fetching?

Cool. I have never used `git remote update` before. It works perfectly fine here.

pdgendt:
Ping @marc-hb @carlescufi, I'd like a second pair of eyes if possible before merging.

@carlescufi you suggested at ZDS that the path directory could be optional, I'm not sure I agree, I'd prefer users of this making a conscious decision where to put the caches. If a "default" is desired, users can run (only needed once)

```shell
$ west config --global update.auto-cache -- <desired cache path>
```

As the cache is likely to be shared between multiple workspaces, or in CI a default location would be harder to choose.

src/west/app/project.py
```diff
@@ -1071,6 +1072,67 @@ def do_add_parser(self, parser_adder):
...
+        -------
+        Projects are typically initialized by fetching from remote URLs. However,
+        they can also be cloned from local caches on the file system. Three types of
+        caches are supported, and they may be used together. They are searched in the
+        following order of priority:
+          - Priority 0: --name-cache
+          - Priority 1: --path-cache
+          - Priority 2: --auto-cache
+
+        Local caches are only used during the initial setup of your workspace.
```
carlescufi: @thorsten-klein this is not quite true in the case of auto-cache, is it? you fetch into the auto-cache from the remote and then fetch from the auto-cache into the actual project folder?

src/west/app/project.py
```diff
@@ -1071,6 +1072,67 @@ def do_add_parser(self, parser_adder):
...
+        Once a repository is cloned from a local cache, its remote URL is updated to
+        point to the actual remote source. As a result, all future updates will pull
+        changes directly from the remote.
```
carlescufi: Again, this is not quite true in the case of auto-cache?

carlescufi (commented):
Really nice work @thorsten-klein, I just have a couple of doc questions

src/west/app/project.py
thorsten-klein: Yes, good finding. This has slightly changed over the PR evolution. I will change to the following after I am back from vacation:

After the initial workspace setup, the remote URL of each repo is switched back to the original remote. This means, for name-cache and path-cache, the local caches are only used during the initial workspace setup. In case of auto-cache, also on subsequent updates the local cache is updated first and then used as the source for fetching changes.

src/west/app/project.py
thorsten-klein: I will change to the following:

This means that for manual updates, as well as when using name-cache or path-cache, changes are pulled directly from the remote. Only with auto-cache, the process updates the local cache first, which then serves as the source for pulling changes into the workspace.

carlescufi:
> @carlescufi you suggested at ZDS that the path directory could be optional, I'm not sure I agree, I'd prefer users of this making a conscious decision where to put the caches. If a "default" is desired, users can run (only needed once)

I think it's debatable, given how west already uses `XDG_HOME` and its equivalents, and how other tools do not require you to set a directory manually (e.g. pip). But this can be changed later actually, modifying the option to allow both empty and parametrized entries.

pdgendt:
Thanks @thorsten-klein !

@carlescufi PTAL

carlescufi:
Thanks @thorsten-klein

PR #830, opened 2025-07-16
README.rst: add `pip3 install branch.zip` example

marc-hb:
As just shared by @pdgendt in (unrelated) #824.

This is extremely useful for testing.


pdgendt:
Also, please don't add my GitHub handle in the commit message, as this creates many more notifications 😃

kartben:
This is the first time I'm seeing it done this way, is there a reason not to use the git+https:// scheme instead? I think it's a bit more idiomatic, and should one ever want to `pip freeze` based off this install they end up doing it for the actual git commit as opposed to the checksum of a zip file that might disappear as soon as the branch gets updated.

```
pip3 install git+https://github.com/zephyrproject-rtos/west.git@v1.4-branch
```

marc-hb:
> pip install https://github.com/zephyrproject-rtos/west/archive/pull/830/head.zip

Good idea.

> is there a reason not to use the git+https:// scheme instead?

I just tried `git+https` and I was not impressed, sorry. It uses git but then it seems to throw away all the git repo and data?! I bet it's very useful for downloading outside Github when the git server does not automatically create a .zip file but we don't have that problem here.

I think git+https would muddy the waters compared to these two, very obvious methods:

1. From git. 2 steps: clone + pip. Everything git offers.
2. Not from git. 1 step: pip install .zip. Nothing git offers.

Simple.

> should one ever want to pip freeze based off this install

I don't think people should "freeze" unreleased versions or even worse: code still being developed. I'm sure they _can_ freeze anything they want if they try hard enough but I don't think we should make that easier. The purpose of this PR is to give bug reporters a super fast way to test a (candidate) fix and then immediately _throw it away_ so we don't have many different versions in use out there by accident.

I have no problem adding a link to https://pip.pypa.io/en/stable/topics/vcs-support/ for completeness but I'm very reluctant to add an example.

marc-hb:
>  the option to install the version in a pull request directly:

Just added.

Also added another commit that moves testing _after_ downloading. This messes up the diff of the entire PR so look at the diff of each commit separately.

PR #831, opened 2025-07-22
allow combining `--name-cache` and `--path-cache` for `west update`

thorsten-klein:
This PR contains preparations for PR #829.

- `--name-cache` and `--path-cache` can be combined and used at the same time (`--name-cache` is preferred over `--path-cache`)
- All tests regarding caching are moved to a separate file
- Existing tests regarding caching are extended to`assert` correct remote url for each cloned repository
- New tests for combined caches are added


tests/test_project_caching.py
```diff
@@ -0,0 +1,265 @@
...
+    # Move the repositories out of the way and test the configuration option.
+    # (We can't use shutil.rmtree here because Windows.)
+    shutil.move(os.fspath(foo), os.fspath(tmpdir))
+    shutil.move(os.fspath(bar), os.fspath(tmpdir))
+    cmd(['config', 'update.path-cache', path_cache_dir])
+    cmd('update')
+    assert foo.check(dir=1)
+    assert bar.check(dir=1)
+    assert rev_parse(foo, 'HEAD') == foo_head
+    assert rev_parse(bar, 'HEAD') == bar_head
```
thorsten-klein, line 158: Here it could also be checked that the remote url of `foo` and `bar` is correctly set to the original remote url (and not the path of the local cache). Shall I add this check here?

tests/test_project.py
```diff
@@ -71,7 +71,7 @@
                  'add']
 
 # Helper string for the same purpose when running west update.
-PROTOCOL_FILE_ALLOW = '--submodule-init-config protocol.file.allow=always'
+PROTOCOL_FILE_ALLOW = '--submodule-init-config=protocol.file.allow=always'
```
marc-hb: Is this really required or just a minor cleanup "in passing"? Either way, tell us in the commit message. Need to know how much it matters.

tests/test_project_caching.py
marc-hb: If it's not a ton of code and you already know how to do it, then just go for it.

You can never have too much testing and too many asserts. You can only waste time[*] and effort on the wrong coverage priorities.

[*] developer or cpu cycles

pdgendt:
> please don't move existing tests AND add new tests to the same file in the same commit.

So just a bit of cleanup in the commits:
- A commit for the behavioral change (and any test that would fail because of it)
- Moving the tests
- Adding new tests

Thanks!

pdgendt:
@thorsten-klein could you add a signed-off to your commits? While not enforced (I think) in CI, this repository tries to follow the same contribution guidelines as Zephyr.

Typical commit style:

```git
area: Subject of change

Detailed description on rationale of the change.

Signed-off-by: Full Name <full.name@email.org>
```

PR #832, opened 2025-08-02
.github: Add job names and don't persist checkout credentials

str4t0m:
I reviewed the workflows and noticed that we could avoid the default behavior of the auth token being persisted in the local git confg for subsequent steps of the git checkout action.
Additionally I added a few names to jobs that didn't have any yet.

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -33,6 +33,8 @@ jobs:
     steps:
     - name: Checkout repository
       uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+      with:
+        persist-credentials: false
 
     - name: Initialize CodeQL
       uses: github/codeql-action/init@60168efe1c415ce0f5521ea06d5c2062adbeed1b # v3.28.17
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -19,6 +19,8 @@ jobs:
     steps:
       - name: 'Checkout Repository'
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
+        with:
+          persist-credentials: false
 
       - name: 'Dependency Review'
         uses: actions/dependency-review-action@8805179dc9a63c54224914839d370dd93bd37b2e # v4
--- a/.github/workflows/dependency-review.yml
+++ b/.github/workflows/dependency-review.yml
@@ -15,6 +15,7 @@ permissions:
 
 jobs:
   dependency-review:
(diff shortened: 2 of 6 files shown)


marc-hb (commented):
Can you please elaborate a bit more in the commit message about the risk of persisting credentials? In general, security-related changes need longer commit messages.

EDIT: a wrong default setting is a "serious accusation" for GitHub :-)

Also, please start the subject with the important change and mention the cosmetic change only second (if any space left in 80 characters; otherwise the commit message is enough)/.

str4t0m:
> Can you please elaborate a bit more in the commit message about the risk of persisting credentials? In general, security-related changes need longer commit messages.

I see it as good practise. A few things need to come together that it becomes a problem. Starting with something trivial like accidentally uploading the .git folder as artifact. Which some definitely managed to do with pre 4.4 upload-artifact actions. See for example https://unit42.paloaltonetworks.com/github-repo-artifacts-leak-tokens/.

 The difference is that in the default cause the .git/config file has an entry with the token like the following until the end of the job.
```
[http "https://github.com/"]
         extraheader = AUTHORIZATION: basic eC6hY2Nlc3MtdGk2Z746Z2hzX1JIdEhWEdhqdVRBeFR29dk2NzF3Tkh==
```

> EDIT: a wrong default setting is a "serious accusation" for GitHub :-) Please substantiate it.
> 
> Also, please start the subject with the important change and mention the cosmetic change only second (if any space left in 80 characters; otherwise the commit message is enough)/.

Sorry didn't intend to make any accusations, and there is also the issue 485 open in the checkout action repo since quite a while. I think they prefer not to break users work-flows depending on this behaviour, over making such a change.
I'll amend a short note to the commit message later.

marc-hb:
> I see it as good practise.

Why not but that's vague. Need to be more specific on which particular door this PR is closing.

> A few things need to come together that it becomes a problem.

That's typical with security issues or accidents in general: it's always a _combination_ of factors. No need to give an entire combination in the commit message, you can just focus on the relevant part. If a full, actual incident helps understand this bit then why not but it's not required. A long example could also be distracting.

> Sorry didn't intend to make any accusations,

Afraid you didn't get my half-serious (my bad) comment. When overriding a default Github setting, then you SHOULD make a Github "accusation" (for the lack of a better word), "accusing" Github of getting the default value wrong (my bet: to avoid breaking existing stuff). In other words, the commit message must explain why the default value chosen by Github is wrong - since you are overriding it. If the default value were not wrong, then why would you override it? (Now erring on the "too verbose" side)

str4t0m:
I added a bit more description in the commit message. And pushed right before I saw your latest message so let me know if you think it's fine or if you prefer better describing the reason for the change.

PR #833, opened 2025-08-06
support environment variables `WEST_PATH_CACHE` and `WEST_NAME_CACHE`

thorsten-klein:
While working on #829 I found it helpful to be able to specify a auto cache directory via environment variable.
 
With this change the configuration of cache paths can be done more flexible via environment variables. This is especially helpful if the cache path is located in the user's home directory and users want to avoid that west config contains any absolute paths.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -1066,11 +1066,15 @@ class Update(_ProjectCommand):
         group.add_argument('--name-cache',
                            help='''cached repositories are in subdirectories
                            matching the names of projects to update. This cache
-                           has highest priority (Prio 0).''')
+                           has highest priority (Prio 0). If not set, falls
+                           back to the value of west config update.name-cache,
+                           or WEST_NAME_CACHE environment variable.''')
         group.add_argument('--path-cache',
                            help='''cached repositories are in the same relative
                            paths as the workspace being updated. This cache has
-                           lower priority (Prio 1).''')
+                           lower priority (Prio 1). If not set, falls back to
+                           the value of west config update.path-cache, or
+                           WEST_PATH_CACHE environment variable.''')
 
         group = parser.add_argument_group(
             title='fetching behavior',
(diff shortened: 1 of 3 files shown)


src/west/app/project.py
```diff
@@ -1151,8 +1155,16 @@ def init_state(self, args):
...
+        self.path_cache = (
+            args.path_cache or
+            config.get('update.path-cache') or
+            os.environ.get('WEST_PATH_CACHE')
+        )
+        self.name_cache = (
+            args.name_cache or
+            config.get('update.name-cache') or
+            os.environ.get('WEST_NAME_CACHE')
+        )
```
pdgendt, line 1167: Hmn, I am a bit conflicted about this change. Is it due to having these defined in some `.bashrc` file that you'd like to use it like this?
You can set the cache options on a `global` configuration, isn't that sufficient?

thorsten-klein: > Is it due to having these defined in some .bashrc file that you'd like to use it like this

Yes exactly. Personally I prefer to use env variables instead of maintaining multiple different config files at scattered places .
If you don't like, it is also completely fine for me to arrange with the global configuration (`$XDG_CONFIG_HOME/west/config`) for specifying the cache directories. 

It was just some idea for an improvement and I wanted to contribute and document it for others.
Feel free to decline this PR (and possibly reopen at a later point of time again).

marc-hb: > Personally I prefer to use env variables instead of maintaining multiple different config files at scattered places .

`west config` can be centralized. 

I find environment variables generally evil (TM) because they are "sneaky": extremely hard to trace. Especially nasty when defined in some secure CI system very few users have access to.

Like global variables but only much worse. I think they should be used only as a very last resort, not just for "convenience".

thorsten-klein:
> Going to put a NAK for the time being

PR is closed according to NAK.

marc-hb:
@thorsten-klein can you check whether a global `west config` solves your problems? We may not agree with this solution but we always care about the user experience!

PR #839, opened 2025-09-17
python: Drop 3.9 support

pdgendt:
Python 3.9 is almost EOL, remove it from the supported versions.

Firstly bump the `ruff` version to make sure we have latest versions of linters.
With the minimum required version set to 3.10, some linter changes are needed, namely:

- Replace `Optional[Type]` with `Type | None` [Ruff UP045](https://docs.astral.sh/ruff/rules/non-pep604-annotation-optional/)
- Replace `Union[X, Y]` with `X | Y` [Ruff UP007](https://docs.astral.sh/ruff/rules/non-pep604-annotation-union/)
- Add `strict` argument to `zip` calls [Ruff B905](https://docs.astral.sh/ruff/rules/zip-without-explicit-strict/)
- Import `Callable` from `collections.abc` instead of `typing` [Ruff UP035](https://docs.astral.sh/ruff/rules/deprecated-import/)

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.10', '3.11', '3.12', '3.13']
     steps:
       - name: Checkout
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 # v4.2.2
--- a/src/west/app/main.py
+++ b/src/west/app/main.py
@@ -26,7 +26,7 @@ from collections import OrderedDict
 from io import StringIO
 from pathlib import Path, PurePath
 from subprocess import CalledProcessError
-from typing import NamedTuple, Optional
+from typing import NamedTuple
 
 import colorama
 
@@ -84,9 +84,9 @@ class EarlyArgs(NamedTuple):
     # Expected arguments:
     help: bool                  # True if -h was given
     version: bool               # True if -V was given
-    zephyr_base: Optional[str]  # -z argument value
+    zephyr_base: str | None     # -z argument value
     verbosity: int              # 0 if not given, otherwise counts
-    command_name: Optional[str]
+    command_name: str | None
 
     # Other arguments are appended here.
     unexpected_arguments: list[str]
--- a/src/west/commands.py
+++ b/src/west/commands.py
@@ -14,11 +14,12 @@ import subprocess
 import sys
 from abc import ABC, abstractmethod
 from collections import OrderedDict
+from collections.abc import Callable
 from dataclasses import dataclass
 from enum import IntEnum
 from pathlib import Path
 from types import ModuleType
-from typing import Callable, NoReturn
+from typing import NoReturn
 
 import colorama
 import pykwalify
--- a/src/west/commands.py
+++ b/src/west/commands.py
@@ -18,7 +18,7 @@ from dataclasses import dataclass
 from enum import IntEnum
 from pathlib import Path
 from types import ModuleType
-from typing import Callable, NoReturn, Optional
+from typing import Callable, NoReturn
 
 import colorama
 import pykwalify
@@ -152,7 +152,7 @@ class WestCommand(ABC):
         self.accepts_unknown_args: bool = accepts_unknown_args
         self.requires_workspace = requires_workspace
         self.verbosity = verbosity
-        self.topdir: Optional[str] = None
+        self.topdir: str | None = None
         self.manifest = None
         self.config = None
         self._hooks: list[Callable[[WestCommand], None]] = []
@@ -170,8 +170,8 @@ class WestCommand(ABC):
 
     def run(self, args: argparse.Namespace, unknown: list[str],
             topdir: PathType,
-            manifest: Optional[Manifest] = None,
-            config: Optional[Configuration] = None) -> None:
+            manifest: Manifest | None = None,
+            config: Configuration | None = None) -> None:
         '''Run the command.
 
         This raises `west.commands.CommandContextError` if the command
@@ -269,7 +269,7 @@ class WestCommand(ABC):
                      'Try "west -vv manifest --validate" to debug.')
         return self._manifest
 
-    def _set_manifest(self, manifest: Optional[Manifest]):
+    def _set_manifest(self, manifest: Manifest | None):
         self._manifest = manifest
 
     # Do not use @property decorator syntax to avoid a false positive
@@ -294,7 +294,7 @@ class WestCommand(ABC):
                      "variables, which were not available.")
         return self._config
 
-    def _set_config(self, config: Optional[Configuration]):
+    def _set_config(self, config: Configuration | None):
         self._config = config
 
     config = property(_get_config, _set_config)
@@ -602,7 +602,7 @@ class WestExtCommandSpec:
     factory: _ExtFactory
 
 def extension_commands(config: Configuration,
-                       manifest: Optional[Manifest] = None):
+                       manifest: Manifest | None = None):
     # Get descriptions of available extension commands.
     #
     # The return value is an ordered map from project paths to lists of
--- a/tests/test_manifest.py
+++ b/tests/test_manifest.py
@@ -393,7 +393,7 @@ def test_project_revisions():
     ''')
     expected = [Project('p1', 'u1', revision='defaultrev'),
                 Project('p2', 'u2', revision='rev')]
-    for p, e in zip(m.projects[1:], expected):
+    for p, e in zip(m.projects[1:], expected, strict=True):
         check_proj_consistency(p, e)
 
     # The default revision, if not given in a defaults section, is
@@ -776,7 +776,7 @@ def test_self_tag():
                 Project('testproject2', 'https://example2.com/testproject2')]
 
     # Check the projects are as expected.
-    for p, e in zip(m.projects, expected):
+    for p, e in zip(m.projects, expected, strict=True):
         check_proj_consistency(p, e)
 
     # With a "self: path:" value, that will be available in the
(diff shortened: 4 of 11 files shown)


.github/workflows/package.yml
```diff
@@ -17,7 +17,7 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
         with:
-          python-version: "3.13"
+          python-version: "3.14.0-rc.2"
```
marc-hb: Still a package noob sorry: shouldn't this be lowest version to maximize backwards compatibility?

marc-hb (commented):
This is just a draft so I imagine you will split this into smaller commits / smaller PRs?

For instance: does 3.9 become EOL at the same second 3.14 gets released? :-)

.github/workflows/package.yml
pdgendt: It shouldn't matter that much, but it also doesn't hurt, will update.

pdgendt:
> This is just a draft so I imagine you will split this into smaller commits / smaller PRs?

This wasn't the initial thought. It's more like "flipping the switch" kind of thing.

> For instance: does 3.9 become EOL at the same second 3.14 gets released? :-)

[More or less](https://devguide.python.org/versions/), dropping support for 3.9 is the "big" thing here, where the changes come from. Supporting 3.14 is simply adding it to the matrix.

tests/test_manifest.py
```diff
@@ -393,7 +393,7 @@ def test_project_revisions():
     ''')
     expected = [Project('p1', 'u1', revision='defaultrev'),
                 Project('p2', 'u2', revision='rev')]
-    for p, e in zip(m.projects[1:], expected):
+    for p, e in zip(m.projects[1:], expected, strict=True):
```
marc-hb, line 396: Can you explain how this is related to the Python version? I find this one hard to guess...

tox.ini
```diff
@@ -33,7 +33,7 @@ deps =
     types-PyYAML
     flake8
     mypy
-    ruff==0.8.1
+    ruff==0.13.0
```
marc-hb, line 36: Why not, but how is this related to Python versions?

.github/workflows/test.yml
```diff
@@ -18,7 +18,7 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
-        python-version: ['3.9', '3.10', '3.11', '3.12', '3.13']
+        python-version: ['3.10', '3.11', '3.12', '3.13', '3.14.0-rc.2']
```
marc-hb: Sorry for being pedantic but I don't see why the latest release candidate should be added in the same commit that removes 3.9.

src/west/app/main.py
marc-hb, line 1: I imagine ~3.9~ 3.14 dropped support for the old syntax? Can you make that explicit in the commit message and subject?

src/west/commands.py
marc-hb, line 1: I miss how ruff warnings are related to python versions.

tests/test_manifest.py
pdgendt: I've updated the commits/PR descriptions. It's needed now that the `requires-python` version in `pyproject.toml` has been bumped to 3.10. The linter applies rules based on the minimum required version.

tox.ini
pdgendt: Removed

.github/workflows/test.yml
pdgendt: Removed the 3.14 version

tox.ini
pdgendt: UPDATE: it _is_ needed, as version 0.8 applies rule [UP038](https://docs.astral.sh/ruff/rules/non-pep604-isinstance/), which has been removed in a later version.

marc-hb (commented):
Thanks for all the details.

I think a more accurate PR title would be: "Drop 3.9 AND leverage new 3.10 features for some cleanups". That would have avoided some confusion of mine.

All these cleanups look good but I don't understand the rush. None of these cleanups looks very useful; they're just cleanups. Python 3.9 is EOL soon but not even yet! Some distros and users lag behind; this could force some users to upgrade Python which can be painful sometimes.

In my ideal world, 3.9 usage gets blocked when there were some major, desirable 3.10 feature(s) bringing some significant improvement (of any kind). But AFAICT this PR is just cleanups.

pdgendt:
> I think a more accurate PR title would be: "Drop 3.9 AND leverage new 3.10 features for some cleanups". That would have avoided some confusion of mine.

Right, sorry about that. The `ruff` linter will not pass in CI because bumping the minimal version will enable newer rules.

> All these cleanups look good but I don't understand the rush. None of these cleanups looks very useful; they're just cleanups. Python 3.9 is EOL soon but not even yet! Some distros and users lag behind; this could force some users to upgrade Python which can be painful sometimes.

There's no rush? But 3.9 is already 5 years old, and I'd like that `west` follows the supported Python versions. If this goes in, the current `west` version will still support Python 3.9, as v1.5.0 that will be coming out soon, too.

> In my ideal world, 3.9 usage gets blocked when there were some major, desirable 3.10 feature(s) bringing some significant improvement (of any kind). But AFAICT this PR is just cleanups.

If users want to keep holding on to an EOL Python version, that's fine, but IMO then they should keep using the version of `west` that had support for it too.

marc-hb:
> There's no rush? But 3.9 is already 5 years old, 

"It's old" is not a technical justification. In any case only the EOL date matters, the first release date does not.

> and I'd like that west follows the supported Python versions. 

OK, but for which specific, west reasons? What advantage does this give to `west` users? So far I only see in this PR minor cleanups for `west` developers and zero value for `west` users (and negative value for 3.9 users)

> but IMO then they should keep using the version of west that had support for it too.

Why should they? Upgrading `west` is a minor change, upgrading Python is a big one. We want as many users as possible testing the latest `west` version, no matter what Python version they use[*]. Policing users and forcing them to upgrade other, EOL software is not part of `west` goals.


[*] unless there's some annoying issue in Python 3.9 triggered by `west`. In such a case then yes let's drop 3.9. 

A reminder that some Zephyr users do NOT use west already: https://github.com/zephyrproject-rtos/zephyr/pull/93944
Let's not give some users one reason to stop using west.

pdgendt:
> > There's no rush? But 3.9 is already 5 years old,
> 
> "It's old" is not a technical justification. In any case only the EOL date matters, the first release date does not.

Outdated then? EOL is just around the corner.

> > and I'd like that west follows the supported Python versions.
> 
> OK, but for which specific, west reasons? What advantage does this give to `west` users? So far I only see in this PR minor cleanups for `west` developers and zero value for `west` users (and negative value for 3.9 users)

I much prefer these smaller incremental bumps, and there is no negative impact for Python 3.9 users yet. We're not releasing this asap? We're still busy releasing v1.5.0, and Python 3.9 users can happily use that, forever.

> > but IMO then they should keep using the version of west that had support for it too.
> 
> Why should they? Upgrading `west` is a minor change, upgrading Python is a big one. We want as many users as possible testing the latest `west` version, no matter what Python version they use[*]. Policing users and forcing them to upgrade other, EOL software is not part of `west` goals.
> 
> [*] unless there's some annoying issue in Python 3.9 triggered by `west`. In such a case then yes let's drop 3.9.

From that point of view, if Python 3.13 had a feature we really wanted, we should drop support for 3.9 to 3.12? I'm obviously over-exaggerating, but keeping a support window of 5 years is already more than most LTS-s.
And who's not to say we couldn't even release v1.5.1 if users find a bug that would impact those left behind?

> A reminder that some Zephyr users do NOT use west already: [zephyrproject-rtos/zephyr#93944](https://github.com/zephyrproject-rtos/zephyr/pull/93944) Let's not give some users one reason to stop using west.

Zephyr users [dropped Python 3.9](https://github.com/zephyrproject-rtos/zephyr/pull/72142) support a long time ago.

marc-hb:
> Outdated then? EOL is just around the corner.

"Outdated" may sound more technical than "old" but I don't think it makes much difference.

> From that point of view, if Python 3.13 had a feature we really wanted, we should drop support ~for~from 3.9 to 3.12? I'm obviously over-exaggerating,

It all depends how many 3.9-3.12 users there are left and at what point in time. It's always a trade-off, this is never a yes/no question.

> I much prefer these smaller incremental bumps, and there is no negative impact for Python 3.9 users yet. We're not releasing this asap? We're still busy releasing v1.5.0, and Python 3.9 users can happily use that, forever.

> And who's not to say we couldn't even release v1.5.1 if users find a bug that would impact those left behind?

> Zephyr users https://github.com/zephyrproject-rtos/zephyr/pull/72142 support a long time ago.

These are all good reasons, thanks. Especially the last one! "3.9 is old" does not do them justice, avoid next time?

src/west/app/main.py
marc-hb: Answering myself based on newer commit messages: no, it's the linters recommending the newer, 3.10 syntax.

pdgendt:
> > From that point of view, if Python 3.13 had a feature we really wanted, we should drop support ~for~from 3.9 to 3.12? I'm obviously over-exaggerating,
> 
> It all depends how many 3.9-3.12 users there are left and at what point in time. It's always a trade-off, this is never a yes/no question.

I agree, here the trade-off is to gradually move forward, and should come to no surprise for users if kept consistent. We [dropped support for Python 3.8](https://github.com/zephyrproject-rtos/west/pull/752) almost one year ago.

> > I much prefer these smaller incremental bumps, and there is no negative impact for Python 3.9 users yet. We're not releasing this asap? We're still busy releasing v1.5.0, and Python 3.9 users can happily use that, forever.
> 
> > And who's not to say we couldn't even release v1.5.1 if users find a bug that would impact those left behind?
> 
> > Zephyr users [zephyrproject-rtos/zephyr#72142](https://github.com/zephyrproject-rtos/zephyr/pull/72142) support a long time ago.
> 
> These are all good reasons, thanks. Especially the last one! "3.9 is old" does not do them justice, avoid next time?

Just for the record, I said "is already 5 years old", which is different 😛. And was as a referral to the [end-of-life](https://devguide.python.org/versions/#:~:text=released%20as%20needed.-,end%2Dof%2Dlife%3A,-Five%20years%20after) cycle.

marc-hb (approved):

kartben (approved):

PR #840, opened 2025-09-18
repo: Patch coverage with subprocess

pdgendt:
Coverage subprocess has been removed in pytest-cov 7.0. To migrate the `patch` option has to be added.
See https://pytest-cov.readthedocs.io/en/latest/subprocess-support.html

@marc-hb the [coverage](https://app.codecov.io/gh/zephyrproject-rtos/west) plummeted to 48% for no apparent reason because, in CI, we simply install the latest versions of all tools. 
As it took quite some time to figure this out, I intend to come back to locking dependency versions in CI. But with a better proposal, stay tuned.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -49,6 +49,7 @@ where = ["src"]
 namespaces = false
 
 [tool.coverage.run]
+patch = ["subprocess"]
 relative_files = true
 omit = [
   "*/tmp/*",


carlescufi (approved):

marc-hb:
> As it took quite some time to figure this out, I intend to come back to locking dependency versions in CI. But with a better proposal, stay tuned.

I lost track of the previous discussion sorry (assuming I ever was on top of it...EDIT: #802?) but why can't this simply be solved with TWO `requirements.txt` files? A "very hardcoded" one used by CI and releases (and anyone else who wants it) and more "relaxed", flexible and "traditional" one with just minimum versions? The latter can be used by developers to avoid virtualenv and to spot regressions like you just did. I did exactly that (2 `requirements-*.txt` files) once for a tricky sphinx build and it worked really well.

> As it took quite some time to figure this out,...

If that's any consolation, someone would have had to spend this time eventually anyway and it most likely would have still been you :-/ However I agree it would be better to be in control of the timing!

PR #841, opened 2025-09-18
Codecov test result reporting

pdgendt:
Upload test reports to Codecov to track pytest results.

Test dashboard: https://app.codecov.io/gh/zephyrproject-rtos/west/tests

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -48,8 +48,14 @@ jobs:
           path: "./check out/.coverage"
           include-hidden-files: true
 
-  coverage-report:
-    name: Report test coverage
+      - name: Upload test reports
+        uses: actions/upload-artifact@ea165f8d65b6e75b540449e92b4886f43607fa02 # v4.6.2
+        with:
+          name: tests-${{ matrix.os }}-${{ matrix.python-version }}
+          path: "./check out/junit.xml"
+
+  test-report:
+    name: Report test results
     runs-on: ubuntu-latest
     needs: ["build"]
     steps:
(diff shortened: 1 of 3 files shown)


tox.ini
```diff
@@ -38,7 +38,7 @@ setenv =
     # For instance: ./.tox/py3/tmp/
     TOXTEMPDIR={envtmpdir}
 commands =
-  python -m pytest --cov-report=html --cov-config=pyproject.toml --cov=west {posargs:tests} --basetemp='{envtmpdir}/pytest with space/'
+  python -m pytest --junitxml=junit.xml -o junit_family=legacy --cov-report=html --cov-config=pyproject.toml --cov=west {posargs:tests} --basetemp='{envtmpdir}/pytest with space/'
```
marc-hb: Mmm... this affects not just CI... Is there a way to move this to some CI-only `pytest` config file? "legacy" sounds bad :-D

pdgendt: I've used the alternative naming `xunit1` as Codecov is compatible with the xunit 1.0 format. As `tox` is both used for local testing and CI, I'd prefer to only change it in 1 place.

.github/workflows/test.yml
```diff
@@ -86,3 +92,10 @@ jobs:
         with:
           token: ${{ secrets.CODECOV_TOKEN }}
           slug: ${{ github.repository_owner }}/west
+
+      - name: Upload test results to Codecov
+        if: github.repository_owner == 'zephyrproject-rtos' && ${{ !cancelled() }}
+        uses: codecov/test-results-action@47f89e9acb64b76debcd5ea40642d25a4adced9f # v1.1.1
+        with:
+          token: ${{ secrets.CODECOV_TOKEN }}
+          slug: ${{ github.repository_owner }}/west
```
kartben: do you have reasons to think the slug should be manually passed?

pdgendt: No, copied that from the previous step. Removed.

PR #845, opened 2025-09-19
De-tox and uv-ify

pdgendt:
Remove the tox dependency, which allows to run the testing directly or with other tasks runners (here using [`poethepoet`](https://poethepoet.natn.io/index.html)).

In CI (can be used locally too) we use [`uv`](https://docs.astral.sh/uv/) to install and manage dependencies. A `uv.lock` file has pinned dependencies.

Some benefits of using `uv`:
- Fast environment setup
- Clean pyproject.toml management
- Reproducible builds with lockfiles
- Easy Python version switching

Check individual commits for details.


tests/test_alias.py
```diff
@@ -2,14 +2,11 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-import os
 import subprocess
 
 import pytest
 from conftest import cmd
 
-assert 'TOXTEMPDIR' in os.environ, "you must run these tests using tox"
```
marc-hb, line 11: Ha, I'm finally understanding the "detox" pun :-)

This would be great: because of too many layers, so far my only solution to run a test in a debugger was `PUDB_TTY`. It worked... only most of the time!

- https://github.com/inducer/pudb/issues/598#issuecomment-1839292342

Don't forget to update b7e091d2e902

marc-hb: Unless... the `uv` replacement merely substitutes another layer of indirection, still getting in the way and still making it hard to run a test in a debugger directly?

Sorry for thinking out loud. Hey it's just a draft PR ;-)

uv.lock
marc-hb, line 1: These hundreds of lines can only have been generated.

Mixing human and generated files in `git` tends to make `git log` and some other basic `git` commands unusable :-(

tests/test_alias.py
pdgendt: I want to reduce the layering too, it should be possible to run `pytest` directly IMO. `uv` is an optional utility, but I do want to make it the default for running in CI.

uv.lock
pdgendt: Definitly! This is very much WIP, I will do cleanup before marking it as ready.

pyproject.toml
```diff
@@ -75,6 +75,8 @@ omit = [
 [tool.coverage.report]
 omit = [
   "*/tmp/*",
+  "net-tools/scripts/test.py",
+  "subdir/Kconfiglib/scripts/test.py",
```
marc-hb, line 79: Has this any dependency with this PR?

.github/workflows/package.yml
```diff
@@ -17,13 +17,14 @@ jobs:
       - name: Set up Python
         uses: actions/setup-python@a26af69be951a213d495a4c3e4e4022e16d87065 # v5.6.0
         with:
-          python-version: "3.13"
-      - name: Install Python dependencies
-        run: |
-          pip3 install build
+          python-version: '3.10'
```
marc-hb: This looks like a separate fix for 38aae907fd065ee69c... can you submit separately?

pdgendt: Yes, will remove from this PR.

pyproject.toml
pdgendt: Running the tests "regularly" with `pytest` these files are next to the source files, but the coverage tool can't resolve them.
This wasn't the case for `tox`.

pdgendt:
@thorsten-klein, as you're actively developing for west, would you mind sharing your opinion (if you have any) on this PR?

thorsten-klein:
@pdgendt Nice! Looking forward towards easier debugging of west tests. It seems that with this change pytest can be directly run 👍

mbolivar (approved):
Didn't test, but LGTM

thorsten-klein:
I had a try today and I was not able to run pytest directly. It seems that following points need to be tackled:

1. An entry in pyproject.toml is necessary to extend the pythonpath with src folder.

2. cmd() is calling 'west' which does not exist in the PATH. It would be nice if this function is reworked so that it can also be debugged.

I will have a closer look tomorrow 👍

.github/workflows/test.yml
```diff
@@ -27,19 +27,21 @@ jobs:
...
 
-      - name: Install tox
-        run: pip3 install tox
+      - name: Setup project
+        run: uv sync --frozen --directory "./check out/"
 
-      - name: Run tox
-        run: tox -c 'check out' -- -W error
+      - name: Run tasks
+        run: uv run --frozen --directory "./check out/" poe all
```
marc-hb, line 44: Could you not run "all" here to provide finer-grained results? And durations.

README.rst
marc-hb, line 1: This new version of the README.rst seems to hesitate between embracing the new `uv` and poe approaches for "advanced" developers and sticking to the old pip ways for more basic stuff... I feel like this could needs more review time but I don't have that now. As long as all the examples were tested and do work, then good enough for now! Can be polished later after merging this. The README.rst tail should not wag the `uv` dog and delay this.

uv.lock
marc-hb, line 1: I hate mixing generated and non-generated files in the same git repo but who am I to question the wisdom of the new Youvee religion?

.github/workflows/test-distros.yml
```diff
@@ -75,8 +75,11 @@ jobs:
       - name: Display Python version
         run: python3 -c "import sys; print(sys.version); print(sys.platform)"
 
-      - name: Install tox
-        run: pip3 install tox --break-system-packages
+      - name: Install uv
+        run: pip3 install uv --break-system-packages
```
marc-hb, line 79: GH runners are ephemeral so it does really matter but... just for better "role-modelling" maybe?
```suggestion
        run: pipx install uv
```
Only if pipx is available by default, otherwise forget it.

marc-hb (approved):
Minor comments that can all be addressed later.

I downloaded this and gave it good try and it felt great.  Running pytest is especially easier. Thanks @pdgendt, not the least for forcing me to discover `uv`. It was worth it. Also, now I feel rejuvenated:  like I belong to the cool kids!

marc-hb:
> I had a try today and I was not able to run pytest directly. It seems that following points need to be tackled...

Maybe not directly yet, but you can `uv run pytest [--trace] -k something` without setting any virtualenv or anything and this works even inside Emacs PDB :-)

README.rst
pdgendt: I wanted to keep options open for developers.

BTW, I intend to re-vamp the readme together with moving the docs from Zephyr to a West site, and create a dedicated page for testing west itself.

.github/workflows/test.yml
pdgendt: This is close to what `tox` did, so I'll leave it for now. I think we should split indeed, but also reduce the matrix a bit. We don't need to run the linter task 15 times. Will follow-up!

pdgendt:
> I had a try today and I was not able to run pytest directly. It seems that following points need to be tackled:
> 
> 1. An entry in pyproject.toml is necessary to extend the pythonpath with src folder.

It's a bit tricky, as we want to test an installment, rather than the sources directly? Not sure if the difference is even noticeable? We can document an `pip install -e .`, but that would need to be done for every `pytest` run, so maybe we need to do your suggestion.

> 2. cmd() is calling 'west' which does not exist in the PATH. It would be nice if this function is reworked so that it can also be debugged.

Sure, we can improve on this, I'm going to go ahead with merging as this is a good basis IMO.

thorsten-klein:
> Sure, we can improve on this, I'm going to go ahead with merging as this is a good basis IMO.

Sounds good. I will then open the follow-up PR with my changes to run pytest directly 👍

.github/workflows/test.yml
marc-hb: Submitted in:
- #858

README.rst
marc-hb: Big doc move WIP in:
- #860

marc-hb (commented):
> I will then open the follow-up PR with my changes to run pytest directly

- #859

PR #846, opened 2025-09-23
fixed `RecursionError` in `die()` method

thorsten-klein:
Fixed `RecursionError` in `die()` method which has occured if the config `_config` is missing. 

Additionally added first tests for the log output methods (e.g. `die`, `err`, `wrn`, `inf`, `dbg`).

### DIFF
--- a/src/west/commands.py
+++ b/src/west/commands.py
@@ -526,7 +526,7 @@ class WestCommand(ABC):
     @property
     def color_ui(self) -> bool:
         '''Should we colorize output?'''
-        return self.config.getboolean('color.ui', default=True)
+        return self.config.getboolean('color.ui', default=True) if self.has_config else True
 
     #
     # Internal APIs. Not for public consumption.
--- a/tests/test_commands.py
+++ b/tests/test_commands.py
@@ -2,7 +2,9 @@
 
 import os
 
-from west.commands import WestCommand
+import pytest
+
+from west.commands import Verbosity, WestCommand
 
 assert 'TOXTEMPDIR' in os.environ, "you must run these tests using tox"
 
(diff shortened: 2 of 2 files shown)


thorsten-klein:
Note: You can revert the change in `commands.py` in order to see the `RecurseError`

src/west/commands.py
```diff
@@ -526,7 +526,7 @@ def die(self, *args, exit_code: int = 1) -> NoReturn:
     @property
     def color_ui(self) -> bool:
         '''Should we colorize output?'''
-        return self.config.getboolean('color.ui', default=True)
+        return self.config.getboolean('color.ui', default=True) if self.has_config else True
```
marc-hb, line 529: I think I would prefer erring on the side of caution and default to `False`. The lack of color is less harmful than a build log polluted by control characters.

marc-hb (commented):
Thanks for the extensive test coverage! Could you try reduce the duplication with something like this? UNTESTED

```python

cmd = WestCommandImpl(name="x", help="y", description="z", verbosity=Verbosity.DBG_EXTREME)

testdata = [
   (cmd.dbg, ""),
   (cmd.wrn, ""\x1b[93mWARNING: "),
   ...
   ]

@pytest.mark.parametrize("loglevel,msgprefix", testdata)
  test_log(capsys, loglevel, msgprefix):

```

I'm also curious about the "real-world" situation that triggered this?

src/west/commands.py
thorsten-klein: I took over the default value from `getboolean`. Shall I change only the default for the case that no config is available?

pdgendt:
> I'm also curious about the "real-world" situation that triggered this?

@thorsten-klein is there a scenario where you ran into this? I'd like to know the impact.

thorsten-klein:
I ran into it while writing tests in Zephyr, where I had to mock the `die` method due to this bug in `west`.  

Ref: [zephyrproject-rtos/zephyr#96435 (diff link)](https://github.com/zephyrproject-rtos/zephyr/pull/96435/files#diff-1647d372e3469c455852fce3eeb2c96a1afb7e77ea3c183a7a8645caabb79a09R86)  

If I understand correctly, this issue occurs when `west` attempts to log but no configuration is available.
I’m not sure if users encounter this or if those cases are covered anyhow else in west — I’ve personally never seen it outside of my test case.

src/west/commands.py
thorsten-klein: I prefer to keep `True` so that behavior remains consistent in both cases. 
Otherwise, users might be confused if the default changes depending on whether a config is present or not.

So if a config exists but `color.ui` is not defined, the default is `True` (keep current behavior)
Newly added: if no config exists at all, the default is also `True`.

pdgendt: Knowing the origin of the issue now, this makes sense to me.

pdgendt:
> I ran into it while writing tests in Zephyr, where I had to mock the `die` method due to this bug in `west`.
> 
> Ref: [zephyrproject-rtos/zephyr#96435 (diff link)](https://github.com/zephyrproject-rtos/zephyr/pull/96435/files#diff-1647d372e3469c455852fce3eeb2c96a1afb7e77ea3c183a7a8645caabb79a09R86)
> 
> If I understand correctly, this issue occurs when `west` attempts to log but no configuration is available. I’m not sure if users encounter this or if those cases are covered anyhow else in west — I’ve personally never seen it outside of my test case.

Right, so the issue occurs when running `west` outside of a workspace then? The fix does look trivial, and the additional tests are welcome, thanks!

tests/test_commands.py
```diff
@@ -16,3 +18,103 @@ def test_parse_git_version():
...
+
+test_str = "This is some test string"
+col_red = "\x1b[91m"
+col_yellow = "\x1b[93m"
+col_off = "\x1b[0m"
+
+expected_default = test_str + '\n'
+expected_warning = col_yellow + 'WARNING: ' + test_str + '\n' + col_off
+expected_error = col_red + 'ERROR: ' + test_str + '\n' + col_off
+expected_fatal_error = col_red + 'FATAL ERROR: ' + test_str + '\n' + col_off
```
pdgendt: Constants typically use uppercase names. Also prefer using f-strings instead of concatenation with `+`

thorsten-klein: Done 👍🏻

pdgendt (approved):

marc-hb (approved):

PR #847, opened 2025-09-26
Support `west config --update <file>`

thorsten-klein:
## Use case
We need an easy way to update the user config (local, global or system) with all options and values from another given config.
This is especially helpful when users want to share configuration files within their repository.

## New flag `-u` (`--update`)
Added support for `west config -u` (`--update`), which can be used to update the user's west config with all options and values from another given config file. The `-u`/`--update` argument can be specified multiple times, whereby values from later files override earlier ones.

## Example

command:
```
west config --global -u shared/config.default
west config -u shared/aliases.default -u shared/aliases.jamie -u shared/aliases.pieter # default: local config
```

Whereby the configs could look like this for example: 
`shared/config.default`
```
[update]
auto-cache = /usr/tmp/west/auto-cache

[build]
cmake-args = -DCMAKE_COLOR_DIAGNOSTICS=ON
```

`shared/aliases.default`
```
[alias]
menuconfig = build --pristine=never --target menuconfig
```

`shared/aliases.jamie`
```
[alias]
jamie-board = build -b sama7g54_ek --pristine=always
```

`shared/aliases.pieter`
```
[alias]
pieter-board = build -b acn52832
```

### DIFF
--- a/src/west/app/config.py
+++ b/src/west/app/config.py
@@ -108,6 +108,9 @@ class Config(WestCommand):
                            help="delete an option everywhere it's set")
         group.add_argument('-a', '--append', action='store_true',
                            help='append to an existing value')
+        group.add_argument('-u', '--update', action='append', metavar='FROM_FILE',
+                           help='''update the config with all options and
+                           values from one or more other config files''')
 
         group = parser.add_argument_group(
             "configuration file to use (give at most one)"
@@ -135,6 +138,9 @@ class Config(WestCommand):
         if args.list:
             if args.name:
                 self.parser.error('-l cannot be combined with name argument')
+        elif args.update:
+            if args.name:
+                self.parser.error('-u cannot be combined with name argument')
         elif not args.name:
             self.parser.error('missing argument name '
                               '(to list all options and values, use -l)')
@@ -146,10 +152,12 @@ class Config(WestCommand):
             self.list(args)
         elif delete:
             self.delete(args)
-        elif args.value is None:
+        elif args.name and args.value is None:
             self.read(args)
         elif args.append:
             self.append(args)
+        elif args.update:
+            self.update(args)
         else:
             self.write(args)
 
@@ -157,6 +165,10 @@ class Config(WestCommand):
         what = args.configfile or ALL
         for option, value in self.config.items(configfile=what):
             self.inf(f'{option}={value}')
+
+    def update(self, args):
+        for path in args.update:
+            self.config.update(other=path, configfile=args.configfile or LOCAL)
 
     def delete(self, args):
         self.check_config(args.name)
--- a/src/west/configuration.py
+++ b/src/west/configuration.py
@@ -73,9 +73,16 @@ class _InternalCF:
     def __init__(self, path: Path):
         self.path = path
         self.cp = _configparser()
+        self._read(path)
+
+    def _read(self, path: Path):
         read_files = self.cp.read(path, encoding='utf-8')
         if len(read_files) != 1:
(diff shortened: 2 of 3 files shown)


pdgendt:
Looks neat, not sure about the `update` naming though. Maybe `-i/--import`?

thorsten-klein:
I considered using `--import`, but naming becomes tricky because `import` is a reserved keyword in python.
Instead, I chose `--update`, drawing an analogy to `dict.update()`.

marc-hb:
Any reason why this needs to be a `west` command? It looks like a pure, generic .ini script could do that.

marc-hb:
Rather than having more and more complex code editing files that were originally supposed to be changed by humans, have you considered the `west/config.d/NNN-multiple.conf` approach? Try `ls -ld /etc/*.d/ /etc/*/*.d/` to find some examples (40+ on my system).

This approach is extremely popular because it requires almost no code: just copy, delete or symlink (no duplication!) the files you want. No code = no bug and unlimited "features". At usage time, it requires very little code either: just concatenate the files to read in some well-defined order and that's it.

"Popular" also means it requires little introduction and documentation.

thorsten-klein:
> Any reason why this needs to be a west command?

I like that I can easily update the different west configs (`--local`, `--system`, `--global`) without having to know where they are really stored. The `update` is analog to multiple `set`s (e.g. `west config x.y z`). These are also just simple ini operations and available as west command.

> have you considered the west/config.d/NNN-multiple.conf approach? 

I know the `config.d` approach and actually it would fit for us as well. So you prefer the `.west/config.d/*` approach?

marc-hb:
> I like that I can easily update the different west configs (--local, --system, --global) without having to know where they are really stored. 

There are two configuration approaches: 1) configuration files edited only manually, versus: 2) a more "opaque" and abstract configuration database that is modified only through APIs and front-ends. I realize `west` configuration is in some uncomfortable middle, which your observation in https://github.com/zephyrproject-rtos/west/issues/848#issuecomment-3339303560 illustrates perfectly. This proposition is pushing more towards 2), creating more distance between the user and the files. I don't think that is a good idea. I would rather push users more towards 1), which means less `west` code and letting users do whatever they want with configuration files directly. 

https://docs.zephyrproject.org/latest/develop/west/why.html#design-constraints

> Conservative about features: no features will be accepted without strong and compelling motivation.

Steering users towards the actual files is more "low-level" but it also gives more freedom to _directly_ do whatever they want with them without being limited to what `west` offers. Every layer of indirection adds restrictions and friction (very common problem in build systems). The `.ini` format was chosen because it is very simple, intuitive, standard and popular, every programming language (and every editor) has a parser for it already. So this idea can probably be implemented as a small, standalone program in many languages. In fact this has probably been implemented a few times already. This PR goes in the direction of duplicating that inside `west`.

> I know the config.d approach and actually it would fit for us as well. So you prefer the .west/config.d/* approach?

Yes I think that would be great and better than this.

marc-hb:
> The .ini format was chosen because it is very simple, intuitive, standard and popular, every programming language (and every editor) has a parser for it already. So this idea can probably be implemented as a small, standalone program in many languages.

Some quotes from http://www.catb.org/~esr/writings/taoup/html/ch01s06.html#id2877537

> Rule of Modularity: Write simple parts connected by clean interfaces.

> Rule of Composition: Design programs to be connected with other programs.

thorsten-klein:
> Yes I think that would be great and better than this.

I will close this PR now and create a new one with the `.west/config.d` approach.

Issue #429, opened 2020-08-22
Support 4th level west -c config.ini file(s)


pabigot:
In September 2025, the title was changed by @marc-hb from the original
"add environment support for defaulting arguments"
to:
"Support 4th level west -c config.ini file(s)"

------
Add environment support for defaulting arguments.

It would be wonderfully cool if west had a facility to infer defaults from environment variables.

I tend to build on a variety of boards.  If I could do:

    WEST_BUILD_BOARD=pca10056

and then have:

    rm -rf build && west build

automatically infer `--board pca10056` things would be easier.

Similarly I always sign with:

    WEST_SIGN_KEY=${HOME}/.sensitive/pab-mcuboot.pem

It would be nice to avoid the extra arguments of:

    west sign --key ${HOME}/.sensitive/pab-mcuboot.pem

or (from zephyrproject-rtos/zephyr#27688) have to use:

    west build ... -- -DCONFIG_MCUBOOT_SIGNATURE_KEY_FILE=\"${HOME}/.sensitive/pab-mcuboot.pem\"

It's fairly common that I run the same sequence of commands on different boards to test them or assess why they're different.  Having board-specific arguments complicates reproducing builds on different platforms.  Taking parameters from the environment would allow me to have multiple shell windows, one for each board, with the commands (held in an editor buffer) being portable between them.

The variable name proposed here is `WEST_<command>_<long-option>` which seems automatable.  Alternatively support an [INI file](https://en.wikipedia.org/wiki/INI_file) that held defaults and could be selected by a single new global parameter that specified where to read them.

cc:
- #849 
- #867
- https://github.com/thesofproject/sof/pull/10307

mbolivar-nordic:
I agree this would be a worthwhile enhancement.

> Alternatively support an [INI file](https://en.wikipedia.org/wiki/INI_file) that held defaults and could be selected by a single new global parameter that specified where to read them.

This would be my preference, due to a slight personal allergy to environment variables. I would go for a new global option `-c / --config`, which like `--verbose` could be given to any command. This option would take an argument, which is a configuration file to parse and include at highest precedence (i.e. its settings would override the [local configuration file's](https://docs.zephyrproject.org/latest/guides/west/config.html#west-configuration-files)).

Example:

```
$ cat myconf.ini
[build]
board = nrf52dk_nrf52832
$ west -c myconf.ini build ... # gets build.board nrf52dk_nrf52832
```

@carlescufi  - worth prioritizing?

[...]

PR #849, opened 2025-09-27
Automatically use additional config files from `.d` directories

thorsten-klein:
Successor of #847

All config files that are placed inside according `.d` folders are applied _in addition_ to the actual config exists.

For example 
- `./west/config.d/*` for local configs if `./west/config` exists
- `~/.westconfig.d/*` for global config if `~/.westconfig` exists
- `etc/westconfig.d/*`/ `/usr/local/etc/westconfig.d/*` / `%PROGRAMDATA%\west\config.d/*` for system config (if the according config file exists)

All files are loaded in lexicographic order.
If the same option/key/value is defined in multiple files, the last one loaded wins (i.e., later file (e.g. `99-conf`) override earlier ones (e.g. `01-conf`)).
In the end, the actual config is applied, so if a option's value is defined in the actual config file it overrides the `.d` value.

The paths are respecting according env variables (`WEST_CONFIG_LOCAL`/`WEST_CONFIG_SYSTEM`/`WEST_CONFIG_GLOBAL`). In this case all files under `<WEST_CONFIG_xxx>.d` directory are automatically imported.

For easier determination of the active config file path, a new `west config [--local|--global|--system] --print-path` argument is added, which prints the currently active config directory.
Note: If no local config can be determined, the command does not output anything (it does _NOT_ fail).


src/west/configuration.py
```diff
@@ -68,15 +68,33 @@ def parse_key(dotted_name: str):
 
     @staticmethod
     def from_path(path: Optional[Path]) -> Optional['_InternalCF']:
-        return _InternalCF(path) if path and path.exists() else None
+        config = _InternalCF(path) if path and path.exists() else None
+        if not config:
+            return None
```
pdgendt: What's the reasoning for requiring a `config` file, next to the `config.d` directory?

thorsten-klein: This is the config file which is edited by the west config commands. And I want to make the change backwards compatible.

Updated answer:
Sorry, got the question wrong. 
For me the config.d is only an extension. Therefore the actual config should always exist.

thorsten-klein: This is the config file which is edited by the west config commands. I want to make the change backwards compatible.

src/west/configuration.py
```diff
@@ -68,15 +68,35 @@ def parse_key(dotted_name: str):
...
+
+        # check if dropin configs exist (in according .d directory)
+        dropins_dir = Path(f'{path}.d')
+        if dropins_dir.exists():
+            # update config with values from dropin config files, whereby the
+            # dropin configs are applied in alphabetical order
+            for conf in sorted(dropins_dir.iterdir()):
+                config.cp.read(dropins_dir / conf, encoding='utf-8')
+            # finally overwrite with values from higher-prior actual config
+            config.cp.read(path, encoding='utf-8')
```
pdgendt: So you need to read the "main" `config` file twice to make sure it has precedence over the `config.d/` files. That feels a bit hacky.

pdgendt (commented):
I know this PR was created based on a suggestion from https://github.com/zephyrproject-rtos/west/pull/847#issuecomment-3339189308, but it does add some complexity to the configuration system.
Besides that, testing now has a lot more edge-cases.

At the very least we require some more documentation bits. But I'm not convinced yet that this addition is in the spirit of:
> **Conservative about features:** no features will be accepted without strong and compelling motivation.

Can you provide some insight into where you "really" need this kind of flexibility?

src/west/configuration.py
thorsten-klein: Would it be okay if I add some argument to the constructor of the config which allows to create the instance without directly reading the file?

thorsten-klein:
> Can you provide some insight into where you "really" need this kind of flexibility?

In our company, only a small group of people maintains the build system and tooling and configs (e.g., west).  
Their goal is to set up the development environment correctly and shield everyone else from pitfalls.  
Similarly to this, also customer repositories may define configurations (on top of the generic configs) that should be shared across their product developers to ensure consistency.  
This leads to the need for maintaining **(generic) configs** and a mechanism to overlay configs. 
In the end, maybe something like a "repository config level" is missing for us. 

I hope, that it is a common enough use case for others, so that west itself should support it.  
(Especially, since the existence of other `west config` commands suggests that users already wanted to set options via the command line.) 

My first idea was adding a command like `west config --update`.  
It would allow us to ensure required config values are set. This could be implemented in only few lines (see #847). 
Without such a mechanism, we would have to write our own tool for config-patching or a script which uses currently supported commands to set each option separately.   

An alternative would be to support a dropin directory like `.west/config.d` directory.  
This would allow us to ship shared configs and place them before the build in `.west/config.d` (or symlink the files/folder) 
Users could still override values in `.west/config`, but in most cases that file would remain empty because all necessary options are maintained in the generic `.west/config.d`.. 

The important thing is: the generic configs may change anytime. In this case the new config values should be automatically applied if the users are using the default configs. If they have set some different value (on their own risk), their value shall be used. This scenario is perfectly fitting to the drop-in mechanism. 

I hope you understand roughly what we do.

src/west/configuration.py
pdgendt: `config.cp.read` actually allows passing [multiple filenames](https://docs.python.org/3/library/configparser.html#configparser.ConfigParser.read), why not pass multiple paths to the `__init__` function?

thorsten-klein: There is a check at later point that the number of read configs is exactly 1. I can remove this check and rework the for loop and instead pass the list of multiple configs.

Nevertheless: what should be the path attribute of the class in case of multiple configs? This is why I thought that requiring the actual config is the best idea.

pdgendt: IMO if you have used multiple paths to load the configuration, it could be the argument for the constructor. It would actually allow the `--print-path` to print all used paths. Because now this is a hidden feature.

thorsten-klein: Good point. I like the idea to use the path of the config as argument for the constructor and within the constructor all according dropin config files are searched and their paths are stored in ` self.dropins`  .

Before I make those adaptions, maybe we can clarify first, if you will accept this feature. See my other comment in the conversation with some explanation what we want to do.

Let's use this conversation for clarification of open questions in order to not spam the issue conversation.

src/west/configuration.py
marc-hb: > > For me the config.d is only an extension. Therefore the actual config should always exist.

I don't remember this with any other `.d/` files. AFAIK it's always possible to use either approach or both. I don't know which one has usually precedence in other tools but for `west` I would give it to the older, file outside `.d/` because that's where `west config set ...` will keep writing to.

See other thread below.

src/west/app/config.py
marc-hb, line 1: You need to update CONFIG_DESCRIPTION --help

src/west/app/config.py
```diff
@@ -100,6 +100,8 @@ def do_add_parser(self, parser_adder):
             "action to perform (give at most one)"
         ).add_mutually_exclusive_group()
 
+        group.add_argument('-p', '--print-path', action='store_true',
+                           help='print path from the active west config file')
```
marc-hb: It does not look like "active" has ever been defined before. It would be difficult to define because most commands consider multiple files. I think you can avoid a new term and definition like this:

```suggestion
                           help='print file path of the config file where a new value would be added. Combine this with --system, --global,...)
```

src/west/app/config.py
```diff
@@ -153,6 +157,11 @@ def do_run(self, args, user_args):
         else:
             self.write(args)
 
+    def print_path(self, args):
+        config_path = self.config.get_path(args.configfile or LOCAL)
+        if config_path:
```
marc-hb, line 226: How could this be None or empty? If it were, something went seriously wrong and it would be better to crash.

src/west/configuration.py
```diff
@@ -68,15 +68,35 @@ def parse_key(dotted_name: str):
...
+        config = _InternalCF(path) if path and path.exists() else None
+        if not config:
+            return None
+
+        # check if dropin configs exist (in according .d directory)
+        dropins_dir = Path(f'{path}.d')
+        if dropins_dir.exists():
+            # update config with values from dropin config files, whereby the
+            # dropin configs are applied in alphabetical order
+            for conf in sorted(dropins_dir.iterdir()):
```
marc-hb: Please only read the files ending with `.conf`, otherwise you will read `.conf~` backup files and maybe other files (used to be a common issue, fixed like this).

src/west/configuration.py
marc-hb: > There is a check at later point that the number of read configs is exactly 1. I can remove this check and rework the for loop and instead pass the list of multiple configs.

Yes please. Also, I think it's better to just add all the code to the `__init__` constructor, not to some convenience wrapper.

> Nevertheless: what should be the path attribute of the class in case of multiple configs? This is why I thought that requiring the actual config is the best idea.

`path` should be the non-`.d/` file. If it does not exist, then maybe it should be None? Not sure about that. Just like before, writes to a config file that does not exist yet should create that file, and deletions or appending in a non-existing main file should fail. If there was no test coverage for these then it should be added because the new "drop-ins" are going to make the corresponding logic a bit more complicated.

Writing to `.d/` files should not be possible through the CLI, way too complicated. This must be documented.

pdgendt: > Before I make those adaptions, maybe we can clarify first, if you will accept this feature. See my other comment in the conversation with some explanation what we want to do.

For me, it mostly depends on the complexity of the feature and how intuitive it's for users. For example,

> Writing to .d/ files should not be possible through the CLI, way too complicated. This must be documented.

What about deletions? An option can come from any of those files, but they can't be deleted if not coming from the main config file? This feels somewhat counter-intuitive.

I'd like some other opinions from @carlescufi or @mbolivar too.

> > There is a check at later point that the number of read configs is exactly 1. I can remove this check and rework the for loop and instead pass the list of multiple configs.
> 
> Yes please. Also, I think it's better to just add all the code to the `__init__` constructor, not to some convenience wrapper.

I agree here.

pdgendt:
> An alternative would be to support a dropin directory like `.west/config.d` directory. This would allow us to ship shared configs and place them before the build in `.west/config.d` (or symlink the files/folder) Users could still override values in `.west/config`, but in most cases that file would remain empty because all necessary options are maintained in the generic `.west/config.d`..
> 
> The important thing is: the generic configs may change anytime. In this case the new config values should be automatically applied if the users are using the default configs. If they have set some different value (on their own risk), their value shall be used. This scenario is perfectly fitting to the drop-in mechanism.
> 
> I hope you understand roughly what we do.

Yes, I understand you want something like:

```
$ tree -a
.
├── .west
│   ├── config
│   └── config.d
│       └── 01-generic.conf -> ../../app/shared/01-generic.conf
├── app
│   └── shared
│       └── 01-generic.conf
└── zephyr
```

And get configuration updates when pulling the latest version of `app`. And I agree this can be powerful feature, but the edge cases need to be well defined.

mbolivar:
Few initial thoughts:

1. Security

> And I agree this can be powerful feature, but the edge cases need to be well defined.

Let's consider the security impact as well please.

I think @pdgendt 's example is a place where I can see how this is different than having an existing global or system config, but it makes me worried. We can assume that a ".d" directory in a home directory and anything in the system locations are controlled well by the user, but the workspace itself is an adversarial environment in my opinion -- that's why we can turn off extensions and why we do not parse extension command files unless the user specifically asks for one.

2. On Linux, I see no mention of XDG_CONFIG_HOME and I think we need to keep up with this standard for where these go.

marc-hb:
> but the workspace itself is an adversarial environment in my opinion

Is the 9f4dcf55c083df5f7550c check not enough? If not, can you elaborate on the "threat model" you have in mind?

> On Linux, I see no mention of XDG_CONFIG_HOME and I think we need to keep up with this standard for where these go.

XDG_CONFIG_HOME sounds like a different feature than this. Are you saying it should come first?

pdgendt:
> Before I make those adaptions, maybe we can clarify first, if you will accept this feature. See my other comment in the conversation with some explanation what we want to do.

@thorsten-klein The more I think about this, the more I dislike how this creates hidden paths for configuration bits. I was more in favor if updating/importing the configuration as was proposed with #847, which IMO has a clear intent and is a one-off action.
It does require more effort for users to keep up-to-date when updates are done to some shared files, but at least the user has to make a conscious decision.

One last alternative we did not bring up yet: we take inspiration from `git`, and `git` too has a way to extend configuration files, by adding [include sections](https://git-scm.com/docs/git-config#_includes). Maybe we can use the same principle here? I am just spitballing here, but it would support your use-case, while maybe keeping things "simpler"?

thorsten-klein:
> It does require more effort for users to keep up-to-date when updates are done to some shared files, but at least the user has to make a conscious decision.

Exactly. That’s the main drawback with the approach from #847: we cannot easily tell whether a config was intentionally changed by the user, or if it’s simply outdated (or unintentionally modified). Therefore I prefer the dropin solution nowadays.

> git too has a way to extend configuration files, by adding [include sections](https://git-scm.com/docs/git-config#_includes)

I’ve considered this as well. But does it really simplify things? Users could even create nested includes, which complicates matters further.  Also, how should include order be handled if an `include` section appears in the middle of a config file? Do later values override those from the included file? This will become very tricky within the implementation I guess.

> this creates hidden paths for configuration bits

I share the same concern. This approach offers users a lot of flexibility, at the cost of potential intransparency.
But since the concept is quite common (specially in Unix world), I think it’s a reasonable compromise and acceptable for west as well.

src/west/app/config.py
thorsten-klein: It can be `None` in case of local configs if there is no `.west` directory, so if west is run outside any workspace. 
This makes sense for this command, e.g. if the user just opens a console in his home directory and runs `west config --print-paths` to see which configs are expected.

Would you expect that it fails in this case, or shall it exit 0 and print no path for the local config (as it is now)?

mbolivar:
> > but the workspace itself is an adversarial environment in my opinion
> 
> Is the [9f4dcf5](https://github.com/zephyrproject-rtos/west/commit/9f4dcf55c083df5f7550c6ddde2910743d1c91e8) check not enough? If not, can you elaborate on the "threat model" you have in mind?

@marc-hb it's not enough.

The threat model is basically the same as extension commands: threat actor wants to insert arbitrary code execution into something that happens when the user runs a west command in a way that is "difficult" to observe.

If you look at the way extension commands work, we only read YAML files from each west project to determine what extension command implementations it provides. We do not parse any Python when loading the manifest. If the user runs an extension command, then we parse just the python required to run it, and run it since we assume the user is opting in to running the command.

In https://github.com/zephyrproject-rtos/west/pull/849#issuecomment-3350222708, @pdgendt shows a west project with a configuration file that I understand is meant to be added to the west.configuration.Configuration object's state when an arbitrary west command, extension or otherwise, is run.

Just to illustrate where this could be problematic, consider if the adversary sets `build.cmake-args` to  "-C pwned.cmake". Or sets `alias.innocent-looking-command` to `forall -c "rm -rf ~"`. Or some other thing we add later and forget to consider in this context if you want a built-in thing and don't feel convinced since the adversary might toss other arbitrary cmake into the zephyr repo (though that is harder to hide IMO).

In summary I think it is a non-starter to allow the workspace itself to set configuration options.

marc-hb:
@mbolivar I have no doubt `workspace/.west/config.d/` would be a great place to do evil things. But my main question was rather: _how does someone evil get there_? Also, how would these new files more vulnerable than the already existing `workspace/.west/config` file, what's the difference? You discussed extensions and YAML files but all these generally come from the _outside world through git_. AFAIK that should not be the case for west config files? I mean not unless the user explicitly copies or imports some specific config files from git into his workspace, on a narrow per user, per workspace, per project and per file basis (which happens to be exactly the feature desired by @thorsten-klein[*]). 


The vast majority of users have their workspace in their home directory. If something evil can write to `workspace/.west/config.d/`, then for the vast majority of users that means it can write anywhere in the home directory and then the user has a much bigger problem on their hand than west config files...?


[*] EDIT: users can _already_ do this sort of "dangerous" imports with a single command like `crudini --merge workspace/.west/config < evil_from_git.ini` or the equivalent PowerShell command. @thorsten-klein only wants to avoid the external dependency.

mbolivar:
>I have no doubt workspace/.west/config.d/

I'm probably missing something -- when I see this:

```
$ tree -a
.
├── .west
│   ├── config
│   └── config.d
│       └── 01-generic.conf -> ../../app/shared/01-generic.conf
├── app
│   └── shared
│       └── 01-generic.conf
└── zephyr
```

I'm guessing you're saying that none of this is intended to create the symlink in config.d to app/shared, and the user has to do it by hand?

marc-hb:
> I'm guessing you're saying that none of this is intended to create the symlink in config.d to app/shared, and the user has to do it by hand?

Right, I would go even further and DISCOURAGE users to manually create such symlinks because they create a permanent bridge between a sensitive area (west configs) and updates from a less trusted one (a git repo). Instead, I would have in documentation something like: "after inspecting it very carefully, you make a (static!) copy of some .ini file into `.west/config.d/`"

Just for the record I think the symlink example came from @pdgendt and not for @thorsten-klein . I think it was a nice explanation shortcut to explain the rationale but with hindsight it may have backfired a little :-) EDIT: I just found @thorsten-klein did mention the symlink idea first. Does not matter.

Of course, nothing would stop "unsafe" users to create symlinks like this anyway but there's generally nothing that can be done for people driving over the speed limit without their seat belt. EDIT: and maybe it's OK if they're doing this only in a private and tightly controlled circuit. Does not matter as long as it's not officially recommended.

src/west/configuration.py
marc-hb: Re-opening this comment because maybe we want both `.conf` and `.ini`?

thorsten-klein:
I’m not entirely sure where the security concerns come from at this point. The existing `west config` commands already allow equivalent behavior. From a user’s perspective, it makes little difference whether a script runs command `west config ...` or `ln -s ...`. And if users invoke commands directly, they should understand in both cases what they are actually doing.

Would it perhaps be better to make the drop-in feature explicitly opt-in?
For example, via a setting like `config.dropins = true`, or even by supporting a string value such as `config.dropins = ../generic-config.d:my-config.d`? This would at least be safer if you are afraid that a repository ships an arbitrary `.west/config.d` folder (whereby I am unsure if `west init` will work if `.west` already exists).

UPDATE:
When I think about it, I like `config.dropins = ../generic-config.d:my-config.d` 👍🏻 
It could also be supported per-type (whereby relative paths are always relative to local workspace):
```
[dropins]
local = ../generic-local-config.d:my-config.d
global = ../generic-global-config.d:another-global-config.d
system = ../generic-system-config.d
``` 

But this adds even more complexity. So maybe we should go with `config.dropins = true`.

marc-hb:
> From a user’s perspective, it makes little difference whether a script runs command west config ... or ln -s ...

There are (at least?) two important differences that I think concern @mbolivar (and myself now)

1. When you run `west config...`, you **see** what you change.  You don't see the change when importing an entire file. In theory you should look at the file content too; but in practice... I'm aware this is in a nutshell the exact feature you want and I'm personally OK with it. But it makes the other difference even more critical:

2. A symbolic link is _dynamic_: the content changes every time you run `west update`, when you totally forgot you created this link! In other words, you surrender control of your home directory to an outside git repo. Granted, there is plenty of other attack surface in a typical build system but `west` is a simpler and hopefully safer tool that runs long before any build command has been issued, so the bar should be much higher. I think the bar for `west` alone should be as high as for `git`.

Thinking of it, this "dynamic" nature is also bad for reproducibility and other build properties. Imagine `git bisect` silently changing the build flags! Or much worse. Again: nothing can stop you from doing that but it should IMHO be discouraged.


You can admittedly symlink `workspace/.west/config` already but that's pretty inconvenient because it forces you to abandon the feature for yourself, so it does not come "naturally" like it would with this PR. You will almost never see that sort of advice.


> So maybe we should go with config.dropins = true

Works for me.

thorsten-klein:
This feature now requires explicit user activation. To enable it, users must set `config.dropins = true` within the config file.
Crucially, this is done on a *per config file level*. 

For example, running `west config --local config.dropins true` will activate drop-in configurations only for the local config (only this west workspace). 
Running `west config --global config.dropins true` will activate only global drop-in configurations (and no local or system ones). 

This design ensures on the hand hand complete transparency: users can clearly understand that only the `.d` folder next to their config file is enabled. 
On the other hand it also ensures security, since a globally enabled option does not automatically lead to the automatic import of local dropin configs (which are probably delivered as part of an evil west workspace).

I hope this feature is now ready 👍🏻

src/west/configuration.py
thorsten-klein: > What about deletions? An option can come from any of those files, but they can't be deleted

Yes. In this case the user must set the value to the default value explicitly. I think this is very transparent as the user can track in the config files the value of each option.  If we allow deleting an option in the config, it might be confusing for users where the different value comes from.

carlescufi:
@thorsten-klein could you please provide an example of tools that already implement something similar to this? I am curious to see how others have done it and whether they've considered the security implications. Thanks!

thorsten-klein:
> please provide an example of tools that already implement something similar

I’ve found a few examples of tools that support drop-in configurations:
- **systemd:** [dropin.c](https://github.com/systemd/systemd/blob/e06b9a1d5265d30061a431cadbdcfce9292c2e15/src/shared/dropin.c#L240)
- **httpd:** [config.c](https://github.com/apache/httpd/blob/trunk/server/config.c#L1928C27-L1928C37)
- **cron**
- **polkit**
- **ssh**
- ...and many others (an AI can list quite a few more Linux tools using similar mechanisms).

However, the main difference between `west` and those tools is that `west` supports defining aliases, which can execute arbitrary commands directly (whereby other tools just allow simple config values) — for example:
```bash
[alias]
update= forall -c "rm -rf ~"
```

So IMO, the real security concern is not the *drop-in config* mechanism itself, but rather the **alias feature**.

Nevertheless, I believe that `alias`es in `west` are a useful and meaningful feature, so I’d like them to be supported in the drop-in configs as well.  
I’m therefore happy with the compromise that drop-in configs must be explicitly enabled in the user’s configuration (at the user’s own risk).

mbolivar:
> So IMO, the real security concern is not the _drop-in config_ mechanism itself, but rather the **alias feature**.

.... **today, that we know of, in built-in west**

Like I tried to explain, that was just:

1. an example off the top of my head
2. for built-in commands (west is extensible!)
3. for today's codebase, to which we want to add features and for which we would have to consider and safely handle the interaction with this feature forever more

thorsten-klein:
@mbolivar / @carlescufi How do you want to move forward with this PR? Should I do anything else? 
I’d be glad to help finalize it for merging.

src/west/app/config.py
marc-hb: If the user requests the local path while outside any workspace, I would just let the tool crash. This is an obvious misuse of the tool, no need to test for that in the code.

I found a `raise MalformedConfig('local configuration file not found')` somewhere else, that's another option.

In any case I would never exit 0, the tool should not succeed.

PS: please do not reply with a question and then hide the thread as "resolved" :-D

src/west/configuration.py
```diff
@@ -178,6 +178,16 @@ def __init__(self, topdir: PathType | None = None):
         self._global = _InternalCF.from_path(self._global_path)
         self._local = _InternalCF.from_path(self._local_path)
 
+    def get_path(self, configfile: ConfigFile = ConfigFile.LOCAL):
+        if configfile == ConfigFile.ALL:
+            raise RuntimeError(f'{configfile} not allowed for get_path')
```
```suggestion
            raise RuntimeError(f'{configfile} not allowed for --print-path')
```

src/west/configuration.py
```diff
@@ -68,63 +68,86 @@ def parse_key(dotted_name: str):
 
     @staticmethod
     def from_path(path: Path | None) -> '_InternalCF | None':
-        return _InternalCF(path) if path and path.exists() else None
+        cf = _InternalCF(path) if path else None
+        if not cf:
+            return None
+        if not cf.path and not cf.dropin_paths:
```
marc-hb, line 80: Why bother creating an object when neither exists?

src/west/configuration.py
```diff
@@ -68,63 +68,86 @@ def parse_key(dotted_name: str):
...
+                # only consider .conf files
+                if conf.suffix.lower() in ['.conf', '.ini']:
+                    self.dropin_paths.append(self.dropin_dir / conf)
+        self._read()
+
+    def _read(self):
+        if self.path:
+            self.cp.read(self.path, encoding='utf-8')
+        if self.dropin_paths:
+            self.dropin_cp.read(self.dropin_paths, encoding='utf-8')
```
marc-hb: Please explain in the `class _InternalCF:` "pydoc" why you need to keep these in separate `configparser` objects. Because appending is tricky maybe?

At first sight, it looks like merging both here (with the appropriate precedence) would save a ton of code below.

Exerting precedence here would be a good place to let the user know with some INFO log statement.

Either way, the `class _InternalCF` becomes a bit more complicated and needs some introduction explaining its data structures.

src/west/configuration.py
```diff
@@ -68,63 +68,86 @@ def parse_key(dotted_name: str):
...
+
+    def _write(self):
+        if not self.path:
+            raise WestNotFound('No config file exists that can be written')
```
marc-hb: If not too difficult, it would be nice to print the missing path in the error message.
Also, if dropins are enabled then it would be nice to print a reminder that writing and appending ignores them.

src/west/configuration.py
```diff
@@ -68,63 +68,86 @@ def parse_key(dotted_name: str):
...
-
-        with open(self.path, 'w', encoding='utf-8') as f:
-            self.cp.write(f)
+        self._write()
 
     def delete(self, option: str):
         section, key = _InternalCF.parse_key(option)
-
-        if section not in self.cp:
+        if (section not in self.cp) or (key not in self.cp[section]):
```
marc-hb: Again, maybe a reminder that dropins are ignored.

src/west/configuration.py
```diff
@@ -362,13 +385,14 @@ def _copy_to_configparser(self, cp: configparser.ConfigParser) -> None:
...
-                for key, value in contents.items():
-                    cp[section][key] = value
+            for cp in [cf.dropin_cp, cf.cp]:
+                for section, contents in cp.items():
+                    if section == 'DEFAULT':
+                        continue
+                    if section not in cp:
+                        cp.add_section(section)
+                    for key, value in contents.items():
+                        cp[section][key] = value
```
marc-hb, line 419: Would you mind expanding a bit the pydoc for this function? It's pretty cryptic to me, sorry. Or just add some relevant references.

src/west/app/config.py
```diff
@@ -73,6 +73,14 @@
...
 To delete <name> everywhere it's set, including the system file:
     west config -D <name>
+
+Additionally to the config file a dropin config directory is considered.
+The directory is named as the according config file, but with a '.d' suffix.
+As a result there are three levels for dropin config directories (local, global
+and system), whereby all '.conf' files from each dropin directory are loaded in
+alphabetical order.
+For example:
+    .west/config.d/basics.conf
```
marc-hb, line 93: I think this falls a bit short and misses at least the fact that writing and appending ignores dropins.

src/west/configuration.py
thorsten-klein: Good point. It can be changed to
```
if not path:
    return None
cf = _InternalCF(path)
```

src/west/configuration.py
thorsten-klein: I think this can be refactored/simplified again since with current concept a config file must always exist (where the dropin config is actually enabled).

You are right, that a reminder would be nice. You prefer it only in case of writing to a config via command line, or always on reading (e.g. on stderr)?

src/west/configuration.py
thorsten-klein: Honestly I am not sure if this function is still neccessary. I have removed it completely (and its usage in `main.py`), and the tests still pass.

pdgendt:
I'm going to be honest here, that I'm not keen on adding these configuration dropins. The configuration mechanism for west developers (and a bit for west users) is a lot more complicated. Future changes need be aware to not break any of this.

> > Can you provide some insight into where you "really" need this kind of flexibility?
> 
> In our company, only a small group of people maintains the build system and tooling and configs (e.g., west). Their goal is to set up the development environment correctly and shield everyone else from pitfalls. Similarly to this, also customer repositories may define configurations (on top of the generic configs) that should be shared across their product developers to ensure consistency. This leads to the need for maintaining **(generic) configs** and a mechanism to overlay configs. In the end, maybe something like a "repository config level" is missing for us.

Mind you there's already a mechanism for extending functionality: [extension commands](https://docs.zephyrproject.org/latest/develop/west/extensions.html). You can simply provide west extension commands that are "simple" wrapper commands.
Correct me if I'm wrong, but I think you've pointed out sharing _alias_ commands, which can be done just as easy with _extension_ commands.

thorsten-klein:
> I'm going to be honest here, that I'm not keen on adding these configuration drop-ins.

Thanks for the honest feedback! 👍🏻  

> Correct me if I'm wrong, but I think you've pointed out sharing alias commands:

I assume we’re not the first ones who’d like to share a company-wide **west** configuration within a repository, are we?

That was just one example that came to mind. For instance, we’d also want to ensure that `auto-cache` is enabled, or define a default `build.board`. 

On the other hand, I also agree that this relatively small use case might not justify adding a large feature like configuration drop-ins. So I am also open for any other solutions for our use case.

---

So, to summarize our use case again:  

At our company, we’d like to maintain a **company-wide west configuration** with default values.  
We want to make sure some config values are always set — either explicitly by the user in their local config, or by falling back to the defaults from our company-wide config.  
Of course, users must always be able to override any company-wide values.

Without a “drop-in configs” feature, here’s how we tried to approach it:

**Approach 1:**  
We could check in our `.west/config` file into the repository to share it across the team.  
The downside is that any local change by a user would appear as a Git diff.  
Additionally, nested `.west` folders don’t work well (for example, if this repository is later integrated into another app).
Result: This is not a good idea.

**Approach 2:**  
We have a bootstrapping script that runs `west init`, `west update`.  This script could also call `west config section.key value` to ensure the expected values are set. However, this command overwrites existing values — no matter if the user has intentionally changed them, or if they simply have outdated values (previous defaults).  
What would be your recommendation for handling this?
Result: This is not a good idea.

---

Do you see the challenge we’re facing? What would your suggested (existing) solution?

marc-hb:
> I’ve found a few examples of tools that support drop-in configurations:

As mentioned before in #847, `sudo find /etc /usr -name '*.d' -type d` shows more than 100 examples on one of my linux systems. This "divide and conquer" configuration approach has quickly become universal; @thorsten-klein's issue is really not special. Most Linux kernel builds are configured with Kconfig "fragments"; the kernel has  `scripts/kconfig/merge_config.sh` and other scripts for that. From https://docs.zephyrproject.org/latest/build/snippets/design.html
> - able to combine multiple types of configuration: snippets make it possible to store multiple different types of build system settings in one place, and apply them all together

Dropins are part of the bigger "configuration as code" approach; they enable it.

While not exactly it, the flexibility to move files in and out of `.d` directories would be a useful alternative to solve the #429 use case.

src/west/configuration.py
```diff
@@ -68,63 +68,95 @@ def parse_key(dotted_name: str):
...
+        if self.cp.getboolean('config', 'dropins', fallback=False):
+            # dropin dir is the config path with .d suffix
+            dropin_dir = Path(f'{path}.d')
+            self.dropin_dir = dropin_dir if dropin_dir.exists() else None
+            if self.dropin_dir:
+                # dropin configs are applied in alphabetical order
+                for conf in sorted(self.dropin_dir.iterdir()):
+                    # only consider .conf files
+                    if conf.suffix in ['.conf', '.ini']:
+                        self.dropin_paths.append(self.dropin_dir / conf)
```
pdgendt, line 104: Why also have `.ini` files included? This is not how it's documented above

```md
All files inside a drop-in directory must use `.conf` extension and are
loaded in **alphabetical order**.
```

src/west/configuration.py
```diff
@@ -68,63 +68,95 @@ def parse_key(dotted_name: str):
...
-        try:
-            return getter(section, key)
-        except (configparser.NoOptionError, configparser.NoSectionError) as err:
-            raise KeyError(option) from err
+        if section in self.cp and key in self.cp[section]:
+            getter = config_getter
+        elif section in self.dropin_cp and key in self.dropin_cp[section]:
+            getter = dropin_getter
+        else:
+            raise KeyError(option)
```
pdgendt, line 144: There is a subtle difference here, but one nonetheless. Before we were able to determine a difference between a missing key or an entire missing section, we no longer have that difference.

src/west/configuration.py
pdgendt: The comment at the start of the function would indicate that west extension commands could rely on it?

```python
        # Internal API for main to use to maintain backwards
        # compatibility for existing extensions using the legacy
        # function-and-global-state APIs.
```

thorsten-klein: Okay. Is it possible that there is no test existing for this? I am not sure how I can guarantee that I do not break something here, since I do not know the previous behavior. 
So does it make sense that I adapt this function? Will such a function be kept forever or will it be removed at some time?

marc-hb: The answers to this sort of questions usually lie in git history. `git blame` works but I tend to prefer `tig blame`. GitHub's blame interface is surprisingly decent and it has direct links to the corresponding pull requests. Of course, this assumes the git history is usable with no too much whitespace pollution :-)

marc-hb: Maybe this? Just spotted by chance:
https://docs.zephyrproject.org/4.2.0/develop/west/west-apis.html#configuration-api

> ... recommended API to use **since west v0.13**.

marc-hb:
> The threat model is basically the same as extension commands: threat actor wants to insert arbitrary code execution into something that happens when the user runs a west command in a way that is "difficult" to observe.

So now I just found that WEST_CONFIG_* accept relative paths #871 ...

PR #854, opened 2025-10-05
support 'west init --topdir'

### LINKED ISSUE #774: `west init` gives wrong config `path` for a manifest file in nested directories

thorsten-klein:
Fixes #774  (Successor of #775)

This PR introduces a new `--topdir` option for `west init`, allowing users to explicitly specify the west workspace directory (topdir) during initialization.  

This is particularly useful when using `--local`, as the current behavior always sets the parent of the given directory as the topdir. In addition, west is currently unable to determine both `manifest.path` and `manifest.file` from a single `directory` argument as the path can be interpreted in multiple ways (e.g. it can be split at various places).

With the new `--topdir` argument, users can now directly control where the west workspace is created.  

Parameterized tests for `west init --topdir` are added as well.

Moreover, the help texts and descriptions are (hopefully) simplified.

EDIT: superseded by:
- #953

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -24,10 +24,8 @@ from urllib.parse import urlparse
 from west import util
 from west.commands import CommandError, Verbosity, WestCommand
 from west.configuration import Configuration
-from west.manifest import MANIFEST_REV_BRANCH as MANIFEST_REV
-from west.manifest import QUAL_MANIFEST_REV_BRANCH as QUAL_MANIFEST_REV
-from west.manifest import QUAL_REFS_WEST as QUAL_REFS
 from west.manifest import (
+    _WEST_YML,
     ImportFlag,
     Manifest,
     ManifestImportFailed,
@@ -35,6 +33,9 @@ from west.manifest import (
     Submodule,
     _manifest_content_at,
 )
+from west.manifest import MANIFEST_REV_BRANCH as MANIFEST_REV
+from west.manifest import QUAL_MANIFEST_REV_BRANCH as QUAL_MANIFEST_REV
+from west.manifest import QUAL_REFS_WEST as QUAL_REFS
 from west.manifest import is_group as is_project_group
 from west.util import expand_path
 
@@ -236,19 +237,58 @@ class Init(_ProjectCommand):
             'init',
             'create a west workspace',
             f'''\
-Creates a west workspace.
+Initialize a west workspace (topdir) from a west manifest (`west.yml`) by
+creating a `.west` directory in the topdir and a local configuration file
+`.west/config`.
+The West manifest can come from either a git repository (that will be cloned
+during workspace initialization) or from an already existing local directory.
 
-With -l, creates a workspace around an existing local repository;
-without -l, creates a workspace by cloning a manifest repository
-by URL.
+Arguments
+---------
+--mf / --manifest-file
+    The relative path to the manifest file within the manifest repository
+    or directory. Config option `manifest.file` will be set to this value.
+    Defaults to '{_WEST_YML}' if not provided.
 
-With -m, clones the repository at that URL and uses it as the
-manifest repository. If --mr is not given, the remote's default
-branch will be used, if it exists.
+-t / --topdir
+    Specifies the directory where west should create the workspace.
+    The `.west` folder will be created inside this directory.
+    If it is an already existing directory, it must not contain a .west folder.
 
-With neither, -m {MANIFEST_URL_DEFAULT} is assumed.
 
-Warning: 'west init' renames and/or deletes temporary files inside the
+1. Using a Manifest Repository (default)
+----------------------------------------
+West clones the given repository (provided via `-m / --manifest-url`).
+Note, that the repository must contain a west manifest.
+
+If no `-m / --manifest-url` is provided, west uses Zephyr URL by default:
+  {MANIFEST_URL_DEFAULT}.
+
+The topdir (where `.west` is created) is determined as follows:
+- argument `-t / --topdir` if provided
+- otherwise: the positional argument `directory` if provided
+- otherwise: the current working directory
+
+If both `-t / --topdir` and `directory` are provided, `-t / --topdir`
+specifies the workspace directory, while positional argument `directory`
+specifies the exact path where the manifest repository is cloned
+(it must be inside the workspace).
+With `--mr`, the revision (branch, tag, or sha) of the repository can be
+specified that will be used. It defaults to the repository's default branch.
+
+2. Using a Local Manifest
+-------------------------
+If `-l / --local` is given, west initializes a workspace from an already
+existing `west.yml`. Its location can be specified in the `manifest_directory`
+positional argument, which defaults to current working directory.
+
+The topdir (where `.west` is created) is determined as follows:
+- argument `-t / --topdir` if provided
+- otherwise: `manifest_directory`'s parent
+
+Known Issues
+------------
+'west init' renames and/or deletes temporary files inside the
 workspace being created. This fails on some filesystems when some
 development tool or any other program is trying to read/index these
 temporary files at the same time. For instance, it is required to stop
@@ -273,9 +313,11 @@ below.
         parser = self._parser(
             parser_adder,
             usage='''
+  init with a repository:
+  %(prog)s [-m URL] [--mr REVISION] [--mf FILE] [-o=GIT_CLONE_OPTION] [-t WORKSPACE_DIR] [directory]
 
-  %(prog)s [-m URL] [--mr REVISION] [--mf FILE] [-o=GIT_CLONE_OPTION] [directory]
-  %(prog)s -l [--mf FILE] directory
+  init with an existing local manifest:
+  %(prog)s -l [-t WORKSPACE_DIR] [--mf FILE] [manifest_directory]
 ''',
         )
 
@@ -284,6 +326,7 @@ below.
         parser.add_argument(
             '-m',
             '--manifest-url',
+            metavar='URL',
             help='''manifest repository URL to clone;
                     cannot be combined with -l''',
         )
@@ -292,6 +335,7 @@ below.
             '--clone-opt',
             action='append',
             default=[],
+            metavar='GIT_CLONE_OPTION',
             help='''additional option to pass to 'git clone'
                     (e.g. '-o=--depth=1'); may be given more than once;
                     cannot be combined with -l''',
@@ -300,21 +344,36 @@ below.
             '--mr',
             '--manifest-rev',
             dest='manifest_rev',
+            metavar='REVISION',
             help='''manifest repository branch or tag name
                     to check out first; cannot be combined with -l''',
-        )
-        parser.add_argument(
-            '--mf', '--manifest-file', dest='manifest_file', help='manifest file name to use'
         )
         parser.add_argument(
             '-l',
             '--local',
             action='store_true',
-            help='''use "directory" as an existing local
-                    manifest repository instead of cloning one from
-                    MANIFEST_URL; .west is created next to "directory"
-                    in this case, and manifest.path points at
-                    "directory"''',
+            help='''initialize workspace from an already existing local
+                    manifest instead of cloning a manifest;
+                    cannot be combined with -m''',
+        )
+        parser.add_argument(
+            '--mf',
+            '--manifest-file',
+            dest='manifest_file',
+            metavar='FILE',
+            help=f'''manifest file to use. It is the relative
+                     path to the manifest file within the manifest_directory.
+                     Defaults to {_WEST_YML}.''',
+        )
+        parser.add_argument(
+            '-t',
+            '--topdir',
+            dest='topdir',
+            metavar='WORKSPACE_DIR',
+            help='''the workspace directory where .west directory
+                    will be created (WORKSPACE_DIR/.west);
+                    the workspace directory may already exist
+                    and WORKSPACE_DIR/.west must not exist''',
         )
         parser.add_argument(
             '--rename-delay',
@@ -330,9 +389,15 @@ below.
             'directory',
             nargs='?',
             default=None,
-            help='''with -l, the path to the local manifest repository;
-            without it, the directory to create the workspace in (defaulting
-            to the current working directory in this case)''',
+            metavar='directory',
+            help='''With --local: the path to a local directory which contains
+                    a west manifest (west.yml);
+                    Otherwise, if no --topdir is provided:
+                    the location of the west workspace where .west directory
+                    will be created (<directory>/.west)
+                    Otherwise:
+                    the location where west will clone the manifest repository
+                    ''',
         )
 
         return parser
@@ -385,7 +450,7 @@ below.
         # Path('..').
         #
         # https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.parent
-        manifest_dir = Path(args.directory or os.getcwd()).resolve()
+        manifest_dir = Path(args.directory or '.').resolve()
 
         try:
             already = util.west_topdir(manifest_dir, fall_back=False)
@@ -393,16 +458,26 @@ below.
         except util.WestNotFound:
             pass
 
-        manifest_filename = args.manifest_file or 'west.yml'
+        manifest_filename = args.manifest_file or _WEST_YML
         manifest_file = manifest_dir / manifest_filename
-        topdir = manifest_dir.parent
-        rel_manifest = manifest_dir.name
-        west_dir = topdir / WEST_DIR
-
         if not manifest_file.is_file():
             self.die(f'can\'t init: no {manifest_filename} found in {manifest_dir}')
 
-        self.banner('Initializing from existing manifest repository', rel_manifest)
+        topdir = Path(args.topdir or manifest_dir.parent).resolve()
+
+        if not manifest_file.is_relative_to(topdir):
+            self.die(f'{manifest_file} must be relative to west topdir {topdir}')
+
+        try:
+            already = util.west_topdir(manifest_dir, fall_back=False)
+            self.die_already(already)
+        except util.WestNotFound:
+            pass
+
+        rel_manifest = manifest_dir.relative_to(topdir)
+        west_dir = topdir / WEST_DIR
+
+        self.banner('Initializing with existing manifest', rel_manifest)
         self.small_banner(f'Creating {west_dir} and local configuration file')
         self.create(west_dir, exist_ok=False)
         os.chdir(topdir)
@@ -413,8 +488,25 @@ below.
         return topdir
 
     def bootstrap(self, args) -> Path:
-        topdir = Path(abspath(args.directory or os.getcwd()))
-        self.banner('Initializing in', topdir)
+        manifest_subdir = Path()
+        if args.topdir:
+            topdir = Path(args.topdir).resolve()
+            if args.directory:
+                directory = Path(args.directory)
+                if not directory.is_absolute():
+                    directory = directory.resolve()
+                if not directory.is_relative_to(topdir):
+                    self.die(
+                        f"directory '{args.directory}' must be inside west topdir '{args.topdir}'"
+                    )
+                manifest_subdir = directory.relative_to(topdir)
+        elif args.directory:
+            # This case will be deprecated later in favor of --topdir.
+            topdir = Path(args.directory)
+        else:
+            topdir = Path()
+        topdir = topdir.resolve()
+        self.banner(f'Initializing in {topdir}')
 
         manifest_url = args.manifest_url or MANIFEST_URL_DEFAULT
         if args.manifest_rev:
@@ -469,7 +561,7 @@ below.
             raise
 
         # Verify the manifest file exists.
-        temp_manifest_filename = args.manifest_file or 'west.yml'
+        temp_manifest_filename = args.manifest_file or _WEST_YML
         temp_manifest = tempdir / temp_manifest_filename
         if not temp_manifest.is_file():
             self.die(
@@ -486,15 +578,17 @@ below.
         manifest = Manifest.from_data(
             temp_manifest.read_text(encoding=Manifest.encoding), import_flags=ImportFlag.IGNORE
         )
+
         if manifest.yaml_path:
-            manifest_path = manifest.yaml_path
+            yaml_path = manifest.yaml_path
         else:
             # We use PurePath() here in case manifest_url is a
             # windows-style path. That does the right thing in that
             # case, without affecting POSIX platforms, where PurePath
             # is PurePosixPath.
-            manifest_path = PurePath(urlparse(manifest_url).path).name
+            yaml_path = PurePath(urlparse(manifest_url).path).name
 
+        manifest_path = manifest_subdir if manifest_subdir != Path() else Path(yaml_path)
         manifest_abspath = topdir / manifest_path
 
         # We already looked for old .west/ pollution earlier but only at the highest
@@ -531,7 +625,7 @@ below.
             self.die(e)
         self.small_banner('setting manifest.path to', manifest_path)
         self.config = Configuration(topdir=topdir)
-        self.config.set('manifest.path', manifest_path)
+        self.config.set('manifest.path', str(manifest_path))
         self.config.set('manifest.file', temp_manifest_filename)
 
         return topdir
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -665,3 +665,15 @@ def check_proj_consistency(actual, expected):
     assert actual.clone_depth == expected.clone_depth
     assert actual.revision == expected.revision
     assert actual.west_commands == expected.west_commands
+
+
+def tree(path: Path, prefix="") -> str:
+    lines = []
+    entries = sorted(path.iterdir(), key=lambda p: (p.is_file(), p.name))
+    for i, entry in enumerate(entries):
+        connector = "└── " if i == len(entries) - 1 else "├── "
+        lines.append(prefix + connector + entry.name)
+        if entry.is_dir():
+            indent = "    " if i == len(entries) - 1 else "│   "
+            lines.extend(tree(entry, prefix + indent))
+    return "\n".join(lines)
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -25,6 +25,7 @@ from conftest import (
     create_repo,
     create_workspace,
     rev_parse,
+    tree,
     yaml_editor,
 )
 
@@ -2038,6 +2039,146 @@ def test_init_local_with_manifest_filename(repos_tmpdir):
     workspace.chdir()
     cmd('update')
 
+
+def _setup_test_init(manifest_repo, local_dir, directory, manifest_file):
+    flags = []
+    if local_dir:
+        # for the test: ensure that manifest_repo is present at local_dir
+        clone(str(manifest_repo), str(local_dir))
+        flags += ['-l']
+    else:
+        # clone from remote manifest
+        flags += ["-m", manifest_repo]
+
+    # extend west init flags according to specified test case
+    if manifest_file:
+        flags += ['--mf', manifest_file]
+    if directory:
+        flags += [directory]
+    return flags
+
+
+TEST_CASES_INIT = [
+    # (local_dir, directory, manifest_file)
+    ######################################
+    # REMOTE MANIFEST
+    ######################################
+    # init from remote repository (without any parameters)
+    (None, None, None),
+    # use deprecated [directory] to specify topdir
+    (None, Path('subdir'), None),
+    ######################################
+    # LOCAL MANIFEST
+    ######################################
+    # init workspace in current working directory (without --topdir)
+    (Path('workspace') / 'zephyr', Path('workspace') / 'zephyr', None),
+]
+
+
+@pytest.mark.parametrize("test_case", TEST_CASES_INIT)
+def test_init(repos_tmpdir, test_case):
+    local_dir, directory, manifest_file = test_case
+
+    repos_tmpdir.chdir()
+    manifest_repo = repos_tmpdir / 'repos' / 'zephyr'
+    flags = _setup_test_init(manifest_repo, local_dir, directory, manifest_file)
+
+    # initialize west workspace
+    cmd(['init'] + flags)
+
+    # cd into the workspace
+    if local_dir:
+        workspace = directory.parent
+    else:
+        workspace = directory or Path.cwd()
+    workspace_abs = workspace.absolute()
+    os.chdir(workspace)
+
+    # check if config option 'manifest.path' is correctly set
+    actual = cmd('config manifest.path')
+    if local_dir:
+        assert Path(actual.rstrip()) == directory.relative_to(workspace)
+    else:
+        if directory:
+            # zephyr is cloned into specified subdirectory in workspace
+            subdir = directory.relative_to(workspace)
+            assert Path(actual.rstrip()) == Path(subdir) / 'zephyr'
+        else:
+            # zephyr is cloned directly to workspace
+            assert Path(actual.rstrip()) == Path('zephyr')
+
+    # check if config option 'manifest.file' is correctly set
+    actual = cmd('config manifest.file')
+    expected = manifest_file or Path('west.yml')
+    assert Path(actual.rstrip()) == Path(expected)
+
+    # re-initialization of a workspace must always fail
+    _, stderr = cmd_raises(['init'] + flags, SystemExit)
+    assert "FATAL ERROR: already initialized" in stderr
+
+
+TEST_CASES_INIT_FAILS = [
+    # (local_dir, directory, manifest_file, expected_error, expected_error)
+    ######################################
+    # REMOTE MANIFEST
+    ######################################
+    # specify a non-existent manifest file
+    (
+        None,
+        None,
+        Path('non-existent.yml'),
+        SystemExit,
+        "FATAL ERROR: can't init: no non-existent.yml found",
+    ),
+    ######################################
+    # LOCAL MANIFEST
+    ######################################
+    # init workspace without a directory
+    (
+        Path('workspace') / 'zephyr',
+        None,
+        None,
+        SystemExit,
+        "FATAL ERROR: can't init: no west.yml found",
+    ),
+    # init workspace from non-existent manifest
+    (
+        Path('workspace') / 'zephyr',
+        Path('non-existent.yml'),
+        None,
+        SystemExit,
+        "FATAL ERROR: can't init: no west.yml found",
+    ),
+]
+
+
+@pytest.mark.parametrize("test_case", TEST_CASES_INIT_FAILS)
+def test_init_fails(repos_tmpdir, test_case):
+    local_dir, directory, manifest_file, expected_error, expected_stderr = test_case
+
+    manifest_repo = repos_tmpdir / 'repos' / 'zephyr'
+    repos_tmpdir.chdir()
+    flags = _setup_test_init(manifest_repo, local_dir, directory, manifest_file)
+    _, stderr = cmd_raises(['init'] + flags, expected_error)
+    assert expected_stderr in stderr
+
+
+def test_init_topdir_again(repos_tmpdir):
+    repos_tmpdir.chdir()
+    workspace = Path(repos_tmpdir) / 'ws'
+    zephyr = repos_tmpdir / 'repos' / 'zephyr'
+
+    # initial clone
+    cmd(['init', 'ws', '-m', zephyr])
+
+    # initialize west workspace again
+    _, stderr = cmd_raises(['init', 'ws', '-m', zephyr], SystemExit)
+    assert "FATAL ERROR: already initialized" in stderr
+
+    # initialize west workspace again
+    os.chdir(workspace)
+    _, stderr = cmd_raises(['init', '--local', 'zephyr'], SystemExit)
+    assert "FATAL ERROR: already initialized" in stderr
 
 def test_init_local_with_empty_path(repos_tmpdir):
     # Test "west init -l ." + "west update".
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -2040,7 +2040,7 @@ def test_init_local_with_manifest_filename(repos_tmpdir):
     cmd('update')
 
 
-def _setup_test_init(manifest_repo, local_dir, directory, manifest_file):
+def _setup_test_init(manifest_repo, local_dir, topdir, directory, manifest_file):
     flags = []
     if local_dir:
         # for the test: ensure that manifest_repo is present at local_dir
@@ -2051,6 +2051,8 @@ def _setup_test_init(manifest_repo, local_dir, directory, manifest_file):
         flags += ["-m", manifest_repo]
 
     # extend west init flags according to specified test case
+    if topdir:
+        flags += ['-t', topdir]
     if manifest_file:
         flags += ['--mf', manifest_file]
     if directory:
@@ -2059,38 +2061,75 @@ def _setup_test_init(manifest_repo, local_dir, directory, manifest_file):
 
 
 TEST_CASES_INIT = [
-    # (local_dir, directory, manifest_file)
+    # (local_dir, topdir, directory, manifest_file)
     ######################################
     # REMOTE MANIFEST
     ######################################
     # init from remote repository (without any parameters)
-    (None, None, None),
+    (None, None, None, None),
+    # specify topdir in current directory
+    (None, Path('.'), None, None),
+    # specify topdir in a subfolder
+    (None, Path('subdir'), None, None),
     # use deprecated [directory] to specify topdir
-    (None, Path('subdir'), None),
+    (None, None, Path('subdir'), None),
+    # specify topdir and [directory]: manifest cloned to the given directory
+    (None, Path('subdir'), Path('subdir') / 'extra' / 'level', None),
+    # specify topdir and [directory]: manifest cloned to the given directory
+    (None, Path('.'), Path('extra') / 'level', None),
     ######################################
     # LOCAL MANIFEST
     ######################################
     # init workspace in current working directory (without --topdir)
-    (Path('workspace') / 'zephyr', Path('workspace') / 'zephyr', None),
+    (Path('workspace') / 'zephyr', None, Path('workspace') / 'zephyr', None),
+    # init workspace in current working directory
+    (Path('workspace') / 'zephyr', Path('.'), Path('workspace') / 'zephyr', None),
+    # init workspace in a subfolder of current working directory
+    (Path('workspace') / 'zephyr', Path('workspace'), Path('workspace') / 'zephyr', None),
+    # init workspace in current working directory by providing a manifest file
+    (Path('workspace') / 'zephyr', Path('.'), Path('workspace'), Path('zephyr') / 'west.yml'),
+    # init workspace in itself by providing manifest file
+    (
+        Path('workspace') / 'zephyr',
+        Path('.'),
+        Path('.'),
+        Path('workspace') / 'zephyr' / 'west.yml',
+    ),
+    # init workspace in a subfolder by providing manifest file
+    (
+        Path('workspace') / 'subdir' / 'zephyr',
+        Path('workspace'),
+        Path('workspace') / 'subdir',
+        Path('zephyr') / 'west.yml',
+    ),
+    # init workspace in a subfolder by providing manifest file
+    (
+        Path('workspace') / 'subdir' / 'zephyr',
+        Path('workspace'),
+        Path('workspace'),
+        Path('subdir') / 'zephyr' / 'west.yml',
+    ),
 ]
 
 
 @pytest.mark.parametrize("test_case", TEST_CASES_INIT)
 def test_init(repos_tmpdir, test_case):
-    local_dir, directory, manifest_file = test_case
+    local_dir, topdir, directory, manifest_file = test_case
 
     repos_tmpdir.chdir()
     manifest_repo = repos_tmpdir / 'repos' / 'zephyr'
-    flags = _setup_test_init(manifest_repo, local_dir, directory, manifest_file)
+    flags = _setup_test_init(manifest_repo, local_dir, topdir, directory, manifest_file)
 
     # initialize west workspace
     cmd(['init'] + flags)
 
     # cd into the workspace
     if local_dir:
-        workspace = directory.parent
+        # topdir is either specified or default (directory.parent)
+        workspace = topdir or directory.parent
     else:
-        workspace = directory or Path.cwd()
+        # topdir is either specified, directory or default (cwd)
+        workspace = topdir or directory or Path.cwd()
     workspace_abs = workspace.absolute()
     os.chdir(workspace)
 
@@ -2099,12 +2138,11 @@ def test_init(repos_tmpdir, test_case):
     if local_dir:
         assert Path(actual.rstrip()) == directory.relative_to(workspace)
     else:
-        if directory:
-            # zephyr is cloned into specified subdirectory in workspace
-            subdir = directory.relative_to(workspace)
-            assert Path(actual.rstrip()) == Path(subdir) / 'zephyr'
+        if topdir and directory:
+            # manifest is cloned to the specified directory within workspace
+            assert Path(actual.rstrip()) == directory.relative_to(topdir)
         else:
-            # zephyr is cloned directly to workspace
+            # manifest is cloned directly to workspace, using the name from the manifest
             assert Path(actual.rstrip()) == Path('zephyr')
 
     # check if config option 'manifest.file' is correctly set
@@ -2115,20 +2153,44 @@ def test_init(repos_tmpdir, test_case):
     # re-initialization of a workspace must always fail
     _, stderr = cmd_raises(['init'] + flags, SystemExit)
     assert "FATAL ERROR: already initialized" in stderr
+    os.chdir(workspace.parent)
+    _, stderr = cmd_raises(['init', '-l', '--topdir', workspace_abs, manifest_repo], SystemExit)
+    assert "FATAL ERROR: already initialized" in stderr
+    _, stderr = cmd_raises(['init', '--topdir', workspace_abs, '-m', manifest_repo], SystemExit)
+    assert "FATAL ERROR: already initialized" in stderr
 
 
 TEST_CASES_INIT_FAILS = [
-    # (local_dir, directory, manifest_file, expected_error, expected_error)
+    # (local_dir, topdir, directory, manifest_file, expected_error, expected_error)
     ######################################
     # REMOTE MANIFEST
     ######################################
     # specify a non-existent manifest file
     (
         None,
+        Path('.'),
         None,
         Path('non-existent.yml'),
         SystemExit,
         "FATAL ERROR: can't init: no non-existent.yml found",
+    ),
+    # specify topdir and [directory] (but directory is not inside topdir)
+    (
+        None,
+        Path('subdir'),
+        Path('sibling'),
+        None,
+        SystemExit,
+        "FATAL ERROR: directory 'sibling' must be inside west topdir 'subdir'",
+    ),
+    # specify topdir and [directory] (but directory is not inside topdir)
+    (
+        None,
+        Path('somedir').absolute(),
+        Path('extra') / 'dir',
+        None,
+        SystemExit,
+        f"FATAL ERROR: directory '{Path('extra') / 'dir'}' must be inside west topdir",
     ),
     ######################################
     # LOCAL MANIFEST
@@ -2136,49 +2198,102 @@ TEST_CASES_INIT_FAILS = [
     # init workspace without a directory
     (
         Path('workspace') / 'zephyr',
+        Path('.'),
         None,
         None,
         SystemExit,
         "FATAL ERROR: can't init: no west.yml found",
     ),
+    # init workspace in a sibling repository path
+    (
+        Path('workspace') / 'zephyr',
+        Path('sibling'),
+        Path('workspace') / 'zephyr',
+        None,
+        SystemExit,
+        "west.yml must be relative to west topdir",
+    ),
     # init workspace from non-existent manifest
     (
         Path('workspace') / 'zephyr',
+        Path('.'),
         Path('non-existent.yml'),
         None,
         SystemExit,
         "FATAL ERROR: can't init: no west.yml found",
+    ),
+    # init workspace from a manifest not inside the workspace
+    (
+        Path('..') / 'zephyr',
+        Path('.'),
+        Path('..') / 'zephyr',
+        None,
+        SystemExit,
+        "west.yml must be relative to west topdir",
     ),
 ]
 
 
 @pytest.mark.parametrize("test_case", TEST_CASES_INIT_FAILS)
 def test_init_fails(repos_tmpdir, test_case):
-    local_dir, directory, manifest_file, expected_error, expected_stderr = test_case
+    local_dir, topdir, directory, manifest_file, expected_error, expected_stderr = test_case
 
     manifest_repo = repos_tmpdir / 'repos' / 'zephyr'
     repos_tmpdir.chdir()
-    flags = _setup_test_init(manifest_repo, local_dir, directory, manifest_file)
+    flags = _setup_test_init(manifest_repo, local_dir, topdir, directory, manifest_file)
     _, stderr = cmd_raises(['init'] + flags, expected_error)
     assert expected_stderr in stderr
+
+
+def test_init_result_in_similar_worktree(repos_tmpdir):
+    zephyr_remote = repos_tmpdir / 'repos' / 'zephyr'
+    repos_tmpdir.chdir()
+    zephyr_subdir = Path('extra') / 'level' / 'zephyr'
+
+    # init a workspace from an existing local manifest
+    topdir_l = Path(repos_tmpdir) / 'topdir_l'
+    clone(str(repos_tmpdir / 'repos' / 'zephyr'), topdir_l / 'extra' / 'level' / 'zephyr')
+    flags_l = ['-l', f'-t={topdir_l}', topdir_l / zephyr_subdir]
+    _ = cmd(['init'] + flags_l)
+
+    # init a workspace from a remote manifest using the same directory argument
+    topdir_r = Path(repos_tmpdir) / 'topdir_r'
+    flags_r = [f'-m={zephyr_remote}', f'-t={topdir_r}', topdir_r / zephyr_subdir]
+    _ = cmd(['init'] + flags_r)
+
+    # Check that both resulting worktrees look the same
+    tree_l = tree(topdir_l)
+    tree_r = tree(topdir_r)
+    assert tree_l == tree_r
 
 
 def test_init_topdir_again(repos_tmpdir):
     repos_tmpdir.chdir()
     workspace = Path(repos_tmpdir) / 'ws'
     zephyr = repos_tmpdir / 'repos' / 'zephyr'
+    err_msg = "FATAL ERROR: already initialized"
 
-    # initial clone
-    cmd(['init', 'ws', '-m', zephyr])
+    # initial clone using --topdir
+    cmd(['init', '--topdir', 'ws', '-m', zephyr])
 
     # initialize west workspace again
     _, stderr = cmd_raises(['init', 'ws', '-m', zephyr], SystemExit)
-    assert "FATAL ERROR: already initialized" in stderr
+    assert err_msg in stderr
+
+    # initialize west workspace again
+    _, stderr = cmd_raises(['init', '--topdir', 'ws', '-m', zephyr], SystemExit)
+    assert err_msg in stderr
 
     # initialize west workspace again
     os.chdir(workspace)
     _, stderr = cmd_raises(['init', '--local', 'zephyr'], SystemExit)
-    assert "FATAL ERROR: already initialized" in stderr
+    assert err_msg in stderr
+
+    # initialize west workspace again
+    os.chdir(workspace.parent)
+    _, stderr = cmd_raises(['init', '-l', '--topdir', 'ws', Path('ws') / 'zephyr'], SystemExit)
+    assert err_msg in stderr
+
 
 def test_init_local_with_empty_path(repos_tmpdir):
     # Test "west init -l ." + "west update".


mbolivar (commented):
Seems reasonable to me

pdgendt (approved):
Looks good!

pdgendt:
@marc-hb I'm not sure if @fouge is very active? I think we can move forward with this?

src/west/app/project.py
```diff
@@ -220,10 +259,11 @@ def do_add_parser(self, parser_adder):
                             background scanner to complete. ''')
 
         parser.add_argument(
-            'directory', nargs='?', default=None,
-            help='''with -l, the path to the local manifest repository;
-            without it, the directory to create the workspace in (defaulting
-            to the current working directory in this case)''')
+            'directory', nargs='?', default=None, metavar='directory',
+            help='''with --local: the path to the local manifest repository
```
carlescufi: I think we should only deprecate this when combined with `-l`, but not when pullling the repo from the remote Git server.

src/west/app/project.py
```diff
@@ -293,8 +337,12 @@ def local(self, args) -> Path:
         return topdir
 
     def bootstrap(self, args) -> Path:
-        topdir = Path(abspath(args.directory or os.getcwd()))
-        self.banner('Initializing in', topdir)
+        if args.directory:
+            self.wrn("positional argument [directory] is deprecated for "
```
carlescufi: Same comment as above, I think we should only deprecate this for `-l`, and what is actually a "remote manifest repository"?

src/west/app/project.py
```diff
@@ -187,31 +218,39 @@ def do_add_parser(self, parser_adder):
...
-                            in this case, and manifest.path points at
-                            "directory"''')
+                            help='''initialize from an already existing local
+                            manifest instead of cloning a remote manifest.''')
+        parser.add_argument('--mf', '--manifest-file', dest='manifest_file',
+                            metavar='FILE',
+                            help=f'''manifest file to use. It is the relative
+                            path of the manifest file within the repository
+                            (remote or local). Defaults to {_WEST_YML}.''')
+        parser.add_argument('-t', '--topdir', dest='topdir',
```
carlescufi: I think this option should only be available when using `-l`, and I don't think you are checking it?

src/west/app/project.py
pdgendt: After reading up again, and discussing with @carlescufi, we shouldn't deprecate this.

It'll be used all over in CI, so this is too much of an ask to remove even in the long run. You could disallow using it together with `--topdir`, though.

pdgendt (changes requested):

src/west/app/project.py
thorsten-klein: I think it makes sense for all `west init` commands, whereby `--topdir` option is even preferred if it is specified.
With `--topdir`, now there is a standard way to define the topdir, no matter if it is a local or remote manifest.
This behavior I even test in the tests, so that `topdir` is preferred over `directory`.

Note: The current `[directory]` has two different meanings, which seems not so nice to me.
It is ok to keep this for now, but IMO it would make sense to mark it `deprecated` for next major release, so that in future only one general way exists to specify a topdir via `west init --topdir`.

pdgendt: But what does the `directory` do when `--topdir` is passed? Running

```shell
$ west init -t . -m https://github.com/zephyrproject-rtos/example-application some/path/to/nowhere
```

The directory argument has no effect...? This is confusing. You should make `topdir` and `directory` mutually exclusive, right?

thorsten-klein: Unfortunately, it is not possible to make a positional argument mutually exclusive. This becomes very tricky here. 

> But what does the `directory` do when `--topdir` is passed?

Currently it is doing exactly like you say: It is overwritten by the topdir argument. 
This is also mentioned in the help of directory.
Therefore I also printed some depreciation warning in this case (removed with latest change) to mention that it is overwritten by topdir. Shall I throw an error instead? 

For me it sounds more intuitive that --topdir always works than having to know that the positional argument (directory) has a different meaning depending on the ``-l` flag. So I would propose to to get rid of the positional argument (directory). 

Is there any reason why `--local` is just a flag and does not take the local directory as a value?
If it was this way, the positional argument (directory) could be used as topdir. But this is a breaking change and will not be accepted I guess.

pdgendt: We can make them mutually exclusive with a manual check, and printing an error for the user. It's not possible with `argparse` probably.

src/west/app/project.py
thorsten-klein: Done.

src/west/app/project.py
thorsten-klein: A "remote manifest repository" is a git repository that contains a west manifest (`west.yml`).
In contrast to this there can be a local manifest.
Would you prefer some other naming?

pdgendt (dismissed):

marc-hb:
I really want to review this one because I already spent a lot of time on #774 and #775 before but @thorsten-klein is just submitting way too much code, I can't keep up!

`west` did not use to be like this, help! :-D

Can you give me a few more days?

thorsten-klein:
> @thorsten-klein is just submitting way too much code, I can't keep up.

Sorry 🤞 

> Can you give me a few more days?

Yes, sure 👍🏻  Thank you in advance for your time and input!

src/west/app/project.py
carlescufi: No, that's fine thanks @thorsten-klein. I misread it the first time.

src/west/app/project.py
carlescufi: We cannot get rid of (not even deprecate!) the positional argument, that would be a massive breaking change. I assume you implemented what @pdgendt suggested above, but personally I still think that `—topdir` should be only used with `-l`

carlescufi:
> @thorsten-klein is just submitting way too much code, I can't keep up.

Seconded, and I'll take this opportunity to thank you for all your contributions @thorsten-klein.

carlescufi (approved):
Even though I'd prefer `--topdir` was only compatible with `-l`, this is not a blocking issue for me. Thanks!

pdgendt:
Let's wait for feedback from @marc-hb, as he requested some more time.

src/west/app/project.py
```diff
@@ -155,19 +156,49 @@ def __init__(self):
...
+- If `--mr` is omitted, the repository's default branch (if available) is used.
 
-With neither, -m {MANIFEST_URL_DEFAULT} is assumed.
+Note: The west workspace will be created in the current working directory, if
+neither --topdir nor [directory] are provided.
 
-Warning: 'west init' renames and/or deletes temporary files inside the
+2. Local Manifest
+--------------
+If `--local` is given, west initializes a workspace from an already existing
```
```suggestion
If `-l / --local` is given, west initializes a workspace from an already existing
```

... otherwise you have to look in 3 places when reading this; inconvenient.

src/west/app/project.py
```diff
@@ -155,19 +156,49 @@ def __init__(self):
...
+Initialize a west workspace.
+
+West supports two ways of creating a workspace:
 
-With -l, creates a workspace around an existing local repository;
-without -l, creates a workspace by cloning a manifest repository
-by URL.
+1. Remote Manifest Repositories
+----------------------------
+If `-m` is given, west clones the manifest repository from the specified URL
```
```suggestion
If `-m / --manifest-url` is given, west git clones the manifest repository from the specified URL
```
For consistency with `-l` below.

src/west/app/project.py
```diff
@@ -155,19 +156,49 @@ def __init__(self):
...
+into the workspace. Config option manifest.path will point at this cloned
```
```suggestion
into a new workspace. Inside that workspace, the config option `manifest.path` will be the relative path to this git clone.
```

src/west/app/project.py
```diff
@@ -155,19 +156,49 @@ def __init__(self):
...
+  manifest repository.
+- If `--mr` is omitted, the repository's default branch (if available) is used.
 
-With neither, -m {MANIFEST_URL_DEFAULT} is assumed.
+Note: The west workspace will be created in the current working directory, if
+neither --topdir nor [directory] are provided.
 
-Warning: 'west init' renames and/or deletes temporary files inside the
+2. Local Manifest
+--------------
```
```suggestion
---------------
```

src/west/app/project.py
```diff
@@ -192,9 +223,10 @@ def do_add_parser(self, parser_adder):
         parser = self._parser(
             parser_adder,
             usage='''
-
-  %(prog)s [-m URL] [--mr REVISION] [--mf FILE] [-o=GIT_CLONE_OPTION] [directory]
-  %(prog)s -l [--mf FILE] directory
+  remote repository:
+  %(prog)s [-m URL] [--mr REVISION] [--mf FILE] [-o=GIT_CLONE_OPTION] [-t WORKSPACE_DIR] [directory]
+  local manifest:
+  %(prog)s -l [-t WORKSPACE_DIR] [--mf FILE] directory
```
```suggestion
  %(prog)s -l [-t WORKSPACE_DIR] [--mf FILE] [manifest_directory]
```

src/west/app/project.py
```diff
@@ -192,9 +223,10 @@ def do_add_parser(self, parser_adder):
         parser = self._parser(
             parser_adder,
             usage='''
-
-  %(prog)s [-m URL] [--mr REVISION] [--mf FILE] [-o=GIT_CLONE_OPTION] [directory]
-  %(prog)s -l [--mf FILE] directory
+  remote repository:
+  %(prog)s [-m URL] [--mr REVISION] [--mf FILE] [-o=GIT_CLONE_OPTION] [-t WORKSPACE_DIR] [directory]
```
marc-hb, line 317:
```suggestion
  %(prog)s [-m URL] [--mr REVISION] [--mf FILE] [-o=GIT_CLONE_OPTION] [-t WORKSPACE_DIR | directory]
```
Otherwise:
`FATAL ERROR: --topdir cannot be combined with positional argument [directory]`

src/west/app/project.py
```diff
@@ -155,19 +156,49 @@ def __init__(self):
...
-manifest repository. If --mr is not given, the remote's default
-branch will be used, if it exists.
+- If `-m` is not provided, west defaults to the Zephyr manifest repository:
+  {MANIFEST_URL_DEFAULT}.
+- With `--mr`, you can specify the revision (branch, tag, or sha) of the
+  manifest repository.
+- If `--mr` is omitted, the repository's default branch (if available) is used.
 
-With neither, -m {MANIFEST_URL_DEFAULT} is assumed.
+Note: The west workspace will be created in the current working directory, if
```
```suggestion
Note: The west workspace will be the current working directory, if
```

I think this help text needs to entirely ban the expression "workspace in X/" because this expression is ambiguous: it can mean either `X/.west/` or `X/Y/.west`



This is even more important with this new `--topdir` option.

marc-hb (changes requested):
> This is particularly useful when using --local

I downloaded this PR and tested the other, --manifest-url option and it does not seem to work for me. I ran out of time and did not try --local yet. I didn't look at the code either.

Describing directory nesting clearly and accurately is **not** easy - it is basically a "specification" except it has to be user-friendly too. So I will look at the code only when that "specification" is complete and unambiguous.

This help looks like the two last ones can be both used at the same time:
```
west init [-m URL] [--mr REVISION] [--mf FILE] [-o=GIT_CLONE_OPTION] [-t WORKSPACE_DIR] [directory]
```

But it's not actually possible:
```
FATAL ERROR: --topdir cannot be combined with positional argument [directory]
```

When using either separately, each one seems to have the same effect. So, there does not seem to be any new feature in the remote case, which means it's still not possible to `west init` a manifest two levels down or more in the remote case? Is this an oversight? The help does not seem to say. I don't see why `--local` would be more flexible than `--manifest-url`. I would hope the `--manifest-url` and `--local` options work the same immediately after the git clone operation.

marc-hb:
> > This is particularly useful when using --local
> 
> I downloaded this PR and tested the other, --manifest-url option and it does not seem to work for me. I ran out of time and did not try --local yet. I didn't look at the code either.
> 
> Describing directory nesting clearly and accurately is **not** easy - it is basically a "specification" except it has to be user-friendly too. So I will look at the code only when that "specification" is complete and unambiguous (EDIT and of course: working for me).
> 
> This help looks like the two last ones can be both used at the same time:
> 
> ```
> west init [-m URL] [--mr REVISION] [--mf FILE] [-o=GIT_CLONE_OPTION] [-t WORKSPACE_DIR] [directory]
> ```
> 
> But it's not actually possible:
> 
> ```
> FATAL ERROR: --topdir cannot be combined with positional argument [directory]
> ```
> 
> When using either separately, each one seems to have the same effect. So, there does not seem to be any new feature in the remote case, which means it's still not possible to `west init` a manifest two levels down or more in the remote case? Is this an oversight? The help does not seem to say. I don't see why `--local` would be more flexible than `--manifest-url`. I would hope the `--manifest-url` and `--local` options work the same immediately after the git clone operation.

thorsten-klein:
Thank you for the review 🙏

> When using either separately, each one seems to have the same effect. So, there does not seem to be any new feature in the remote case, which means it's still not possible to west init a manifest two levels down or more in the remote case? Is this an oversight? The help does not seem to say. I don't see why --local would be more flexible than --manifest-url. I would hope the --manifest-url and --local options work the same immediately after the git clone operation.

Yes, exactly. `--topdir` and positional argument `directory` do exactly the same in remote case.
I have proposed to get rid of this positional argument at all, but it will not be accepted for now because it is a breaking change (what I fully understand). 

Nevertheless, in remote case those arguments that do the same are mutually exclusive (see error message you got).
I have applied your suggested change, which totally make sense to point out in the help that they both cannot be combined (`[--topdir DIR | directory]`) 👍🏻 

> it's still not possible to west init a manifest two levels down or more in the remote case?

This has already worked before in remote case, since the positional argument `[directory]` already exists and it does exactly the same as `--topdir` (see above). The main value in this change lays in the local case (and is the pre-work to potentially remove the positional argument `[directory]` in future)

Thank you 👍🏻

thorsten-klein:
I think I have resolved all review comments in the code👍🏻

src/west/app/project.py
```diff
@@ -155,19 +156,54 @@ def __init__(self):
...
+Initialize a west workspace.
+
+West supports two ways of creating a workspace:
+
+1. Remote Manifest Repositories
+-------------------------------
+If `-m / --manifest-url` is given, west clones the manifest repository from the
+specified URL into a new workspace. Inside that workspace, the local config
+option `manifest.path` will be set to point at this cloned directory (relative
+path from the git clone to the workspace).
```
```suggestion
path to the git clone from the workspace).
```

marc-hb:
> > it's still not possible to west init a manifest two levels down or more in the remote case?

> This has already worked before in remote case, ...

Has it? Please share one example of a `west init --manifest-url ...` command that can produce this:

```
workspace
├── extra
│   └── level
│       └── manifest_clone
│           └── .git
└── .west
```

thorsten-klein:
> it's still not possible to west init a manifest two levels down

Yes, you are right. I will push my changes where I have added support for this and also some test.

src/west/app/project.py
```diff
@@ -155,19 +156,61 @@ def __init__(self):
...
+    Specifies the directory where west should create the workspace.
+    The `.west` folder will be created inside this directory.
+
+
+1. Remote Manifest Repository (default)
+---------------------------------------
+West clones a remote repository (provided via `-m / --manifest-url`) into a new
+workspace. The repositoy has to contain a west manifest.
+Inside the new workspace, the local config option `manifest.path` is set to
+point at the cloned directory (relative path to the workspace).
```
marc-hb: Maybe that's because I'm not a native speaker (anyone?) but I still read this as "the path (that leads) to the workspace", which obviously does not sound right.


```suggestion
point at the cloned directory (relative path from the workspace).
```

src/west/app/project.py
```diff
@@ -155,19 +156,61 @@ def __init__(self):
...
+  default branch (if available).
+
+2. Local Manifest
+-----------------
+If `-l / --local` is given, west initializes a new workspace from an already
+existing local manifest. The directory containing the manifest can be provided
+in positional argument `[manifest_directory]` (it defaults to current working
+directory if no value is provided).
+Inside the new workspace, the local config option `manifest.path` is set to
+point at the directory containing the manifest (relative path to the workspace)
```
marc-hb: This is duplicated across 1. and 2. I would move it at the top, just before the "Arguments" section.

src/west/app/project.py
```diff
@@ -155,19 +156,61 @@ def __init__(self):
...
+If no `-m / --manifest-url` is provided, west uses Zephyr URL by default:
+  {MANIFEST_URL_DEFAULT}.
+
+The west workspace (topdir) is determined as follows:
+- `-t / --topdir` if provided
+- `[directory]` if provided
+- the current working directory (if neither `--topdir` nor `[directory]` are
+  provided)
+
+Note:
```
```suggestion
```
A "Note" sounds like something not important to know. These seem important to me, it's what this whole PR is about? And the whole section does not seem long enough for "notes".

src/west/app/project.py
```diff
@@ -155,19 +156,61 @@ def __init__(self):
...
+2. Local Manifest
+-----------------
+If `-l / --local` is given, west initializes a new workspace from an already
+existing local manifest. The directory containing the manifest can be provided
+in positional argument `[manifest_directory]` (it defaults to current working
+directory if no value is provided).
+Inside the new workspace, the local config option `manifest.path` is set to
+point at the directory containing the manifest (relative path to the workspace)
+
+Note: The west workspace is created in the `manifest_directory`'s parent
```
```suggestion
The west workspace created is the parent of `manifest_directory' if ...
```

Again, I really think the phrasing "workspace is ~**in**~ X" must be avoided everywhere because that can be understood as either `X/.west/`, or `X/ws/.west`. Such ambiguity is obviously critical in the context of this PR and help text.

This is not a "Note" either, this information is central.

src/west/app/project.py
```diff
@@ -155,19 +156,61 @@ def __init__(self):
...
+workspace. The repositoy has to contain a west manifest.
+Inside the new workspace, the local config option `manifest.path` is set to
+point at the cloned directory (relative path to the workspace).
+
+If no `-m / --manifest-url` is provided, west uses Zephyr URL by default:
+  {MANIFEST_URL_DEFAULT}.
+
+The west workspace (topdir) is determined as follows:
+- `-t / --topdir` if provided
+- `[directory]` if provided
```
```suggestion
- `directory` if provided
```

The brackets are only to show _optional_ arguments in the "usage" section. Take a look at "man \<anything\>".

Same issue below.

src/west/app/project.py
```diff
@@ -155,19 +156,61 @@ def __init__(self):
...
-
-With neither, -m {MANIFEST_URL_DEFAULT} is assumed.
-
-Warning: 'west init' renames and/or deletes temporary files inside the
+Arguments
+---------
+--mf
+    The relative path to the manifest file within the manifest repository
+    (remote or local). Config option manifest.file will be set to this value.
+    Defaults to `{_WEST_YML}` if not provided.
```
```suggestion
    Defaults to '{_WEST_YML}' if not provided.
```
Help texts generally don't use backquotes. I had a look and I don't think `west` has been so far.

Same issue below.

src/west/app/project.py
```diff
@@ -249,9 +307,10 @@ def do_add_parser(self, parser_adder):
             'directory',
             nargs='?',
             default=None,
-            help='''with -l, the path to the local manifest repository;
-            without it, the directory to create the workspace in (defaulting
-            to the current working directory in this case)''',
+            metavar='directory',
+            help='''with --local: the path to the local manifest repository
+                    which contains a west.yml;
+                    otherwise: the directory to create the workspace in''',
```
```suggestion
otherwise it depends whether --topdir is used or not, see above.
```

src/west/app/project.py
```diff
@@ -155,19 +156,61 @@ def __init__(self):
...
+-----------------
+If `-l / --local` is given, west initializes a new workspace from an already
+existing local manifest. The directory containing the manifest can be provided
+in positional argument `[manifest_directory]` (it defaults to current working
+directory if no value is provided).
+Inside the new workspace, the local config option `manifest.path` is set to
+point at the directory containing the manifest (relative path to the workspace)
+
+Note: The west workspace is created in the `manifest_directory`'s parent
+      if no other directory is provided in `--topdir`.
```
marc-hb: I think this section 2. can be shortened and clarified significantly while not losing any information. Let me give it a try; two short sentences:

If `-l / --local` is given, west initializes a new workspace from a
manifest already cloned and found in the '?manifest_?directory' argument. This defaults to the current directory.
The `--topdir` argument is the new workspace created. It defaults to the parent of the manifest directory.

marc-hb (dismissed):
I gave the new version a try and I think I found two confirmed bugs, see below. Reproducing and describing these bugs was very time-consuming so I ran out of time again and I didn't look at the code again :-(

I love the energy but please be much more thorough in testing edge cases and addressing review comments, otherwise the limited review bandwidth cannot keep up. It's a bit counter-intuitive but in this case "slowing down" means finishing sooner.

There seems to be a large quantity of test code being added which is great! However the "quality" of the test code matters even more than its quantity. I mean being imaginative and trying hard to "break" the production code by reaching as many edge cases as possible is much more important than being repetitive. Too much test repetition only makes running the test suite longer.


**Bug 1: spurious level in remote --topdir**

```
# west init  --topdir ws/   ws/manifest_dir/

$ tree -a -L 3 ws

ws
├── manifest_dir
│   └── zephyr      # <=== !!
│     ├── .git
│     
└── .west
    └── config
```

I don't think this extra "zephyr" level matches the current --help text. Either way, I don't think the extra level is the correct behavior because the extra level is inconsistent with how the --local behavior treat the positional argument.

Funny enough, this rejoins my other observation that writing "workspace/manifest created '**in**' X/" is ambiguous and a bad idea. Does "in" mean "X/.git" or "X/manifest/.git"? Unclear, avoid "in".


**Bug 2: re-initialization is not always caught**

I suspect the check that catches the re-initialization needs to be "upgraded" for the new --topdir


```
west init ws

cd ws
west init --local zephyr
FATAL ERROR: already initialized in ...                       # Good!

cd ..
west init --local --topdir ws/  ws/zephyr/
=== Initializing from existing manifest repository zephyr     # What?
--- Creating ws/.west and local configuration file
=== Initialized. Now run "west update" inside ws.
```

Interestingly enough, .west/config does not seem to be modified - unlike what the logs pretend. So there may be **more than one bug there**. I mean check why the logs are lying BEFORE fixing the check.

thorsten-klein:
> I suspect the check that catches the re-initialization needs to be "upgraded" for the new --topdir

@marc-hb I’ve tried with the code to fail if the topdir is already initialized (when using `--local`).
However, it seems that some tests (e.g. `test_update_some_with_imports`) rely on the current behavior and perform a re-initialization of some already initialized workspace.
What shall I do? 
It looks like this behavior might have been intentional, so I make it fail only when the new argument `--topdir` is used.

thorsten-klein:
> Has it? Please share one example of a west init --manifest-url ... command that can produce this
```
workspace
├── extra
│   └── level
│       └── manifest_clone
│           └── .git
└── .west
```

You requested that west is able to create such a layout. The command for this is now:
`west init  --topdir ws   ws/extra/level`

All projects will be cloned under `ws/extra/level`.
Is this correct behavior or what would be your expectation when both `args.topdir` and `directory` are provided?
If they are mutually exclusive (as requested above), your requested feature will not be possible.
Currently. they are not mutually exclusive anymore in order to support your use case.

src/west/app/project.py
thorsten-klein: I think there are already some backslashes in the help of ManifestCommand (in this same file)

```
This is similar to `git status` for untracked files
```

Do you nevertheless prefer single quotes?

src/west/app/project.py
thorsten-klein: I will shorten it, but I would like to keep the info that it is by default the manifest_directory's parent.

src/west/app/project.py
thorsten-klein: I chose naming analogue to `pathlib` where it also says `.relative_to(other)` 
Ref: https://docs.python.org/3/library/pathlib.html

src/west/app/project.py
```diff
@@ -219,21 +264,33 @@ def do_add_parser(self, parser_adder):
...
             help='''manifest repository branch or tag name
                     to check out first; cannot be combined with -l''',
         )
-        parser.add_argument(
-            '--mf', '--manifest-file', dest='manifest_file', help='manifest file name to use'
-        )
         parser.add_argument(
             '-l',
             '--local',
             action='store_true',
```
thorsten-klein, line 354: The more I think about this PR, the more I prefer that this argument takes a value, for example: `--local manifest_dir`. 
In my opinion, this would make the command’s usage much more consistent, as it removes the need for a positional argument that behaves differently depending on which other options are provided.  
Instead the positional argument could be the workspace directory (topdir) that is initialized.

If you are afraid of the breaking change, a new command could be introduced to implement this behavior (with a new argparser and therefore without breaking compatibility), maybe something like
```
west workspace -l manifest_dir [topdir]
west workspace -m manifest_url [topdir]
```

---

EDIT: This is maybe some follow up discussion, since this change could be planned for some west 2.x. 
So let's continue with this branch as it does not make anything worse.

carlescufi:
@marc-hb could you please follow-up on this PR?

marc-hb:
Sorry I was on vacation and then it fell through the cracks. Will take another look ASAP

src/west/app/project.py
```diff
@@ -156,19 +157,60 @@ def __init__(self):
...
-
-With neither, -m {MANIFEST_URL_DEFAULT} is assumed.
-
-Warning: 'west init' renames and/or deletes temporary files inside the
+Initialize a west workspace (topdir) around a west manifest (`west.yml`) by
+creating a `.west` directory in the topdir and a local configuration file
+`.west/config`.
+The West manifest can come from either a remote repository (that will be cloned
+during workspace initialization) or from an already existing local directory.
+Config option `manifest.path` is set accordingly (see below).
```
marc-hb: This is too early for that implementation detail, you can drop that line here.

To be honest I'm not sure this name and implementation detail should be mentioned at all in the help text. But for sure not that early.

src/west/app/project.py
```diff
@@ -193,9 +235,11 @@ def do_add_parser(self, parser_adder):
         parser = self._parser(
             parser_adder,
             usage='''
+  init around a remote repository:
```
```suggestion
  init after cloning a manifest repository:
```
"around" something that does not exist yet sounds weird. "remote" sounds like it has to go over the network. I don't think it has.

src/west/app/project.py
```diff
@@ -204,14 +248,16 @@ def do_add_parser(self, parser_adder):
         parser.add_argument(
             '-m',
             '--manifest-url',
-            help='''manifest repository URL to clone;
+            metavar='URL',
+            help='''remote manifest repository URL to clone;
```
marc-hb: "remote" sounds like this must go over the network. I don't think it has to. I don't think this line needs to change. It already says "clone" so it's not ambiguous.

src/west/app/project.py
```diff
@@ -220,21 +266,35 @@ def do_add_parser(self, parser_adder):
...
             '-l',
             '--local',
             action='store_true',
-            help='''use "directory" as an existing local
-                    manifest repository instead of cloning one from
-                    MANIFEST_URL; .west is created next to "directory"
-                    in this case, and manifest.path points at
-                    "directory"''',
+            help='''initialize workspace around an already existing local
+                    manifest instead of cloning a remote manifest;
```
```suggestion
                    manifest instead of cloning a manifest;
```

src/west/app/project.py
```diff
@@ -220,21 +266,35 @@ def do_add_parser(self, parser_adder):
...
+                    cannot be combined with -m''',
+        )
+        parser.add_argument(
+            '--mf',
+            '--manifest-file',
+            dest='manifest_file',
+            metavar='FILE',
+            help=f'''manifest file to use. It is the relative
+                     path of the manifest file within the remote repository
+                     respectively the local manifest_directory.
```
marc-hb: I don't think both cases need to be listed, it's the same thing in both cases. This just the "relative path TO the manifest file within the manifest directory" and that's IMHO enough. It does not matter whether that manifest directory is remote, local, or being cloned from a local disk. All that dilutes the information.

src/west/app/project.py
```diff
@@ -220,21 +266,35 @@ def do_add_parser(self, parser_adder):
...
+                     Defaults to {_WEST_YML}.''',
+        )
+        parser.add_argument(
+            '-t',
+            '--topdir',
+            dest='topdir',
+            metavar='WORKSPACE_DIR',
+            help='''the location of the west workspace where .west directory
```
marc-hb: typo
```suggestion
            help='''the workspace directory where the .west/ directory
```
Also add: "WORKSPACE_DIR may already exist, `WORKSPACE_DIR/.west/` may not."

src/west/app/project.py
```diff
@@ -250,9 +310,15 @@ def do_add_parser(self, parser_adder):
             'directory',
             nargs='?',
             default=None,
-            help='''with -l, the path to the local manifest repository;
-            without it, the directory to create the workspace in (defaulting
-            to the current working directory in this case)''',
+            metavar='directory',
+            help='''With --local: the path to a local directory which contains
+                    a west manifest (west.yml);
+                    Otherwise (if -m is used and no --topdir is provided):
```
```suggestion
                    Otherwise, if -m is used and no --topdir is provided:
```
Parenthesis mean "can be skipped without much loss". This bit is very important, you don't want readers to skip it. Same below.

src/west/app/project.py
```diff
@@ -156,19 +157,60 @@ def __init__(self):
...
+
+Arguments
+---------
+--mf / --manifest-file
+    The relative path to the manifest file within the manifest repository
+    or directory. Config option `manifest.file` will be set to this value.
+    Defaults to `{_WEST_YML}` if not provided.
+
+-t / --topdir
+    Specifies the directory where west should create the workspace.
```
marc-hb, line 254:
```suggestion
    Specifies the directory where west should create the workspace. It may already exist.
```

src/west/app/project.py
```diff
@@ -156,19 +157,60 @@ def __init__(self):
...
+    The `.west` folder will be created inside this directory.
+
+
+1. Remote Manifest Repository (default)
```
```suggestion
1. Cloning a Manifest Repository (default)
```

src/west/app/project.py
```diff
@@ -156,19 +157,60 @@ def __init__(self):
...
+---------------------------------------
+West clones the given remote repository (provided via `-m / --manifest-url`).
```
```suggestion
West clones the given repository (provided via `-m / --manifest-url`).
```
It can be cloned from the same disk, does not have to be remote. "clones" is the keyword that is enough to disambiguate and the only difference that matters. It's tempting to think that the opposite of "local" is "remote" but that's not really true for `west init`.

The previous version used the word "remote" only once and only in a specific git context. This PR is bumping that to 8 while changing nothing in that respect. Please search and drop "remote" almost everywhere.

src/west/app/project.py
```diff
@@ -156,19 +157,60 @@ def __init__(self):
...
+  {MANIFEST_URL_DEFAULT}.
+
+The topdir (where `.west` is created) is determined as follows:
+- argument `-t / --topdir` if provided
+- otherwise: argument `directory` if provided
+- otherwise: the current working directory
+
+If both `-t / --topdir` and `directory` are provided, `-t / --topdir`
+specifies the workspace directory, while `directory` specifies the subfolder
+within the workspace where the remote repository is cloned during
```
```suggestion
within the workspace where the manifest repository is cloned during
```

src/west/app/project.py
```diff
@@ -156,19 +157,60 @@ def __init__(self):
...
+1. Remote Manifest Repository (default)
+---------------------------------------
+West clones the given remote repository (provided via `-m / --manifest-url`).
+Config option `manifest.path` is set to point to this git clone.
+Note, that the repository must contain a west manifest.
+
+If no `-m / --manifest-url` is provided, west uses Zephyr URL by default:
+  {MANIFEST_URL_DEFAULT}.
+
+The topdir (where `.west` is created) is determined as follows:
```
marc-hb, line 267: (It really sucks we have to keep such a complicated logic for backwards compatibility)

src/west/app/project.py
```diff
@@ -156,19 +157,60 @@ def __init__(self):
...
+specifies the workspace directory, while `directory` specifies the subfolder
+within the workspace where the remote repository is cloned during
+initialization (e.g. by default west clones to <directory>/zephyr/.git).
+With `--mr`, the revision (branch, tag, or sha) of the remote repository can be
+specified that will be used. It defaults to the repository's default branch.
+
+2. Local Manifest
+-----------------
+If `-l / --local` is given, west initializes a workspace around an already
+existing `west.yml`. Its location can be specified in `manifest_directory`
```
```suggestion
existing `west.yml`. Its location can be specified in the `manifest_directory` positional argument
```

src/west/app/project.py
```diff
@@ -156,19 +157,60 @@ def __init__(self):
...
+(defaults to current working directory). Config option `manifest.path` is set
```
```suggestion
It defaults to current working directory.
```
No parenthesis.

Is mentioning `manifest.path` really required at all in this help text? I think the west documentation should be enough.

src/west/app/project.py
```diff
@@ -405,6 +496,7 @@ def bootstrap(self, args) -> Path:
             # is PurePosixPath.
             manifest_path = PurePath(urlparse(manifest_url).path).name
 
+        manifest_path = str(Path(subdir) / manifest_path)
```
marc-hb: Why does need to be converted to a string? The comment just above warns about this needing to be a `PurePath` in some cases...

src/west/app/project.py
```diff
@@ -303,16 +369,28 @@ def local(self, args) -> Path:
...
-        west_dir = topdir / WEST_DIR
-
         if not manifest_file.is_file():
             self.die(f'can\'t init: no {manifest_filename} found in {manifest_dir}')
 
-        self.banner('Initializing from existing manifest repository', rel_manifest)
+        if args.topdir:
+            topdir = Path(args.topdir).resolve()
+            try:
+                already = util.west_topdir(topdir, fall_back=False)
```
```suggestion
                already = util.west_topdir(manifest_dir, fall_back=False)
```
We want to maximize the chances of finding an existing workspace, not minimize them.

Also: move the `manifest_file.is_relative_to(topdir):` check first.

marc-hb (commented):
Partial review, I didn't get a chance to look at the test code yet, sorry. Later this week.


> I’ve tried with the code to fail if the topdir is already initialized (when using --local). However, it seems that some tests (e.g. test_update_some_with_imports) rely on the current behavior and perform a re-initialization of some already initialized workspace.
> What shall I do? It looks like this behavior might have been intentional, so I make it fail only when the new argument --topdir is used.
> 

I'm amazed that such "reinit" situations exist, is it still the case? If yes, can you elaborate and provide a simple reproduction? This looks like a serious bug. Does it predate this PR? Does it simply deletes the existing `.west/` directory? I can't believe there is code doing that.

> You requested that west is able to create such a layout. 

I only requested feature parity between --local and not --local. This is not "my" use case, this is the same use case than --local but when cloning.

> Is this correct behavior or what would be your expectation when both args.topdir and directory are provided?
> If they are mutually exclusive (as requested above), your requested feature will not be possible.

I don't remember requesting these they should be mutually exclusive, did I? I only remember experiencing that they were mutually exclusive in a previous version while the --help presented them as not mutually exclusive. So I requested to correct the help: https://github.com/zephyrproject-rtos/west/pull/854#discussion_r2434378346
Obviously, that's not relevant any more.


I find the current user interface confusing in the cloning case but:
- it seems consistent with --local
- it seems backwards-compatible
- I can't think of a better one at this point.

src/west/app/project.py
thorsten-klein: To me it was a helpful detail since I can directly see as a user what's the impact of this command.
Especially it helped to point out the difference between local and remote.

But I will remove 👍

src/west/app/project.py
thorsten-klein: I think it was some naming I have read anywhere and took over. 
I would prefer having some shrt preposition.
Maybe one of those?
- `init workspace FROM an existing local manifest`
- `init workspace WITH an existing local manifest`
- `init workspace FOR an existing local manifest`

I prefer `WITH`.

src/west/app/project.py
thorsten-klein: For setting the option the value have to be `str`: `TypeError: option values must be strings`
I have moved to conversion to the function call now 👍

src/west/app/project.py
thorsten-klein: ~~I think this is not correct.
E.g. `--topdir workspace --local /path/to/my/manifest` should only fail if `topdir` is already a workspace, not if any parent of the manifest is already a workspace.~~

Ok.

src/west/app/project.py
marc-hb: Backquotes are useful in .rst docs but they look weird. Not important.

src/west/app/project.py
```diff
@@ -156,19 +157,57 @@ def __init__(self):
...
+  {MANIFEST_URL_DEFAULT}.
+
+The topdir (where `.west` is created) is determined as follows:
+- argument `-t / --topdir` if provided
+- otherwise: argument `directory` if provided
+- otherwise: the current working directory
+
+If both `-t / --topdir` and `directory` are provided, `-t / --topdir`
+specifies the workspace directory, while `directory` specifies the subfolder
+within the workspace where the repository is cloned during initialization
```
```suggestion
within the workspace where the manifest repository is cloned during initialization
```

src/west/app/project.py
```diff
@@ -156,19 +157,57 @@ def __init__(self):
...
+---------------------------------------
+West clones the given repository (provided via `-m / --manifest-url`).
+Note, that the repository must contain a west manifest.
+
+If no `-m / --manifest-url` is provided, west uses Zephyr URL by default:
+  {MANIFEST_URL_DEFAULT}.
+
+The topdir (where `.west` is created) is determined as follows:
+- argument `-t / --topdir` if provided
+- otherwise: argument `directory` if provided
```
```suggestion
- otherwise: the positional `directory` argument if provided without --topdir. Deprecated: prefer --topdir.
```

src/west/app/project.py
```diff
@@ -156,19 +157,57 @@ def __init__(self):
...
+- otherwise: the current working directory
+
+If both `-t / --topdir` and `directory` are provided, `-t / --topdir`
+specifies the workspace directory, while `directory` specifies the subfolder
```
```suggestion
specifies the workspace directory, while the positional directory argument specifies the manifest subfolder
```

src/west/app/project.py
marc-hb: To simplify the logic, we can just deprecate using a positional argument instead of the new --topdir. This will make the local and cloning case more consistent with each other.

src/west/app/project.py
```diff
@@ -250,9 +306,15 @@ def do_add_parser(self, parser_adder):
...
             default=None,
-            help='''with -l, the path to the local manifest repository;
-            without it, the directory to create the workspace in (defaulting
-            to the current working directory in this case)''',
+            metavar='directory',
+            help='''With --local: the path to a local directory which contains
+                    a west manifest (west.yml);
+                    Otherwise (if -m is used and no --topdir is provided):
+                    the location of the west workspace where .west directory
+                    will be created (<directory>/.west)
```
marc-hb, line 397:
```suggestion
                    will be created (<directory>/.west). Deprecated: use --topdir.
```

src/west/app/project.py
marc-hb: Wait: `-m` is not relevant here, is it?
```suggestion
                    Otherwise, if no --topdir is provided:
```

src/west/app/project.py
```diff
@@ -303,16 +365,26 @@ def local(self, args) -> Path:
...
-        west_dir = topdir / WEST_DIR
-
         if not manifest_file.is_file():
             self.die(f'can\'t init: no {manifest_filename} found in {manifest_dir}')
 
-        self.banner('Initializing from existing manifest repository', rel_manifest)
+        topdir = Path(args.topdir or manifest_dir.parent).resolve()
+
+        if not manifest_file.is_relative_to(topdir):
+            self.die(f'{manifest_file} must be relative to west topdir')
```
```suggestion
            self.die(f'{manifest_file} must be relative to west topdir {topdir}')
```

src/west/app/project.py
```diff
@@ -323,8 +395,21 @@ def local(self, args) -> Path:
...
+            topdir = Path(abspath(args.topdir))
+            if args.directory:
+                if not Path(abspath(args.directory)).is_relative_to(topdir):
+                    self.die(
+                        f"directory '{args.directory}' must be relative "
+                        f"to west topdir '{args.topdir}'"
+                    )
+                subdir = os.path.relpath(args.directory, topdir)
+        elif args.directory:
+            topdir = Path(abspath(args.directory))
```
```suggestion
            self.wrn(f"DEPRECATED, use --topdir {args.directory} instead")
            topdir = Path(abspath(args.directory))
```

src/west/app/project.py
```diff
@@ -323,8 +395,21 @@ def local(self, args) -> Path:
         return topdir
 
     def bootstrap(self, args) -> Path:
-        topdir = Path(abspath(args.directory or os.getcwd()))
-        self.banner('Initializing in', topdir)
+        subdir = '.'
+        if args.topdir:
+            topdir = Path(abspath(args.topdir))
+            if args.directory:
+                if not Path(abspath(args.directory)).is_relative_to(topdir):
```
marc-hb: `abspath(args.directory)` does not make sense because `args.directory` is meant to be appended to topdir first. And then the former is automatically relative to the latter. So this check does not make sense to me.

src/west/app/project.py
```diff
@@ -396,15 +481,17 @@ def bootstrap(self, args) -> Path:
...
+            mf_path = manifest.yaml_path
         else:
             # We use PurePath() here in case manifest_url is a
             # windows-style path. That does the right thing in that
             # case, without affecting POSIX platforms, where PurePath
             # is PurePosixPath.
-            manifest_path = PurePath(urlparse(manifest_url).path).name
+            mf_path = PurePath(urlparse(manifest_url).path).name
 
+        manifest_path: Path = Path(subdir) / mf_path
```
marc-hb: This looks like a misunderstanding... `manifest.path` comes either from the command line, or from `self.path` in the manifest file, but it's never a combination of both. If both are defined, the code should fail if they don't match.

If none is defined, keeping falling back on basename(url) like before.

marc-hb (changes requested):
> manifest_path: Path = Path(subdir) / mf_path

This line does not make sense to me, see below.

src/west/app/project.py
thorsten-klein: I would propose to add the `--topdir` feature here, and do this deprecation in a separate PR.

src/west/app/project.py
thorsten-klein: I think `if provided without --topdir` is not relevant (same as your comment above where I removed `if -m is used`).
Also here I would propose to do the Deprecation in a separate PR.

src/west/app/project.py
thorsten-klein: I would propose to do the deprecation in a separate PR.

src/west/app/project.py
thorsten-klein: I would propose to do the deprecation in a separate PR.

src/west/app/project.py
thorsten-klein: Without prepending `subdir` the test fails when `topdir` is `Path('subdir')` and the positional argument `directory` is `Path('subdir') / 'extra' / 'level'`)

As per definition:
```
If both `-t / --topdir` and `directory` are provided, `-t / --topdir`
specifies the workspace directory, while positional argument `directory`
specifies the subfolder within the workspace where the manifest repository is
cloned during initialization
```

Therefore the manifest will be located under `<topdir>/<subdir>/<yml_path>`

UPDATE: I have locally merged your PR (https://github.com/zephyrproject-rtos/west/pull/906) and all tests still pass 👍🏻

src/west/app/project.py
marc-hb: Agreed for a separate PR, glad you're ok with the deprecation idea. Can you please add a comment in the code now? "# This case will be deprecated later". And in the corresponding test code if any.

src/west/app/project.py
marc-hb: > Without prepending subdir the test fails...

Tests are only a means to an end, if they don't reflect the desired behavior then they need to change.

> As per definition

In the --local case, the positional argument becomes `manifest.path`[*]. For consistency, I think we want the same in the cloning case. If that clashes with self.path then we should either error or warn, not sure which.


[*] Not "literally" because absolute/relative but they do point at the same directory.

thorsten-klein: I’m not sure we’re talking about the same thing. Let’s align.

When `west init --topdir subdir subdir/extra/level` is invoked, my expectation is that:

- The workspace is created under `./subdir`, i.e. `./subdir/.west` is created.
- The Zephyr repository is cloned to `<topdir>/extra/level/zephyr/.git` (which is `./subdir/extra/level/zephyr/.git`).
- The `manifest.path` config option points to `extra/level/zephyr`.
- The `manifest.file` config option is set to `west.yml`.

Does this match your expectations? Am I incorrect about any of these points, or missing something important?

tests/test_project.py
```diff
@@ -1876,6 +1870,186 @@ def test_init_local_with_manifest_filename(repos_tmpdir):
     cmd('update')
 
 
+TEST_CASES_INIT = [
```
pdgendt, line 2063: I don't think we should be mixing test cases that result in errors with successful cases.

Ideally the cases that result in errors have the error message checked (a test function per expected different error case).

tests/test_project.py
```diff
@@ -1876,6 +1870,197 @@ def test_init_local_with_manifest_filename(repos_tmpdir):
...
+
+
+@pytest.mark.parametrize("test_case", TEST_CASES_INIT_FAILS)
+def test_init_fails(repos_tmpdir, test_case):
+    local_dir, topdir, directory, manifest_file, expected_error = test_case
+
+    manifest_repo = repos_tmpdir / 'repos' / 'zephyr'
+    repos_tmpdir.chdir()
+    flags = _setup_test_init(manifest_repo, local_dir, topdir, directory, manifest_file)
+    cmd_raises(['init'] + flags, expected_error)
```
pdgendt: Can we add an expected error message too? You can make it optional, but we should narrow down where relevant.

```suggestion
    _, stderr = cmd_raises(['init'] + flags, expected_error)
    if expected_error_message is not None:
        assert expected_error_message in stderr
```

thorsten-klein: Done

tests/test_project.py
thorsten-klein: Done

src/west/app/project.py
```diff
@@ -302,17 +367,27 @@ def local(self, args) -> Path:
...
+        try:
+            already = util.west_topdir(manifest_dir, fall_back=False)
+            self.die_already(already)
+        except util.WestNotFound:
+            pass
+
+        rel_manifest = manifest_dir.relative_to(topdir)
+        west_dir = topdir / WEST_DIR
+
+        self.banner('Initializing around existing manifest', rel_manifest)
```
```suggestion
        self.banner('Initializing with existing manifest', rel_manifest)
```

src/west/app/project.py
```diff
@@ -396,15 +485,17 @@ def bootstrap(self, args) -> Path:
...
+            yaml_path = manifest.yaml_path
         else:
             # We use PurePath() here in case manifest_url is a
             # windows-style path. That does the right thing in that
             # case, without affecting POSIX platforms, where PurePath
             # is PurePosixPath.
-            manifest_path = PurePath(urlparse(manifest_url).path).name
+            yaml_path = PurePath(urlparse(manifest_url).path).name
 
+        manifest_path: Path = Path(manifest_subdir) / yaml_path
```
pdgendt: No need for explicit typing?

```suggestion
        manifest_path = Path(manifest_subdir) / yaml_path
```

src/west/app/project.py
marc-hb: > The Zephyr repository is cloned to <topdir>/extra/level/zephyr/.git

No this does not match my expectations. For consistency with `west init -l topdir/extra/level/zephyr`, I'm expecting `west init --topdir topdir topdir/extra/level/zephyr` + an error or a warning when `self.path` is present and different from `extra/level/zephyr`. Long story short, all these arguments should always point at the manifest directory for consistency. Some arguments may be absolute, others like `self.path` must be relative but they should all point to the same directory in the end for simplicity. In case of a failure becaus of a conflict with `self.path`, the error message could require the users to 1. clone 2. west init in two steps.

Please prove me wrong and demonstrate that there is a more user-friendly UI design, but again it would have been more efficient to have this very discussion about the user interface and design first in #774 BEFORE writing the corresponding code, rather then "reserve-engineering" the code at code review time to discover that the UI design was not formalized.

This PR is 100% about nesting subdirectories, an area where "off by one" and similar errors are extremely common[*]. So it is critical to pay great attention to such nesting details. This is also why I repeatedly warned about phrasing like "thing gets created inside dir0" in the help text: does it mean "thing=dir0" or "dir0/thing"? No one knows. Less ambiguous language is required.

[*] examples:
- https://en.wikipedia.org/wiki/Tar_(computing)#Tarbomb
- https://unix.stackexchange.com/questions/402555/why-add-a-trailing-slash-after-an-rsync-destination
- Run `cp -r dir0/ dir1/` twice, starting with non-existent dir1. Results depend on the operating system!

marc-hb:
From https://github.com/zephyrproject-rtos/west/pull/854#discussion_r2706082963

> but again it would have been more efficient to have this very discussion about the user interface and design first BEFORE writing the corresponding code, rather then "reserve-engineering" the code at code review time to discover that the UI design was not formalized.

Also, it's very time-consuming and error prone to have high-level design discussions buried in in tens of comments about typos and other, low-level implementation details. Let's discuss user interface and other high-level design points in issue #774 rather than in this PR. #774 is a bit noisy because of some initial misunderstandings but it's not as noisy as here.

carlescufi:
> From [#854 (comment)](https://github.com/zephyrproject-rtos/west/pull/854#discussion_r2706082963)
> 
> > but again it would have been more efficient to have this very discussion about the user interface and design first BEFORE writing the corresponding code, rather then "reserve-engineering" the code at code review time to discover that the UI design was not formalized.
> 
> Also, it's very time-consuming and error prone to have high-level design discussions buried in in tens of comments about typos and other, low-level implementation details. Let's discuss user interface and other high-level design points in issue #774 rather than in this PR. #774 is a bit noisy because of some initial misunderstandings but it's not as noisy as here.

In an ideal world, sure. But I think in this particular case we are already too far ahead in the implementation and design to change the location of the discussions in my opinion. Reading your comments, I don't think we are that far from agreeing to a solution here, so @thorsten-klein perhaps we can try to resolve the last points (specifically [this comment](https://github.com/zephyrproject-rtos/west/pull/854#discussion_r2706082963)).

src/west/app/project.py
carlescufi: Let's try to figure out the actual disagreement here, step by step. First of all, let's use proper full commands. @marc-hb  and @thorsten-klein in the examples you are providing above with `west init —topdir` you are not including `-l` nor `-m`. Can we start by defining the full command-line arguments of the command that you @marc-hb think should work differently?

marc-hb:
> But I think in this particular case we are already too far ahead in the implementation and design to change the location of the discussions in my opinion

It's not really a "change" because all the `--topdir` user interface and high-level design discussion happened in #774 in the first place and the majority of it is still there. Even after this gets merged, someone wondering about the user interface would rather stick to #774 and not have to get lost in typos and other implementation details to find rationale(s).

BTW I'm writing "high-level design" as in "looking at the --help, not the code". This is only about defining what the command-line parameters mean, it's not advanced maths! It's just faster to discuss and define this before writing and reviewing the code, that's all.

>  I don't think we are that far from agreeing to a solution here

Agreed, and I also love being proven wrong. I'm only upset about the time-consuming way to get there.

EDIT, trying to be fair: @thorsten-klein did not initially intend to implement `--topdir` and cloning at the same time and graciously accepted to add that during the review. But that entire discussion should still have happened before writing any code.

src/west/app/project.py
marc-hb: Just found this, funny enough: https://interrupt.memfault.com/blog/practical_zephyr_west

> Notice that we no longer specify the current directory in the call to west init using “.”. In fact, the directory - optionally passed as the last argument to west init - is interpreted differently by West when using the --local flag or the -m arguments:
>
> With --local, the directory specifies the path to the local manifest repository.
Without the --local flag, the directory refers to the topdir and thus the folder in which to create the workspace (defaulting to the current working directory in this case).
> Awkward, but this approach is probably used due to legacy reasons. If no --local flag is used and no repository is specified using -m, West defaults to using the Zephyr repository.

src/west/app/project.py
```diff
@@ -326,8 +398,22 @@ def local(self, args) -> Path:
         return topdir
 
     def bootstrap(self, args) -> Path:
-        topdir = Path(abspath(args.directory or os.getcwd()))
-        self.banner('Initializing in', topdir)
+        manifest_subdir = Path()
+        if args.topdir:
+            topdir = Path(args.topdir)
+            if args.directory:
+                if not Path(args.directory).is_relative_to(topdir):
```
marc-hb: So you just removed `abspath()` from the earlier version ` if not Path(abspath(args.directory)).is_relative_to` following my comment in https://github.com/zephyrproject-rtos/west/pull/854#discussion_r2700537680 but that did not fix relative directory arguments. Now relative directory arguments all fail because `is_relative_to()` fails silently with relative arguments:
```python
>>> Path('d1/d2').is_relative_to('/')
False
>>> Path('/d1/d2').is_relative_to('/')
True
```
It's confusing to have relative arguments not being relative (!) but it's documented in Pathlib. There is a test coverage gap for this use case (and you changed that code without testing manually).


This raises a subtle question: are relative directory arguments relative to topdir, or relative to the current directory? I would say the former but I don't have a strong opinion. Either way it needs to be clearly specified.

src/west/app/project.py
```diff
@@ -326,8 +398,22 @@ def local(self, args) -> Path:
...
+                    self.die(
+                        f"directory '{args.directory}' must be relative "
+                        f"to west topdir '{args.topdir}'"
+                    )
+                manifest_subdir = Path(os.path.relpath(args.directory, topdir))
```
marc-hb: Please avoid `os.path` unless absolutely necessary. It's complicated enough to deal with one Python module (`Pathlib`), please don't mix both.

src/west/app/project.py
marc-hb: I could be wrong, but for `args.directory` consistency, I expected these two commands to converge to the exact same state. I just tested and they produce something different right now:

```bash
west init -l                      -t ~/tmp/topdir ~/tmp/topdir/extra/level/zephyr/
# ... should end in the same final state as
west init -m http://..../zephyr/  -t ~/tmp/topdir ~/tmp/topdir/extra/level/zephyr/
```

Once the behavior for absolute paths is finalized, there is the delicate question of relative paths. When args.directory is relative, what is it relative to?  Is it relative to topdir? Or is it relative to the current directory where the command is run? Either way, this command always fails right now:

```bash
cd ~/tmp/topdir/
west init -m  ....    -t ~/tmp/topdir extra/level
FATAL ERROR: directory 'extra/level' must be relative to west topdir '~/tmp/topdir'
```
That's because Pathlib is_relative_to() is not intuitive at all, see my other comment about that.

marc-hb (commented):
All the combinatorial explosion and complexity was clearly and vastly underestimated. I think the only way to get back on track is to follow the "top-down", more formal approach that should have been used in the first place:

1. Make an extensive list of all user input possibilities. NO CODE here.
2. Complete and unambiguous definition of all resulting behaviors that can immediately answer all questions like the ones above. NO CODE here.
3. "Tests first" approach with extensive coverage.
4. Code LAST (design first)

Steps 3. and 4. should be PAUSED until 1. and 2. are discussed and agreed: PLEASE DO NOT PUSH until they are. Prototyping is great. Reviewing prototypes while wrongly believing they are close to final has been crazy time-consuming, let's please stop?

thorsten-klein:
>west init -l                      -t ~/tmp/topdir ~/tmp/topdir/extra/level/zephyr/
>... should end in the same final state as
> west init -m http://..../zephyr/  -t ~/tmp/topdir ~/tmp/topdir/extra/level/zephyr/

I almost agree (only the last argument has to omit `/zephyr`):
```
west init -l                      -t ~/tmp/topdir ~/tmp/topdir/extra/level/zephyr
# ... should end in the same final state as
west init -m http://..../zephyr/  -t ~/tmp/topdir ~/tmp/topdir/extra/level
```

I have added a test to prove that the commands really result in the same final state (= same tree).

marc-hb:
> I almost agree (only the last argument has to omit /zephyr):
>
> ```
> west init -l                      -t ~/tmp/topdir ~/tmp/topdir/extra/level/zephyr
> # ... should end in the same final state as
> west init -m http://..../zephyr/  -t ~/tmp/topdir ~/tmp/topdir/extra/level
> ```

Can you give some reason(s) why you prefer this variant? Not only it make these two commands above inconsistent with each other, this also makes the second one inconsistent with `git clone`:
```
git clone http://..../zephyr/                   ~/tmp/topdir/extra/level
ls ~/tmp/topdir/extra/level/
west.yml README ...
```
Also, this makes it impossible to rename without an extra rename action after cloning (which is probably why `git clone` made that choice)

This is not an obscure implementation detail that could easily be changed later, this is literally defining the main user interface of this new feature. After a gazillion of prototypes force-pushed that I manually tested and reviewed :-((( Once that user interface choice is made and people start using this feature, there is no way back (except for a painful deprecation, best avoided).

> > Steps 3. and 4. should be PAUSED until 1. and 2. are discussed and agreed: PLEASE DO NOT PUSH until they are.

So, in answer to this request of mine, "I prefer this variant" without any elaboration was your only plain English reply AND you force-pushed at least twice, piling up on the already out of control "[evolog](https://docs.jj-vcs.dev/latest/cli-reference/#jj-evolog)" and untrackable progress[1,2]. Did you miss or misunderstand what I asked there or is it really impossible to change your approach?

[1] I did not look at these new force pushes yet. I bet even more will come.
[2] compounded by https://github.com/zephyrproject-rtos/zephyr/issues/39194

marc-hb:
> [1] I did not look at these new force pushes yet. I bet even more will come.

There will be more revisions because the latest version https://github.com/zephyrproject-rtos/west/commit/cb9714f5d038299e8fd0129323c69af5d5ec8db9 fails on Windows:

https://github.com/zephyrproject-rtos/west/actions/runs/22708722014/job/65841352565?pr=854

```
       flags = _setup_test_init(manifest_repo, local_dir, topdir, directory, manifest_file)
        _, stderr = cmd_raises(['init'] + flags, expected_error)
>       assert expected_stderr in stderr
E       assert "FATAL ERROR: directory 'extra/dir' must be inside west topdir" in "FATAL ERROR: directory 'extra\\dir' must be inside west topdir 'D:\\a\\west\\west\\check out\\somedir'\n"
```

As a coincidence, I just had to work on similar portability issues and found a way to test Windows without Windows:
- #923

thorsten-klein:
> Can you give some reason(s) why you prefer this variant

I do not prefer this as well. Until now I thought that this was existing behavior. But it seems to be not. 
So I agree that it should be as you suggest 👍🏻

marc-hb:
> Once the behavior for absolute paths is finalized, there is the delicate question of relative paths. When `args.directory` is relative, what is it relative to? Is it relative to topdir? Or is it relative to the current directory where the command is run? 
>
> `cd some_dir/`
> `west init  ....    -t ~/tmp/topdir manifest/dir`

Ping? Is it `some_dir/manifest/dir` or is it `~/tmp/topdir/manifest/dir`? The latter "looks" better but the former allows completion... Which one is more consistent with other and existing cases and how?

thorsten-klein:
I’ve separated the tests into individual commits so it’s easier to verify that existing behavior remains unchanged.
That said, I’m unsure how to proceed with this PR and would appreciate some guidance.

marc-hb:
> That said, I’m unsure how to proceed with this PR and would appreciate some guidance.

1. First things first: finish _defining the user interface_ (in the `help` string). I sincerely hope the question of what relative arguments are relative to is the last question that is lacking an answer, see my "ping" just above. Don't worry about the current test failures yet. Fixing them could be a waste of time before the user interface is completely defined. Especially fixing failures like "X should be inside Y" when relative paths have not been defined yet.

2. Try to "use it in anger" and try break it by interactively testing all the user inputs you can think of. This is how I kept  finding bugs, missing features and unspecified behaviors the whole time. Make a list of edge cases than need test coverage.

3. As a last step, look at existing test failures and add new tests for any missing coverage. I can probably investigate and help with Windows-specific failures but not before the user interface is completely defined.

tests/conftest.py
```diff
@@ -650,3 +650,15 @@ def check_proj_consistency(actual, expected):
     assert actual.clone_depth == expected.clone_depth
     assert actual.revision == expected.revision
     assert actual.west_commands == expected.west_commands
+
+
+def tree(path: Path, prefix="") -> str:
```
marc-hb, line 670: I think https://anytree.readthedocs.io/en/latest/intro.html provides this for free and it is already a Zephyr requirement.


Don't get me wrong: I am NOT suggesting to add new west dependency but only to add a new west TEST dependency

marc-hb:
> > I’ve tried with the code to fail if the topdir is already initialized (when using --local). However, it seems that some tests (e.g. test_update_some_with_imports) rely on the current behavior and perform a re-initialization of some already initialized workspace.

> I'm amazed that such "reinit" situations exist, is it still the case? If yes, can you elaborate and provide a simple reproduction? This looks like a serious bug. 

This was a `west init` bug (and inconsistency). Filed in #935 and fix submitted in #936

> Does it predate this PR? 

This bug looks like it has been there forever. No one noticed before you because:
- users probably don't re-initialize workspaces by mistake much.
- as long as you nothing had changed, the re-initialization was harmless

> Does it simply deletes the existing .west/ directory?

It resets two values in `.west/config` without even changing the rest.

> > What shall I do?

Help review #936 :-)

> It looks like this behavior might have been intentional, so I make it fail only when the new argument --topdir is used.

It wasn't intentional because it depended on the current directory `west init` was run from. That's too inconsistent to be intentional.

marc-hb:
Fix #936 merged, @thorsten-klein please rebase. 

Also left TODO:

> - Once the behavior for absolute paths is finalized, there is the delicate question of relative paths.

- Using `anytree` in the test code (if needed)

I will investigate Windows-specific test failure if any left after rebase.

tests/test_project.py
```diff
@@ -1876,6 +1871,261 @@ def test_init_local_with_manifest_filename(repos_tmpdir):
...
+    repos_tmpdir.chdir()
+    workspace = Path(repos_tmpdir) / 'ws'
+    zephyr = repos_tmpdir / 'repos' / 'zephyr'
+
+    # initial clone
+    cmd(['init', 'ws', '-m', zephyr])
+
+    # initialize west workspace again
+    _, stderr = cmd_raises(['init', 'ws', '-m', zephyr], SystemExit)
+    assert "FATAL ERROR: already initialized" in stderr
```
marc-hb, line 2112: De-duplicate using a constant like in the existing tests
```suggestion
    err_msg = "FATAL ERROR: already initialized"
    assert err_msg in stderr
```

tests/test_project.py
```diff
@@ -1876,6 +1871,261 @@ def test_init_local_with_manifest_filename(repos_tmpdir):
...
+    assert tree_l == tree_r
+
+
+def test_init_topdir_again(repos_tmpdir):
+    repos_tmpdir.chdir()
+    workspace = Path(repos_tmpdir) / 'ws'
+    zephyr = repos_tmpdir / 'repos' / 'zephyr'
+
+    # initial clone
+    cmd(['init', 'ws', '-m', zephyr])
```
marc-hb, line 2108: Could use `--topdir` here too for a little bit more test coverage?

tests/test_project.py
```diff
@@ -1876,6 +1871,261 @@ def test_init_local_with_manifest_filename(repos_tmpdir):
...
+
+    # initialize west workspace again
+    _, stderr = cmd_raises(['init', 'ws', '-m', zephyr], SystemExit)
```
marc-hb, line 2111: Does this one add coverage compared to the existing `init_again` test?

thorsten-klein:
Apologies for the delay on this PR 👎🏻 This PR got out of control with all those (outdated) comments, commits and conversations.
I've updated it now and I hope I have tackled most concerns.

To keep things moving, I'd be happy if we can merge it as it is (if there are no bigger doubts).
Future comments we can tackle in a follow-up PR.

Alternatively, I'd propose a fresh start of this whole PR, so I will close this PR and submit a new one (so review process starts from scratch).

marc-hb:
Thanks @thorsten-klein for not giving up, much appreciated.

> Alternatively, I'd propose a fresh start of this whole PR, so I will close this PR and submit a new one (so review process starts from scratch).

I think this is probably the best next step at this stage. We can (and should) still refer to this older PR if needed, it's not going to disappear! But it will be easier for the new PR to focus on the (hopefully) very few important things left to discuss.

Also, please, please focus on a "top-down" approach in such a new PR, with a very heavy focus on all the _user interface_ details, corner cases and help text. It really think a "bottoms-up" approach was the main factor drowning this PR, where I kept trying to "reverse-engineer" the (sometimes unintentional) behavior from the code. Incredibly time-consuming.

Notably this (hopefully last) user interface question: https://github.com/zephyrproject-rtos/west/pull/854#issuecomment-4003176128

> To keep things moving, I'd be happy if we can merge it as it is (if there are no bigger doubts). Future comments we can tackle in a follow-up PR.

This approach works only when we are 100% sure that such follow-up PRs are clearly scoped, small and self-contained/isolated. Not when we are still discussing _user interface_ details. So not in this particular case I'm afraid.

thorsten-klein:
> Also, please, please focus on a "top-down" approach in such a new PR

@marc-hb I'll open a new PR with the current commits as they are now. Ok?

marc-hb:
> @marc-hb I'll open a new PR with the current commits as they are now. Ok?

Thanks, that should be a good starting point.

EDIT:
- #953

PR #855, opened 2025-10-07
test_project.py: fix test not to require a 500MB download.

marc-hb:
Fix "test_init_with_clone_option_depth_one(repos_tmpdir)" to actually use the local repos_tmpdir fixture instead of requiring a network connection to github.com/zephyr and using it to download rougly 500 MB (as of today).

This saves considerable time on slow connnections and makes it possible to run the whole test suite offline (when all dependencies are already available).

Fixes commit 1007557b92cad ("Add test cases for west init -o= option")

This wasn't obvious and was missed in code review because github.com/zephyr is the default manifest for "west init".

(and yes this was found on a plane...)

cc: @thorsten-klein

### DIFF
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -1773,8 +1773,9 @@ def test_init_with_clone_option_depth_one(repos_tmpdir):
     # Test that 'west init -o=--depth=1' only clones depth 1
 
     west_tmpdir = repos_tmpdir / 'workspace'
+    mnft_url = str(repos_tmpdir / 'repos' / 'zephyr')
 
-    cmd(['init', '-o=--depth=1', west_tmpdir])
+    cmd(['init', '-o=--depth=1', '-o=--no-local', '-m', mnft_url, west_tmpdir])
     assert 1 == int(subprocess.check_output([GIT, 'rev-list', '--count', '--max-count=5', 'HEAD'],
                                             cwd=west_tmpdir / 'zephyr').decode().strip())
 


pdgendt (approved):

thorsten-klein (approved):

thorsten-klein:
Did you consider disconnecting internet access in CI during tests?
This could help to ensure that tests run offline and don’t download external resources (as you found out by accident now). 

For instance, I tried locally by setting `http_proxy` and `https_proxy` to an unavailable proxy:
```
http_proxy=x https_proxy=y uv run poe all
```

With this I get same result that `tests/test_project.py::test_init_with_clone_option_depth_one` fails.

Maybe there is some better (not so hacky) solution available in Github Actions to cut internet access?

marc-hb:
I found that `uv` has some  interesting `--offline` but it was not transitive, applying it only to itself. At least not transitive for this. Such "transitivity" is admittedly a very different challenge technically.

I found some pytest-offline plugin that blocks all socket usage. That should do the trick but it seemed orphaned? Then I stopped searching. I didn't search hard, maybe there is something. If there is something decent, then we should definitely use it in pytest configuration or CI.

marc-hb:
> I tried locally by setting http_proxy and https_proxy to an unavailable proxy:

Misclick sorry.

marc-hb:
> I tried locally by setting http_proxy and https_proxy to an unavailable proxy

Too hackish for a default, shared configuration; but good enough for `.github/workflows` IMHO!

marc-hb:
> Too hackish for a default, shared configuration; but good enough for .github/workflows IMHO!

- #861

PR #858, opened 2025-10-07
ci: Split 'all' into smaller tasks

pdgendt:
Run testing on the different supported versions while running the linter and type checks only on 1 combination.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -12,13 +12,27 @@ concurrency:
 
 jobs:
   build:
-    name: Build python tests
+    name: Run tasks
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         python-version: ['3.10', '3.11', '3.12', '3.13']
+        task: [test]
+        # The include map here is actually used to extend the matrix.
+        # By passing all keys used in the matrix we append new unique combinations.
+        # For more information see:
+        # https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-adding-configurations
+        include:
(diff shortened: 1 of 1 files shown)


mbolivar (approved):

.github/workflows/test.yml
```diff
@@ -12,13 +12,23 @@ concurrency:
...
-    name: Build python tests
+    name: Run tasks
     runs-on: ${{ matrix.os }}
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, macos-latest, windows-latest]
         python-version: ['3.10', '3.11', '3.12', '3.13']
+        task: [test]
+        include:
```
marc-hb, line 27: I find the ways to extend GH matrices super obtuse; I have to go back to the docs every time. Can you please use comment(s) to warn the reader about that? I mean this is one of those rare cases where readers are deceived into thinking "How hard to understand could this simple syntax be? I really don't need to spend time finding the documentation".

But yes they do. Otherwise they waste a lot more time.

- It's really not intuitive that `include` _extends_!
- It's not intuitive either that `test` does not get overwritten and still runs for 3.13.

So please mention these, and also strongly recommend to open the documentation. It's currently at
https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#example-adding-configurations but maybe just the section name in the doc is more "permanent" and enough.


Otherwise everything looks good.

pdgendt: Added a comment with a link to the docs

marc-hb (approved):

marc-hb:
EDIT: this comment was meant for (related) #861 

-----

Passing without the revert https://github.com/zephyrproject-rtos/west/actions/runs/18323255742/job/52181378153?pr=861

This will conflict with #858 which should go first. Will rebase.

PR #859, opened 2025-10-07
support running and debugging `pytest` for local tree

thorsten-klein:
### Why?
I'm not sure how tests for the local source tree are currently debugged.
In my setup, I’m unable to run `pytest` directly on the local tree, and debugging is difficult because it’s not possible to step through since `subprocess` is used in the current test setup.

### Proposed Changes
This change allows running `pytest` directly, instead of only being able to run it via `uv run poe test`.
`pytest` can test either the installed West package or the local source tree, depending on how it’s invoked.
This simplifies test execution and improves compatibility with `pytest` integrations (e.g. in IDEs).

With this setup, `pytest` is fully in charge for setting up the Python environment, so testing becomes straightforward:

```bash
# Test the installed West package
pytest

# Test the local source tree
pytest -o pythonpath=src
# or
PYTHONPATH=src pytest
```

### Background
During this work, I faced that many tests used `subprocess`, which interfered with testing the local copy.
Subprocesses do not inherit the full Python environment configured by `pytest`, causing a mix between modules from the installed West package and those from the local source tree. Therefore I removed the use of subprocesses in tests wherever possible. Where subprocesses are still used, the python module (`west/app/main.py`) is called instead of `west` whereby it is ensured now that its correct modules are used by prepending correct PYTHONPATH. With those changes only one Python environment managed by `pytest` is used, ensuring consistent behavior and enabling proper debugging.

### Additional infos
`cmd()` captures only Python-level stdout and does not include output from any subprocesses launched internally.

If the code under test starts one or more subprocesses and the test verifies this combined stdout, the older `cmd_subprocess()` behavior must be used, as this function captures all stdout, both from Python itself and from any spawned subprocesses.

To simplify invoking west with corresponding modules, the `main.py` script can now be executed directly. 
This approach is used in tests by `cmd_subprocess()`, but also end users can also invoke `main.py` directly in their local tree if needed.

### DIFF
--- a/src/west/app/main.py
+++ b/src/west/app/main.py
@@ -30,6 +30,13 @@ from typing import NamedTuple
 
 import colorama
 
+if __name__ == "__main__":
+    # Prepend the west src directory to sys.path so that running this script
+    # directly in a local tree always uses the according local 'west' modules
+    # instead of any installed modules.
+    src_dir = Path(__file__).resolve().parents[2]
+    sys.path.insert(0, os.fspath(src_dir))
+
 import west.configuration
 from west import log
 from west.app.config import Config
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -3,6 +3,7 @@
 # SPDX-License-Identifier: Apache-2.0
 
 import contextlib
+import io
 import os
 import platform
 import shutil
@@ -13,6 +14,8 @@ import uuid
 from pathlib import Path, PurePath
 
 import pytest
+
+from west.app import main
 
 GIT = shutil.which('git')
 
@@ -366,48 +369,74 @@ def check_output(*args, **kwargs):
     return out_bytes.decode(sys.getdefaultencoding())
 
 
-def cmd(cmd, cwd=None, stderr=None, env=None):
-    # Run a west command in a directory (cwd defaults to os.getcwd()).
-    #
-    # This helper takes the command as a string.
-    #
-    # This helper relies on the test environment to ensure that the
-    # 'west' executable is a bootstrapper installed from the current
-    # west source code.
-    #
-    # stdout from cmd is captured and returned. The command is run in
-    # a python subprocess so that program-level setup and teardown
-    # happen fresh.
+def _cmd(cmd, cwd=None, env=None):
+    # Executes a west command by invoking the `main()` function with the
+    # provided command arguments.
+    # Parameters:
+    #   cwd: The working directory in which to execute the command.
+    #   env: A dictionary of extra environment variables to apply temporarily
+    #        during execution.
 
-    # If you have quoting issues: do NOT quote. It's not portable.
-    # Instead, pass `cmd` as a list.
-    cmd = ['west'] + (cmd.split() if isinstance(cmd, str) else cmd)
+    # ensure that cmd is a list of strings
+    cmd = cmd.split() if isinstance(cmd, str) else cmd
+    cmd = [str(c) for c in cmd]
 
-    print('running:', cmd)
-    if env:
-        print('with non-default environment:')
-        for k in env:
-            if k not in os.environ or env[k] != os.environ[k]:
-                print(f'\t{k}={env[k]}')
-        for k in os.environ:
-            if k not in env:
-                print(f'\t{k}: deleted, was: {os.environ[k]}')
-    if cwd is not None:
-        cwd = os.fspath(cwd)
-        print(f'in {cwd}')
-    try:
-        return check_output(cmd, cwd=cwd, stderr=stderr, env=env)
-    except subprocess.CalledProcessError:
-        print('cmd: west:', shutil.which('west'), file=sys.stderr)
-        raise
+    # run main()
+    with (
+        chdir(cwd or Path.cwd()),
+        update_env(env or {}),
+    ):
+        try:
+            main.main(cmd)
+        except SystemExit as e:
+            if e.code:
+                raise e
+        except Exception as e:
+            print(f'Uncaught exception type {e}', file=sys.stderr)
+            raise e
 
 
-def cmd_raises(cmd_str_or_list, expected_exception_type, cwd=None, env=None):
-    # Similar to 'cmd' but an expected exception is caught.
-    # Returns the output together with stderr data
-    with pytest.raises(expected_exception_type) as exc_info:
-        cmd(cmd_str_or_list, stderr=subprocess.STDOUT, cwd=cwd, env=env)
-    return exc_info.value.output.decode("utf-8")
+def cmd(cmd: list | str, cwd=None, stderr: io.StringIO | None = None, env=None):
+    # Same as _cmd(), but it captures and returns combined stdout and stderr.
+    # Optionally stderr can be captured separately into given stderr.
+    # Note that this function does not capture any stdout or stderr from an
+    # internally invoked subprocess.
+    stdout_buf = io.StringIO()
+    stderr_buf = stderr or stdout_buf
+    with contextlib.redirect_stdout(stdout_buf), contextlib.redirect_stderr(stderr_buf):
+        _cmd(cmd, cwd, env)
+    return stdout_buf.getvalue()
+
+
+def cmd_raises(cmd: list | str, expected_exception_type, stdout=None, cwd=None, env=None):
+    # Similar to '_cmd' but an expected exception is caught.
+    # The exception is returned together with stderr.
+    # Optionally stdout is captured into given stdout (io.StringIO)
+    stdout_buf = stdout or sys.stdout
+    stderr_buf = io.StringIO()
+    with (
+        contextlib.redirect_stdout(stdout_buf),
+        contextlib.redirect_stderr(stderr_buf),
+        pytest.raises(expected_exception_type) as exc_info,
+    ):
+        _cmd(cmd, cwd=cwd, env=env)
+    return exc_info, stderr_buf.getvalue()
+
+
+def cmd_subprocess(cmd: list | str, *args, **kwargs):
+    # This function behaves similarly to `cmd()`, but executes the command in a
+    # separate Python subprocess, capturing all stdout output.
+    # The captured stdout includes both Python-level output and the output of
+    # any subprocesses spawned internally. This makes the function particularly
+    # useful in test cases where the code under test launches subprocesses and
+    # the combined stdout needs to be verified.
+    # The main drawback is that it cannot be debugged within Python, so it
+    # should only be used sparingly in tests.
+    cmd = cmd if isinstance(cmd, list) else cmd.split()
+    cmd = [sys.executable, main.__file__] + cmd
+    print('running (subprocess):', cmd)
+    ret = check_output(cmd, *args, **kwargs)
+    return ret
 
 
 def create_workspace(workspace_dir, and_git=True):
--- a/tests/test_alias.py
+++ b/tests/test_alias.py
@@ -2,10 +2,8 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
-import subprocess
-
 import pytest
-from conftest import cmd
+from conftest import cmd, cmd_raises
 
 
 @pytest.fixture(autouse=True)
@@ -50,10 +48,8 @@ def test_alias_infinite_recursion():
     cmd('config alias.test2 test3')
     cmd('config alias.test3 test1')
 
-    with pytest.raises(subprocess.CalledProcessError) as excinfo:
-        cmd('test1', stderr=subprocess.STDOUT)
-
-    assert 'unknown command "test1";' in str(excinfo.value.stdout)
+    exc, _ = cmd_raises('test1', SystemExit)
+    assert 'unknown command "test1";' in str(exc.value)
 
 
 def test_alias_empty():
@@ -62,10 +58,8 @@ def test_alias_empty():
     # help command shouldn't fail
     cmd('help')
 
-    with pytest.raises(subprocess.CalledProcessError) as excinfo:
-        cmd('empty', stderr=subprocess.STDOUT)
-
-    assert 'empty alias "empty"' in str(excinfo.value.stdout)
+    exc, _ = cmd_raises('empty', SystemExit)
+    assert 'empty alias "empty"' in str(exc.value)
 
 
 def test_alias_early_args():
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -1,3 +1,8 @@
+import runpy
+import sys
+from pathlib import Path
+
+import pytest
 from conftest import cmd, cmd_subprocess
 
 import west.version
@@ -21,3 +26,25 @@ def test_main():
 
     # output must be same in both cases
     assert output_subprocess.rstrip() == output_directly.rstrip()
+
+
+def test_module_run(tmp_path, monkeypatch):
+    actual_path = ['initial-path']
+
+    # mock sys.argv and sys.path
+    monkeypatch.setattr(sys, 'path', actual_path)
+    monkeypatch.setattr(sys, 'argv', ['west', '--version'])
+
+    # ensure that west.app.main is freshly loaded
+    sys.modules.pop('west.app.main', None)
+
+    # run west.app.main as module
+    with pytest.raises(SystemExit) as exit_info:
+        runpy.run_module('west.app.main', run_name='__main__')
+
+    # check that exit code is 0
+    assert exit_info.value.code == 0
+
+    # check that that the sys.path was correctly inserted
+    expected_path = Path(__file__).parents[1] / 'src'
+    assert actual_path == [f'{expected_path}', 'initial-path']
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -1,5 +1,4 @@
-import subprocess
-import sys
+from conftest import cmd, cmd_subprocess
 
 import west.version
 
@@ -11,7 +10,14 @@ def test_main():
     # sane (i.e. the actual version number is printed instead of
     # simply an error message to stderr).
 
-    output_as_module = subprocess.check_output([sys.executable, '-m', 'west', '--version']).decode()
-    output_directly = subprocess.check_output(['west', '--version']).decode()
-    assert west.version.__version__ in output_as_module
-    assert output_as_module == output_directly
+    expected_version = west.version.__version__
+
+    # call west executable directly
+    output_directly = cmd(['--version'])
+    assert expected_version in output_directly
+
+    output_subprocess = cmd_subprocess('--version')
+    assert expected_version in output_subprocess
+
+    # output must be same in both cases
+    assert output_subprocess.rstrip() == output_directly.rstrip()
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -10,6 +10,7 @@ import textwrap
 from pathlib import Path, PurePath
 
 import pytest
+import yaml
 from conftest import (
     GIT,
     WINDOWS,
@@ -19,6 +20,7 @@ from conftest import (
     check_proj_consistency,
     cmd,
     cmd_raises,
+    cmd_subprocess,
     create_branch,
     create_repo,
     create_workspace,
@@ -148,10 +150,10 @@ def test_list(west_update_tmpdir):
     abs_inside = cmd(_list_f('{name}') + [klib_abs], cwd=klib_abs).strip()
     assert abs_inside == 'Kconfiglib'
 
-    with pytest.raises(subprocess.CalledProcessError):
+    with pytest.raises(SystemExit):
         cmd('list NOT_A_PROJECT', cwd=klib_abs)
 
-    with pytest.raises(subprocess.CalledProcessError):
+    with pytest.raises(SystemExit):
         cmd('list NOT_A_PROJECT')
 
 
@@ -235,7 +237,7 @@ def test_list_groups(west_init_tmpdir):
         ['foo .foo-group-1,foo-group-2. foo inactive', 'bar .. path-for-bar active'],
     )
 
-    err_msg = cmd_raises('list -i foo bar', subprocess.CalledProcessError)
+    _, err_msg = cmd_raises('list -i foo bar', SystemExit)
     assert '-i cannot be combined with an explicit project list' in err_msg
 
     cmd('config manifest.group-filter +foo-group-1')
@@ -548,7 +550,7 @@ def test_compare(config_tmpdir, west_init_tmpdir):
     assert 'foo' in actual
 
     # --exit-code should work for the manifest repository too.
-    with pytest.raises(subprocess.CalledProcessError):
+    with pytest.raises(SystemExit):
         cmd('compare --exit-code')
 
     # Remove the file and verify compare output is empty again.
@@ -577,7 +579,7 @@ def test_compare(config_tmpdir, west_init_tmpdir):
     cmd('config -d manifest.group-filter')
 
     # Verify --exit-code works as advertised, and clean up again.
-    with pytest.raises(subprocess.CalledProcessError):
+    with pytest.raises(SystemExit):
         cmd('compare --exit-code')
     os.unlink(bar)
     assert cmd('compare --exit-code') == ''
@@ -639,7 +641,7 @@ def test_forall(west_init_tmpdir):
 
     # 'forall' with no projects cloned shouldn't fail
 
-    assert cmd(['forall', '-c', 'echo foo']).splitlines() == [
+    assert cmd_subprocess(['forall', '-c', 'echo foo']).splitlines() == [
         '=== running "echo foo" in manifest (zephyr):',
         'foo',
     ]
@@ -647,7 +649,7 @@ def test_forall(west_init_tmpdir):
     # Neither should it fail after cloning one or both projects
 
     cmd('update net-tools')
-    assert cmd(['forall', '-c', 'echo foo']).splitlines() == [
+    assert cmd_subprocess(['forall', '-c', 'echo foo']).splitlines() == [
         '=== running "echo foo" in manifest (zephyr):',
         'foo',
         '=== running "echo foo" in net-tools (net-tools):',
@@ -658,7 +660,7 @@ def test_forall(west_init_tmpdir):
 
     env_var = "%WEST_PROJECT_NAME%" if WINDOWS else "$WEST_PROJECT_NAME"
 
-    assert cmd(['forall', '-c', f'echo {env_var}']).splitlines() == [
+    assert cmd_subprocess(['forall', '-c', f'echo {env_var}']).splitlines() == [
         f'=== running "echo {env_var}" in manifest (zephyr):',
         'manifest',
         f'=== running "echo {env_var}" in net-tools (net-tools):',
@@ -666,7 +668,7 @@ def test_forall(west_init_tmpdir):
     ]
 
     cmd('update Kconfiglib')
-    assert cmd(['forall', '-c', 'echo foo']).splitlines() == [
+    assert cmd_subprocess(['forall', '-c', 'echo foo']).splitlines() == [
         '=== running "echo foo" in manifest (zephyr):',
         'foo',
         '=== running "echo foo" in Kconfiglib (subdir/Kconfiglib):',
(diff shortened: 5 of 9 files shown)


pdgendt:
I'm not sure I like where this is going. Testing `west` should focus on the package. I think we could add some documentation bits for people who _really, really_ want to use `pytest` directly that they should install an editable version with

```shell
$ python -m venv .venv
$ source .venv/bin/activate
$ pip install -e .
```

thorsten-klein:
Note: A few other tests invoke the hardcoded `west` executable directly, rather than using `cmd()`.  
As a result, they only work if `west` is installed and available in `PATH`.  

For consistency and robustness, these tests should be refactored so that pytest passes even when `west` is not installed.  
Would you like me to address this in the current PR, or handle it in a follow-up PR?

tests/conftest.py
```diff
@@ -13,7 +13,24 @@
...
+WEST_MAIN = main.__file__
+
+# installed west executable (only set if it is correct version)
+WEST_EXE = shutil.which('west')
+if WEST_EXE:
+    # ensure that installed west exe has expected version
+    expected = west.version.__version__
+    actual = subprocess.check_output([WEST_EXE,'--version'], text=True)
+    if expected not in actual:
+        WEST_EXE = None
```
pdgendt: This test isn't sufficient IMO, if you ran `pip install -e .` before, the installed `west` executable will probably be used, while not desired.

pdgendt (commented):
I recommend you check https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#tests-outside-application-code too.

thorsten-klein:
> I'm not sure I like where this is going. Testing west should focus on the package. I think we could add some documentation bits for people who really, really want to use pytest directly that they should install an editable version with

Installation in editable mode does not work for me:
```
ERROR: Project file:///home/user/work/GIT/west has a 'pyproject.toml' and
its build backend is missing the 'build_editable' hook. Since it does not have
a 'setup.py' nor a 'setup.cfg', it cannot be installed in editable mode.
Consider using a build backend that supports PEP 660.
```
Am I doing something wrong?

> I recommend you check https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#tests-outside-application-code too.

In this documentation it says:
```
If you don’t use an editable install and are relying on the fact that Python by default 
puts the current directory in sys.path to import your package, you can execute 
python -m pytest to execute the tests against the local copy directly, without using pip.
```

So this is exactly what I want to do. I want to run `python -m pytest` and execute the tests against the local copy.
When the tests are run in tox, then they run against the installed package.
Therefore I adapted the tests to prefer the installed package (if `west` in correct version exists), and otherwise fall back to the local copy (`main.py`). 

In many other Python open-source projects, it’s possible to run `pytest` directly, and I find that very convenient.

>  if you ran pip install -e . before, the installed west executable will probably be used, while not desired.

If the installed west version matches, then the tests should test against this installed package.
So when installed with `pip install -e .` the west version should always be up-to-date.

The main problem seems to lay in current tests that call `subprocess`, which does not inherit the `sys.path` from the test environment. Therefore pytest does not work as expected, since different west modules are imported in the test versus the called `west` process.

Running `pytest` directly becomes particularly useful when debugging in an IDE, where for example some paths from `~/.bashrc` are not set when IDE is started via GUI. 
Currently, I have to either start my IDE inside a prepared virtual environment or I have to install the tool I want to test into my IDE environment. That feels wrong. 

---

~Out of curiosity: Have you ever considered distinguishing between different test levels and handling them different (e.g., unit tests vs. integration tests)?~ (Does not matter)

tests/conftest.py
thorsten-klein: Maybe I get you wrong. 
But when installed with `pip install -e .` the installed west executable is up-to-date, so it is desired to use this?

pdgendt:
> Installation in editable mode does not work for me:
> 
> ```
> ERROR: Project file:///home/user/work/GIT/west has a 'pyproject.toml' and
> its build backend is missing the 'build_editable' hook. Since it does not have
> a 'setup.py' nor a 'setup.cfg', it cannot be installed in editable mode.
> Consider using a build backend that supports PEP 660.
> ```
> 
> Am I doing something wrong?

So what I do to test (requires an environment!):

```shell
$ python3.10 -m venv .venv
$ source .venv/bin/activate
$ pip install -e .
```

> In many other Python open-source projects, it’s possible to run `pytest` directly, and I find that very convenient.

I get that, but I also want to make sure developers are running the version they're expecting to run. Maybe we should check [`POE_ACTIVE`](https://poethepoet.natn.io/env_vars.html#internal-environment-variables) and in that case require an installed version.

> Running `pytest` directly becomes particularly useful when debugging in an IDE, where for example some paths from `~/.bashrc` are not set when IDE is started via GUI. Currently, I have to either start my IDE inside a prepared virtual environment or I have to install the tool I want to test into my IDE environment. That feels wrong.

Right, but I'd like to avoid having to mess with the system path if possible. There should be some recommended practices for this, no? There are so many packages out there, we can't be the only with this problem.

I guess that's why the myriad of task runners exist.

tests/conftest.py
pdgendt: It is, but the following would become confusing:

```shell
$ pip install -e .
# adjust some stuff in the code
$ pytest ...
```

You'd need to run `pip install -e .` every time you adjust something, but forgetting that will still happily run the tests with the previous installed one.

thorsten-klein:
> Right, but I'd like to avoid having to mess with the system path if possible. There should be some recommended practices for this, no? There are so many packages out there, we can't be the only with this problem.

Wouldn't it make sense to get rid of this subprocess completely and instead use `WestApp.run()`?
This is at least how conan seems to do it:
Ref:
- https://github.com/conan-io/conan/blob/develop2/test/integration/command/test_build.py (using `TestClient.run()` which is analog to west's `cmd()`)
- https://github.com/conan-io/conan/blob/develop2/conan/test/utils/tools.py#L569 (implementation of `TestClient.run` calls  `_run_cli()` )
- https://github.com/conan-io/conan/blob/develop2/conan/test/utils/tools.py#L553 (implementation of the `_run_cli` method without any `subprocess`)

With this, pytest can completely ensure which modules are tested, because there are no other processes that have different PYTHONPATH. 
This is also the benefit of the src layout, as described in your mentioned link (https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#tests-outside-application-code)

```
pytest # test installed package
PYTHONPATH=src pytest # test local package
```

The benefit of this would also be that the tests become fully debuggable. 
Currently. `subprocess` cannot be fully debugged, since the whole west logic happens in this other process, which is only one python call.

pdgendt:
> Wouldn't it make sense to get rid of this subprocess completely and instead use `WestApp.run()`? This is at least how conan seems to do it: Ref:

Yes please, but I guess that this is even more work than the current proposal?

> With this, pytest can completely ensure which modules are tested, because there are no other processes that have different PYTHONPATH. This is also the benefit of the src layout, as described in your mentioned link ([docs.pytest.org/en/7.1.x/explanation/goodpractices.html#tests-outside-application-code](https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#tests-outside-application-code))
> 
> ```
> pytest # test installed package
> PYTHONPATH=src pytest # test local package
> ```
> 
> The benefit of this would also be that the tests become fully debuggable. Currently. `subprocess` cannot be fully debugged, since the whole west logic happens in this other process, which is only one python call.

Getting rid of the `subprocess` would be nice, we can also get rid of the patch for coverage.

thorsten-klein:
I will give it a try 👍

mbolivar:
> Wouldn't it make sense to get rid of this subprocess completely and instead use `WestApp.run()`?

Just FYI that's technically an abstraction violation; nothing in `west.app` is API

pdgendt:
> > Wouldn't it make sense to get rid of this subprocess completely and instead use `WestApp.run()`?
> 
> Just FYI that's technically an abstraction violation; nothing in `west.app` is API

But that doesn't mean we can't rely on it for testing, right? From a practical point of view it would solve some of the issues we have with testing/coverage.

mbolivar:
No, it's just white-box testing

thorsten-klein:
I use `main.main()` now for testing, which takes `argv` as argument. This argv is the west command line API which should be stable, so it can be used for testing 😇

tests/conftest.py
thorsten-klein: I think with `pip install -e .` local changes are applied automatically. At least on Linux there is some symlink magic.

marc-hb:
> These changes allow running pytest directly, simplifying test execution and making debugging even easier.

Shorter commands are always nicer but it's IMHO not enough to justify a PR that big. Is there some bigger problem that this PR solves for you? I remember I used to struggle to use `pudb` with `tox` but now this seems enough for me:

```
uv run pytest -k narrow -s
```

marc-hb:
>  Is there some bigger problem that this PR solves for you? 

I scanned the comments and found a couple problem descriptions that should have been in commit messages. And then some more:

> > Wouldn't it make sense to get rid of this subprocess completely and instead use WestApp.run()? 

> Yes please, but I guess that this is even more work than the current proposal?

Prototyping and experimenting is great, but once a PR leaves the draft status it should be clear 1. what is the problem solved 2. why and how that solution was chosen.

For earlier stage discussions please use drafts, issues, discord,...

Some good inspiration:
https://github.com/zephyrproject-rtos/zephyr/blob/67435e394f0d025b8/.github/ISSUE_TEMPLATE/002_enhancement.yml

thorsten-klein:
> 1. what problem does this solve

As a developer, I want to easily run the project’s tests using `pytest` directly in my local environment.  
Although the `README.md` mentions that tests can be executed with `pytest`, this currently doesn’t work because certain prerequisites must be met (for example, `west` must be installed into the environment).
I understand that this setup may be a leftover from earlier times when `tox` was used, but now it should be possible to run `pytest` on the local copy without having to install the project I want to test.

This is matching to what pytest proposes under "Benefits".  
Ref: https://docs.pytest.org/en/7.1.x/explanation/goodpractices.html#tests-outside-application-code  

Debugging is also challenging under the current approach since many tests rely on `subprocess` calls.  
Those subprocesses run in a separate environment (potentially even with different module versions), which makes it difficult to inspect variables or step through the code during debugging.  

I’m not sure how you currently debug tests in your IDE, but most IDEs offer native `pytest` integration — something that doesn’t work well with the current test structure.

> 2. why and how this solution was chosen

The proposed change minimizes or eliminates the use of `subprocess` in tests, allowing them to run directly in the current Python process — which makes debugging far easier.

Additionally, it introduces flexibility by allowing developers to choose what is tested using standard `pytest` options:
```
# test the installed west 
pytest

# test the local copy 
pytest -o pythonpath=src
PYTHONPATH=src pytest
```

---

I hope these arguments are convincing.  

PS: Up to now, I’ve been developing `west` using only `print` statements for debugging — which has been quite painful.  
Often I had to use `print('test', file=sys.stderr)` so that I can see my output at all within tests, because of the redirected `subprocess` stdout. I’ve never been able to get proper debugging working in my IDE. 

That’s why I _really, really_ hope you’ll accept this Pull Request — it is state-of-the-art and it will make development and debugging so much easier.

tests/conftest.py
```diff
@@ -317,48 +361,53 @@ def check_output(*args, **kwargs):
...
+            with contextlib.redirect_stdout(stdout_buf), \
+                 contextlib.redirect_stderr(stderr or sys.stderr):
+                try:
+                    main.main(cmd)
+                except SystemExit as e:
+                    if e.code:
+                        raise e
+                except Exception as e:
+                    print(f'Uncaught exception type {e}', file=sys.stderr)
+                    raise e
```
pdgendt: Dedent this a bit more

```python
    with (
        chdir(cwd or Path.cwd()),
        update_env(env or {}),
        contextlib.redirect_stdout(stdout_buf),
        contextlib.redirect_stderr(stderr or sys.stderr),
    ):
        try:
            main.main(cmd)
        except SystemExit as e:
            if e.code:
                raise e
        except Exception as e:
            print(f'Uncaught exception type {e}', file=sys.stderr)
            raise e
```

tests/conftest.py
```diff
@@ -317,48 +361,53 @@ def check_output(*args, **kwargs):
...
+    # Returns the stderr data.
+    stderr_buf = io.StringIO()
     with pytest.raises(expected_exception_type) as exc_info:
-        cmd(cmd_str_or_list, stderr=subprocess.STDOUT, cwd=cwd, env=env)
-    return exc_info.value.output.decode("utf-8")
+        cmd(cmd_str_or_list, stderr=stderr_buf, cwd=cwd, env=env)
+    return exc_info, stderr_buf.getvalue()
+
+
+def cmd_all_stdout(cmd, cwd=None, stderr=None, env=None):
```
pdgendt: `stderr` and `env` aren't used?

pdgendt (commented):
Could you cleanup your commits a bit, please?

- Use the body to describe the change and its rationale
- Don't add stuff that is removed in the next commit (`tool.pytest.ini_options` in pyproject.toml)
- Split test commits from other changes

marc-hb:
> I’m not sure how you currently debug tests in your IDE, but most IDEs offer native pytest integration — something that doesn’t work well with the current test structure.

Thanks, I guess this is how this PR should have all started. Please write a more detailed commit message based on that and start a fresh, clean PR (referring to this one). So far this PR looks more like brainstorming. Brainstorming is great! It's just confusing when it happens in a non-draft PR.

> Could you cleanup your commits a bit, please?

See above :-)


> Debugging is also challenging under the current approach since many tests rely on subprocess calls.
Those subprocesses run in a separate environment (potentially even with different module versions), which makes it difficult to inspect variables or step through the code during debugging.

I suspect this is a different issue that requires a different solution. If this can be solved at a different time, then it should. Should it? Clearly state which way each PR is going.

FYI I've been using PUDB_TTY to avoid that problem
https://documen.tician.de/pudb/starting.html#debugging-from-a-separate-terminal

marc-hb:
> I’m not sure how you currently debug tests in your IDE, but most IDEs offer native pytest integration — something that doesn’t work well with the current test structure.

It's not "pytest integration" and you may not even call this an "IDE" but FWIW this works fine with Emacs right now:

```
Esc-x pdb
uv run pytest -k narrow --trace
```

It works even remotely over ssh/Tramp.

However this is only debugging the pytest code, NOT the actual west code for the subprocess reasons already discussed above.

pdgendt:
I'm having some trouble with getting this to work locally. My assumption was that I should now be able to do:

```shell
$ python3 -m venv .venv
$ source .venv/bin/activate
$ pip install -e .
$ pip install pytest
$ pytest
```

However this fail, for example with a single test: (same with `-o pythonpath=src`)
```shell
$ pytest tests/test_help.py::test_extension_help_and_dash_h
============================= test session starts ==============================
platform linux -- Python 3.12.11, pytest-8.4.2, pluggy-1.6.0
rootdir: /home/pdgendt/west
configfile: pyproject.toml
collected 1 item

tests/test_help.py F                                                     [100%]

=================================== FAILURES ===================================
________________________ test_extension_help_and_dash_h ________________________

west_init_tmpdir = local('/tmp/pytest-of-pdgendt/pytest-9/test_extension_help_and_dash_h0/workspace')

    def test_extension_help_and_dash_h(west_init_tmpdir):
        # Test "west help <command>" and "west <command> -h" for extension
        # commands (west_init_tmpdir has a command with one).
    
>       cmd('update')

/home/pdgendt/west/tests/test_help.py:28: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/home/pdgendt/west/tests/conftest.py:401: in cmd
    _cmd(cmd, cwd, env)
/home/pdgendt/west/tests/conftest.py:386: in _cmd
    raise e
/home/pdgendt/west/tests/conftest.py:383: in _cmd
    main.main(cmd)
/home/pdgendt/west/src/west/app/main.py:1205: in main
    app.run(argv or sys.argv[1:])
/home/pdgendt/west/src/west/app/main.py:284: in run
/home/pdgendt/west/src/west/app/main.py:588: in run_command
    self.run_builtin(args, unknown)
/home/pdgendt/west/src/west/app/main.py:702: in run_builtin
    self.cmd.run(args, unknown, self.topdir,
/home/pdgendt/west/src/west/commands.py:201: in run
    self.do_run(args, unknown)
/home/pdgendt/west/src/west/app/project.py:1208: in do_run
    self.update_all()
/home/pdgendt/west/src/west/app/project.py:1274: in update_all
    self.update(project)
/home/pdgendt/west/src/west/app/project.py:1522: in update
    self.update_submodules(project)
/home/pdgendt/west/src/west/app/project.py:1423: in update_submodules
    self.die(
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <west.app.project.Update object at 0x7a3b414c5e50>, exit_code = 1
args = ('Submodule status failed for project: Kconfiglib.',)

    def die(self, *args, exit_code: int = 1) -> NoReturn:
        '''Print a fatal error using err(), and abort the program.
    
        :param args: sequence of arguments to print.
        :param exit_code: return code the program should use when aborting.
    
        Equivalent to ``die(*args, fatal=True)``, followed by an attempt to
        abort with the given *exit_code*.'''
        self.err(*args, fatal=True)
        if self.verbosity >= Verbosity.DBG_EXTREME:
            raise RuntimeError("die with -vvv or more shows a stack trace. "
                               "exit_code argument is ignored.")
        else:
>           sys.exit(exit_code)
E           SystemExit: 1

/home/pdgendt/west/src/west/commands.py:525: SystemExit
---------------------------- Captured stdout setup -----------------------------
initializing session repositories in /tmp/pytest-of-pdgendt/pytest-9/session_repos0
Initialized empty Git repository in /tmp/pytest-of-pdgendt/pytest-9/session_repos0/Kconfiglib/.git/
[master (root-commit) 58e308a] initial c41dc662-4a8a-4334-9138-b53799fdadce
Initialized empty Git repository in /tmp/pytest-of-pdgendt/pytest-9/session_repos0/tagged_repo/.git/
[master (root-commit) e97846e] initial d762cb97-fd56-4abe-a5cb-1db47d829573
Initialized empty Git repository in /tmp/pytest-of-pdgendt/pytest-9/session_repos0/net-tools/.git/
[master (root-commit) 78efb82] initial baeb89cf-ef96-4f14-890d-525a290af83e
Initialized empty Git repository in /tmp/pytest-of-pdgendt/pytest-9/session_repos0/zephyr/.git/
[master (root-commit) 5368f34] initial f7d2e202-c0a3-435a-8bbe-cb6ac04b1adc
[master f340070] base zephyr commit
 3 files changed, 2 insertions(+)
 create mode 100644 CODEOWNERS
 create mode 100644 include/header.h
 create mode 100644 subsys/bluetooth/code.c
[zephyr 7e0c488] test kconfiglib commit
 1 file changed, 1 insertion(+)
 create mode 100644 kconfiglib.py
[master ef483c3] tagged_repo commit
 1 file changed, 1 insertion(+)
 create mode 100644 test.txt
[master 3375f40] test net-tools commit
 3 files changed, 18 insertions(+)
 create mode 100644 qemu-script.sh
 create mode 100644 scripts/test.py
 create mode 100644 scripts/west-commands.yml
finished initializing session repositories
[master 0f868d9] add manifest
 1 file changed, 26 insertions(+)
 create mode 100644 west.yml
---------------------------- Captured stderr setup -----------------------------
Switched to branch 'zephyr'
Cloning into 'Kconfiglib'...
done.
Cloning into 'tagged_repo'...
done.
Cloning into 'net-tools'...
done.
Cloning into 'zephyr'...
done.
Cloning into '/tmp/pytest-of-pdgendt/pytest-9/test_extension_help_and_dash_h0/workspace/.west/manifest-tmp'...
done.
----------------------------- Captured stderr call -----------------------------
Cloning into bare repository '/home/pdgendt/.cache/zephyrproject/Kconfiglib/0aa789fdcd0cd5b2a14b654f054b65ae'...
done.
Cloning into '/tmp/pytest-of-pdgendt/pytest-9/test_extension_help_and_dash_h0/workspace/subdir/Kconfiglib'...
done.
From /home/pdgendt/.cache/zephyrproject/Kconfiglib/0aa789fdcd0cd5b2a14b654f054b65ae
 * branch            zephyr     -> FETCH_HEAD
HEAD is now at 7e0c488 test kconfiglib commit
FATAL ERROR: Submodule status failed for project: Kconfiglib.
=========================== short test summary info ============================
FAILED tests/test_help.py::test_extension_help_and_dash_h - SystemExit: 1
============================== 1 failed in 0.24s ===============================
```

Running with `uv` also results in the same errors.

EDIT: I actually have this on `main` 🤔 need to look into this.

pdgendt:
> Running with `uv` also results in the same errors.
> 
> EDIT: I actually have this on `main` 🤔 need to look into this.

@thorsten-klein see #862

tests/test_alias.py
```diff
@@ -50,10 +48,9 @@ def test_alias_infinite_recursion():
     cmd('config alias.test2 test3')
     cmd('config alias.test3 test1')
 
-    with pytest.raises(subprocess.CalledProcessError) as excinfo:
-        cmd('test1', stderr=subprocess.STDOUT)
-
-    assert 'unknown command "test1";' in str(excinfo.value.stdout)
+    with pytest.raises(SystemExit) as excinfo:
+        cmd('test1')
+    assert 'unknown command "test1";' in str(excinfo.value.code)
```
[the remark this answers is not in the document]
pdgendt: This works as expected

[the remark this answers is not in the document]
thorsten-klein: I was suprised as well that in case of SystemExit the exception.value.code is a string with the message, not the exist code. Is it related to how west raises the SystemExit?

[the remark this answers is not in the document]
pdgendt: The exit code has the following type:
```python
_ExitCode: TypeAlias = str | int | None
```

So a string is valid.

pdgendt (approved):

thorsten-klein:
Gentle ping :-) 

I have adopted the PR description to explain the "Why".

Is there anything else you’d like me to change on this branch?
I’d really appreciate if this PR could be merged soon, as it allows me to fully debug the tests.

pdgendt:
Needs a rebase

thorsten-klein:
> Needs a rebase

Done 👍🏻

carlescufi (approved):
This looks reasonable to me altogether. I don't love the addition to `main.py`, but I have read the comments and it seems the better way of handling this.

tests/conftest.py
```diff
@@ -366,48 +369,69 @@ def check_output(*args, **kwargs):
...
+    return exc_info, stderr_buf.getvalue()
+
+
+def cmd_subprocess(cmd, *args, **kwargs):
+    # This function is similar to `cmd()`, but runs the command in a separate
+    # Python subprocess. This ensures that both Python-level stdout and any
+    # subprocess output invoked by `main` are captured together in a single
+    # string and also, that program-level setup and teardown in `main` happen
+    # fresh for each call.
+    cmd = cmd.split() if isinstance(cmd, str) else cmd
```
marc-hb:
```suggestion
   cmd = cmd if isinstance(cmd, list) else cmd.split()
```
2 reasons:
- more obvious that `list` is the preferred option
- if `cmd` is neither list or str by mistake, fail faster.

tests/conftest.py
```diff
@@ -366,48 +369,69 @@ def check_output(*args, **kwargs):
...
+        _cmd(cmd_str_or_list, cwd=cwd, env=env)
+    return exc_info, stderr_buf.getvalue()
+
+
+def cmd_subprocess(cmd, *args, **kwargs):
+    # This function is similar to `cmd()`, but runs the command in a separate
+    # Python subprocess. This ensures that both Python-level stdout and any
+    # subprocess output invoked by `main` are captured together in a single
+    # string and also, that program-level setup and teardown in `main` happen
+    # fresh for each call.
```
marc-hb: Please mention briefly that this is best avoided because it makes it impossible or harder to use a debugger, etc. That's the one of the main purposes of this PR!

tests/conftest.py
```diff
@@ -366,48 +369,69 @@ def check_output(*args, **kwargs):
...
+    cmd = cmd.split() if isinstance(cmd, str) else cmd
+    cmd = [sys.executable, main.__file__] + cmd
+    print('running (subprocess):', cmd)
```
marc-hb, line 437: Naive question sorry: isn't there some logging function around here?

tests/conftest.py
```diff
@@ -366,48 +369,69 @@ def check_output(*args, **kwargs):
...
+    ):
+        _cmd(cmd_str_or_list, cwd=cwd, env=env)
+    return exc_info, stderr_buf.getvalue()
+
+
+def cmd_subprocess(cmd, *args, **kwargs):
+    # This function is similar to `cmd()`, but runs the command in a separate
+    # Python subprocess. This ensures that both Python-level stdout and any
+    # subprocess output invoked by `main` are captured together in a single
+    # string and also, that program-level setup and teardown in `main` happen
```
marc-hb: `stderr` and `stdout` always race with each other, so this is a bad idea. Was it already like this? Either way, this deserves an elaboration: explaining that it is generally bad, then explaining why it is not too bad in this particular context.

tests/test_project.py
```diff
@@ -2279,19 +2306,23 @@ def do_run(self, args, ignored):
     ]
 
     # Expect output from the built-in command, not its Kconfiglib duplicate.
-    actual = cmd('list zephyr -f {name}', stderr=subprocess.STDOUT).splitlines()
-    assert actual == expected_warns + ['manifest']
+    stderr = io.StringIO()
+    actual = cmd('list zephyr -f {name}', stderr=stderr).splitlines()
+    assert actual == ['manifest']
+    assert stderr.getvalue().splitlines() == expected_warns
```
marc-hb: This split between stdout and stderr looks unrelated to this commit. Can you please submit it in a separate PR?

src/west/app/main.py
```diff
@@ -30,6 +30,12 @@
 
 import colorama
 
+if __name__ == "__main__":
+    # Prepend the west src directory to sys.path so that running this script
+    # always uses the local 'west' modules instead of any installed ones.
+    src_dir = Path(__file__).resolve().parents[2]
+    sys.path.insert(0, os.fspath(src_dir))
```
marc-hb, line 38: The comment must explain what is the effect of this when people run `west` from a `pip3` `pipx`/`uv`/`uvx` install or from their distro.

`sys.path` is security-sensitive, all use cases must be carefully considered.

You made this a separate commit. I love smaller commits and smaller PRs [*] but I wonder what happens when I try to use the previous, big refactoring without this commit?

[*] https://docs.zephyrproject.org/latest/contribute/contributor_expectations.html#defining-smaller-prs

tests/test_main.py
marc-hb, line 1: Please correct me: this 3rd commit and test addition is only to test the previous, `sys.path` commit? In other words, this is test code to test... code required for testing? Test coverage is great but this feels overkill... if the previous commit is really required for testing, then it's already going to be covered regularly, no? If not, then maybe that previous commit is not required that much.

Please correct me.

marc-hb (dismissed):
I have no problem with the general idea but I have a few requests, see below.

Also, as the original author, I think we really need @mbolivar to approve the general idea. Not to review every line but at least approve the conceptual changes.

Finally, I didn't have time to check every code change yet, sorry.

tests/conftest.py
thorsten-klein: I don't think so. There are print statements at other places as well in `conftest.py`.

tests/conftest.py
thorsten-klein: It was always this case, and I think the "racing" is not problematic as west does not invoke async subprocesses.
I will adapt the text little bit to point out the rare case when this function shall be used.

tests/test_project.py
thorsten-klein: I have adapter the refactored `cmd()` function to capture combined stdout and stderr.

src/west/app/main.py
thorsten-klein: There is no effect when people run `west`. As the comment says it has only an effect when `running this script`. I will adapt the comment little bit to point out that it only affects if `running this script directly`

tests/test_main.py
thorsten-klein: Yes, this is just a test for the second commit.

Without the second commit, the `cmd_subprocess` does not use the correct local modules.

I am not sure if I see this as a test code. With this change also users can invoke west by running the `main.py` script directly. No need for them to install west or setting `PYTHONPATH`.

I will add this to the PR description.

pdgendt:
> I have no problem with the general idea but I have a few requests, see below.
> 
> Also, as the original author, I think we really need @mbolivar to approve the general idea. Not to review every line but at least approve the conceptual changes.

Ping @mbolivar, we have the necessary approvals, but I agree with @marc-hb, I'd like to know your opinion.

mbolivar:
@pdgendt if you're happy with it, I don't object. I did it this way in large part because of #149 -- the global state made it really tricky to try to do everything from one pytest process

marc-hb:
Is #149 not a problem anymore? AFAICT it's still open because of `west.log`.

mbolivar:
> Is #149 not a problem anymore? AFAICT it's still open because of `west.log`

You're going to hate this answer...

At least where tests are concerned, west.log isn't a problem, because we have no tests for it in west/tests :)

marc-hb:
Thanks! Considering it's deprecated, fine by me. It bet its logic was not very complicated and not crying for test coverage either.

src/west/app/main.py
marc-hb: This is apparently not enough, see fixup:
- #883

PR #860, opened 2025-10-07
Move west documentation to west repo

pdgendt:
WIP


doc/Makefile
```diff
@@ -0,0 +1,20 @@
+# Minimal makefile for Sphinx documentation
+#
+
+# You can set these variables from the command line, and also
+# from the environment for the first two.
+SPHINXOPTS    ?=
+SPHINXBUILD   ?= sphinx-build
```
mbolivar: I assume you'll rewrite this in CMake at some point?

mbolivar (commented):
Cool! I saw y'all had plans about this in @carlescufi's talk at ZDS, excited to see it.

doc/Makefile
pdgendt: Not sure if it is really worth it, the `make` command is supported by sphinx and a simple wrapper.

marc-hb: The super annoying thing with all build systems for C/C++ is their (apparently unavoidable) TWO steps:
- configure what you want with an often very long list of arguments
- build

I've seen a lot of wrappers and other creative ways to... make that a more user-friendly experience. To be honest, a very short makefile like this is very far from the worst solution. Old-style makefiles have a lot of flaws, but funny enough they're still great for providing a simple "make what_I_want" style of CLI interface.

Granted, depending on Make isn't great on Windows but: 1) Make on Windows should still be possible 2) as in most projects it should always be possible to bypass this convenience wrapper (it is possible now) 3) how many doc contributors use Windows? Sorry, I mean: how many doc contributors? 4) one change at a time?

PS: I have not looked at the PR yet.

pdgendt: The file was generated with
```shell
$ sphinx-quickstart
```

I think we can even drop the Makefile, and run the actual `sphinx` build command in CI. But then again, it's nice to have `make html` at my fingertips.

mbolivar: Yeah, I'm certainly not going to die on the hill that we need to be able to build these docs on windows, but that is why the main zephyr docs build system was rewritten in cmake from the makefile that was autogenerated by sphinx-quickstart back in ancient times.

doc/index.rst
```diff
@@ -0,0 +1,55 @@
+.. _west:
+
+West (Zephyr's meta-tool)
```
mbolivar: If part of the reason for this is that west is big enough to deserve its own documentation separate from zephyr, perhaps it's finally time to revisit this tag line.

Personally, I think "West: an extensible workspace manager" is more descriptive for the standalone tool itself. What do you think @marc-hb @pdgendt ?

marc-hb:
We do not have to lose the history, attribution, git blame etc. See below.

The `--allow-unrelated-histories` concept is a bit "advanced" but I don't think that bar is too high for a project the purpose of which is to manage git repositories :-)

```shell
pipx install git-filter-repo # or https://github.com/newren/git-filter-repo/ - I used today's main which I already had for other reasons.
cd /tmp
git clone zephyr
git clone west

cd zephyr/
time ~/git-filter-repo/git-filter-repo --path doc/develop/west --path doc/guides/west --path doc/tools/west --path doc/west 
# => less than 10 seconds

git mv doc/develop/west doc/; git commit ...

cd ../west
du -sh .git/ # 21M
git fetch ../zephyr/ main
du -sh .git/ # 21M => near zero disk cost. Only 217 commits with just plain text docs.
git merge --allow-unrelated-histories FETCH_HEAD
```

marc-hb:
> git merge --allow-unrelated-histories FETCH_HEAD

Just in case anyone finds this git feature scary: I did exactly this in  https://github.com/thesofproject/sof/commit/fd497c906a41 and https://github.com/thesofproject/sof/pull/8297 and there was no issue whatsoever. Most people did not even notice.

doc/_static/css/custom.css
```diff
@@ -0,0 +1,1119 @@
...
+/* Navigation bar logo and search */
+
+.logo {
+    opacity: var(--logo-opacity);
+}
+
+.wy-side-nav-search > a img.logo {
+    /* Fixed size to prevent reflows and support hiDPI displays */
+    /* A 5 pixel margin is added on each side. The logo itself displays at 200×100 at 100% scaling. */
+    width: 210px;
```
kartben: logo is distorted otherwise - i think just the fixed height is enough, width is auto.
prob need to apply the same cleanup in docs.zephyrproject.org

```suggestion
```

Issue #615, opened 2022-12-21
Allow modifying remote/project urls of imported manifests


jobasto:
### Introduction
We use west to checkout all source files necessary for building a firmware and use the import feature as pictured below. This way we don't need to specify the exact version of the modules we want to use because that information is taken from the imported manifests.
```
project_repository (internal git server)
├─ west.yml
    ⇓ imports (filtered by name-allowlist)
    company_library_repository (internal git server)
    ├─ west.yml
        ⇓ imports
        zephyr_repository (https://github.com/zephyrproject-rtos/zephyr)
        ├─ west.yml
             ⇓ imports
             zephyr_module_1 (https://github.com/zephyrproject-rtos/zephyr_module_1)
             zephyr_module_2 (https://github.com/zephyrproject-rtos/zephyr_module_2)
             zephyr_module_3 (https://github.com/zephyrproject-rtos/zephyr_module_3)
```
### Requirement
To ensure long term reproducibility we want to mirror all externally hosted repositories on our internal git server and use this mirror server for cloning.

#### User/system specific git config
One way to achieve this would be to use the git [url.\<base\>.insteadOf](https://git-scm.com/docs/git-config#Documentation/git-config.txt-urlltbasegtinsteadOf) config option (see also https://github.com/zephyrproject-rtos/west/issues/28#issuecomment-478816075). However, this option is user and/or system specific but we want that the actually used servers are deducible by examining the project_repository and its imports.
#### Specifying git config options in manifest
Another way would be to introduce a new manifest file key that allows setting arbitrary git config options. However, west probably shouldn't set user or system wide config options so it would have to set the config option in all imported repositories.
#### Adding a `rewrite-url` key to the manifest
Probably the cleanest way would be to introduce a new key to the manifest file, e.g. `rewrite-url`. An example could look like
```
manifest:
  defaults:
    remote: zephyr
    rewrite-url:
      - from: "default from regexp"
        to: "default to regexp "

  remotes:
    - name: zephyr
      url-base: https://github.com/zephyrproject-rtos

  projects:
    - name: zephyr
      revision: xxxxxx
      path: modules/zephyr
      import:
        rewrite-url:
          - from: "from regexp 1"
            to: "to regexp 1"
          - from: "from regexp 2"
             to: "to regexp 2"
    - name: module
      revision: yyyyyyyyyyy
      path: modules/lib/module
      import:
         true # uses default rewrite-url
```
[...]

PR #863, opened 2025-10-09
app: cache: Expand tilde to user home in cache paths

pdgendt:
Allow passing a cache path starting with a '~' and expand to the user's home directory.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -2352,7 +2352,7 @@ class Grep(_ProjectCommand):
 
         config_option = self.config.get(f'grep.{tool}-path')
         if config_option:
-            return config_option
+            return str(expand_path(config_option))
 
         if tool == 'git-grep':
             self.die_if_no_git()
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -36,6 +36,7 @@ from west.manifest import (
     _manifest_content_at,
 )
 from west.manifest import is_group as is_project_group
+from west.util import expand_path
 
 #
 # Project-related or multi-repo commands, like "init", "update",
@@ -1833,14 +1834,14 @@ class Update(_ProjectCommand):
         # Note: The url is hashed and used as a subfolder to accomodate
         # changes in the manifest.
         subdir_hash = hashlib.md5(project.url.encode('utf-8')).hexdigest()
-        return os.fspath(Path(self.auto_cache) / basename(project.url) / subdir_hash)
+        return os.fspath(expand_path(self.auto_cache) / basename(project.url) / subdir_hash)
 
     def project_cache(self, project):
         # Find the absolute path to a pre-existing local clone of a project
         # and return it. If the search fails, return None.
 
         if self.name_cache is not None:
-            maybe = Path(self.name_cache) / project.name
+            maybe = expand_path(self.name_cache) / project.name
             if maybe.is_dir():
                 self.dbg(
                     f'found {project.name} in --name-cache {self.name_cache}',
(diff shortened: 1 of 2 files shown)


thorsten-klein (approved):
Makes totally sense!

marc-hb (commented):
This looks useful but will users get confused because it expands and works in one place but not in others?

PS: `west` aside, I've seen people frequently confused when `~` does not expand. Because they don't realize it's generally done in the shell and not at the OS or filesystem level.

mbolivar:
> This looks useful but will users get confused because it expands and works in one place but not in others?

yeah I tend to agree this might lead to a game of whack-a-mole...

pdgendt:
> > This looks useful but will users get confused because it expands and works in one place but not in others?
> 
> yeah I tend to agree this might lead to a game of whack-a-mole...

Looking at https://docs.zephyrproject.org/latest/develop/west/config.html#built-in-configuration-options, I think these are the only configuration options where it makes sense?

The reason why I find this feature compelling is when dotfiles are shared between multiple devices. For example your home and work computers, where the username can be different, but you want to apply the same logic of caching locations.

thorsten-klein:
> The reason why I find this feature compelling is when dotfiles are shared between multiple devices. For example your home and work computers, where the username can be different, but you want to apply the same logic of caching locations.

I’m currently using `/var/tmp/west/auto-cache`, since it’s the only directory outside of `~` where users have write permissions by default. So this path can be used in a shared config 👍

Out of curiosity: How do you share your config file between multiple devices?

pdgendt:
> I’m currently using `/var/tmp/west/auto-cache`, since it’s the only directory outside of `~` where users have write permissions by default. So this path can be used in a shared config 👍

Sure, but the `~/.cache` directory is a common used place for that, and it makes it easy to do a cleanup every once in a while.

```shell
$ rm -rf ~/.cache/*
```

> Out of curiosity: How do you share your config file between multiple devices?

I followed https://www.atlassian.com/git/tutorials/dotfiles, there are maybe fancier ways, but it works 🤷

mbolivar:
> Looking at https://docs.zephyrproject.org/latest/develop/west/config.html#built-in-configuration-options, I think these are the only configuration options where it makes sense?

to be clear, I'm not trying to block anything!

carlescufi (approved):
Thanks!

marc-hb (changes requested):
> > This looks useful but will users get confused because it expands and works in one place but not in others?

> Looking at https://docs.zephyrproject.org/latest/develop/west/config.html#built-in-configuration-options, I think these are the only configuration options where it makes sense?

There's also `zephyr.base` and `grep.tool`. It would also make sense in `WEST_CONFIG_LOCAL` too.  `url-base` in the manifest is debatable.

It could make sense in some future https://github.com/zephyrproject-rtos/west/issues/615 too.

Last but not least, all `west` extensions could want to be consistent and do the same expansion.

===> So, instead of "manually hard-coding" `.expanduser()` in 3 places, I think there should be at very least be some sort of centralized and re-usable `west_path_expand()` level of indirection.

pdgendt:
> There's also `zephyr.base` and `grep.tool`. It would also make sense in `WEST_CONFIG_LOCAL` too. `url-base` in the manifest is debatable.

`zephyr.base` is relative to the `topdir`, so expanding doesn't really make sense. There's also the environment variable cases, but I didn't want to complicate it for now.
Also note that `zephyr.base` is some remaining coupled west <-> zephyr infrastructure, that will be removed in West v2.

> It could make sense in some future #615 too.
> 
> Last but not least, all `west` extensions could want to be consistent and do the same expansion.
> 
> ===> So, instead of "manually hard-coding" `.expanduser()` in 3 places, I think there should be at very least be some sort of centralized and re-usable `west_path_expand()` level of indirection.

Added a utility function `west.util.expand_path` that can be adopted/used by extensions.

marc-hb (approved):

marc-hb (approved):

PR #864, opened 2025-10-09
avoid using actual user configurations during tests

### LINKED ISSUE #862: Running test suite accesses user and system configuration files
For example with

```shell
$ cat ~/.westconfig
[update]
auto-cache = /home/pdgendt/.cache/west
```

The testing fails when running

```shell
$ uv run pytest
```

thorsten-klein:
Fixes #862 

A new pytest fixture is added which is automatically applied (`autouse=True`). 
This fixture sets the west environment variables `WEST_CONFIG_SYSTEM` and `WEST_CONFIG_GLOBAL` during each test, in order to prevent it from using or modifying the user's actual global or system configuration.

Note: The environment variables are inherited from `subprocess`, so that also the `subprocess` calls do not use or touch the user's real configs.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -2,6 +2,7 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 
+import contextlib
 import os
 import platform
 import shutil
@@ -12,8 +13,6 @@ import uuid
 from pathlib import Path, PurePath
 
 import pytest
-
-from west import configuration as config
 
 GIT = shutil.which('git')
 
@@ -62,7 +61,56 @@ manifest:
 WINDOWS = platform.system() == 'Windows'
 
 #
-# Test fixtures
+# Contextmanager
+#
+
+
+@contextlib.contextmanager
+def update_env(env: dict[str, str | None]):
+    """
+    Temporarily update the process environment variables.
+    This context manager updates `os.environ` with the key-value pairs
+    provided in the `env` dictionary for the duration of the `with` block.
+    The existing environment is preserved and fully restored when the block
+    exits. If the value is set to None, the environment variable is unset.
+    """
+    env_bak = dict(os.environ)
+    env_vars = {}
+    for k, v in env.items():
+        # unset if value is None
+        if v is None and k in os.environ:
+            del os.environ[k]
+        # set env variable to new value only if v is not None
+        elif v is not None:
+            env_vars[k] = v
+    # apply the new environment
+    os.environ.update(env_vars)
+    try:
+        yield
+    finally:
+        # reset to previous environment
+        os.environ.clear()
+        os.environ.update(env_bak)
+
+
+@contextlib.contextmanager
+def chdir(path):
+    """
+    Temporarily change the current working directory.
+    This context manager changes the current working directory to `path`
+    for the duration of the `with` block. After the block exits, the
+    working directory is restored to its original value.
+    """
+    oldpwd = os.getcwd()
+    os.chdir(path)
+    try:
+        yield
+    finally:
+        os.chdir(oldpwd)
+
+
+#
+# Test fixtures (autouse=True)
 #
 
 
@@ -88,6 +136,35 @@ def _check_git_capabilities(tmpdir_factory):
         GIT_INIT_HAS_BRANCH = True
     except subprocess.CalledProcessError:
         pass
+
+
+@pytest.fixture(autouse=True)
+def independent_global_and_system_config(tmpdir_factory):
+    # Fixture to ensure that the user's actual global and system configuration
+    # files are neither used nor touched during test.
+    #
+    # We also set ZEPHYR_BASE (to avoid complaints in subcommand
+    # stderr), but to a spurious location (so that attempts to read
+    # from inside of it are caught here).
+    #
+    tmpdir = Path(tmpdir_factory.mktemp("test-configs"))
+
+    # config paths
+    system = tmpdir / 'config.system'
+    glbl = tmpdir / 'config.global'
+
+    # run with environment variables set
+    with update_env({
+        'WEST_CONFIG_SYSTEM': str(system),
+        'WEST_CONFIG_GLOBAL': str(glbl),
+        'ZEPHYR_BASE': str(tmpdir / 'no-zephyr-here'),
+    }):
+        yield
+
+
+#
+# Test fixtures
+#
 
 
 @pytest.fixture(scope='session')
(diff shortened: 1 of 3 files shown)


thorsten-klein:
@pdgendt We could also create some invalid user config in the CI system (at the default path) to ensure that this corrupt config is never used in CI. Otherwise tests could potentially touch real configs again in future (e.g. when they use `del os.environ['WEST_CONFIG_GLOBAL']`)

tests/conftest.py
```diff
@@ -62,7 +61,65 @@
...
+    os.environ.update(env_vars)
+    try:
+        yield
+    finally:
+        os.environ.clear()
+        os.environ.update(env_bak)
+
+
+
+
```
pdgendt: Do we need 4 newlines?

pdgendt (approved):
Thanks for the fast fix! This looks good IMO.

pdgendt:
> @pdgendt We could also create some invalid user config in the CI system (at the default path) to ensure that this corrupt config is never used in CI. Otherwise tests could potentially touch real configs again in future (e.g. when they use `del os.environ['WEST_CONFIG_GLOBAL']`)

I like the idea, would even be better if we could verify that locally before CI, but not sure if that's easy.

tests/conftest.py
```diff
@@ -62,7 +61,63 @@
...
+    provided in the `env` dictionary for the duration of the `with` block.
+    Any existing environment variables are preserved and restored when
+    the block exits.
+    If the value is set to None, the environment variable is unset.
+    The value can also be either appended (append=True) or prepended
+    (prepend=True) to existing, using the given separator (sep).
+    """
+    env_bak = dict(os.environ)
+    env_vars = {}
+    for k,v in env.items():
```
pdgendt:
```suggestion
    for k, v in env.items():
```

tests/conftest.py
```diff
@@ -86,6 +141,34 @@ def _check_git_capabilities(tmpdir_factory):
...
+    # config paths
+    system = tmpdir / 'config.system'
+    glbl = tmpdir / 'config.global'
+
+    # run with environment variables set
+    with update_env({
+        'WEST_CONFIG_SYSTEM' : str(system),
+        'WEST_CONFIG_GLOBAL' : str(glbl),
+        'ZEPHYR_BASE' : str(tmpdir / 'no-zephyr-here'),
+    }):
```
pdgendt, line 161:
```suggestion
    with update_env(
        {
            'WEST_CONFIG_SYSTEM': str(system),
            'WEST_CONFIG_GLOBAL': str(glbl),
            'ZEPHYR_BASE': str(tmpdir / 'no-zephyr-here'),
        }
    ):
```

tests/conftest.py
```diff
@@ -237,66 +320,20 @@ def west_init_tmpdir(repos_tmpdir):
...
-                                 topdir=str(td)) ==
-                str(local))
-        td.remove(rec=1)
-        assert not td.exists()
-
-        assert not local.exists()
-
-        # All clear: switch to the temporary directory and run the test.
+    # run the test within tmpdir and with env variable set
+    with chdir(tmpdir), update_env({'WEST_CONFIG_LOCAL' : local_config}):
```
pdgendt:
```suggestion
    with chdir(tmpdir), update_env({'WEST_CONFIG_LOCAL': local_config}):
```

tests/test_config.py
```diff
@@ -213,19 +213,18 @@ def test_local_creation_with_topdir():
...
-    assert not system.exists()
-    assert not glbl.exists()
-    assert not local.exists()
-    assert topdir_config.exists()
-
-    assert cfg(f=ALL, topdir=str(topdir))['pytest']['key'] == 'val'
-    assert 'pytest' not in cfg(f=SYSTEM)
-    assert 'pytest' not in cfg(f=GLOBAL)
-    assert cfg(f=LOCAL, topdir=str(topdir))['pytest']['key'] == 'val'
+    with update_env({'WEST_CONFIG_LOCAL':None}):
```
pdgendt:
```suggestion
    with update_env({'WEST_CONFIG_LOCAL': None}):
```

tests/test_manifest.py
```diff
@@ -164,12 +165,9 @@ def test_manifest_from_file_with_fall_back(manifest_repo):
         ''')
     repo_abspath = Path(str(manifest_repo))
     os.chdir(repo_abspath.parent.parent)  # this is the tmp_workspace dir
-    try:
-        os.environ['ZEPHYR_BASE'] = os.fspath(manifest_repo)
+    with update_env({'ZEPHYR_BASE' : os.fspath(manifest_repo)}):
```
pdgendt:
```suggestion
    with update_env({'ZEPHYR_BASE': os.fspath(manifest_repo)}):
```

pdgendt (commented):
Some styling fixes

thorsten-klein:
I have added this new method `update_env()` which can be safely used in tests in order to unset env variables (instead of `del os.environ`). Maybe there is some way to forbid via `conftest.oy` that `del os.environ` is used?

tests/test_config.py
thorsten-klein: Wasn't this checked earlier while linting?

pdgendt: Linter doesn't do formatting, dropped https://github.com/zephyrproject-rtos/west/pull/865 🫣

tests/conftest.py
```diff
@@ -86,6 +141,36 @@ def _check_git_capabilities(tmpdir_factory):
...
+    # config paths
+    system = tmpdir / 'config.system'
+    glbl = tmpdir / 'config.global'
+
+    # run with environment variables set
+    with update_env(
+        {
+            'WEST_CONFIG_SYSTEM' : str(system),
+            'WEST_CONFIG_GLOBAL' : str(glbl),
+            'ZEPHYR_BASE' : str(tmpdir / 'no-zephyr-here'),
```
pdgendt: This is why we need a formatter 😆 

```suggestion
            'WEST_CONFIG_SYSTEM': str(system),
            'WEST_CONFIG_GLOBAL': str(glbl),
            'ZEPHYR_BASE': str(tmpdir / 'no-zephyr-here'),
```

thorsten-klein:
I will rebase once your PR #865 is merged and run `poe format` 👍

pdgendt:
> I will rebase once your PR #865 is merged and run `poe format` 👍

Update this PR instead, so we can go ahead with this one. I'm afraid the formatting PR will need some more eyes and time.

marc-hb:
I haven't looked at the code yet but it reminded me 6cf0a0236b10665... was this lost in the transition to `uv`?

pdgendt:
> I haven't looked at the code yet but it reminded me 6cf0a0236b10665... was this lost in the transition to `uv`?

I don't think so, it just wasn't noticed when new tests were added.

pdgendt:
> > I haven't looked at the code yet but it reminded me [6cf0a02](https://github.com/zephyrproject-rtos/west/commit/6cf0a0236b1066562ec8e58e0dfb11cc742a6e15)... was this lost in the transition to `uv`?
> 
> I don't think so, it just wasn't noticed when new tests were added.

Can confirm this issue exists in v1.5.0 running with `tox`.

thorsten-klein:
> We could also create some invalid user config in the CI system (at the default path) to ensure that this corrupt config is never used in CI 

Let's tackle this in a follow-up PR 👍

Once a second reviewer has approved, this PR can be merged, correct?

tests/conftest.py
```diff
@@ -237,66 +322,20 @@ def west_init_tmpdir(repos_tmpdir):
...
-    start_dir = os.getcwd()
-    tmpdir.chdir()
+    # determine a local config in tmp dir
+    local_config = tmpdir / 'config.local'
 
-    try:
-        assert config._location(config.ConfigFile.SYSTEM) == str(system)
-        assert config._location(config.ConfigFile.GLOBAL) == str(glbl)
-        td = tmpdir / 'test-topdir'
-        td.ensure(dir=True)
```
thorsten-klein, line 288: I have removed this block since in my opinion it does not need to run before each test.

tests/conftest.py
```diff
@@ -237,66 +322,20 @@ def west_init_tmpdir(repos_tmpdir):
...
-        (td / '.west').ensure(dir=True)
-        (td / '.west' / 'config').ensure(file=True)
-        assert config._location(config.ConfigFile.LOCAL) == str(local)
-        assert (config._location(config.ConfigFile.LOCAL,
-                                 topdir=str(td)) ==
-                str(local))
-        td.remove(rec=1)
-        assert not td.exists()
-
-        assert not local.exists()
```
thorsten-klein, line 296: Does it make sense to keep this assert? I mean, we create a new tmpdir and so this local config will never exist?!

tests/conftest.py
```diff
@@ -62,7 +61,63 @@
 WINDOWS = (platform.system() == 'Windows')
 
 #
-# Test fixtures
+# Contextmanager
+#
+
+@contextlib.contextmanager
+def update_env(env: dict, append=False, prepend=False, sep=os.pathsep):
```
thorsten-klein: would it make sense to move `update_env` and `chdir` to `src`? Those could be useful in productive code as well, e.g. `update_env` could be used here: https://github.com/zephyrproject-rtos/west/blame/main/src/west/app/project.py#L1964

pdgendt: Not necessarily, there the original environment doesn't need to be reset.

pdgendt: The `append`, `prepend` and `sep` aren't used AFAICS? We should keep it simple IMO and leave those out.

thorsten-klein: Ok, I will remove the `append`, `prepend` and `sep` parameters

tests/conftest.py
```diff
@@ -62,7 +61,56 @@
 WINDOWS = (platform.system() == 'Windows')
 
 #
-# Test fixtures
+# Contextmanager
+#
+
+@contextlib.contextmanager
+def update_env(env: dict):
```
pdgendt: I think we should be more strict at the input here, like Copilot suggests

```suggestion
def update_env(env: dict[str, str]):
```

tests/conftest.py
```diff
@@ -62,7 +61,56 @@
...
+    """
+    env_bak = dict(os.environ)
+    env_vars = {}
+    for k, v in env.items():
+        # unset if value is None
+        if v is None and k in os.environ:
+            del os.environ[k]
+        # set env variable to new value
+        else:
+            env_vars[k] = f'{v}'
```
pdgendt: We can simplify this

```suggestion
            env_vars[k] = v
```

tests/conftest.py
```diff
@@ -237,66 +315,20 @@ def west_init_tmpdir(repos_tmpdir):
...
-                                 topdir=str(td)) ==
-                str(local))
-        td.remove(rec=1)
-        assert not td.exists()
-
-        assert not local.exists()
-
-        # All clear: switch to the temporary directory and run the test.
+    # run the test within tmpdir and with env variable set
+    with chdir(tmpdir), update_env({'WEST_CONFIG_LOCAL': local_config}):
```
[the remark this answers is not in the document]
thorsten-klein: This conversion is done in update_env. Shall I enforce that the dict given to update_env is dict(str:str)?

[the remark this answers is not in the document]
pdgendt: Yes, see https://github.com/zephyrproject-rtos/west/pull/864#pullrequestreview-3324533867

tests/conftest.py
```diff
@@ -62,7 +61,56 @@
 WINDOWS = (platform.system() == 'Windows')
 
 #
-# Test fixtures
+# Contextmanager
+#
+
+@contextlib.contextmanager
+def update_env(env: dict[str, str | None]):
```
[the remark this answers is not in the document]
pdgendt: This is fine

pdgendt:
@mbolivar PTAL, as you're the original author of the environment mechanisms in tests.

I think the fix is elegant and more Pythonic.

mbolivar (approved):

tests/conftest.py
```diff
@@ -62,7 +61,57 @@
...
+#
+
+
+@contextlib.contextmanager
+def update_env(env: dict[str, str | None]):
+    """
+    Temporarily update the process environment variables.
+    This context manager updates `os.environ` with the key-value pairs
+    provided in the `env` dictionary for the duration of the `with` block.
+    Any existing environment variables are preserved and restored when
```
marc-hb: Nit
```suggestion
    The existing environment is preserved and fully restored when
```

marc-hb (commented):
I had a quick look and spotted nothing wrong, thanks for the fix.

PR #866, opened 2025-10-14
configuration.py: add types to _whence() function, other cosmetic fixes

marc-hb:
Add typing information to the _whence() function that I found difficult to read. Other cosmetic fixes.

### DIFF
--- a/src/west/configuration.py
+++ b/src/west/configuration.py
@@ -58,6 +58,8 @@ class _InternalCF:
     # For internal use only; convenience interface for reading and
     # writing INI-style [section] key = value configuration files,
     # but presenting a west-style section.key = value style API.
+    # Unlike the higher level and public "Configuration" class, there is
+    # one _InternalCF object for each location: LOCAL, GLOBAL,...
 
     @staticmethod
     def parse_key(dotted_name: str):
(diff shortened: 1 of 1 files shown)


marc-hb:
I struggled a bit while trying to review #849 

cc: @thorsten-klein

src/west/configuration.py
```diff
@@ -236,21 +239,21 @@ def _get(self, getter, default, configfile):
 
         return default
 
-    def _whence(self, configfile):
-        if configfile == ConfigFile.ALL:
+    def _whence(self, locations: ConfigFile) -> list[_InternalCF | None]:
+        if locations == ConfigFile.ALL:
```
marc-hb, line 243: I think a large part of the readability issue is: `ConfigFile` is not a configuration file but an Enum. This is one way to mitigate this problem (without the massive churn of renaming the Enum)

thorsten-klein: The name `config` for a `ConfigFile` object is used at many more places. It is always this enum.
So I would prefer if the naming is consistent. 
Would do you think about changing the variable name (and maybe also the type name) at all places?

marc-hb: I don't think local variables of a given type have to be named the same across the board.

I submitted a change in this particular place because I found it especially confusing. I had a very quick look at other places and while the name `configfile` is still not good, they generally did not seem as confusing. So not worth the git churn IMHO. If you find some other confusing places then by all means submit another PR.

pdgendt (approved):

mbolivar (approved):
fine with me if it makes it more readable for you!

PR #867, opened 2025-10-15
Support multiple west config files per config level

thorsten-klein:
This is (another) alternative to PR #849. It is easier to discuss this solution while there is already some implementation existing.

# Why?
In a big company there are often shared configs for build system and tooling (e.g., west and its configs).
Those generic configs must be applied first and contain some default or fallback values, if the user does not specify any value in any user config.
Depending on the company and team size, multiple different configs need to be stacked.
Remember, that users must always be able to overwrite values from those generic config files within their config.

# Proposed Solution
Multiple config files can be specified in the according environment variable (`WEST_CONFIG_SYSTEM`, `WEST_CONFIG_GLOBAL`, `WEST_CONFIG_LOCAL`) for each config level (separated by `os.pathsep`).
The config files are applied in the same order as they are specified, whereby values from later files override earlier ones.

I personally like configuration through env variables, but I fear you don't.
Nevertheless I have prepared this proposal. Maybe it convinces you.

In a next step, this mechanism could also be combined with the approach from #849, so that multiple configs per config level are supported with a dropin config directory each. This would be the ultimative solution.

### DIFF
--- a/src/west/app/config.py
+++ b/src/west/app/config.py
@@ -54,8 +54,11 @@ file locations can be specified. To do so, set according environment variable
 to contain all paths (separated by 'os.pathsep', which is ';' on Windows or
 ':' otherwise): Latter configuration files have precedence in such lists.
 
-The following command prints a list of all configuration files currently
-considered and existing (listed in the order as they are loaded):
+To list all configuration files searched by west config, in the order as they
+are looked up:
+    west config --list-search-paths
+
+To list only existing configs (listed in the order as they are loaded):
     west config --list-paths
 
 To get the value for config option <name>, type:
@@ -118,7 +121,12 @@ class Config(WestCommand):
         group.add_argument(
             '--list-paths',
             action='store_true',
-            help='list all config files that are currently considered by west config',
+            help='list paths of existing config files currently used by west config',
+        )
+        group.add_argument(
+            '--list-search-paths',
+            action='store_true',
+            help='list search paths for west config files',
         )
         group.add_argument(
             '-l', '--list', action='store_true', help='list all options and their values'
@@ -171,10 +179,19 @@ class Config(WestCommand):
 
     def do_run(self, args, user_args):
         delete = args.delete or args.delete_all
-        if args.list:
+        if any([args.list, args.list_paths, args.list_search_paths]):
             if args.name:
-                self.parser.error('-l cannot be combined with name argument')
-        elif not args.name and not args.list_paths:
+                self.parser.error(
+                    (
+                        '-l'
+                        if args.list
+                        else '--list-paths'
+                        if args.list_paths
+                        else '--list-search-paths'
+                    )
+                    + ' cannot be combined with name argument'
+                )
+        elif not args.name:
             self.parser.error('missing argument name (to list all options and values, use -l)')
         elif args.append:
             if args.value is None:
@@ -182,6 +199,8 @@ class Config(WestCommand):
 
         if args.list_paths:
             self.list_paths(args)
+        elif args.list_search_paths:
+            self.list_search_paths(args)
         elif args.list:
             self.list(args)
         elif delete:
@@ -197,6 +216,11 @@ class Config(WestCommand):
         config_paths = self.config.get_existing_paths(args.configfile or ALL)
         for config_path in config_paths:
             self.inf(config_path)
+
+    def list_search_paths(self, args):
+        search_paths = self.config.get_search_paths(args.configfile or ALL)
+        for search_path in search_paths:
+            self.inf(search_path)
 
     def list(self, args):
         what = args.configfile or ALL
--- a/src/west/app/config.py
+++ b/src/west/app/config.py
@@ -49,6 +49,11 @@ All `west config` commands can be applied for a specific configuration level by
 providing one of the following arguments:
     --local | --system | --global
 
+For each configuration level (local, global, and system) multiple config
+file locations can be specified. To do so, set according environment variable
+to contain all paths (separated by 'os.pathsep', which is ';' on Windows or
+':' otherwise): Latter configuration files have precedence in such lists.
+
 The following command prints a list of all configuration files currently
 considered and existing (listed in the order as they are loaded):
     west config --list-paths
@@ -189,7 +194,7 @@ class Config(WestCommand):
             self.write(args)
 
     def list_paths(self, args):
-        config_paths = self.config.get_paths(args.configfile or ALL)
+        config_paths = self.config.get_existing_paths(args.configfile or ALL)
         for config_path in config_paths:
             self.inf(config_path)
 
--- a/src/west/configuration.py
+++ b/src/west/configuration.py
@@ -70,15 +70,30 @@ class _InternalCF:
         return section_child
 
     @staticmethod
-    def from_path(path: Path | None) -> '_InternalCF | None':
-        return _InternalCF(path) if path and path.exists() else None
+    def from_paths(paths: list[Path] | None) -> '_InternalCF | None':
+        if not paths:
+            return None
+        paths = [p for p in paths if p.exists()]
+        return _InternalCF(paths) if paths else None
 
-    def __init__(self, path: Path):
-        self.path = path
+    def __init__(self, paths: list[Path]):
+        # Latter elements in the list have precedence as documented by configparser.
         self.cp = _configparser()
-        read_files = self.cp.read(path, encoding='utf-8')
-        if len(read_files) != 1:
-            raise FileNotFoundError(path)
+        self.paths = paths
+        read_files = self.cp.read(self.paths, encoding='utf-8')
+        if len(read_files) != len(self.paths):
+            readable = [str(p) for p in paths]
+            raise MalformedConfig(f"Error while reading one of '{readable}'")
+
+    def _check_single_config(self):
+        assert self.paths
+        if len(self.paths) > 1:
+            raise ValueError(f'Cannot write if multiple configs in use: {self.paths}')
+
+    def _write(self):
+        assert len(self.paths) == 1  # callers are expected to _check_single_config()
+        with open(self.paths[0], 'w', encoding='utf-8') as f:
+            self.cp.write(f)
 
     def __contains__(self, option: str) -> bool:
         section, key = _InternalCF.parse_key(option)
@@ -106,6 +121,7 @@ class _InternalCF:
             raise KeyError(option) from err
 
     def set(self, option: str, value: Any):
+        self._check_single_config()
         section, key = _InternalCF.parse_key(option)
 
         if section not in self.cp:
@@ -113,10 +129,10 @@ class _InternalCF:
 
         self.cp[section][key] = value
 
-        with open(self.path, 'w', encoding='utf-8') as f:
-            self.cp.write(f)
+        self._write()
 
     def delete(self, option: str):
+        self._check_single_config()
         section, key = _InternalCF.parse_key(option)
 
         if section not in self.cp:
@@ -126,8 +142,7 @@ class _InternalCF:
         if not self.cp[section].items():
             del self.cp[section]
 
-        with open(self.path, 'w', encoding='utf-8') as f:
-            self.cp.write(f)
+        self._write()
 
 
 class ConfigFile(Enum):
@@ -173,25 +188,33 @@ class Configuration:
         :param topdir: workspace location; may be None
         '''
 
-        local_path = _location(ConfigFile.LOCAL, topdir=topdir, find_local=False) or None
+        self._local_locs = [Path(p) for p in _location(ConfigFile.LOCAL, topdir, find_local=False)]
+        self._global_locs = [Path(p) for p in _location(ConfigFile.GLOBAL, topdir)]
+        self._system_locs = [Path(p) for p in _location(ConfigFile.SYSTEM, topdir)]
 
-        self._system_path = Path(_location(ConfigFile.SYSTEM, topdir=topdir))
-        self._global_path = Path(_location(ConfigFile.GLOBAL, topdir=topdir))
-        self._local_path = Path(local_path) if local_path is not None else None
+        self._system = _InternalCF.from_paths(self._system_locs)
+        self._global = _InternalCF.from_paths(self._global_locs)
+        self._local = _InternalCF.from_paths(self._local_locs)
 
-        self._system = _InternalCF.from_path(self._system_path)
-        self._global = _InternalCF.from_path(self._global_path)
-        self._local = _InternalCF.from_path(self._local_path)
-
-    def get_paths(self, location: ConfigFile = ConfigFile.ALL) -> list[Path]:
+    def get_search_paths(self, location: ConfigFile = ConfigFile.ALL) -> list[Path]:
         ret = []
-        if self._system and location in [ConfigFile.SYSTEM, ConfigFile.ALL]:
-            ret.append(self._system.path)
-        if self._global and location in [ConfigFile.GLOBAL, ConfigFile.ALL]:
-            ret.append(self._global.path)
-        if self._local and location in [ConfigFile.LOCAL, ConfigFile.ALL]:
-            ret.append(self._local.path)
+        if location in [ConfigFile.SYSTEM, ConfigFile.ALL]:
+            ret.extend(self._system_locs)
+        if location in [ConfigFile.GLOBAL, ConfigFile.ALL]:
+            ret.extend(self._global_locs)
+        if location in [ConfigFile.LOCAL, ConfigFile.ALL]:
+            ret.extend(self._local_locs)
         return ret
+
+    def get_existing_paths(self, location: ConfigFile = ConfigFile.ALL) -> list[Path]:
+        paths = []
+        if location in [ConfigFile.SYSTEM, ConfigFile.ALL]:
+            paths.extend(self._system.paths if self._system else [])
+        if location in [ConfigFile.GLOBAL, ConfigFile.ALL]:
+            paths.extend(self._global.paths if self._global else [])
+        if location in [ConfigFile.LOCAL, ConfigFile.ALL]:
+            paths.extend(self._local.paths if self._local else [])
+        return paths
 
     def get(
         self, option: str, default: str | None = None, configfile: ConfigFile = ConfigFile.ALL
@@ -287,36 +310,40 @@ class Configuration:
             # We need a real configuration file; ALL doesn't make sense here.
             raise ValueError(configfile)
         elif configfile == ConfigFile.LOCAL:
-            if self._local_path is None:
+            if not self._local_locs:
                 raise ValueError(
                     f'{configfile}: file not found; retry in a workspace or set WEST_CONFIG_LOCAL'
                 )
-            if not self._local_path.exists():
-                self._local = self._create(self._local_path)
-            if TYPE_CHECKING:
-                assert self._local
-            self._local.set(option, value)
+
+        # ensure that only one config is in use
+        configs = self.get_search_paths(configfile)
+        if len(configs) > 1:
+            raise ValueError(f'Cannot set value if multiple configs in use: {configs}')
+        assert len(configs) == 1, f'{configfile}: no config file in use'
+        config_path = configs[0]
+
+        # get internal attribute name (_local/_global/_system)
+        assert configfile in [ConfigFile.SYSTEM, ConfigFile.GLOBAL, ConfigFile.LOCAL]
+        if configfile == ConfigFile.SYSTEM:
+            attr = '_system'
         elif configfile == ConfigFile.GLOBAL:
-            if not self._global_path.exists():
-                self._global = self._create(self._global_path)
-            if TYPE_CHECKING:
-                assert self._global
-            self._global.set(option, value)
-        elif configfile == ConfigFile.SYSTEM:
-            if not self._system_path.exists():
-                self._system = self._create(self._system_path)
-            if TYPE_CHECKING:
-                assert self._system
-            self._system.set(option, value)
-        else:
-            # Shouldn't happen.
-            raise AssertionError(configfile)
+            attr = '_global'
+        elif configfile == ConfigFile.LOCAL:
+            attr = '_local'
+
+        # create config file if it does not already exist and use it
+        if not getattr(self, attr, None):
+            config = self._create(config_path)
+            setattr(self, attr, config)
+
+        # set config value
+        getattr(self, attr).set(option, value)
 
     @staticmethod
     def _create(path: Path) -> _InternalCF:
         path.parent.mkdir(parents=True, exist_ok=True)
         path.touch(exist_ok=True)
-        ret = _InternalCF.from_path(path)
+        ret = _InternalCF.from_paths([path])
         if TYPE_CHECKING:
             assert ret
         return ret
@@ -441,6 +468,13 @@ def _deprecated(old_function):
     )
 
 
+def _ensure_one_file_max_configs(topdir):
+    for loc in [ConfigFile.SYSTEM, ConfigFile.GLOBAL, ConfigFile.LOCAL]:
+        config = _location(loc, topdir=topdir, find_local=False)
+        if len(config) > 1:
+            raise MalformedConfig(f"{loc}: Switch to newer API to use multiple configs: '{config}'")
+
+
 def read_config(
     configfile: ConfigFile | None = None,
     config: configparser.ConfigParser = config,
@@ -471,6 +505,7 @@ def read_config(
     :param topdir: west workspace root to read local options from
     '''
     _deprecated('read_config')
+    _ensure_one_file_max_configs(topdir)
 
     if configfile is None:
         configfile = ConfigFile.ALL
@@ -503,6 +538,7 @@ def update_config(
       one is not found).
     '''
     _deprecated('update_config')
+    _ensure_one_file_max_configs(topdir)
 
     if configfile == ConfigFile.ALL:
         # Not possible to update ConfigFile.ALL, needs specific conf file here.
@@ -552,20 +588,23 @@ def delete_config(
       one is not found).
     '''
     _deprecated('delete_config')
+    _ensure_one_file_max_configs(topdir)
 
     stop = False
+    considered_locations = []
     if configfile is None:
-        to_check = [_location(x, topdir=topdir) for x in [ConfigFile.LOCAL, ConfigFile.GLOBAL]]
+        considered_locations = [ConfigFile.LOCAL, ConfigFile.GLOBAL]
         stop = True
     elif configfile == ConfigFile.ALL:
-        to_check = [
-            _location(x, topdir=topdir)
-            for x in [ConfigFile.SYSTEM, ConfigFile.GLOBAL, ConfigFile.LOCAL]
-        ]
+        considered_locations = [ConfigFile.SYSTEM, ConfigFile.GLOBAL, ConfigFile.LOCAL]
     elif isinstance(configfile, ConfigFile):
-        to_check = [_location(configfile, topdir=topdir)]
+        considered_locations = [configfile]
     else:
-        to_check = [_location(x, topdir=topdir) for x in configfile]
+        considered_locations = configfile
+    assert isinstance(considered_locations, list)
+
+    # thanks to _ensure_one_file_max_configs() above, we don't actually have 2 levels here.
+    to_check = [f for cfg in considered_locations for f in _location(cfg, topdir=topdir)]
 
     found = False
     for path in to_check:
@@ -597,7 +636,9 @@ def _rel_topdir_to_abs(p: PathType, topdir: PathType | None) -> str:
     return str(os.path.join(topdir, p))
 
 
-def _location(cfg: ConfigFile, topdir: PathType | None = None, find_local: bool = True) -> str:
+def _location(
+    cfg: ConfigFile, topdir: PathType | None = None, find_local: bool = True
+) -> list[str]:
     # Return the WEST_CONFIG_x environment variable if defined, or the
     # OS-specific default value. Anchors relative paths to
     # "topdir". Does _not_ check whether the file exists or if it is
@@ -616,27 +657,34 @@ def _location(cfg: ConfigFile, topdir: PathType | None = None, find_local: bool 
     # any cases which might run on Windows, then call fspath() on the
     # final result. This lets the standard library do the work of
     # producing a canonical string name.
+
+    def parse_paths(paths: str | None, sep: str = os.pathsep) -> list[Path]:
+        """Split a string into a list of Path objects using the given separator."""
+        paths = paths or ""
+        return [Path(p) for p in paths.split(sep) if p]
+
     env = os.environ
 
     if cfg == ConfigFile.ALL:
         raise ValueError('ConfigFile.ALL has no location')
     elif cfg == ConfigFile.SYSTEM:
         if 'WEST_CONFIG_SYSTEM' in env:
-            return _rel_topdir_to_abs(env['WEST_CONFIG_SYSTEM'], topdir)
+            paths = parse_paths(env['WEST_CONFIG_SYSTEM'])
+            return [_rel_topdir_to_abs(p, topdir) for p in paths]
 
         plat = platform.system()
 
         if plat == 'Linux':
-            return '/etc/westconfig'
+            return ['/etc/westconfig']
 
         if plat == 'Darwin':
-            return '/usr/local/etc/westconfig'
+            return ['/usr/local/etc/westconfig']
 
         if plat == 'Windows':
-            return os.path.expandvars('%PROGRAMDATA%\\west\\config')
+            return [os.path.expandvars('%PROGRAMDATA%\\west\\config')]
 
         if 'BSD' in plat:
-            return '/etc/westconfig'
+            return ['/etc/westconfig']
 
         if 'CYGWIN' in plat or 'MSYS_NT' in plat:
             # Cygwin can handle windows style paths, so make sure we
@@ -647,29 +695,31 @@ def _location(cfg: ConfigFile, topdir: PathType | None = None, find_local: bool 
             # See https://github.com/zephyrproject-rtos/west/issues/300
             # for details.
             pd = PureWindowsPath(os.environ['ProgramData'])
-            return os.fspath(pd / 'west' / 'config')
+            return [os.fspath(pd / 'west' / 'config')]
 
         raise ValueError('unsupported platform ' + plat)
     elif cfg == ConfigFile.GLOBAL:
         if 'WEST_CONFIG_GLOBAL' in env:
-            return _rel_topdir_to_abs(env['WEST_CONFIG_GLOBAL'], topdir)
+            paths = parse_paths(env['WEST_CONFIG_GLOBAL'])
+            return [_rel_topdir_to_abs(p, topdir) for p in paths]
 
         if platform.system() == 'Linux' and 'XDG_CONFIG_HOME' in env:
-            return os.path.join(env['XDG_CONFIG_HOME'], 'west', 'config')
+            return [os.path.join(env['XDG_CONFIG_HOME'], 'west', 'config')]
 
-        return os.fspath(Path.home() / '.westconfig')
+        return [os.fspath(Path.home() / '.westconfig')]
     elif cfg == ConfigFile.LOCAL:
         if 'WEST_CONFIG_LOCAL' in env:
-            return _rel_topdir_to_abs(env['WEST_CONFIG_LOCAL'], topdir)
+            paths = parse_paths(env['WEST_CONFIG_LOCAL'])
+            return [_rel_topdir_to_abs(p, topdir) for p in paths]
 
         if topdir:
-            return os.fspath(Path(topdir) / WEST_DIR / 'config')
+            return [os.fspath(Path(topdir) / WEST_DIR / 'config')]
 
         if find_local:
             # Might raise WestNotFound!
-            return os.fspath(Path(west_dir()) / 'config')
+            return [os.fspath(Path(west_dir()) / 'config')]
         else:
-            return ''
+            return []
     else:
         raise ValueError(f'invalid configuration file {cfg}')
 
@@ -677,15 +727,15 @@ def _location(cfg: ConfigFile, topdir: PathType | None = None, find_local: bool 
 def _gather_configs(cfg: ConfigFile, topdir: PathType | None) -> list[str]:
     # Find the paths to the given configuration files, in increasing
     # precedence order.
-    ret = []
+    ret: list[str] = []
 
     if cfg == ConfigFile.ALL or cfg == ConfigFile.SYSTEM:
-        ret.append(_location(ConfigFile.SYSTEM, topdir=topdir))
+        ret.extend(_location(ConfigFile.SYSTEM, topdir=topdir))
     if cfg == ConfigFile.ALL or cfg == ConfigFile.GLOBAL:
-        ret.append(_location(ConfigFile.GLOBAL, topdir=topdir))
+        ret.extend(_location(ConfigFile.GLOBAL, topdir=topdir))
     if cfg == ConfigFile.ALL or cfg == ConfigFile.LOCAL:
         try:
-            ret.append(_location(ConfigFile.LOCAL, topdir=topdir))
+            ret.extend(_location(ConfigFile.LOCAL, topdir=topdir))
         except WestNotFound:
             pass
 
@@ -695,11 +745,12 @@ def _gather_configs(cfg: ConfigFile, topdir: PathType | None) -> list[str]:
 def _ensure_config(configfile: ConfigFile, topdir: PathType | None) -> str:
     # Ensure the given configfile exists, returning its path. May
     # raise permissions errors, WestNotFound, etc.
-    loc = _location(configfile, topdir=topdir)
-    path = Path(loc)
-
+    configs: list[str] = _location(configfile, topdir=topdir)
+    assert configs, 'No configfile found'
+    assert len(configs) == 1, f'Multiple config files in use: {configs}'
+    path = Path(configs[0])
     if path.is_file():
-        return loc
+        return os.fspath(path)
 
     path.parent.mkdir(parents=True, exist_ok=True)
     path.touch(exist_ok=True)
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -69,6 +69,22 @@ WINDOWS = platform.system() == 'Windows'
 
 
 @contextlib.contextmanager
+def tmp_west_topdir(path: str | Path):
+    """
+    Temporarily create a west topdir for the duration of the `with` block by
+    creating a .west directory at given path. The directory is removed again
+    when the `with` block exits.
+    """
+    west_dir = Path(path) / '.west'
+    west_dir.mkdir(parents=True)
+    try:
+        yield
+    finally:
+        # remove the directory (must be empty)
+        west_dir.rmdir()
+
+
+@contextlib.contextmanager
 def update_env(env: dict[str, str | None]):
     """
     Temporarily update the process environment variables.
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -9,9 +9,10 @@ import textwrap
 from typing import Any
 
 import pytest
-from conftest import chdir, cmd, cmd_raises, update_env
+from conftest import WINDOWS, chdir, cmd, cmd_raises, tmp_west_topdir, update_env
 
 from west import configuration as config
+from west.configuration import MalformedConfig
 from west.util import PathType, WestNotFound
 
 SYSTEM = config.ConfigFile.SYSTEM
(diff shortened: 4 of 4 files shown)


marc-hb:
> This is (another) alternative to PR https://github.com/zephyrproject-rtos/west/pull/849.

I don't see how this is better than #849 dropins, sorry. It seems to have the same complexity (e.g.: how do you append to an existing value?) but with a less convenient and less standard user interface.

thorsten-klein:
Perfectly fine for me. I just wanted to show it to you for completeness.
~~I will close this PR now 👍🏻~~

marc-hb:
You could have waited for other opinions! They could still come but it's bit less likely in a closed PR.

src/west/configuration.py
```diff
@@ -68,20 +68,38 @@ def parse_key(dotted_name: str):
...
+        with open(self.paths[0], 'w', encoding='utf-8') as f:
+            self.cp.write(f)
 
     def __contains__(self, option: str) -> bool:
         section, key = _InternalCF.parse_key(option)
 
-        return section in self.cp and key in self.cp[section]
+        if section in self.cp and key in self.cp[section]:
+            return True
+        return False
```
pdgendt: Revert this change, no need to do:

```python
if some_boolean_result:
    return True
return False
# just do
return some_boolean_result
```

pdgendt (commented):
Honestly, I like the simplicity here. A user that sets the environment variable, to change a config file location, already needs to know what they are doing.

Allowing to pass in multiple files is in line with the ConfigParser's functionality.

src/west/app/config.py
```diff
@@ -170,6 +196,11 @@ def do_run(self, args, user_args):
         else:
             self.write(args)
 
+    def list_paths(self, args):
+        config_paths = Configuration().get_paths(args.configfile or ALL)
+        for config_path in config_paths:
+            print(config_path)
```
pdgendt: Should use `self.inf` here.

pdgendt (commented):
@thorsten-klein could you split the `--list-paths` into it's own PR? Both this and the dropins PR have this addition, we could maybe move ahead with this simpler addition and clean up this and the dropins PR.

carlescufi (approved):
I am fine with this, thanks @thorsten-klein.

tests/test_config.py
```diff
@@ -590,6 +590,81 @@ def test_config_precedence():
...
+
+    # create some configs with
+    # - some individual option per config file
+    # - the same option defined in multiple configs
+    write_config(config_s1, 'sec', 's', 1, 's1', 1)
+    write_config(config_s2, 'sec', 's', 2, 's2', 2)
+    write_config(config_g1, 'sec', 'g', 1, 'g1', 1)
+    write_config(config_g2, 'sec', 'g', 2, 'g2', 2)
+    write_config(config_l1, 'sec', 'l', 1, 'l1', 1)
+    write_config(config_l2, 'sec', 'l', 2, 'l2', 2)
```
marc-hb: Please add whitespace, `\` and a maximum of other special characters in these test cases. Anything that is allowed by Windows filesystems (Windows is the most restrictive because [Unix allows literally anything](https://dwheeler.com/essays/fixing-unix-linux-filenames.html))

```suggestion
    write_config(config_l2, '*sec*', 'l', 2, 'l 2', 2)
```

You could even try newlines. That will probably expose existing bugs in `configparser` though. Only then you give up.

tests/test_config.py
```diff
@@ -590,6 +590,81 @@ def test_config_precedence():
...
+            conf.write(content)
+
+    # config file paths
+    config_dir = pathlib.Path(config_tmpdir) / 'configs'
+    config_s1 = config_dir / 'system1'
+    config_s2 = config_dir / 'system2'
+    config_g1 = config_dir / 'global1'
+    config_g2 = config_dir / 'global2'
+    config_l1 = config_dir / 'local1'
+    config_l2 = config_dir / 'local2'
```
```suggestion
    config_l1 = config_dir / 'local* 1'
    config_l2 = config_dir / 'local  2'
```
Etc., see below. Test code must always try to break things.

src/west/configuration.py
```diff
@@ -70,19 +70,34 @@ def parse_key(dotted_name: str):
...
+        self.paths = paths
+        read_files = self.cp.read(self.paths, encoding='utf-8')
+        if len(read_files) != len(self.paths):
+            raise FileNotFoundError(paths)
+
+    def _write(self):
+        if not self.paths:
+            raise WestNotFound('No config file exists that can be written')
+        if len(self.paths) > 1:
+            raise WestNotFound('Cannot write if multiple configs in use.')
```
```suggestion
            raise WestNotFound('Cannot write if multiple configs in use:' + os.pathsep.join(self.paths))
```

src/west/app/config.py
```diff
@@ -49,6 +49,11 @@
 providing one of the following arguments:
     --local | --system | --global
 
+For each configuration level (local, global, and system) also multiple config
+file locations can be specified. To do so, set according environment variable
+to contain all paths (separated by 'os.pathsep', which is ';' on Windows or
+':' otherwise)
```
marc-hb:
```suggestion
':' otherwise): Latter configuration files have precedence in such lists.
```

src/west/configuration.py
```diff
@@ -70,19 +70,34 @@ def parse_key(dotted_name: str):
...
+
+    def __init__(self, paths: list[Path]):
         self.cp = _configparser()
-        read_files = self.cp.read(path, encoding='utf-8')
-        if len(read_files) != 1:
-            raise FileNotFoundError(path)
+        self.paths = paths
+        read_files = self.cp.read(self.paths, encoding='utf-8')
+        if len(read_files) != len(self.paths):
+            raise FileNotFoundError(paths)
```
marc-hb: `FileNotFoundError` is not correct here. When a file in the list does not exist, things should still work and they apparently do (make sure you have a test for that, tests are all about edge cases).

On the other hand, while testing and trying to break this I found that a non-readable file throws this exception.

I also found that this WEST_CONFIG_LOCAL also throws this misleading exception: export WEST_CONFIG_LOCAL=~/.westconfig:$EMPTY_BECAUSE_TYPO

An empty path is apparently converted to "."? That seems wrong too.

Printing a stack trace is confusing here because it hides `configparser` which is at the root cause.

Finally, you should replace `paths` with `os.pathsep.join(path)` because this does not look great:
```
FileNotFoundError: [PosixPath('fu'), PosixPath('.'), PosixPath('bar')]
```
So I think you probably want to throw a WestNotFound (no stacktrace?) with a more generic message.

This is not cosmetic: users _will_ misconfigure WEST_CONFIG_LOCAL at some point or another and `west` should give decent feedback when that happens.

src/west/configuration.py
```diff
@@ -186,11 +195,11 @@ def __init__(self, topdir: PathType | None = None):
...
         ret = []
         if self._global and location in [ConfigFile.GLOBAL, ConfigFile.ALL]:
-            ret.append(self._global.path)
+            ret += self._global.paths
         if self._system and location in [ConfigFile.SYSTEM, ConfigFile.ALL]:
-            ret.append(self._system.path)
+            ret += self._system.paths
         if self._local and location in [ConfigFile.LOCAL, ConfigFile.ALL]:
-            ret.append(self._local.path)
+            ret += self._local.paths
```
marc-hb: Is this related to this new feature? If not, separate commit please to reduce mental load. Reviewers don't want to guess.

src/west/configuration.py
```diff
@@ -282,7 +291,9 @@ def set(self, option: str, value: Any, configfile: ConfigFile = ConfigFile.LOCAL
         :param value: value to set option to
         :param configfile: type of config file to set the value in
         '''
-
+        paths = self.get_paths(configfile)
+        if len(paths) > 1:
+            raise WestNotFound('Cannot write if multiple configs in use.')
```
marc-hb: Simple and smart! Nice.

marc-hb (changes requested):
I found a couple issues while testing and reviewing but otherwise I agree this is definitely a minimal way to achieve the objective, nice!

marc-hb:
EDIT: moved to new issue #871. I think #871 should be decided before this #867 PR gets merged.

-------

I just thought about another test case: relative paths. As it stands, this PR allows relative paths in `WEST_CONFIG_LOCAL` and others. This means that the `west config` **depends on the current directory!!**

Everyone please correct me but this looks like a really bad idea to me. I can easily imagine users wasting hours or even days because of subtle compilation differences depending on where they compiled the same thing. There are also serious security implications. So I don't think this should be allowed.

This "bug" predates this PR. So it should probably be fixed before it.

src/west/configuration.py
thorsten-klein: The list is normalized above, so that only paths remain in the list that really exist.
If the count of read files is not matching, something went wrong. I will adapt to raise some `WestNotFound` instead.
I will also extend the test to use a non-readable file and catch this error.

src/west/configuration.py
thorsten-klein: Yes, because now there is a list of paths stored (not a single path).
Therefore I cannot append, but I have to use `extend` or `+=`.

src/west/configuration.py
thorsten-klein: We have this same check at two places. Shall I keep both or only keep one? If yes, which one shall I keep?

pdgendt: I prefer to have the check at the "lowest level", so keep the check in `_write` at least.

This can be used as a fast return path, or we let it fail later on, both are fine IMO.

src/west/configuration.py
pdgendt: For clarity it might be better to use `extend` here, not blocking though.

thorsten-klein:
This PR now also contains a check that the config file paths must be absolute paths (see second commit).
I hope this is fine for you. 
Otherwise I remove the commit from this branch again, so that it can be merged first (only first commit).
I will then immediately open a separate PR for the second commit.

src/west/configuration.py
```diff
@@ -70,19 +70,38 @@ def parse_key(dotted_name: str):
...
-        self.path = path
+    def from_path(path: Path | list[Path] | None) -> '_InternalCF | None':
+        if not path:
+            return None
+
+        # Normalize to a list of existing paths
+        paths = path
+        if isinstance(paths, Path):
+            paths = [Path(p) for p in str(paths).split(os.pathsep)]
+        if not all([p.is_absolute() for p in paths]):
```
pdgendt: The `all` can be optimized to use a generator expression instead.

```suggestion
        if not all(p.is_absolute() for p in paths):
```

src/west/configuration.py
thorsten-klein: We need at both levels since otherwise `_create` is called with the `path1;path2` which of course results in some errors.
We must catch it before we call this function.

marc-hb:
> Otherwise I remove the commit from this branch again, so that it can be merged first (only first commit).
> I will then immediately open a separate PR for the second commit.


Yes please because I would like more people to have a look at and discuss this. It's hopefully harmless but it is a backwards-incompatible change strictly speaking. I suspect I'm not the only struggling to cope with the volume of code you submit and a smaller PR may have more eyes on it. This PR is already too large for anyone to have a "quick look".

https://docs.zephyrproject.org/latest/contribute/contributor_expectations.html#defining-smaller-prs

I also think it's best for bug fixes to stand out in a separate PR when possible: easier to spot or to find - for instance if that broke someone's workspace.


BTW your current fix in https://github.com/zephyrproject-rtos/west/pull/867/commits/620351a8b8bc0b5e5feea54e8c1a8141fba35ca7 in this PR does not point to the bug number #871. I think it also ignores my "middle-ground" suggestion in https://github.com/zephyrproject-rtos/west/issues/871#issuecomment-3440386329 Generally speaking, please try to be much more thorough; quality over quantity. This will make code reviews considerably faster.

src/west/configuration.py
marc-hb: > > WEST_CONFIG_LOCAL=~/.westconfig:$EMPTY_VAR_BECAUSE_TYPO. An empty path is apparently converted to "."? That seems wrong too.

> The list is normalized above, so that only paths remain in the list that really exist.

I don't know where the "normalization" code was but it apparently didn't mind one of the paths to be the current directory "." (which always exists!)

I didn't test any newer version of this PR yet.

src/west/configuration.py
```diff
@@ -70,19 +70,38 @@ def parse_key(dotted_name: str):
...
-    def __init__(self, path: Path):
-        self.path = path
+    def from_path(path: Path | list[Path] | None) -> '_InternalCF | None':
+        if not path:
+            return None
+
+        # Normalize to a list of existing paths
+        paths = path
+        if isinstance(paths, Path):
+            paths = [Path(p) for p in str(paths).split(os.pathsep)]
```
marc-hb: This is confusing because it works only because a single `Path` is not always a real `Path`, sometimes it's a string with a list of paths shoehorned in a single, abused `Path` object! Which is why you convert here this fake `Path` into a string, split it and then convert back to a list of actual `Path`.

Rather than doing this, why not preserve correct typing and move the split to up in the callers? Right now the `string` -> `Path` conversion is performed at this point:

```python
        self._system_path = Path(_location(ConfigFile.SYSTEM))
        self._global_path = Path(_location(ConfigFile.GLOBAL))
        ...
```
After this PR, `self._global_path` can become something like this: `PosixPath('/a/b/c:/d/e/f')`. This is not a real `Path` and very misleading.

Instead you can probably do something like:
```python
        self._system_paths = [Path(p) for p in _locations(ConfigFile.SYSTEM)).split(os.pathsep)]
       ...
```

(this could be a function like `paths_to_str()`)

This would keep strings as strings, every `Path` as an actual, single `Path` and use a proper `Path` list elsewhere.

You can still perform the sanity checks in the common `from_path()`

Changing the name of these to a plural `self._system_paths` and switching to  `self._system_paths[0]` when writing helps makes sure no place ever forgets to check that the length is 1.

src/west/configuration.py
```diff
@@ -70,19 +70,38 @@ def parse_key(dotted_name: str):
         return section_child
 
     @staticmethod
-    def from_path(path: Path | None) -> '_InternalCF | None':
-        return _InternalCF(path) if path and path.exists() else None
-
-    def __init__(self, path: Path):
-        self.path = path
+    def from_path(path: Path | list[Path] | None) -> '_InternalCF | None':
```
marc-hb: This is never actually called with a `list` right now, is it?

I don't think any function or place in the code needs to accept alternatives like this, see below.

src/west/configuration.py
marc-hb: BTW it is this line that converts `:$EMPTY_VAR:` into the current directory.

src/west/configuration.py
```diff
@@ -282,7 +294,9 @@ def set(self, option: str, value: Any, configfile: ConfigFile = ConfigFile.LOCAL
         :param value: value to set option to
         :param configfile: type of config file to set the value in
         '''
-
+        paths = self.get_paths(configfile)
+        if len(paths) > 1:
+            raise WestNotFound(f'Cannot write if multiple configs in use: {paths_to_str(paths)}')
```
marc-hb: Nit: I think the `configfile == ConfigFile.ALL:` check should be first because it's an internal bug.

BTW the `elif` on line 303 should be an `if`.

marc-hb (commented):
(I would still like #871 to be discussed first and fixed separately)

src/west/configuration.py
```diff
@@ -173,24 +179,26 @@ def __init__(self, topdir: PathType | None = None):
...
 
-        self._system = _InternalCF.from_path(self._system_path)
-        self._global = _InternalCF.from_path(self._global_path)
-        self._local = _InternalCF.from_path(self._local_path)
+        self._system = _InternalCF.from_paths(self._system_paths)
+        self._global = _InternalCF.from_paths(self._global_paths)
+        self._local = _InternalCF.from_paths(self._local_paths)
 
-    def get_paths(self, location: ConfigFile = ConfigFile.ALL):
+    def get_paths(self, location: ConfigFile = ConfigFile.ALL, only_existing=True) -> list[Path]:
```
pdgendt: I think `only_existing=False` should be the default to not change the function behavior. Also `existing_only` reads a bit nicer, IMO?

src/west/configuration.py
```diff
@@ -637,18 +671,21 @@ def _location(cfg: ConfigFile, topdir: PathType | None = None, find_local: bool
...
     elif cfg == ConfigFile.GLOBAL:
         if 'WEST_CONFIG_GLOBAL' in env:
-            return env['WEST_CONFIG_GLOBAL']
+            path = env['WEST_CONFIG_GLOBAL']
+            return path
 
         if platform.system() == 'Linux' and 'XDG_CONFIG_HOME' in env:
-            return os.path.join(env['XDG_CONFIG_HOME'], 'west', 'config')
+            path = os.path.join(env['XDG_CONFIG_HOME'], 'west', 'config')
+            return path
```
pdgendt: Leftover changes?

pdgendt:
> > Otherwise I remove the commit from this branch again, so that it can be merged first (only first commit).
> > I will then immediately open a separate PR for the second commit.
> 
> Yes please because I would like more people to have a look at and discuss this.

@thorsten-klein is this still in the works? For me it can go either way, as we've reduced the security concern raised in #871 in https://github.com/zephyrproject-rtos/west/issues/871#issuecomment-3452712021

This PR still only allows absolute paths, while @marc-hb is a proponent of having it relative to the workspace. I don't really have a strong opinion here.

src/west/configuration.py
```diff
@@ -70,19 +70,32 @@ def parse_key(dotted_name: str):
...
-        return _InternalCF(path) if path and path.exists() else None
-
-    def __init__(self, path: Path):
-        self.path = path
+    def from_paths(paths: list[Path] | None) -> '_InternalCF | None':
+        paths = paths or []
+        if not paths:
+            return None
+        if not all(p.is_absolute() for p in paths):
+            raise WestNotFound(f"config file path(s) must be absolute: '{paths_to_str(paths)}'")
```
marc-hb: I don't think this is the right place to do this because `_location()` is used in a number of other places.

I'm working on fixing #871.

marc-hb:
I just submitted #877 to fix the relative path issue #871.

The fix is totally different from the one here (EDIT: cause `_location()` is used in multiple places) and I think it must go first. Please help review, thanks!

src/west/util.py
```diff
@@ -53,6 +53,19 @@ class WestNotFound(RuntimeError):
...
+
+
+def str_to_paths(paths: str | None, sep: str = os.pathsep) -> list[Path]:
+    paths = paths or ""
+    return [Path(p) for p in paths.split(sep) if p]
+
+
+def str_list_to_paths(paths: list[str]) -> list[Path]:
+    return [Path(p) for p in paths]
+
```
pdgendt: Functions added here, are considered public API AFAIK. We should add docstrings in that case.

src/west/configuration.py
```diff
@@ -647,59 +688,66 @@ def _location(cfg: ConfigFile, topdir: PathType | None = None, find_local: bool
...
-            return os.fspath(Path(topdir) / WEST_DIR / 'config')
+            return [os.fspath(Path(topdir) / WEST_DIR / 'config')]
 
         if find_local:
             # Might raise WestNotFound!
-            return os.fspath(Path(west_dir()) / 'config')
+            return [os.fspath(Path(west_dir()) / 'config')]
         else:
-            return ''
+            return []
```
pdgendt, line 722: I'm trying to figure out the impact of the change on this line. Because the result is different, no?

tests/conftest.py
```diff
@@ -68,6 +68,22 @@
 #
 
 
+@contextlib.contextmanager
+def tmp_west_topdir(path: str | Path):
+    """
+    Temporarily create a west topdir for the duration of the `with` block by
+    creating a .west directory at given path. The diectory is removed again
```
pdgendt:
```suggestion
    creating a .west directory at given path. The directory is removed again
```

src/west/configuration.py
thorsten-klein: As far as I can see it was only used to match the return type (str). It seems that it does not break any test.

src/west/configuration.py
```diff
@@ -70,19 +78,30 @@ def parse_key(dotted_name: str):
         return section_child
 
     @staticmethod
-    def from_path(path: Path | None) -> '_InternalCF | None':
-        return _InternalCF(path) if path and path.exists() else None
-
-    def __init__(self, path: Path):
-        self.path = path
+    def from_paths(paths: list[Path] | None) -> '_InternalCF | None':
+        paths = paths or []
```
marc-hb: In which case does this line make a difference?

src/west/configuration.py
```diff
@@ -70,19 +78,30 @@ def parse_key(dotted_name: str):
...
-        if len(read_files) != 1:
-            raise FileNotFoundError(path)
+        self.paths = paths
+        read_files = self.cp.read(self.paths, encoding='utf-8')
+        if len(read_files) != len(self.paths):
+            raise WestNotFound(f"Error while reading one of '{list_to_str(paths)}'")
+
+    def _write(self):
+        if not self.paths:
+            raise WestNotFound('No config file exists that can be written')
```
marc-hb: I think this looks like an internal error, not something that could even happen.
```suggestion
      assert(self.paths)
```

src/west/configuration.py
```diff
@@ -99,35 +118,28 @@ def getfloat(self, option: str):
 
     def _get(self, option, getter):
         section, key = _InternalCF.parse_key(option)
-
```
marc-hb, line 102: Don't mix code changes with whitespace

https://wiki.openstack.org/wiki/GitCommitMessages#Things_to_avoid_when_creating_commits

src/west/configuration.py
marc-hb: This is the code used when west is outside of a workspace. This could also have been `None` because the callers were so far too "accommodating"?

I noticed that we have small test coverage gap (outside this PR) because right now you can replace this line with `return "whatever"` and all the tests pass. A lot of tests fail if you replace it with `return "/etc/passwd"` so it's not so bad.

src/west/configuration.py
```diff
@@ -647,59 +688,66 @@ def _location(cfg: ConfigFile, topdir: PathType | None = None, find_local: bool
...
+        paths = _location(ConfigFile.GLOBAL, topdir=topdir)
+        ret.extend(str_list_to_paths(paths))
     if cfg == ConfigFile.ALL or cfg == ConfigFile.LOCAL:
         try:
-            ret.append(_location(ConfigFile.LOCAL, topdir=topdir))
+            paths = _location(ConfigFile.LOCAL, topdir=topdir)
+            ret.extend(str_list_to_paths(paths))
         except WestNotFound:
             pass
-
```
marc-hb, line 691: Distracting change, remove. This is a complex enough feature.

src/west/util.py
```diff
@@ -53,6 +53,22 @@ class WestNotFound(RuntimeError):
     '''Neither the current directory nor any parent has a west workspace.'''
 
 
+def list_to_str(in_list: list, sep: str = os.pathsep) -> str:
+    """Convert a list of items to a single string joined by the given separator."""
+    return sep.join([str(x) for x in in_list])
```
marc-hb: This one-liner saves not much typing and obscures a very simple thing.

I see this is used mostly in exception strings. I'm not sure it's a good idea to make these "pretty", when something goes wrong I think it's better to show all real data behind the scenes like this:

```python
l = [Path('a'), Path('b')]
print(f"error {l}")

error: [PosixPath('a'), PosixPath('b')]
```

So if there's been some parsing problem then the stacktrace shows it.

Another option is to save the user input in the class for later use instead of doing this round-trip.

Either way this is too basic to belong to `util.py`. Should probably be a _InternalCF method.

src/west/util.py
```diff
@@ -53,6 +53,22 @@ class WestNotFound(RuntimeError):
...
+
+def str_to_paths(paths: str | None, sep: str = os.pathsep) -> list[Path]:
+    """Split a string into a list of Path objects using the given separator."""
+    paths = paths or ""
+    return [Path(p) for p in paths.split(sep) if p]
+
+
+def str_list_to_paths(paths: list[str]) -> list[Path]:
+    """Convert a list of path strings into a list of Path objects."""
+    return [Path(p) for p in paths]
```
marc-hb: This function looks the most basic and least useful of the three. Definitely does not belong to `util.py`.

src/west/util.py
```diff
@@ -53,6 +53,22 @@ class WestNotFound(RuntimeError):
...
 
+def list_to_str(in_list: list, sep: str = os.pathsep) -> str:
+    """Convert a list of items to a single string joined by the given separator."""
+    return sep.join([str(x) for x in in_list])
+
+
+def str_to_paths(paths: str | None, sep: str = os.pathsep) -> list[Path]:
+    """Split a string into a list of Path objects using the given separator."""
+    paths = paths or ""
+    return [Path(p) for p in paths.split(sep) if p]
```
marc-hb: This one has a bit more value but not enough value to be moved to util.py. It should be easy to find in the same file. Maybe it should be static _InternalCF method, parsing feels like it belongs there.

src/west/configuration.py
```diff
@@ -647,59 +688,66 @@ def _location(cfg: ConfigFile, topdir: PathType | None = None, find_local: bool
...
     return ret
 
 
 def _ensure_config(configfile: ConfigFile, topdir: PathType | None) -> str:
     # Ensure the given configfile exists, returning its path. May
     # raise permissions errors, WestNotFound, etc.
-    loc = _location(configfile, topdir=topdir)
-    path = Path(loc)
-
+    config_file = str_list_to_paths(_location(configfile, topdir=topdir))
```
marc-hb: We can't have both `config_file` and `configfile` in the same function... `configs` maybe?

src/west/configuration.py
```diff
@@ -647,59 +688,66 @@ def _location(cfg: ConfigFile, topdir: PathType | None = None, find_local: bool
...
+    if not config_file:
+        raise ValueError(f"no config found for configfile '{configfile}'")
+    if len(config_file) > 1:
+        raise ValueError(f'Cannot write if multiple configs in use: {list_to_str(config_file)}')
```
marc-hb: Rather than converting back, simply save the `_location()` output for later.

src/west/app/config.py
```diff
@@ -49,6 +49,11 @@
 providing one of the following arguments:
     --local | --system | --global
 
+For each configuration level (local, global, and system) also multiple config
+file locations can be specified. To do so, set according environment variable
+to contain all paths (separated by 'os.pathsep', which is ';' on Windows or
+':' otherwise): Latter configuration files have precedence in such lists.
```
marc-hb, line 55: This belongs to the previous commit.

src/west/app/config.py
```diff
@@ -189,7 +194,7 @@ def do_run(self, args, user_args):
             self.write(args)
 
     def list_paths(self, args):
-        config_paths = self.config.get_paths(args.configfile or ALL)
+        config_paths = self.config.get_paths(args.configfile or ALL, existing_only=True)
```
marc-hb: This looks like a fix for the previous commit? 

Please correct me:
- before this PR, `--list-paths` shows only existing files.
- after the previous commit, `--list-paths` shows some non-existing files.
- thanks to this commit `--list-paths` goes back to showing only existing files.

If correct then just squash with the previous commit and avoid the intermediate state.

tests/conftest.py
marc-hb, line 1: Missing space in ` 'tmp_west_topdir'function` commit subject.

tests/test_config.py
marc-hb, line 1: This must be part of the main commit because all the tests must pass in every commit.

marc-hb (changes requested):
Thanks, I think we are getting close! I didn't review every line yet sorry but I think I already have enough comments to share for a first pass.

Note: all the tests must pass in every commit.

tests/test_config.py
pdgendt: I prefer adding new tests in a separate commit, after the commits with functional changes.

tests/conftest.py
pdgendt: This comment is still valid.

src/west/configuration.py
```diff
@@ -130,6 +130,19 @@ def delete(self, option: str):
             self.cp.write(f)
 
 
+class Converter:
```
pdgendt: I think we can make it non-public, typically by prefixing with an underscore.

src/west/configuration.py
```diff
@@ -660,45 +703,50 @@ def _location(cfg: ConfigFile, topdir: PathType | None = None, find_local: bool
...
+            return [os.fspath(Path(west_dir()) / 'config')]
         else:
-            return ''
+            return []
     else:
         raise ValueError(f'invalid configuration file {cfg}')
 
 
-def _gather_configs(cfg: ConfigFile, topdir: PathType | None) -> list[str]:
+def _gather_configs(cfg: ConfigFile, topdir: PathType | None) -> list[Path]:
```
pdgendt: Not sure if we need to change the return type here? You can simply replace `append` with `extend` in this function and be done with it.

tests/test_config.py
marc-hb: Adding new tests in a later commit: no problem of course.
But not breaking existing tests and fixing them later, that's different.

For obvious git reasons, it must be possible to run and pass tests on every commit.

src/west/configuration.py
```diff
@@ -130,6 +130,19 @@ def delete(self, option: str):
             self.cp.write(f)
 
 
+class _Converter:
```
marc-hb: I miss the value of wrapping two small static methods in this new class. Why not just functions? Or inside some existing class.


https://steve-yegge.blogspot.com/2006/03/execution-in-kingdom-of-nouns.html

src/west/configuration.py
```diff
@@ -130,6 +130,19 @@ def delete(self, option: str):
...
+    @staticmethod
+    def str_to_paths(paths: str | None, sep: str = os.pathsep) -> list[Path]:
+        """Split a string into a list of Path objects using the given separator."""
```
marc-hb: I'm concerned pydocs automatically make these part of the API, is that correct? They're too basic and unrelated to the API to be part of it.

src/west/configuration.py
```diff
@@ -130,6 +130,19 @@ def delete(self, option: str):
             self.cp.write(f)
 
 
+class _Converter:
+    @staticmethod
+    def str_to_paths(paths: str | None, sep: str = os.pathsep) -> list[Path]:
```
```suggestion
    def parse_paths(paths: str | None, sep: str = os.pathsep) -> list[Path]:
```

src/west/configuration.py
```diff
@@ -70,19 +75,28 @@ def parse_key(dotted_name: str):
...
-        self.path = path
+    def __init__(self, paths: list[Path]):
         self.cp = _configparser()
-        read_files = self.cp.read(path, encoding='utf-8')
-        if len(read_files) != 1:
-            raise FileNotFoundError(path)
+        self.paths = paths
+        read_files = self.cp.read(self.paths, encoding='utf-8')
+        if len(read_files) != len(self.paths):
+            raise WestNotFound(f"Error while reading one of '{paths}'")
```
```suggestion
            raise MalformedConfig(f"Error while reading one of '{paths}'")
```
All `self.paths` have been verified to `.exists()`, so "NotFound" is IMHO misleading.

src/west/configuration.py
```diff
@@ -113,21 +127,19 @@ def set(self, option: str, value: Any):
 
         self.cp[section][key] = value
 
-        with open(self.path, 'w', encoding='utf-8') as f:
-            self.cp.write(f)
+        self._write()
```
marc-hb, line 132: Sorry but I don't like the fact that you "tentatively" change the value in memory and fail only at the last moment before actually writing. This means the _InternatCF state in memory is "impossible" for some duration. It's OK now but then it's reasonably likely to become a problem in the future whenever someone refactoring happens for some other feature or re-org.

I think you can avoid most of the duplication like this:

```python

def set() / append() / etc.:
    self.single_file_check() # raises exception
    ...
    ...
    self._write()

def _write(self):
     assert len(self.paths) == 1 # (in case some function forgot single_file_check)
     with open(self.paths[0], 'w', encoding='utf-8') as f:
            self.cp.write(f)

def single_file_check():
   assert paths
   len(self.paths) != 1
     raise ValueError(f'Cannot write if multiple configs in use: {self.paths}')
```

src/west/configuration.py
```diff
@@ -113,21 +127,19 @@ def set(self, option: str, value: Any):
...
 
     def delete(self, option: str):
         section, key = _InternalCF.parse_key(option)
 
-        if section not in self.cp:
+        if option not in self:
```
marc-hb: What is the reason for this change? (If it's unrelated to this PR, please submit separately)

src/west/configuration.py
```diff
@@ -186,24 +198,30 @@ def __init__(self, topdir: PathType | None = None):
         :param topdir: workspace location; may be None
         '''
 
-        local_path = _location(ConfigFile.LOCAL, topdir=topdir, find_local=False) or None
+        local_paths = _location(ConfigFile.LOCAL, topdir=topdir, find_local=False) or []
```
marc-hb: I suspect the `or None` was a bit of a hack to simplify the `mypy` of the `_location()` function. 

Is the `or []` still required? If not drop it.

src/west/configuration.py
```diff
@@ -610,7 +650,9 @@ def _rel_topdir_to_abs(p: PathType, topdir: PathType | None) -> str:
     return str(os.path.join(topdir, p))
 
 
-def _location(cfg: ConfigFile, topdir: PathType | None = None, find_local: bool = True) -> str:
+def _location(
```
marc-hb, line 639:
```
def _locations(
```

The comment needs some plural too.

src/west/configuration.py
```diff
@@ -708,11 +753,12 @@ def _gather_configs(cfg: ConfigFile, topdir: PathType | None) -> list[str]:
 def _ensure_config(configfile: ConfigFile, topdir: PathType | None) -> str:
```
marc-hb, line 745: Please add a quick word that this function is used only once by one deprecated function. I lost some time wondering why it seemed to duplicate other code.

tests/conftest.py
marc-hb: ping

tests/test_config.py
```diff
@@ -614,6 +614,127 @@ def test_config_precedence():
...
+    # - some individual option per config file
+    # - the same option defined in multiple configs
+    write_config(config_s1, 'sec', 's', '1 !"$&/()=?', 's1', '1 !"$&/()=?')
+    write_config(config_s2, 'sec', 's', '2', 's2', '2')
+    write_config(config_g1, 'sec', 'g', '1', 'g1', '1')
+    write_config(config_g2, 'sec', 'g', '2', 'g2', '2')
+    write_config(config_l1, 'sec', 'l', '1', 'l1', '1')
+    write_config(config_l2, 'sec', 'l', '2', 'l2', '2')
+
+    # use a non-readable config file (does not work on Windows)
```
```suggestion
    # config file without read permission (does not work on Windows)
```

tests/test_config.py
```diff
@@ -614,6 +614,127 @@ def test_config_precedence():
...
+    if not WINDOWS:
+        config_non_readable = config_dir / 'non-readable'
+        config_non_readable.touch()
+        config_non_readable.chmod(0o000)
+        os.environ["WEST_CONFIG_GLOBAL"] = f'{config_g1}{os.pathsep}{config_non_readable}'
+        _, stderr = cmd_raises('config --global some.section', WestNotFound)
```
marc-hb: "NotFound" is not right...

tests/test_config.py
```diff
@@ -614,6 +614,127 @@ def test_config_precedence():
...
+            conf.write(content)
+
+    # config file paths
+    config_dir = pathlib.Path(config_tmpdir) / 'configs'
+    config_s1 = config_dir / 'system 1'
+    config_s2 = config_dir / 'system 2'
+    config_g1 = config_dir / 'global 1'
+    config_g2 = config_dir / 'global 2'
+    config_l1 = config_dir / 'local 1'
+    config_l2 = config_dir / 'local 2'
```
marc-hb, line 738: Since you already imported Windows, try to break things using the "other" `os.pathsep`

```
  if WINDOWS:
     other_sep = ':'
  else
     other_sep = ';'

   config_l2 = config_dir / 'local 2' + other_sep
```

tests/test_config.py
```diff
@@ -614,6 +614,127 @@ def test_config_precedence():
...
+    stdout = cmd('config --system sec.s2').rstrip()
+    assert stdout == '2'
+    stdout = cmd('config --global sec.g1').rstrip()
+    assert stdout == '1'
+    stdout = cmd('config --global sec.g2').rstrip()
+    assert stdout == '2'
+    stdout = cmd('config --local sec.l1').rstrip()
+    assert stdout == '1'
+    stdout = cmd('config --local sec.l2').rstrip()
+    assert stdout == '2'
```
marc-hb: That's a lot of repetition. Use an array of key,value.

tests/test_config.py
```diff
@@ -614,6 +614,127 @@ def test_config_precedence():
...
+    # The paths may be relative relative paths, which are always anchored to
+    # west topdir. For the test, the cwd is changed to another cwd to ensure
+    # that relative paths are anchored correctly.
+    # - Step 1: if no west topdir exists it should fail as relative path
+    #           cannot be resolved
+    # - Step 2: relative paths are correctly resolved (anchored to west topdir)
+    msg = "'{file}' is relative but 'west topdir' is not defined"
+    cwd = pathlib.Path('any cwd')
+    cwd.mkdir()
+    with chdir(cwd):
```
marc-hb, line 831: I think you could have more coverage with less test code by:
- having some relative files in the test mix, always
- having some top-level test function that runs the ENTIRE test suite twice in two (or three) different directories. These tests do not involve git so I believe they are really quick to run.

marc-hb (commented):
I think you are almost there.

src/west/configuration.py
```diff
@@ -186,24 +198,30 @@ def __init__(self, topdir: PathType | None = None):
...
-        if self._local and location in [ConfigFile.LOCAL, ConfigFile.ALL]:
-            ret.append(self._local.path)
+        if location in [ConfigFile.GLOBAL, ConfigFile.ALL]:
+            ret.extend(self._global_paths)
+        if location in [ConfigFile.SYSTEM, ConfigFile.ALL]:
+            ret.extend(self._system_paths)
+        if location in [ConfigFile.LOCAL, ConfigFile.ALL]:
+            ret.extend(self._local_paths)
+        if existing_only:
+            ret = [p for p in ret if p.exists()]
```
marc-hb: I don't see in which case this can make a difference. The `self._*_paths` have already been filtered in the `_InternalCF.from_paths()` (it's the reason why `from_paths()` exists). I understand why you want this but it does not seem to be happening. Do you have test coverage with missing files? Sorry I did not review the whole test code yet.


EDIT: this is probably one of the places where the existing/non-existing confusion hit me.

src/west/configuration.py
```diff
@@ -296,28 +314,43 @@ def set(self, option: str, value: Any, configfile: ConfigFile = ConfigFile.LOCAL
...
+        def check_configfile(location: ConfigFile) -> Path:
+            '''
+            check that exactly one configfile is in use (even if it not exists yet).
+            Return its path.
+            '''
+            configs = self.get_paths(location)
+            if not configs:
+                raise WestNotFound(f'{configfile}: Cannot determine any config file')
+            if len(configs) > 1:
+                raise ValueError(f'Cannot set value if multiple configs in use: {configs}')
```
marc-hb: This exception is raised in multiple places: raise it from a single "check_single" very small function.

src/west/configuration.py
```diff
@@ -567,18 +600,25 @@ def delete_config(
...
+        considered_locations = [configfile]
     else:
-        to_check = [_location(x, topdir=topdir) for x in configfile]
+        considered_locations = configfile
+
+    # ensure that only one config file is in use
+    all_config_files = [_location(cfg, topdir=topdir) for cfg in considered_locations]
+    for config_files in all_config_files:
+        if len(config_files) > 1:
+            raise ValueError(f'Error: Multiple config paths in use: {config_files}')
```
marc-hb: Is this too restrictive? Let's say I have two GLOBAL files and a single LOCAL file. Now this looks like this blocks me from deleting things in my single LOCAL file, doesn't it?

Maybe this can be done like this:

1. Look for the value in every "location" (GLOBAL,...) requested. Remember where it was found. Do NOT delete anything yet!
2. For each location where it was actually found in step 1, make sure there is single file in the definition of that location. Raise an exception if not. Do not delete anything yet.
3. Finally and atomically delete it from all the location where it was found (and requested).

4. Clearly explain this (or whatever better algorithm you find) in a comment :-)
5. Write some good tests for it.

src/west/configuration.py
```diff
@@ -296,28 +314,43 @@ def set(self, option: str, value: Any, configfile: ConfigFile = ConfigFile.LOCAL
...
         '''
 
+        def check_configfile(location: ConfigFile) -> Path:
+            '''
+            check that exactly one configfile is in use (even if it not exists yet).
+            Return its path.
+            '''
+            configs = self.get_paths(location)
+            if not configs:
+                raise WestNotFound(f'{configfile}: Cannot determine any config file')
```
marc-hb: You passed existing_only=False (the default), so I don't see how this could be raised. This looks like an assert, not an exception.

marc-hb (commented):
EDIT: this was DONE.

- On one hand you want to keep filtering non-existing paths early, at `_InternalCF()` construction time because it's very convenient.
- On the other hand, now you need to remember non-existing files because you (rightfully) want to block modifications of `WEST_CONFIG_x=existing_fileA:non_existing_fileB`. 

So I think that's why you added an new `existing_only` boolean parameter to the `get_paths()` method but that confused me a lot because it has a default value. Also, depending on that parameter `get_paths()` now perform again the `.exists()` filtering that was already performed when constructing the `_InternalCF()` objects... a bit confusing.

I think it would be much clearer to have two different functions:

- `get_path_definitions()` that includes non-existing files 
- `get_existing_paths()` that returns only existing files from already filtered `_InternalCF()` objects.

I bet they won't have any code in common.

Build/compute things only once, name them really well and then re-use them wherever they are needed.

BTW do you have a test that makes sure changes to `WEST_CONFIG_x=existing_fileA:non_existing_fileB` are blocked?

src/west/configuration.py
thorsten-klein: Sorry, but I don't see that they belong to class _InternalCF which says 

```
class _InternalCF:
    # For internal use only; convenience interface for reading and
    # writing INI-style [section] key = value configuration files,
    # but presenting a west-style section.key = value style API.
    # Unlike the higher level and public "Configuration" class, there is
    # one _InternalCF object for each location: LOCAL, GLOBAL,...
```

For understanding: Why are just functions better than functions within a (private) class?

src/west/configuration.py
thorsten-klein: It restricts the configs only for `considered_locations`, which is `[LOCAL]` if `--local` was given. Nevertheless I have added a test for this to ensure that writing of a config is possible (even if the other "levels" use multiple configs)

thorsten-klein:
Found bug #879 while adding support for `--list-search-paths`.
This new flag is extremely helpful when using multiple configs (therefore I have added it as part of this PR), to find out as a user the paths where west does a lookup for config files. In contrast `--list-paths` lists only the existing config files.
I have decided to add it within this PR (but of course separate commit for the feature and the tests) as it is related to this PR. I hope it is fine. Otherwise I will create a separate follow-up PR for this.

src/west/configuration.py
marc-hb: > Why are just functions better than functions within a (private) class?

Because every piece of code must have some purpose? The purpose of a class is to define objects that combine data (fields) and code using that data (methods), that's OoO in a nutshell. If there's no data and just two static methods, why is there a class? This is not Java, classes are not mandatory. BTW top-level functions is one of the many things Kotlin fixed Java: https://kotlinlang.org/docs/functions.html

I have not really thought about what is the best place for these functions, but their _Converter "container" seemed pointless.

> Sorry, but I don't see that they belong to class _InternalCF which says

Maybe I missed something but it looked like these functions were only used with `_InternalCF` data.

thorsten-klein: You are right. The functions are only used inside one other function each. So I have now moved each function to the place where it is used. If there are more places in future, the function(s) can be extracted (to a separate function or class).

src/west/configuration.py
```diff
@@ -70,15 +70,28 @@ def parse_key(dotted_name: str):
...
-        return _InternalCF(path) if path and path.exists() else None
+    def from_paths(paths: list[Path] | None) -> '_InternalCF | None':
+        if not paths:
+            return None
+        paths = [p for p in paths if p.exists()]
+        return _InternalCF(paths) if paths else None
 
-    def __init__(self, path: Path):
-        self.path = path
+    def __init__(self, paths: list[Path]):
```
marc-hb, line 79:
```suggestion
    def __init__(self, paths: list[Path]):
    # Latter elements in the list have precedence as documented by configparser.
```

This is a very important precedence assumption about the configparser here and also in the argument of `__init__` and `__from_paths`. It must be clearly spelled out.

src/west/configuration.py
```diff
@@ -70,15 +70,28 @@ def parse_key(dotted_name: str):
...
+        if len(read_files) != len(self.paths):
+            raise MalformedConfig(f"Error while reading one of '{paths}'")
+
+    def _check_single_config(self):
+        assert self.paths
+        if len(self.paths) > 1:
+            raise ValueError(f'Cannot write if multiple configs in use: {self.paths}')
+
+    def _write(self):
+        assert len(self.paths) == 1  # (in case some function forgot to check)
```
```suggestion
        assert len(self.paths) == 1  # callers are expected to _check_single_config()
```

src/west/configuration.py
```diff
@@ -173,26 +186,36 @@ def __init__(self, topdir: PathType | None = None):
...
+            ret.extend(self._system_paths)
+        if location in [ConfigFile.GLOBAL, ConfigFile.ALL]:
+            ret.extend(self._global_paths)
+        if location in [ConfigFile.LOCAL, ConfigFile.ALL]:
+            ret.extend(self._local_paths)
         return ret
 
+    def get_existing_paths(self, location: ConfigFile = ConfigFile.ALL) -> list[Path]:
+        paths = self.get_search_paths(location)
+        return [p for p in paths if p.exists()]
```
marc-hb: The `_InternalCF` objects have already filtered this, why not simply concatenate their `.paths` fields`? That keeps the filter in a single place and is a good reminder of what `_InternalCF` are.

src/west/configuration.py
```diff
@@ -173,26 +186,36 @@ def __init__(self, topdir: PathType | None = None):
...
+
+    @staticmethod
+    def get_location_paths(
+        configfile: ConfigFile, topdir: PathType | None, find_local: bool = True
```
```suggestion
               loc: ConfigFile, topdir: PathType | None, find_local: bool = True
```

marc-hb: This is a static method with a nice looking name, but in practice this only converting  a list of strings to a list of paths and it's used in only one place. Also, this wrapper "shadows" a little bit the great description just below `def _location()` Can you give this a less official / API looking name and embed it inside `def __init__()`, just before the only place where it is used? Similar to how your place `parse_paths()`

src/west/configuration.py
```diff
@@ -287,36 +310,40 @@ def set(self, option: str, value: Any, configfile: ConfigFile = ConfigFile.LOCAL
...
                 raise ValueError(
                     f'{configfile}: file not found; retry in a workspace or set WEST_CONFIG_LOCAL'
                 )
-            if not self._local_path.exists():
-                self._local = self._create(self._local_path)
-            if TYPE_CHECKING:
-                assert self._local
-            self._local.set(option, value)
+
+        # ensure that onle one config is in use
```
```suggestion
        # ensure that only one config is in use
```

src/west/configuration.py
```diff
@@ -287,36 +310,40 @@ def set(self, option: str, value: Any, configfile: ConfigFile = ConfigFile.LOCAL
...
-                assert self._system
-            self._system.set(option, value)
-        else:
-            # Shouldn't happen.
-            raise AssertionError(configfile)
+            attr = '_global'
+        elif configfile == ConfigFile.LOCAL:
+            attr = '_local'
+
+        # create config file if it does not already exist and use it
```
marc-hb, line 334:
```suggestion
        # create config file and _InternalCF if it does not already exist and use it
```
Why not check `if hasattr(self, attr)` directly?

Once you've created objects, made measurements or established facts, use, re-use and abuse them again. Do not re-compute, re-filter or re-create the same thing again. This avoids duplication and divergence, reduces races and makes the objects' role more obvious.

src/west/configuration.py
```diff
@@ -287,36 +310,40 @@ def set(self, option: str, value: Any, configfile: ConfigFile = ConfigFile.LOCAL
...
-            if not self._local_path.exists():
-                self._local = self._create(self._local_path)
-            if TYPE_CHECKING:
-                assert self._local
-            self._local.set(option, value)
+
+        # ensure that onle one config is in use
+        configs = self.get_search_paths(configfile)
+        if len(configs) > 1:
+            raise ValueError(f'Cannot set value if multiple configs in use: {configs}')
```
marc-hb, line 321: Rather than searching again, why not check if `_InternalCF` already exists and then call its `_check_single_config()`? This seems to be duplicating both the search and the exception.

src/west/configuration.py
marc-hb: ping?

src/west/configuration.py
```diff
@@ -554,18 +581,20 @@ def delete_config(
...
     elif isinstance(configfile, ConfigFile):
-        to_check = [_location(configfile, topdir=topdir)]
+        considered_locations = [configfile]
     else:
-        to_check = [_location(x, topdir=topdir) for x in configfile]
+        considered_locations = configfile
+    assert isinstance(considered_locations, list)
+
+    # ensure that only one config file is in use for the considered_locations
+    to_check = [f for cfg in considered_locations for f in _location(cfg, topdir=topdir)]
```
marc-hb, line 607: Please correct me but `to_check` is flattening the hierarchy, collapsing all files from all locations into a single flat list, right?

This sounds like losing the information needed to address my comment in https://github.com/zephyrproject-rtos/west/pull/867#discussion_r2505938505

> Let's say I have two GLOBAL files and a single LOCAL file. Now this looks like this blocks me from deleting things in my single LOCAL file, doesn't it?

EDIT: nevermind, I missed this is the deprecated API. It does not need to support multiple files.

marc-hb (commented):
Still one major concern about `delete_config()` + some other style issues.

The `delete_config()` API has unfortunately pretty complicated semantics. Fortunately: they are well documented - but for a single file per location. I don't think you can get away with a bottoms-up, "code first" approach for this one. `delete_config()` definitely needs a step and a clear, high-level definition of the new algorithm.

marc-hb:
> Still one major concern about delete_config() + some other style issues.  The delete_config() API has unfortunately pretty complicated semantics.

I just realized `delete_config()` is the _deprecated_ API, sorry!

The newer and current `delete()` API did not require any change? Magic! Proof it's a better design than the deprecated one :-) The lack of change in `delete()` threw me off.

So, now realizing that `delete_config()` is deprecated, then I'm more than happy to raise an exception if ANY `_InternalCF` has multiple files - even one location that was not requested. In other words: people who want to use your new feature (multiple files per location) MUST stop using `delete_config()`. That's a great incentive to move people off the deprecated API. 

Assuming this makes the code simpler (I did not check), you can just invoke `_check_single_config()` on every `_InternalCF` that exists and then keep the rest of the code mostly unchanged. Keep it simple, minimal change?

src/west/app/config.py
```diff
@@ -169,14 +182,16 @@ def do_run(self, args, user_args):
         if args.list:
             if args.name:
                 self.parser.error('-l cannot be combined with name argument')
-        elif not args.name and not args.list_paths:
+        elif not any([args.name, args.list_paths, args.list_search_paths]):
```
pdgendt: Is this even correct (for `args.list_paths` too)? I don't think `args.name` can be used with `--list-paths` or `--list-search-paths`?

If so, these need to be move to the previous `if`.

src/west/configuration.py
thorsten-klein: I think it is fine. No search happens here, but only the list of search paths is returned. I cannot get this work with `_check_single_config`.

src/west/app/config.py
```diff
@@ -166,17 +179,19 @@ def do_add_parser(self, parser_adder):
 
     def do_run(self, args, user_args):
         delete = args.delete or args.delete_all
-        if args.list:
+        if any([args.list, args.list_paths, args.list_search_paths]):
             if args.name:
                 self.parser.error('-l cannot be combined with name argument')
```
pdgendt: The error message isn't accurate in all cases 😕

src/west/app/config.py
```diff
@@ -166,17 +179,25 @@ def do_add_parser(self, parser_adder):
...
+            if args.list:
+                flag = '-l'
+            elif args.list_paths:
+                flag = '--list-paths'
+            elif args.list_search_paths:
+                flag = '--list-search-paths'
             if args.name:
-                self.parser.error('-l cannot be combined with name argument')
-        elif not args.name and not args.list_paths:
+                self.parser.error(f'{flag} cannot be combined with name argument')
```
pdgendt: Move the flag afte the `if args.name:` or maybe

```python
        if any([args.list, args.list_paths, args.list_search_paths]):
            if args.name:
                self.parser.error(
                    (
                        '-l'
                        if args.list
                        else '--list-paths'
                        if args.list_paths
                        else '--list-search-paths'
                    )
                    + ' cannot be combined with name argument'
                )

```

pdgendt (approved):

src/west/configuration.py
marc-hb: You are correct: you could have multiple paths at this point but no file and no `_InternalCF` object yet.

src/west/configuration.py
```diff
@@ -173,26 +187,34 @@ def __init__(self, topdir: PathType | None = None):
...
         '''
 
-        local_path = _location(ConfigFile.LOCAL, topdir=topdir, find_local=False) or None
-
-        self._system_path = Path(_location(ConfigFile.SYSTEM, topdir=topdir))
-        self._global_path = Path(_location(ConfigFile.GLOBAL, topdir=topdir))
-        self._local_path = Path(local_path) if local_path is not None else None
+        self._local_paths = [Path(p) for p in _location(ConfigFile.LOCAL, topdir, find_local=False)]
+        self._global_paths = [Path(p) for p in _location(ConfigFile.GLOBAL, topdir)]
+        self._system_paths = [Path(p) for p in _location(ConfigFile.SYSTEM, topdir)]
```
marc-hb: `self._system_paths` and `self._system.paths` have only one character difference, this keeps misleading me. It just requires too much attention to detail. Maybe it stopped being a problem after investing a lot of time in that code and your brain gets trained to it but I think casual readers deserve a more obvious difference. It was less of a problem before because we did not have `-list-search-paths`, multiple files etc., the world was much simpler. Also, you are already searching/replacing these names anyway, so it is the right time to make a clearer difference. How about... `self._system_locs`? Or, anything better as long as it has more a couple characters difference.

src/west/app/config.py
```diff
@@ -49,6 +49,11 @@
 providing one of the following arguments:
     --local | --system | --global
 
+For each configuration level (local, global, and system) also multiple config
```
marc-hb: Nit:
```suggestion
For each configuration level (local, global, and system) multiple config
```

src/west/configuration.py
```diff
@@ -441,6 +467,13 @@ def _deprecated(old_function):
     )
 
 
+def _ensure_single_config(topdir):
```
marc-hb: 1. This name misses a plural 's'.
```suggestion
def _ensure_one_file_max_configs(topdir):
```
... or any better name with a plural configs.

2. It's not "single" but "1 max"

src/west/configuration.py
```diff
@@ -552,20 +587,23 @@ def delete_config(
...
+        considered_locations = [ConfigFile.SYSTEM, ConfigFile.GLOBAL, ConfigFile.LOCAL]
     elif isinstance(configfile, ConfigFile):
-        to_check = [_location(configfile, topdir=topdir)]
+        considered_locations = [configfile]
     else:
-        to_check = [_location(x, topdir=topdir) for x in configfile]
+        considered_locations = configfile
+    assert isinstance(considered_locations, list)
+
+    # ensure that only one config file is in use for the considered_locations
```
```suggestion
    # thanks to _ensure_one_file_max_configs() above, we don't actually have 2 levels here.
```

tests/test_config.py
```diff
@@ -19,6 +19,18 @@
...
+    SYSTEM: 'WEST_CONFIG_SYSTEM',
+    GLOBAL: 'WEST_CONFIG_GLOBAL',
+    LOCAL: 'WEST_CONFIG_LOCAL',
+}
+
+west_flag = {
+    SYSTEM: '--system',
+    GLOBAL: '--global',
+    LOCAL: '--local',
+}
```
marc-hb, line 33: I've been wondering why we don't have more dicts like this in the production code too... but that's for another day!

marc-hb (commented):
I don't have any major concern anymore, thank you!

I would really like `_system.paths` and `_system_paths` to have more than one character difference for readability.

Other minor comments and suggestions below but less important.

marc-hb:
Forgot to say: I reviewed all my past comments and "resolved" as many as possible. I left a few open, either because they were still valid, or because I wasn't sure. Please take one last look, there should be a really small number left open.

tests/test_config.py
thorsten-klein: I don't understand. Maybe some topic for follow-up or refactoring at some later point.

src/west/configuration.py
thorsten-klein: It is outdated, right?

marc-hb:
> `self._system_paths` and `self._system.paths` have only one character difference, this keeps misleading me. [...] How about... `self._system_locs?` Or, anything better as long as it has more than a couple characters difference.

```python
        self._local_paths = [Path(p) for p in _location(ConfigFile.LOCAL, topdir, find_local=False)]
        self._global_paths = [Path(p) for p in _location(ConfigFile.GLOBAL, topdir)]
        self._system_locs = [Path(p) for p in _location(ConfigFile.SYSTEM, topdir)]
```

Come on... same thing with `self._local_paths` vs `self._local.paths`, and `self._global_paths` vs `self._global.paths` :-( I didn't say because it seemed obvious to me.

You didn't see the inconsistency of replacing just one? Did you search/replace without looking at the code at all?

Am I actually spending more time reading your code than you are? Writing it does not count, now I'm not sure you realize that.

marc-hb (approved):

PR #869, opened 2025-10-17
Support remapping of west projects urls

thorsten-klein:
Proposal for #615

Support is added for a new `remapping` section in the west manifest.

Under the `url` key, users can define search-and-replace patterns (whereby `old` will be replaced with `new`).
Those patterns are applied to all project URLs during project import (recursively).
This allows downstream projects to modify remote URLs while keeping all other metadata (e.g. original `revision`).
This is extremely helpful when downstream projects just want to use mirrored repositories.

## Usage
In general you can specify search-replace patterns for `url` (or substrings from `url`) via `west.yml`.
It is a simple search-replace (no regex).
This means, for example you can add following remapping-block to your `west.yml` in case you have mirrored all zephyr repositories into your own "mirrors" organization in your company Github instance:

```
manifest:
  remapping:
    url:
      - old: github.com/zephyr-rtos
        new: github.company.com/mirrors
```

If the repositories are mirrored to different Github organizations or instances, you can apply multiple (more fine-granular) remapping-patterns.

```
manifest:
  remapping:
    url:
      - old: github.com/foo/libfoo
        new: github.company.com/foo/mirrored-libfoo
      - old: github.com/foo/libbar
        new: github.othercompany.com/forked/libbar
```

### DIFF
--- a/src/west/manifest-schema.yml
+++ b/src/west/manifest-schema.yml
@@ -66,6 +66,25 @@ mapping:
             required: true
             type: str
 
+  # allow some remapping of values by downstream projects
+  remapping:
+    required: false
+    type: map
+    mapping:
+      # search-replace within URLs (e.g. to use mirror URLs)
+      url:
+        required: false
+        type: seq
+        sequence:
+          - type: map
+            mapping:
+              old:
+                required: true
+                type: str
+              new:
+                required: true
+                type: str
+
   # The "projects" key specifies a sequence of "projects", each of which has a
   # remote, and may specify additional configuration.
   #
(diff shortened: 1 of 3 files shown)


marc-hb:
Do you think you could focus on the `west config` approach first? Because it should be smaller  and easier to experiment with, review, test and use in anger etc., while sharing the code with any manifest-based approach later. As discussed in #615, imports can be very tricky and very error-prone.

Obviously, your #867 will make this even easier.

thorsten-klein:
I'm closing this PR for now. We'll let the feature mature in our fork. I will reopen this proposal again once it has proven its stability.

marc-hb:
New attempt:
- #994

marc-hb:
This #869 replaced by #995.

PR #870, opened 2025-10-21
support 'west config --list-paths'

thorsten-klein:
Extracted from #867

# Why?

It is helpful for users to see which configuration files are currently being used by `west`, and in which order they are applied.  
This makes it easier to debug configuration issues and understand where specific settings originate from.

# Proposal

Introduce a new command-line option `--list-paths`, which prints all configuration files that are currently considered and exist. 
The files are listed in the order they are loaded, as later files override values from earlier ones.

This option can be combined with `--local`, `--global`, or `--system` to list only the path for a specific configuration level.

### DIFF
--- a/src/west/app/config.py
+++ b/src/west/app/config.py
@@ -35,17 +35,28 @@ Local files:
 
 - Linux, macOS, Windows: <workspace-root-directory>/.west/config
 
-You can override these files' locations with the WEST_CONFIG_SYSTEM,
-WEST_CONFIG_GLOBAL, and WEST_CONFIG_LOCAL environment variables.
-
 Configuration values from later configuration files override configuration
 from earlier ones. Local values have highest precedence, and system values
 lowest.
 
-To get a value for <name>, type:
+The config file location for each according config level can be changed with
+environment variables:
+- WEST_CONFIG_SYSTEM
+- WEST_CONFIG_GLOBAL
+- WEST_CONFIG_LOCAL
+
+All `west config` commands can be applied for a specific configuration level by
+providing one of the following arguments:
+    --local | --system | --global
+
+The following command prints a list of all configuration files currently
+considered and existing (listed in the order as they are loaded):
+    west config --list-paths
+
+To get the value for config option <name>, type:
     west config <name>
 
-To set a value for <name>, type:
+To set a value for config option <name>, type:
     west config <name> <value>
 
 To append to a value for <name>, type:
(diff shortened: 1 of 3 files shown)


pdgendt:
The command does not appear to work correctly in my Zephyr workspace (my local config file is not printed):

```shell
$ cat .west/config
[manifest]
path = zephyr
file = west.yml
group-filter = +optional

[zephyr]
base = zephyr

$ west config --list-paths
/home/pdgendt/.westconfig

$ west config --list-paths --local

$ west config manifest.path
zephyr
```

src/west/app/config.py
```diff
@@ -170,6 +188,11 @@ def do_run(self, args, user_args):
         else:
             self.write(args)
 
+    def list_paths(self, args):
+        config_paths = Configuration().get_paths(args.configfile or ALL)
```
pdgendt: I think you need this:

```suggestion
        config_paths = self.config.get_paths(args.configfile or ALL)
```

thorsten-klein: Thank you! The tests only covered the case using environment variables.
I’ve now extended it to also catch the scenario where the local config file is determined from the file system.

src/west/app/config.py
```diff
@@ -35,17 +35,28 @@
...
 Configuration values from later configuration files override configuration
 from earlier ones. Local values have highest precedence, and system values
 lowest.
 
-To get a value for <name>, type:
+The config file location for each according config level can be changed with
+environment variables:
+- WEST_CONFIG_SYSTEM,
+- WEST_CONFIG_GLOBAL
+- WEST_CONFIG_LOCAL
```
pdgendt:
```suggestion
- WEST_CONFIG_SYSTEM
- WEST_CONFIG_GLOBAL
- WEST_CONFIG_LOCAL
```

pdgendt (commented):
Looks good, small nit

pdgendt (approved):

src/west/configuration.py
```diff
@@ -181,6 +181,16 @@ def __init__(self, topdir: PathType | None = None):
         self._global = _InternalCF.from_path(self._global_path)
         self._local = _InternalCF.from_path(self._local_path)
 
+    def get_paths(self, configfile: ConfigFile = ConfigFile.ALL):
```
```suggestion
    def get_paths(self, locations: ConfigFile = ConfigFile.ALL):
```

:-)

marc-hb (approved):

Issue #852, opened 2025-10-01
Feature Request: use a local west project (without any remote)


thorsten-klein:
### Problem  
The `west update` command currently performs **two coupled steps** for all projects listed in the manifest:  
1. Clones the repository into the specified location.  
2. Sets up and integrates the project.  
  
This design assumes that every project must be cloned by `west`. However, in some workflows, the project repository may already exist at the target path—e.g. because it is managed by other tools (e.g. Git submodules, Google’s `repo`, or a monorepo workflow).  
  
### Expected Behavior  
`west` should allow using an already-present project directory without requiring to clone it.  
  
### Proposal  
Introduce support for `remote: local` in the manifest (or a new option `local: True`).  
- When set, `west` skips the cloning step and directly integrates the existing project in-place.  
- This makes `west` more flexible in environments where projects are not exclusively managed by `west`.  
  
### Motivation / Use Cases  
1. **Other manifest tools in use**  
   Some teams already use external tools (e.g. Git submodules, Google’s `repo`) to manage multiple repositories. For them, `west` should integrate with these existing setups instead of forcing a re-clone with west.  
  
2. **Monorepo development**  
   During development, west projects may live inside a monorepo instead of separate Git repositories. First at release time, the monorepo is split into individual repositories (e.g. with `copybara`) that are then consumed as `west` projects (as they are supported now). But for developing the west projects within the monorepo this new feature is required.  
  
3. **Full Platform Monorepo** (A mix of 1 and 2) 
   A large software platform may be developed as a single repository containing everything from the platform (code, configs, scripts, and Dockerfile). Customers integrate this platform repo as git submodule because they need some files. E.g. they need build scripts or Docker container from the platform so that they can run `west`. In this case, `west` should support to use the local west projects which are part of the full platform.  

What do you think about this?
Will you accept such a contribution?
Is there anything I should have in mind while implementing it (in order to avoid too many review rounds)?

pdgendt:
One question comes to mind. If the project is a repository and the revision matches with the manifest, it's already possible to keep this all local. Isn't this sufficient?

thorsten-klein:
When we develop in a monorepo, the project is not a repository. 
Therefore we can neither use a revision nor a remote url in the west manifest.

The only revision we have is the one of the monorepo, and this revision changes from commit to commit, so we cannot put a fix revision in the manifest.

I would expect that we can write in our `west.yml`:
```
    - name: our-platform
      path: ../platform
```

But it fails with:
[...]

PR #872, opened 2025-10-26
Add support for local subprojects in west manifest

thorsten-klein:
# Why?

This PR is a follow-up to west’s [issue #852](https://github.com/zephyrproject-rtos/west/issues/852).

We are developing multiple Zephyr projects within a single repository, which we call our **_platform_**.
Some directories might also be git submodules.
All in all this means, our platform consists of several west projects (let's name them **_local subprojects_**), developed and maintained together in one repository.

Why do we use one repository for multiple subprojects?
- During platform development, we want to have atomic pull requests which may contain changes across multiple subprojects. We want to avoid, that we have to sync multiple pull-requests in various repositories during development.
- We want to ensure that there is only one consistent platform during development. By developing all subprojects in one and the same repository we avoid different combinatorics during development. All subprojects are tightly coupled and integrated during development. We have exactly one combination of subprojects, which we test during development.
- Nevertheless, our downstream projects want full flexibility. Therefore we extract each subproject from platform into a separate repository on platform releases. See later.

## Platform Structure

```
├── apps
│   ├── app-1
│   │   ├── CMakeLists.txt
│   │   ├── prj.conf
│   │   └── Kconfig
│   └── app-2
│       ├── CMakeLists.txt
│       ├── prj.conf
│       └── Kconfig
├── subprojects
│   ├── subproject_a
│   │   ├── west.yml
│   │   ├── prj.conf
│   │   ├── CMakeLists.txt
│   │   └── zephyr
│   │       └── module.yml
│   └── subproject_b
│       ├── west.yml
│       ├── prj.conf
│       ├── CMakeLists.txt
│       └── zephyr
│           └── module.yml
├── west.yml
└── zephyr
    └── module.yaml
```

## Subproject extraction

When releasing the platform, we extract each local subproject (`subproject_a` and `subproject_b`) into separate repositories (e.g. by using `copybara`). This allows downstream projects to consume platform in two ways:
- Consume the full platform as a single entity (recommended combination of subprojects), **or**
- Combine subprojects individually (on own risk as they deviate from the recommended combination of subprojects)

## Expected Platform Integration
Downstream projects typically integrate the platform in one of the following ways:

1. **Using west:** (platform as single entity)
Downstream projects add the platform as a project in their `west.yml` (via remote url from platform).
`west update` clones the platform, which contains all the subprojects.
The dependencies of platform subprojects are cloned accordingly.
On demand, downstream project can override a single subproject individually.

2. **Using west:** (individual combination of subprojects)
Downstream projects add each subproject as a project in their `west.yml` (via remote url from extracted repository).
`west update` clones each subproject and its dependencies accordingly.

3. **Using other means (e.g. git submodules):** (advanced use case, just to mention here)
Downstream projects place the platform source code directly in their repository, e.g. as `git submodule`.
They work with the platform as if it is a local subproject.
Downstream projects still rely on using west commands, since they for example use `west update` to fetch the dependencies of the platform.

### The Problem

Currently, we cannot work with multiple local subprojects in our platform repository.
We cannot declare them as west projects in `west.yml`, since they lack a remote or URL.

## First try
We can use submanifests (`self: import: ...`) to import the `west.yml` of each subproject.
But when building in zephyr, this does not fully work, since we must manually enable each subproject as extra Zephyr module by one of the following methods:
- Setting `ZEPHYR_EXTRA_MODULES` in the environment
- Passing `-DZEPHYR_EXTRA_MODULES` directly via `west build -- ...`
- Defining `ZEPHYR_EXTRA_MODULES` within each sample’s `CMakeLists.txt`

This approach is cumbersome, especially when subprojects are stacked.
Additionally it is not possible to override such submanifests in downstream projects.

## Second try
Currently, extra modules in zephyr are read only from environment variables.
Reference: [PR #91101](https://github.com/zephyrproject-rtos/zephyr/pull/91101)

A possible solution of the problem could be to extend zephyr (not west!) to support such subprojects.
The extra zephyr modules could be declared in a file, which is automatically recognized by `west build` (respectively script `zephyr_module.py`):
- Introduce a new file, such as `zephyr.yml`, **or**
- Extend the existing `west.yml` with a new section for extra modules.

**Example:**
```yaml
zephyr:
  extra-modules:
  - name: module-a
    path: modules/module-a
  - name: module-b
    path: modules/module-b
```

Zephyr would then consider those directories as extra modules (additional to the ones specified in environment variables).
Challenges:
- Zephyr must process all those modules recursively (so that also the extra modules from extra modules are enabled).
- The `west.yml` from each subproject is not reused as the subproject is just consumed as a zephyr module.
  This means: `west update` will not automatically clone the west projects from such extra modules, but all those projects must be added to the downstream `west.yml`.

As `west update` should work out-of-the-box, I come to the conclusion, that this feature should be added in `west` (not `zephyr`).

## Final Proposal (This PR): Support local subprojects in west.yml

West could supports such local subprojects (developed within the same repository) in a standard, integrated way.
`west.yml` could allow to declare *local* projects using a special remote identifier `remote = local`.

- These local projects would be handled like any other west project.
- The only difference: west skips cloning of such a local project (since it already exists locally).
- The project tree contains local subprojects, e.g. when `west list` is called.
  - This approach would also allow local projects to be **individually overridden** downstream
  - The local subprojects are contained in `manifest.get_projects()` which is the function called by zephyr).
  - All dependencies (even from local subprojects) are automatically cloned if they have a remote.
- It does not break existing behavior.
- The built-in `west` functions that are based on `git` may ignore those local subprojects (`west diff`, `west grep`, `west status`, ...).
- `west forall` should be configurable. It may consider all projects (default) or only local projects (`--local` / `--no-remote`) or only projects with a remote (`--remote` / `--no-local`)

### DIFF
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -103,6 +103,8 @@ def test_workspace(west_update_tmpdir):
     # places.
     wct = west_update_tmpdir
     assert wct.check(dir=1)
+    assert wct.join('.west').check(dir=1)
+    assert wct.join('.west', 'config').check(file=1)
     assert wct.join('subdir', 'Kconfiglib').check(dir=1)
     assert wct.join('subdir', 'Kconfiglib', '.git').check(dir=1)
     assert wct.join('subdir', 'Kconfiglib', 'kconfiglib.py').check(file=1)
@@ -117,6 +119,174 @@ def test_workspace(west_update_tmpdir):
     assert wct.join('zephyr', 'CODEOWNERS').check(file=1)
     assert wct.join('zephyr', 'include', 'header.h').check(file=1)
     assert wct.join('zephyr', 'subsys', 'bluetooth', 'code.c').check(file=1)
+
+
+def test_workspace_local(repos_tmpdir):
+    remote_repos = repos_tmpdir / 'repos'
+    projects_dir = repos_tmpdir / 'projects'
+    projects_dir.mkdir()
+
+    # create a local base project with a west.yml
+    project_base = projects_dir / 'base'
+    project_base.mkdir()
+    with open(project_base / 'west.yml', 'w') as f:
+        f.write(
+            textwrap.dedent(f'''\
+            manifest:
+              remotes:
+                - name: upstream
+                  url-base: {os.path.dirname(remote_repos / 'zephyr')}
+              projects:
+              - name: zephyr
+                remote: upstream
+                path: zephyr-rtos
+                import: True
+        ''')
+        )
+
+    # create another project with another west.yml (stacked on base)
+    project_middle = projects_dir / 'middle'
+    project_middle.mkdir()
+    with open(project_middle / 'west.yml', 'w') as f:
+        f.write(
+            textwrap.dedent('''\
+            manifest:
+              projects:
+              - name: base
+                remote: local
+                path: ../base
+                import: true
+        ''')
+        )
+
+    # create another project with another west.yml (stacked on base)
+    project_app = projects_dir / 'app'
+    project_app.mkdir()
+    with open(project_app / 'west.yml', 'w') as f:
+        f.write(
+            textwrap.dedent('''\
+            manifest:
+              projects:
+              - name: middle
+                remote: local
+                path: ../middle
+                import: true
+        ''')
+        )
+
+    # init workspace in projects_dir (project_app's parent)
+    cmd(['init', '-l', project_app])
+
+    # update workspace in projects_dir
+    cmd('update', cwd=projects_dir)
+
+    ws = projects_dir
+    # zephyr projects from base are cloned
+    for project_subdir in [
+        Path('subdir') / 'Kconfiglib',
+        'tagged_repo',
+        'net-tools',
+        'zephyr-rtos',
+    ]:
+        assert (ws / project_subdir).check(dir=1)
+        assert (ws / project_subdir / '.git').check(dir=1)
+
+    # list projects
+    stdout = cmd('list', cwd=projects_dir)
+    stdout_split = [line.split() for line in stdout.splitlines()]
+    expected = [
+        ['manifest', 'app', 'HEAD', 'N/A'],
+        ['middle', 'middle', 'N/A', 'N/A'],
+        ['base', 'base', 'N/A', 'N/A'],
+        ['zephyr', 'zephyr-rtos', 'master', f'{remote_repos}/zephyr'],
+        ['Kconfiglib', 'subdir/Kconfiglib', 'zephyr', f'{remote_repos}/Kconfiglib'],
+        ['tagged_repo', 'tagged_repo', 'v1.0', f'{remote_repos}/tagged_repo'],
+        ['net-tools', 'net-tools', 'master', f'{remote_repos}/net-tools'],
+    ]
+    assert str(stdout_split) == str(expected)
+
+    # -----------------------------
+    # Test "forall"
+    # -----------------------------
+    exp_forall_remote = [
+        '=== running "" in zephyr (zephyr-rtos):',
+        '=== running "" in Kconfiglib (subdir/Kconfiglib):',
+        '=== running "" in tagged_repo (tagged_repo):',
+        '=== running "" in net-tools (net-tools):',
+    ]
+    exp_forall_local = [
+        '=== running "" in middle (middle):',
+        '=== running "" in base (base):',
+    ]
+
+    # run forall (default)
+    stdout = cmd(['forall', '-c', ''], cwd=projects_dir)
+    assert stdout.splitlines() == exp_forall_local + exp_forall_remote
+
+    # run forall --no-remote
+    stdout = cmd(['forall', '--no-remote', '-c', ''], cwd=projects_dir)
+    assert stdout.splitlines() == exp_forall_local
+
+    # run forall --no-local
+    stdout = cmd(['forall', '--no-local', '-c', ''], cwd=projects_dir)
+    assert stdout.splitlines() == exp_forall_remote
+
+    # run forall --local
+    stdout = cmd(['forall', '--local', '-c', ''], cwd=projects_dir)
+    assert stdout.splitlines() == exp_forall_local
+
+    # run forall --remote
+    stdout = cmd(['forall', '--remote', '-c', ''], cwd=projects_dir)
+    assert stdout.splitlines() == exp_forall_remote
+
+    # run forall --no-remote --no-local
+    stdout = cmd(['forall', '--no-remote', '--no-local', '-c', ''], cwd=projects_dir)
+    assert stdout.splitlines() == []
+
+    # -----------------------------
+    # Test "status"
+    # -----------------------------
+    stdout = cmd(['status'], cwd=projects_dir)
+    assert stdout == ''
+
+    # -----------------------------
+    # Test "compare"
+    # -----------------------------
+    stdout = cmd(['compare'], cwd=projects_dir)
+    assert stdout == ''
+
+    # -----------------------------
+    # Test "diff"
+    # -----------------------------
+    stdout = cmd(['diff'], cwd=projects_dir)
+    assert stdout.rstrip() == 'Empty diff in 4 projects.'
+
+    # -----------------------------
+    # Test "grep"
+    # -----------------------------
+    stdout = cmd(['grep', 'dummy string to grep'], cwd=projects_dir)
+    assert stdout == ""
+
+    # -----------------------------
+    # Test "get_projects()"
+    # -----------------------------
+    manifest = Manifest.from_topdir(topdir=ws)
+    projects = manifest.get_projects([])
+    assert len(projects) == 7
+    mp, p1, p2, p3, p4, p5, p6 = projects
+    for p in projects:
+        assert Path(p.abspath).exists()
+    assert mp.name == 'manifest'
+    assert p1.name == 'middle'
+    assert p2.name == 'base'
+    assert p3.name == 'zephyr'
+    assert p4.name == 'Kconfiglib'
+    assert p5.name == 'tagged_repo'
+    assert p6.name == 'net-tools'
+    for p in [p1, p2]:
+        assert p.is_local()
+    for p in [mp, p3, p4, p5, p6]:
+        assert not p.is_local()
 
 
 def test_list(west_update_tmpdir):


marc-hb (changes requested):
The core feature of west is to _manage multiple git repos_. Yet this PR is about supporting monorepos?! This does not make more sense to me than #852

> This means, our platform consists of several west projects (let's name them local subprojects), developed and maintained together in one repository.

No: by definition, a west project _is_ a git repo. 
https://docs.zephyrproject.org/latest/develop/west/manifest.html#projects
We will never understand each other if you start by changing the meaning of the most basic west vocabulary. Do you mean "Zephyr module" instead?

BTW I find your name "local" quite misleading considering all code seems to come from the same, "remote" monorepo.


> we want to have atomic pull requests which may contain changes across multiple subprojects. 
> [...]
> By developing all subprojects in one and the same repository we avoid different combinatorics during development. All subprojects are tightly coupled and integrated during development. We have exactly one combination of subprojects, which we test during development.

"exactly one combination" is what a manifest does already. You configure all the exact versions you want in the manifest and that's it: you've locked down the combinatorial. You don't need monorepo a control versions. `west` exists precisely to control multiple versions _without_ a monorepo. This is the main and most important west feature.

Or, if you prefer a monorepo, then you don't need a manifest and you don't need west. Do not use west with monorepos, invoke the build system directly. Problem solved.

https://docs.zephyrproject.org/latest/develop/west/without-west.html

> When releasing the platform, we extract each local subproject (subproject_a and subproject_b) into separate repositories (e.g. by using copybara).

Moving code across git repos is not a normal use case, it's a dirty hack. Once code is shared, git history is not supposed to change:

https://static.lwn.net/kerneldoc/maintainer/rebasing-and-merging.html#rebasing

> History that has been exposed to the world beyond your private system should usually not be changed. 

https://git-scm.com/docs/git-tag#_on_re_tagging

> The insane thing. You really want to call the new version "X" too, even though others have already seen the old one. 

Etc.


> The Problem
>
> Currently, we cannot work with multiple local subprojects in our platform repository.
> We cannot declare them as west projects in west.yml, since they lack a remote or URL.

> We can use submanifests (self: import: ...) to import the west.yml of each subproject.

"import" _for what reason_? To get what data and fix which problem? This is the most important question but you keep avoiding it and I keep having to guess https://github.com/zephyrproject-rtos/west/issues/852#issuecomment-3357863870


What you named "The problem" is actually "my solution X does not work" but it's still never 100% clear WHY you want to do X in the first place. What happens when you do NOT do X? What breaks and how? 

The essence of a manifest is to be a list of git versions. If you have a monorepo, then you don't need manifests. If not git versions, then what ELSE do you want from manifests? What data? What is missing? That is the "real" problem.



west is a git tool that is designed to be independent of Zephyr. `west` is not a build tool. It supports extensions that you can use to build but that's different. In the monorepo case, all your code has been downloaded and west's job is done. You don't need to import manifest files that do not make any sense in the monorepo case.



> Setting ZEPHYR_EXTRA_MODULES ...
> This approach is cumbersome, especially when subprojects are stacked.

I don't find this cumbersome at all compared to the enormous magnitude of the problem you created for yourself by constantly shuffling code across repos. It's actually surprising you can solve the problem so simply, with just ZEPHYR_EXTRA_MODULES.

Are you just trying to restore the output of `west list` that you've lost by switching to a monorepo? `west list` is NOT a list of things to build, it is not a list of Zephyr modules. `west` does not know about Zephyr, it's trying to be as independent from it as possible. `west list` is a list of _git_ repositories_. Zephyr's build system has made the assumption that these two lists are the same. If you don't like that and you want these lists to be different, then you must evolve the Zephyr build system, not ask the tool that manages git repos to lie about the list of git repos.

Or is there anything else you want besides asking `west list` to show Zephyr modules?



> Additionally it is not possible to override such submanifests in downstream projects.

You lost me there. Override what data exactly and for what purpose?

thorsten-klein:
> Do you mean "Zephyr module" instead?

No. I really mean west projects.
My definition of a project: a repository or a directory that contains a `west.yml` and optionally a `zephyr/modules.yml`.
In contrast to this a zephyr module only contains a `zephyr/modules.yml`.

> No: by definition, a west project is a git repo.

At your mentioned link I cannot read any sentence that says that a west project MUST be a git repository.

I totally agree, that west's key feature is to help manageíng multiple west projects that are separate repositories. 
But for me there is no reason for the other way around: Why does a west project always have to be a separate git repository?

> you don't need west

As you know we use zephyr, and zephyr provides some west extensions, that we want to reuse.

If you mean `west update`:
We need to run `west update` in order to resolve the dependencies from the west projects (`west.yml`) we depend on, e.g. zephyr's `west.yml`.

> Moving code across git repos is not a normal use case

Aha. So I guess you never worked with OEMs, right? In this case, we exactly don't want to provide the full git history as it contains developer names. So there is a valid use case for this.

> "import" for what reason?

Did you have a look at the test of this PR? We may have multiple `west.yml` that are stacked.
We need to anyhow _import_ the information from the other west projects.

> If you have a monorepo, then you don't need manifests

That's only partly true. 
Even in a monorepo you can have external dependencies. These dependencies are managed in a manifest.
So our monorepo does only contain the code from our company, but all external code (e.g. zephyr) should be resolved via manifest. This must work transitively, as external code may again have other 3rd party dependencies (e.g. Kconfiglib, hals, mcuboot, ...).

> you can solve the problem so simply, with just ZEPHYR_EXTRA_MODULES

So you recommend that every downstream project must dig in the platform for all available extra-modules and maintain an own list?
I would expect that west offers a way to forward this information (transitively) -- same as west does with west projects.
There the users also do not have to specify all dependencies, but west resolves the project tree transitively.

> west list is a list of git repositories

This is not true. 
`west list` is just a list of west projects (`west.yml`) that are considered by west when building the project tree.
This is also what the documentation says: 
Ref: https://docs.zephyrproject.org/latest/develop/west/built-in.html
`west list: print a line of information about each project in the manifest, according to a format string`
So there is no word about that each project must be a git repository.

---
There is a test added which fully showcases our directory/repository tree.
Explained in short: We have multiple `west.yml` files in one repository which can be stacked. When initializing the workspace, we don't want to provide information again which is actually already contained in the `west.yml`s of the dependencies (this means: Not providing `ZEPHYR_EXTRA_MODULES` manually).

Instead of only tearing down my proposal:
Can you please provide an alternative solution so that my test passes?

nordicjm:
> At your mentioned link I cannot read any sentence that says that a west project MUST be a git repository.

"You can specify what Git remote URLs to use when cloning and fetching the projects, what revisions to track, and where the project should be stored on the local file system"
Also right at the top: https://docs.zephyrproject.org/latest/develop/west/manifest.html "The history of the manifest repository is the line of Git commits which is “floating” on top of the gray plane"

> Aha. So I guess you never worked with OEMs, right? In this case, we exactly don't want to provide the full git history as it contains developer names. So there is a valid use case for this.

then squash it into a single commit and submit it to an internal git server

> Did you have a look at the test of this PR? We may have multiple west.yml that are stacked.
> We need to anyhow import the information from the other west projects.

that's what west does already so nothing here is needed for that

> >    west list is a list of git repositories
> This is not true.
> west list is just a list of west projects (west.yml) that are considered by west when building the project tree.
> This is also what the documentation says:

No it is true, west deals with **git** repositories

>  Can you please provide an alternative solution so that my test passes?

you've been told, the extra modules or use west properly as it was designed, which is what other vendors all over the world are doing today. Or, alternatively (and I have already suggested this) use zephyr without west, it's fully possible and yes some vendors do this already for whatever reasons (some for trust of code issues, some because of fitting in with existing infrastructure) - you do **not** need to use west, to use zephyr, it makes things much easier but you are free to replace it entirely. Someone even made an alternative tool named "east", I don't remember the specifics but they showed it off years ago, so you are not tied to west in any way. Found it: https://github.com/IRNAS/irnas-east-software

And if you really have an aversion to git then use patches https://docs.zephyrproject.org/latest/develop/west/zephyr-cmds.html#working-with-patches-west-patch which I would say the original west author (who stepped down prior) was very much against having support for patch support years before it was implemented

marc-hb:
> My definition of a west project...

The definition of a "west project" is obviously central and essential.

You seem to enjoy writing a lot of code but... submitting a hundred lines-long, complete-looking PR with tests and everything before discussing _the redefinition of that essential concept_ is really "putting the card before the horse" (EDIT: despite multiple, earlier warnings in #852 and https://github.com/zephyrproject-rtos/zephyr/issues/96936 ). Prototyping is extremely useful of course but this PR went way beyond prototyping. I feel like that approach is good neither for you nor for reviewers. It's not good for reviewers because it looks like a lot of review work (again...) and because it can make us feel guilty and uncomfortable discarding a lot of your  work. I suspect it's not good for you either due to the time spent and frustration. Unless writing all that code was more pleasure than work? Great, but then please hide the size of that effort in your backpocket until after a design discussion.


> This is also what the documentation says:
> Ref: https://docs.zephyrproject.org/latest/develop/west/built-in.html
> west list: print a line of information about each project in the manifest, according to a format string
> So there is no word about that each project must be a git repository.

But then what does https://docs.zephyrproject.org/latest/develop/west/manifest.html say? That a manifest is almost entirely about... git repositories!

This is either bad faith or delusion, I hope the latter.


> Aha. So I guess you never worked with OEMs, right? In this case, we exactly don't want to provide the full git history as it contains developer names. So there is a valid use case for this.

I have been exposed multiple times to the challenges of publishing parts of private git repositories. I have seen various approaches to solve that problem and I have even participated in some. But I had never seen so far an attempt to use a multi-repo tool on a monorepo.

> then squash it into a single commit and submit it to an internal git server

BTW the reason this approach is very popular (despite the loss of history) is simply because it's much _safer_: now you have only one version to scan when you want to make sure nothing undesired escaped - whether that's for legal or just reputation reasons (low quality code, inappropriate comments, secrets etc.) Off-topic sorry.

> Even in a monorepo you can have external dependencies. These dependencies are managed in a manifest.

Sure, but that seems quite off-topic.


> > you can solve the problem so simply, with just ZEPHYR_EXTRA_MODULES

> So you recommend that every downstream project must dig in the platform for all available extra-modules and maintain an own list?

Yes - because that's precisely the information that you lost by moving to a monorepo. So you should definitely store and maintain that information in your monorepo. It's not even changing frequently. Then you may want to re-use it to convert back.


> As you know we use zephyr, and zephyr provides some west extensions, that we want to reuse.

This should have been the starting point of a "top-down" discussion. But that point was difficult to see, obscured by the apparent "readiness" and volume of the solution. This looks like a variation of https://en.wikipedia.org/wiki/XY_problem

> We need to anyhow import the information from the other west projects.

You are still not clearly stating _what_ exact information you want to re-use... Again, this should be a starting point.


> Or, alternatively (and I have already suggested this) use zephyr without west, it's fully possible and yes some vendors do this already for whatever reasons
> you do not need to use west, to use zephyr, it makes things much easier but you are free to replace it entirely.

In case some example was needed, people are even using twister to build and run tests without west:
- https://github.com/zephyrproject-rtos/zephyr/pull/93944


To be really honest with you, I'm not sure it was a good idea for Zephyr to use the same tool for versioning _and_ as a front-end to project-specific extensions doing everything else. This could/should have been two different tools. As @nordicjm suggested, maybe you should join forces with "east" or other alternatives not using `west` or using west _only_ for multi-repo versioning which is the central and essential feature! Definitely not an optional feature.

I bet some of those other tools are already able to read and parse some of the information you want. Because west tries to be backwards-compatible, that shouldn't be too hard.

PR #876, opened 2025-11-04
tests: conftest.py: add WEST_CONFIG_LOCAL protection

marc-hb:
I see no reason to exclude WEST_CONFIG_LOCAL. People defining WEST_CONFIG_LOCAL may be rare and test bugs affecting it too but they still deserve some isolation. This also makes the code more "regular" hence raising fewer questions.

Fixes commit 4596dcac05d6 ("Add a new pytest fixture to isolate test environment from user configs")

cc: @thorsten-klein


thorsten-klein:
This makes totally sense 👍

thorsten-klein (approved):

pdgendt (approved):

PR #877, opened 2025-11-04
configuration: anchor relative WEST_CONFIG_x paths to "west topdir"

### LINKED ISSUE #871: WEST_CONFIG_* accept relative paths
I accidentally found this while testing #867 but it predates #867 

I think this is a serious issue because:

- `west config` and (build) output can depend on the current directory. I think they never should because that can lead to hours or even days of troubleshooting. For instance imagine compiler flags that "silently" come and go.
- More severely, I think this is a significant attack surface very similar to what @mbolivar described in https://github.com/zephyrproject-rtos/west/pull/849#issuecomment-3356795725

So, I think relative paths in `WEST_CONFIG_*` should be ~rejected~ or anchored, see below. However, this is technically a backwards-incompatible change. I mean some users could have actually relied on that to "dynamically" change their configuration. They shouldn't have, but they might have...

marc-hb:
Fixes #871.

This could strictly speaking be considered as a backwards-compatible
change, but I don't think it should because: 1. who in their sane mind
would have intentionally relied on the behavior of their west commands
to change depending on the current directory!?

I think it is still very unlikely, but: 2. it's a bit more likely that
some people always run west commands from their topdir directory _and_
leveraged that to saved a little bit of typing with a relative
WEST_CONFIG_x definition(s). Or, maybe even used the same relative
$WEST_CONFIG_x value across workspaces, i.e: topdir1/$WEST_CONFIG_x and
topdir2/$WEST_CONFIG_x, etc. ?

This fix breaks 1. but preserves 2.

To be clear: this commit still does not "condone" relative WEST_CONFIG_x
values; I don't think we should document that. But it makes their behavior
_deterministic_ which is what matters.

### DIFF
--- a/src/west/configuration.py
+++ b/src/west/configuration.py
@@ -175,15 +175,15 @@ class Configuration:
 
         local_path = _location(ConfigFile.LOCAL, topdir=topdir, find_local=False) or None
 
-        self._system_path = Path(_location(ConfigFile.SYSTEM))
-        self._global_path = Path(_location(ConfigFile.GLOBAL))
+        self._system_path = Path(_location(ConfigFile.SYSTEM, topdir=topdir))
+        self._global_path = Path(_location(ConfigFile.GLOBAL, topdir=topdir))
         self._local_path = Path(local_path) if local_path is not None else None
 
         self._system = _InternalCF.from_path(self._system_path)
         self._global = _InternalCF.from_path(self._global_path)
         self._local = _InternalCF.from_path(self._local_path)
 
-    def get_paths(self, location: ConfigFile = ConfigFile.ALL):
+    def get_paths(self, location: ConfigFile = ConfigFile.ALL) -> list[Path]:
         ret = []
         if self._global and location in [ConfigFile.GLOBAL, ConfigFile.ALL]:
             ret.append(self._global.path)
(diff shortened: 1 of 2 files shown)


src/west/configuration.py
```diff
@@ -587,7 +587,22 @@ def delete_config(
         raise KeyError(f'{section}.{key}')
 
 
+def _rel_topdir_to_abs(p: PathType, topdir: PathType | None) -> str:
```
thorsten-klein, line 590:
```suggestion
def _rel_topdir_to_abs(p: Path, topdir: Path | None) -> Path:
```

Maybe just a personal favor, but I have started to use `Path` everywhere. This helps to get rid of the `os.path` usages, e.g. also below `os.path.isabs(p)` could be replaced with `p.is_absolute()`, or `os.path.join(topdir, p)` with `(topdir / p).resolve()`

thorsten-klein, line 590: This function could be moved to `util.py`, right?

```
def make_absolute(p: Path, base_dir: Path | None) -> Path:
```

marc-hb: I prefer to use the new pathlib too and I considered it, but this small function is used from only one place and none of its inputs or output are pathlib objects, so they would have to be all converted back and forth either here or by the caller. So this wouldn't save code and the type juggling would be harder to follow (`PathType` is surprising enough already).

marc-hb: The `west topdir` exception seemed quite specific to me; are you aware of other examples?

pdgendt: There actually is a difference, see https://docs.python.org/3/library/pathlib.html#comparison-to-the-os-and-os-path-modules

I tend to prefer working with `Path` objects too, but since this is internal and trivial, either is fine.

pdgendt (approved):

src/west/configuration.py
marc-hb: > There actually is a difference, see https://docs.python.org/3/library/pathlib.html#comparison-to-the-os-and-os-path-modules

Thanks @pdgendt , this reference is really good to know. The differences are very subtle which makes them look "dangerous": triggering bugs in the obscure edge cases mentioned there. So I really feel like converting back and forth can only make things even more subtle and even more complicated in case something goes wrong. I mean we should have parts of the code either fully migrated to `pathlib` or not at all. Also, I'm trying to keep this bug fix minimal so it could be easily cherry-picked / backported if needed.

thorsten-klein (approved):

PR #879, opened 2025-11-08
fix: `--list-paths` should print config files in correct order

thorsten-klein:
The order of config files which is currently listed in wrong order.
Local config must have highest precedence, and system config lowest.


pdgendt (approved):

marc-hb (approved):
Note the "last one wins" is the most common precedence rule but it is not 100% universal; in some cases the "first one wins".

I'm approving not so much for precedence reasons but because this is consistent with the documentation: https://docs.zephyrproject.org/4.2.0/develop/west/config.html and `west config -h` 6cf0a0236b10

(the documentation does define the precedence of course)

PR #880, opened 2025-11-15
Fix python-xdist failure and run tests 6 times faster

marc-hb:
Generally speaking, tests sensitive to the current directory should not
run from a random directory.

The current directory of test_import_project_release_dir() was either
the current directory or wherever the previous test left it, which could
be inside a random west workspace. This crashed the `west init -l
/tmp/pytest/...` command used by the test.

Found thanks to pytest-xdist and pytest -n 20, but there is more
"reliable" way to reproduce: move west.git instead a west workspace and
run:

   pytest -k test_import_project_release_dir

This fails with the "already initialized" error.

### DIFF
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -2675,6 +2675,12 @@ def test_import_project_release_dir(tmpdir):
     # Tests for a workspace that imports a directory from a project
     # at a fixed release.
 
+    # Don't run in a random cwd, this could be an existing west
+    # workspace and would fail the `west init -l ...` below.  We don't
+    # want this test to depend on any special cwd either so the root is
+    # a reasonable choice.
+    os.chdir(tmpdir.parts()[0])
+
     remotes = tmpdir / 'remotes'
     empty_project = remotes / 'empty_project'
     create_repo(empty_project)


marc-hb:
So, this makes running tests 6 times faster with `xdist` on my system but I don't think we should use `xdist` in CI because:
- Github runners don't have many cores. Not a lot to gain.
- If tests fail in CI because of some concurrency issue, it will be a confusing nightmare to reproduce and debug.

On the other hand, it would still be nice to catch future `xdist` regressions. For this, running tests in a _random order_ would help a lot. 

I spotted https://pypi.org/project/pytest-random-order/ but I did not try it. Ideally, it could print a re-usable, pseudo random seed that can be copied from CI.


cc: @thorsten-klein

thorsten-klein:
Good catch. I also stumbled upon this when I have introduced running pytest directly.
Unfortunately I don't like the solution of this PR so much, as it does only solve symptoms, not the root cause.
Therefore I have opened #881 which solves the real root cause by ensuring that environment and cwd are restored after each test.

marc-hb:
> Unfortunately I don't like the solution of this PR so much, as it does only solve symptoms, not the root cause.

~#881 does for every test what this does for one test~... afraid we don't have the same definition of "root cause" :-)

marc-hb:
> ~#881 does for every test what this does for one test~... afraid we don't have the same definition of "root cause" :-)

I was initially confused by #881, sorry. We still don't have the same definition of "root cause" :-)

#881 makes every test cleanup **after** itself, great idea! I'm all for it. Both should fix the failure of this particular test but I don't think this #880 is mutually exclusive. This PR #880 PR makes `test_import_project_release_dir()` run in a deterministic directory **before** it's run.

Generally speaking, I think tests should cleanup **after** themselves _and_ make sure they setup a good and deterministic environment **before** they run. It's not mutually exclusive because if something goes wrong in either one, then the other gives you a second chance. And of course things regularly go wrong in tests, otherwise you're not testing hard enough :-)

pdgendt:
> Generally speaking, I think tests should cleanup **after** themselves _and_ make sure they setup a good and deterministic environment **before** they run. It's not mutually exclusive because if something goes wrong in either one, then the other gives you a second chance. And of course things regularly go wrong in tests, otherwise you're not testing hard enough :-)

Agree that they are mutually exclusive, but if we can't assume a clean start for each test, isn't that worse? We should be able to randomize the test order, I don't think this needs "fixing".

marc-hb:
> but if we can't assume a clean start for each test, isn't that worse?

- We _generally_ can, but not always. When some previous test fails, you can never be 100% sure of what happened and how the cleanup went.

- Cleanups from previous tests are irrelevant when... there is no previous test! That is: when the first test is run or when running a _single_ test with `pytest -k test_import_project_release_dir`. That's the obvious reason why tests should not rely on previous tests: there can be none. So, each test must make a reasonable effort to be deterministic on its own.


I tested f1151f8390544162f3f0 from #881 (EDIT: changed now) and I verified it does not fix `test_import_project_release_dir` when it is run alone.

I adjusted the source comments and commit message.

pdgendt:
> > but if we can't assume a clean start for each test, isn't that worse?
> 
> * We _generally_ can, but not always. When some previous test fails, you can never be 100% sure of what happened and how the cleanup went.

The pytest fixtures should handle this, if there are tests that do not cleanup temporary fixtures, those need to be looked at, no?

> * Cleanups from previous tests are irrelevant when... there is no previous test! That is: when the first test is run or when running a _single_ test with `pytest -k test_import_project_release_dir`. That's the obvious reason why tests should not rely on previous tests: there can be none. So, each test must make a reasonable effort to be deterministic on its own.
> 
> I tested [f1151f8](https://github.com/zephyrproject-rtos/west/commit/f1151f8390544162f3f08e54e9fc6a61138b50df) from #881 (EDIT: changed now) and I verified it does not fix `test_import_project_release_dir` when it is run alone.
> 
> I adjusted the source comments and commit message.

But there isn't an issue with running the single test, _if_ others cleanup correctly.

marc-hb:
> But there isn't an issue with running the single test, if others cleanup correctly.

By definition, there is no "other" tests when running a single test.

pdgendt:
> > But there isn't an issue with running the single test, if others cleanup correctly.
> 
> By definition, there is no "other" tests when running a single test.

You wrote:

> The current directory of test_import_project_release_dir() was either
> the current directory or wherever the previous test left it, which could
> be inside a random west workspace.
> ...
> pytest -k test_import_project_release_dir
> 
> This fails with the "already initialized" error.

You are running a single test, after other tests. So IMO it's still up to the previous tests to clean up.

marc-hb:
> You are running a single test, after other tests

No, not when invoking `pytest -k test_import_project_release_dir`, which runs a single test. "single" = "no other".

pdgendt:
_I think we are talking past each other, so let's try to circle back to the original issue._

The test can't run if the current working directory where you start the test has a west `topdir`, as the `Init` command verifies here:

https://github.com/zephyrproject-rtos/west/blob/bb1c7bf3937391010f7e26dbb0c49a6460c5e492/src/west/app/project.py#L261-L279

So a few things come to mind:

1. Some sequence (previous tests, manual actions, ...) leaves us inside a west workspace
2. The `init -l <path>` command doesn't allow _any_ absolute `path`, if called from within a workspace, even though they are not relative to each other

For 1. what's the reason/origin? Can this be tackled?

For 2. shouldn't this be allowed/tested?

pdgendt:
Maybe a hammer fix could be to add the auto used fixture in `conftest.py`

```python
@pytest.fixture(autouse=True)
def start_in_tmpdir(tmpdir):
    with chdir(tmpdir):
        yield
```

EDIT: this is already suggested in #881

marc-hb:
Superseded by #881

PR #881, opened 2025-11-16
Improve test isolation with autouse pytest fixtures

thorsten-klein:
Extend and rename current autouse pytest fixture to
`setup_teardown_test_environment`.

The fixture ensures an isolated test environment.
It creates a new temporary directory which is used as working directory.
This ensures a clean start for each test and prevents tests from
affecting another one through changes to the working directory.

The fixture ensures that the user's actual configuration files are
neither used nor touched during test, as WEST_CONFIG_* env variables are
set, whereby no config files are created at these locations.

The fixture sets ZEPHYR_BASE (to avoid complaints in subcommand stderr),
but to a spurious location (so that attempts to read from inside of it
are caught here).

The fixture also ensures that any environment modifications made by a
test do not leak into subsequent tests, as the environment is restored
when the `update_env` with-block exits.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -142,27 +142,43 @@ def _check_git_capabilities(tmpdir_factory):
 
 
 @pytest.fixture(autouse=True)
-def independent_global_and_system_config(tmpdir_factory):
-    # Fixture to ensure that the user's actual configuration
-    # files are neither used nor touched during test.
-    #
-    # We also set ZEPHYR_BASE (to avoid complaints in subcommand
-    # stderr), but to a spurious location (so that attempts to read
-    # from inside of it are caught here).
-    #
+def setup_teardown_test_environment(tmpdir_factory):
+    """
+    The fixture ensures an isolated test environment.
+
+    It creates a new temporary directory which is used as working directory.
+    This ensures a clean start for each test and prevents tests from affecting
+    another one through changes to the working directory.
+
+    The fixture ensures that the user's actual configuration files are neither
+    used nor touched during test, as WEST_CONFIG_* env variables are set,
+    whereby no config files are created at these locations.
+
+    The fixture sets ZEPHYR_BASE (to avoid complaints in subcommand stderr),
+    but to a spurious location (so that attempts to read from inside of it are
+    caught here).
+
+    The fixture also ensures that any environment modifications made by a test
+    do not leak into subsequent tests, as the environment is restored when the
+    `update_env` with-block exits.
+    """
     tmpdir = Path(tmpdir_factory.mktemp("test-configs"))
+    tmp_cwd = tmpdir_factory.mktemp("tmp-cwd")
 
     # config paths
     system = tmpdir / 'config.system'
     glbl = tmpdir / 'config.global'
 
     # run with environment variables set
-    with update_env({
-        'WEST_CONFIG_SYSTEM': str(system),
-        'WEST_CONFIG_GLOBAL': str(glbl),
-        'WEST_CONFIG_LOCAL': None,
-        'ZEPHYR_BASE': str(tmpdir / 'no-zephyr-here'),
-    }):
+    with (
+        chdir(tmp_cwd),
+        update_env({
+            'WEST_CONFIG_SYSTEM': str(system),
+            'WEST_CONFIG_GLOBAL': str(glbl),
+            'WEST_CONFIG_LOCAL': None,
+            'ZEPHYR_BASE': str(tmpdir / 'no-zephyr-here'),
+        }),
+    ):
         yield
 
 


tests/conftest.py
```diff
@@ -117,6 +117,29 @@ def chdir(path):
...
 
 
+@pytest.fixture(autouse=True)
+def initial_cwd(tmpdir):
+    """
+    Ensure each test runs inside its own temporary working directory.
+    This prevents tests from interfering with one another through
+    shared or persistent working-directory state.
+    """
+    cwd = tmpdir.mkdir("initial-cwd")
```
thorsten-klein: In a first draft a new separate tmpdir was created for each test. But I think this is too overkill. So I have changed to:

```
original_cwd = Path.cwd() 
with chdir(original_cwd):
```

The with-block ensures that the cwd is reset to current cwd (which is same as `original_cwd`) when it exits.

pdgendt (approved):
I would have thought this was `pytest`'s default behavior, I do prefer this over #880

tests/conftest.py
```diff
@@ -117,6 +117,31 @@ def chdir(path):
...
+@pytest.fixture(autouse=True)
+def reset_cwd():
+    """
+    Reset the current working directory after each test.
+    This prevents tests from affecting one another through changes to the
+    working directory. The original directory is restored automatically
+    when the `chdir` context manager exits.
+    """
+    original_cwd = Path.cwd()
+    with chdir(original_cwd):
```
marc-hb: I think the code is correct but the name `original_cwd` does not look right, which got me very confused.

The argument of `chdir` is a _temporary_ directory, it's not an "original" directory that is saved for later.

This is a somewhat unexpected use of `chdir()` because you're not using it to change to a temporary directory at all. You're using it only to save the current directory, which is NOT the argument either! Very confusing.

Considering `chdir()` is only 5 lines long and you're using only the second half of it, it seems much simpler to just do it directly (untested):

```python

   saved_cwd = Path.cwd()
   try:
      yield
   finally:
      os.chdir(saved_cwd)
```
As in: https://docs.pytest.org/en/7.1.x/how-to/fixtures.html#yield-fixtures-recommended

It's only 2 lines longer.

tests/conftest.py
```diff
@@ -117,6 +117,31 @@ def chdir(path):
...
+@pytest.fixture(autouse=True)
+def reset_env():
+    """
+    Reset environment variables after each test.
+    This ensures that any environment modifications made by a test do not
+    leak into subsequent tests. The environment is restored when the
+    `update_env` context manager exits.
+    """
+    with update_env({}):
+        yield
```
marc-hb: There is no `original_cwd` confusion in this one but again there is IMHO not much value in re-using only the last third of `update_env()` which is itself not very long. The layer of indirection makes is harder to read cause now you have to know or guess about `update_env()` Why not just:

```python
   env_bak = dict(os.environ) # copy
   try:
     yield
   finally:
     os.environ.clear()
     os.environ.update(env_bak)
```
It's barely longer and it's direct and IMHO much simpler and easier to read.


https://go-proverbs.github.io/
"A little copying is better than a little dependency."

marc-hb (commented):
Thanks! I like the idea, it's great that each test automatically cleans up after itself. I'm not sure it's mutually exclusive with #880 though.  However I find the implementation not easy to read, see below.

Also: I could be wrong but I suspect using many context managers is what makes the stack very large on failure or in the debugger. Less is more...

thorsten-klein:
We could also adapt the fixture to set cwd to a new temp directory by default (this is how I had it before but then I thought it might be overkill. However, it seems to be necessary).

tests/conftest.py
```diff
@@ -117,6 +117,30 @@ def chdir(path):
...
+@pytest.fixture(autouse=True)
+def temp_cwd(tmpdir):
+    """
+    Fixture to create new temporary directory and use it as working directory.
+    This ensures a clean start for each test and prevents tests from affecting
+    another one through changes to the working directory.
+    """
+    cwd = tmpdir.mkdir('temp-cwd')
+    with chdir(cwd):
+        yield
```
pdgendt: No need to create an additional subdirectory, the `tmpdir` will be empty. Also the fixture name could be a bit more descriptive.

```python
def setup_teardown_cwd_tmpdir(tmpdir):
    """
    Fixture to change the working directory to a temporary directory.
    This ensures a clean start for each test and prevents tests from affecting
    another one through changes to the working directory.
    """
    with chdir(tmpdir):
        yield
```

tests/conftest.py
pdgendt: Either is fine IMO, but the fixture name could also use a more descriptive name

```python
@pytest.fixture(autouse=True)
def setup_teardown_clean_env():
    ....
```

pdgendt (commented):
Alternatively we can combine both fixtures

```python
@pytest.fixture(autouse=True)
def setup_teardown_test_environment(tmpdir):
    """
    Fixture to change the working directory to a temporary directory with a clean environment.
    This ensures a clean start for each test and prevents tests from affecting another
    one through changes to the working directory or environment updates.
    """
    with chdir(tmpdir), update_env({}):
        yield

**Comment by thorsten-klein** [comment]
> Alternatively we can combine both fixtures

I like this idea! I have directly updated.

**Review by pdgendt — APPROVED** [review]

File: tests/conftest.py [hunk 5]
```diff
@@ -117,6 +117,21 @@ def chdir(path):
 #
 
 
+@pytest.fixture(autouse=True)
+def setup_teardown_test_environment(tmpdir):
```
pdgendt: Come to think of it, we can even merge this fixture with `independent_global_and_system_config`.

thorsten-klein: However you prefer. I think it is personal favor if you want multiple smaller autouse fixtures or fewer bigger ones.
I can combine them if you want me to 👍

pdgendt: Two `autouse` fixtures that change the environment does not make sense IMO.

pdgendt (approved):
LGTM, thanks!

marc-hb (approved):
Thanks @thorsten-klein , it's nice to have everything in one place. Much easier to debug and/or tinker if needed.

PR #883, opened 2025-11-17
conftest.py: test the local west, not the installed one

marc-hb:
Discovered by chance after uninstall west and trying to run pytest directly. This was hopefully not a problem when running test with "uv"?

This issue was probably introduced by #859. Commit 879bb019276b91 seems relevant too.

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -15,6 +15,9 @@ from pathlib import Path, PurePath
 
 import pytest
 
+# https://chrisyeh96.github.io/2017/08/08/definitive-guide-python-imports.html#case-3-importing-from-parent-directory
+# Test the local west and not some other, installed version.
+sys.path.insert(1, os.path.join(sys.path[0], '../src'))
 from west.app import main
 
 GIT = shutil.which('git')


thorsten-klein:
How did you run pytest? In the README it says to run ``  pytest -o pythonpath=src`  which actually already does this pythonpath extension. Did this not work for you?

If not, I would prefer to set this option as tools.pytest.ini_option in the pyproject.toml instead of sys.path.insert.

marc-hb:
I missed the `-o pythonpath=src`. After #859 I just tried `pytest` and it seemed to work! For a long time, until I realized this. I'm afraid people look at documentation only when things go wrong :-( 

> If not, I would prefer to set this option as tools.pytest.ini_option in the pyproject.toml instead of sys.path.insert.

As long as it stops people from testing (parts of?) the installed version by accident, fine by me! Would like me to do it? I'm not sure how.

thorsten-klein:
In PR #859, I considered adding that `tools.pytest.ini_option`, but I think the preferred behavior from west maintainers is to test the installed version with `uv` to ensure that the resulting pip package works. 
I was very happy to see PR #859 accepted at all, and `pytest -o pythonpath=src` feels like a good compromise for switching between testing the installed version and the local worktree -- even if it is only mentioned in the README.

If more contributors run into your issue, we could consider making `pythonpath=src` the default.
However, in CI we should still test the installed version to ensure that the pip package works.

CC: @pdgendt

marc-hb:
> However, in CI we should still test the installed version to ensure that the pip package works.

Agreed. 

> I considered adding that tools.pytest.ini_option, but I think the preferred behavior from west maintainers is to test the installed version with uv to ensure that the resulting pip package works.

So are you saying we cannot have both out of the box?

> If more contributors run into your issue,...

The real, serious issue here is: people running "pytest" alone do NOT realize they're probably testing a different version. So we'll never know how many people are affected. I realized only by chance after a long time. That's why documentation is not enough in this case. It would be preferable to FAIL rather than silently testing the wrong thing. Once it fails, then everyone checks the documentation. Or asks for help; but either way they don't keep testing the wrong version.

pdgendt:
> > I considered adding that tools.pytest.ini_option, but I think the preferred behavior from west maintainers is to test the installed version with uv to ensure that the resulting pip package works.
> 
> So are you saying we cannot have both out of the box?

I think this is a common issue for python packages... We could use the `poe` tasks to our advantage here maybe, and add an argument to the `test` task to select the installed or local version?

If we want developers to test the local version by default, we maybe should set the `ini_option`, but force the installed version in CI. (we can just add a separate Github specific task like for linting/formatting)

thorsten-klein:
> I think this is a common issue for python packages... We could use the poe tasks to our advantage here maybe, and add an argument to the test task to select the installed or local version?
>If we want developers to test the local version by default, we maybe should set the ini_option, but force the installed version in CI. (we can just add a separate Github specific task like for linting/formatting)

+1

marc-hb:
#884 looks better

PR #884, opened 2025-11-19
test local worktree with pytest by default

thorsten-klein:
Successor of #883 where this was discussed.

Enable pytest `pythonpath=src` by default to run tests against the local worktree. `poe test` now also tests the local worktree. 
In CI, the newly added `poe gh-test` is run in order to test the locally installed package (installed via `uv`). 
`gh-test` explicitly unsets pytest option `pythonpath` and creates a coverage report.

During manual tests it was observed that `uv` always appends the `src` directory to `sys.path`. Luckily, this is not a problem as the path is appended and the path of the installed west comes first.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -147,28 +147,30 @@ First, install the dependencies::
   # Recommended in an active virtual environment
   pip3 install --group dev
 
-Then, run the test suite locally from the top level directory::
 
-  # Using uv
+Then, run the tests (from the project's top level directory)::
+
+  # Run the full test suite with uv (recommended)
   uv run poe all
 
-  # Using poe
-  # Recommended in an active virtual environment
+  # Or run the full test suite with poe directly.
+  # (recommended in an active virtual environment)
   poe all
 
-  # Manually (test the installed west version)
+  # Run only the pytest tests (against the local worktree)
   pytest
 
-  # Manually (test the local copy)
-  pytest -o pythonpath=src
+The ``all`` target from ``poe`` runs multiple tasks sequentially. Use ``poe -h``
+to get a list of all available tasks.
+You can pass arguments to any ``poe`` task, which are forwarded to the
+underlying command. This is especially helpful when you want to run
+only specific tests to save time. Examples::
 
-The ``all`` target from ``poe`` runs multiple tasks sequentially. Run ``poe -h``
-to get the list of configured tasks.
-You can pass arguments to the task running ``poe``. This is especially useful
-on specific tests and save time. Examples::
-
-  # Run a subset of tests
+  # Run a subset of tests (file)
   poe test tests/test_project.py
+
+  # Run a single test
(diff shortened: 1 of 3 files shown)


pdgendt (approved):

README.rst
```diff
@@ -156,11 +156,11 @@ Then, run the test suite locally from the top level directory::
...
   poe all
 
-  # Manually (test the installed west version)
+  # Manually (test the local copy)
   pytest
 
-  # Manually (test the local copy)
-  pytest -o pythonpath=src
+  # Manually (test the installed west version)
+  pytest -o pythonpath=
```
marc-hb: Is this really needed? Packaging issues should already be found by `uv`.

Unless there's value in testing and older version with newer tests or vice-versa? Very "advanced" then. Maybe move this trick outside the README.rst, how about a comment just above the `pythonpath = "src"` line? So people don't assume they have to change that line.

pyproject.toml
```diff
@@ -110,10 +110,14 @@ preview = true
 mypy_path = "src"
 ignore_missing_imports = true
 
+[tool.pytest.ini_options]
+pythonpath = "src"
+
 [tool.poe.tasks.test]
-cmd = "python -m pytest -v -W error --junitxml=junit.xml -o junit_family=xunit1 --cov-report=html --cov-config=pyproject.toml --cov=west"
+cmd = "python -m pytest -W error"
```
marc-hb, line 119: Do we need to lose the coverage options on this task?

marc-hb (commented):
No major concern, see below.

marc-hb:
I know we don't have a lot of pull requests but for the record and for reviewers who only give a hand from time to time, don't forget to mention that this is a better alternative to #883 where relevant discussion already happened.

pyproject.toml
thorsten-klein: @pdgendt has proposed this change. 
Maybe you missed it: This command is moved to `uv run poe gh-test` as those coverage files are most likely only needed in CI.
Do you see any benefit of creating the coverage files always on the user's local system on each `uv run poe test`?

README.rst
thorsten-klein: Yes, this is very advanced. It is a good idea to move it to `pyproject.toml`

pyproject.toml
pdgendt: Arguments passed on the CLI are still possible, it's just not generated by default.

```shell
$ uv run poe test --cov-report=html --cov-config=pyproject.toml --cov=west
```

I don't tend to look at coverage locally that often, so I like that it's now opt-in. We could provide a `poe` [config option](https://poethepoet.natn.io/guides/args_guide.html) for it, but it's a bit overkill IMO.

pdgendt:
@marc-hb any open comments here?

marc-hb (approved):

PR #888, opened 2025-11-28
ci: Test distro builds

pdgendt:
A test commit to see issue in CI.

PR #889, opened 2025-11-28
Extract unreadable-file test to avoid failures on Windows and root

thorsten-klein:
Successor of #888 

Creating a non-readable config file is not possible on Windows or when running as root (since root user can always read all files). 
Since there is some west CI workflow on main branch (not on PRs) executing tests as root, this test case has failed.

This change moves the test into a separate suite which is skipped on Windows or when running as root.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -703,6 +703,30 @@ def test_config_precedence():
     assert cfg(f=LOCAL)['pytest']['precedence'] == 'local'
 
 
+@pytest.mark.skipif(
+    WINDOWS or (os.geteuid() == 0),
+    reason="Non-readable files do not exist on Windows, and root user can always read files",
+)
+def test_config_non_readable_file(config_tmpdir):
+    # test to read a config file without read permission
+    cwd = pathlib.Path.cwd()
+
+    # create a readable file
+    config_readable = cwd / 'readable'
+    config_readable.touch()
+
+    # create a non-readable file
+    config_non_readable = cwd / 'non-readable'
+    config_non_readable.touch()
+    config_non_readable.chmod(0o000)
+
+    # trying to use a non-readable config should result in according error
+    with update_env({'WEST_CONFIG_GLOBAL': f'{config_readable}{os.pathsep}{config_non_readable}'}):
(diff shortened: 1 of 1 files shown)


tests/test_config.py
```diff
@@ -703,6 +703,33 @@ def test_config_precedence():
     assert cfg(f=LOCAL)['pytest']['precedence'] == 'local'
 
 
+def test_config_non_readable_file(config_tmpdir):
+    # test to read a config file without read permission
+    # Note: non-readable files do not work on Windows
+    if WINDOWS:
+        return
```
```suggestion
        pytest.skip("... Windows...")
```

pdgendt (approved):

tests/test_config.py
```diff
@@ -703,6 +703,33 @@ def test_config_precedence():
...
+    # Note: as root user a file is always readable
+    if os.geteuid() == 0:
+        return
```
```suggestion
        pytest.skip(".... as root ...")
```

marc-hb (commented):
I don't think we should encourage using `west` as root but ok, we should not block it either...

marc-hb:
> Since some additional CI on main branch executes tests as root,...

I find this a bit cryptic... could you elaborate and provide a short but more complete problem statement? (that does not depend on a test PR #888?)

tests/test_config.py
```diff
@@ -703,6 +703,33 @@ def test_config_precedence():
     assert cfg(f=LOCAL)['pytest']['precedence'] == 'local'
 
 
+def test_config_non_readable_file(config_tmpdir):
+    # test to read a config file without read permission
+    # Note: non-readable files do not work on Windows
+    if WINDOWS:
```
marc-hb: Off-topic sorry but I just found in https://docs.pytest.org/en/6.2.x/skipping.html#skipping-test-functions that `pytest` uses a different, "fuzzier" test to determine if the OS is Windows... worth looking into? In a different PR.

marc-hb (approved):
I found two distinct error messages was better and "non-readable does not exist on Windows" sounds oversimplified but none of that is important.

Also the commit message says "fails in CI" but this is all green:
https://github.com/zephyrproject-rtos/west/actions

Did you mean _future_ CI or something? Maybe it would have been clearer not to mention CI at all.

PR #890, opened 2025-11-30
auto-cache: fetch remote updates only if necessary

thorsten-klein:
This PR enhances the performance of the auto-cache feature.

#### Changes

- **Speed-up performance**  
  If a tag or commit is used in the manifest, and the revision is already available in the auto-cache, no remote fetch is performed (unless `--fetch=always` is specified). This allows `west update` to finish as fast as possible.

## Benefits

- Seamless offline usage  
- Faster `west update`

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -1177,9 +1177,9 @@ class Update(_ProjectCommand):
             '--auto-cache',
             help='''automatically setup local cache repositories
                     in a flat folder hierarchy, but with an additional
-                    subfolder (hashed name) for different remote URLs.
-                    Each local cache repository is automatically cloned
-                    on first usage and synced on subsequent clones.
+                    subfolder (hashed name) for different remote URLs. Each
+                    local cache repository is automatically cloned on first
+                    usage and synced on subsequent fetches (if necessary).
                     This cache has the lowest priority (Prio 2).''',
         )
 
@@ -1203,7 +1203,8 @@ class Update(_ProjectCommand):
         workspace setup.
         Only in case of auto-cache the 'west update' process updates the local
         caches first, which then serve as the source for pulling changes into
-        the workspace.
+        the workspace. Thereby, the auto-cache only fetches updates from remote
+        if the specified revision is not already present in the local cache.
 
         Example: Assume your manifest describes this workspace structure:
           (workspace)
@@ -1233,7 +1234,8 @@ class Update(_ProjectCommand):
           folder hierarchy is setup automatically. Each repository is stored under a
           directory named after the basename of its remote URL. To prevent conflicts
           between repos with same name, a hash of the remote URL is used as subfolder.
-          Note: Each local cache repo is automatically synced on subsequent updates.
+          Note: Each local cache repo is automatically synced on subsequent updates on
+          demand (if the requested revision is not already contained within the cache).
             (auto cache directory)
             ├── bar.git
             │   ├── <hash>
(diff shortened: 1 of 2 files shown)


src/west/app/project.py
```diff
@@ -1757,13 +1761,67 @@ def handle_auto_cache(self, project):
...
+            def check_git(cmd) -> bool:
+                '''
+                Helper function to run git command in the auto-cache directory
+                with suppressed output. If the command was successful (returncode 0),
+                return 'True', otherwise False.
+                '''
+                p = project.git(
+                    cmd, cwd=cache_dir, capture_stdout=True, capture_stderr=True, check=False
+                )
+                return not p.returncode
```
```suggestion
                return p.returncode == 0
```

src/west/app/project.py
```diff
@@ -1757,13 +1761,67 @@ def handle_auto_cache(self, project):
...
+
+            # Check if the auto-cache is corrupt
+            corrupt = not check_git(['cat-file', '-e', 'HEAD'])
```
pdgendt: When this happens, something is wrong with the cache, I'd prefer not to automatically handle this case. A user/CI should be notified about the issue instead.

I'd suggest just removing this as something will probably fail further down anyway.

src/west/app/project.py
```diff
@@ -1757,13 +1761,67 @@ def handle_auto_cache(self, project):
...
+
+            # Check if the revision is already contained in the auto-cache.
+            contains_rev = check_git(['cat-file', '-e', f'{project.revision}^{{commit}}'])
```
pdgendt: This is a nice feature to speedup the update, it should be the default to return early if the revision (_not branch!_) is already present.

src/west/app/project.py
```diff
@@ -1757,13 +1761,67 @@ def handle_auto_cache(self, project):
...
+
+            if not corrupt:
+                # return if auto-cache is intact, but remote is not reachable (even if
+                # fetch=always is specified), since we know that remote update will fail.
+                if not reachable:
+                    self.dbg(
+                        f'{project.name}: remote ({project.url}) not reachable. '
+                        'Skip auto-cache update with remote.'
+                    )
+                    return
```
pdgendt: If you return early when the revision is already present, this should be an error instead, but this will already become one in `project.git(['remote', 'update', '--prune'], cwd=cache_dir, check=False)`, no?

So maybe `reachable` can be dropped too?

src/west/app/project.py
```diff
@@ -1757,13 +1759,27 @@ def handle_auto_cache(self, project):
...
+            # already contained in the auto-cache.
+            # Only if it is an already available tag or a commit, the remote
+            # update can be skipped. Otherwise the auto-cache must be updated.
+            rev_type = _rev_type(project, cwd=cache_dir)
+            if self.fs != 'always' and rev_type in ('tag', 'commit'):
+                self.dbg(
+                    f'{project.name}: auto-cache remote update is skipped '
+                    f'as it already contains {rev_type} {project.revision}'
+                )
+                return
```
pdgendt: No need to get the `rev_type` if `self.fs` is set to `always`

```suggestion
            if self.fs != 'always' and (rev_type := _rev_type(project, cwd=cache_dir)) in ('tag', 'commit'):
                self.dbg(
                    f'{project.name}: auto-cache remote update is skipped '
                    f'as it already contains {rev_type} {project.revision}'
                )
                return
```

thorsten-klein: This variable is used in the message. I will rework to put it inside the `if`.

pdgendt: Use the walrus operator in that case (updated my suggestion)

thorsten-klein: With the walrus operator the statement looks very complicated. Therefore I have adopted to use a nested `if` instead, which is much easier to understand.

pdgendt (approved):

pdgendt:
Ping @ reviewers

carlescufi (approved):

PR #891, opened 2025-12-01
tests: config: Split the skip cases

pdgendt:
Have the actual reason for skipping the test in the output.

```shell
$ uv run poe test -v tests/test_config.py::test_config_non_readable_file
...
tests/test_config.py::test_config_non_readable_file SKIPPED (root user can always read files)
```

```powershell
PS D:\west> uv run poe test -v tests/test_config.py::test_config_non_readable_file
...
tests/test_config.py::test_config_non_readable_file SKIPPED (Non-readable files do not exist on Windows)                                                                                                                                                                                                                                                                            
```

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -703,9 +703,10 @@ def test_config_precedence():
     assert cfg(f=LOCAL)['pytest']['precedence'] == 'local'
 
 
+@pytest.mark.skipif(WINDOWS, reason="chmod is limited on Windows")
 @pytest.mark.skipif(
-    WINDOWS or (os.geteuid() == 0),
-    reason="Non-readable files do not exist on Windows, and root user can always read files",
+    hasattr(os, "geteuid") and os.geteuid() == 0,
+    reason="root user can always read files",
 )
 def test_config_non_readable_file(config_tmpdir):
     # test to read a config file without read permission


thorsten-klein (approved):

tests/test_config.py
```diff
@@ -703,9 +703,10 @@ def test_config_precedence():
     assert cfg(f=LOCAL)['pytest']['precedence'] == 'local'
 
 
+@pytest.mark.skipif(WINDOWS, reason="Non-readable files do not exist on Windows")
```
```suggestion
@pytest.mark.skipif(WINDOWS, reason="chmod is not available on Windows")
```
Pretty sure NTFS can stop some users from reading some files!

pdgendt: [It is available](https://docs.python.org/3/library/os.html#os.chmod), but you can only make files read-only.

pdgendt: @marc-hb do you still prefer that I change to your suggestion? I'm fine either way.

marc-hb: Sorry for the delay I missed your reply.
How about "chmod is limited on Windows"?

marc-hb: Just FYI, NOT a request to do anything with this:

https://stackoverflow.com/questions/896638/checking-folder-file-ntfs-permissions-using-python

pdgendt: Thanks, updated

marc-hb (approved):

PR #896, opened 2025-12-08
util: quote sh list safely

yvesll:
subprocess accepts bytes, PathLike obj and str but shlex.quote only accept str type. So need a normalize api.

### DIFF
--- a/src/west/util.py
+++ b/src/west/util.py
@@ -7,6 +7,7 @@
 import os
 import shlex
 import textwrap
+from collections.abc import Iterable
 from pathlib import Path
 
 # What west's APIs accept for paths.
@@ -39,9 +40,9 @@ def escapes_directory(path: PathType, directory: PathType) -> bool:
     return ret
 
 
-def quote_sh_list(cmd: list[str]) -> str:
+def quote_sh_list(cmd: Iterable[str | os.PathLike]) -> str:
     '''Transform a command from list into shell string form.'''
-    return ' '.join(shlex.quote(s) for s in cmd)
+    return ' '.join(shlex.quote(str(s)) for s in cmd)
 
 
 def wrap(text: str, indent: str) -> list[str]:


src/west/util.py
```diff
@@ -38,10 +39,18 @@ def escapes_directory(path: PathType, directory: PathType) -> bool:
         ret = True
     return ret
 
-
-def quote_sh_list(cmd: list[str]) -> str:
-    '''Transform a command from list into shell string form.'''
-    return ' '.join(shlex.quote(s) for s in cmd)
+def norm_sh_item(item: Union[str, bytes, os.PathLike]) -> str:
```
pdgendt: Union types should use the `|` operator.

```suggestion
def norm_sh_item(item: str | bytes | os.PathLike) -> str:
```

src/west/util.py
```diff
@@ -38,10 +39,18 @@ def escapes_directory(path: PathType, directory: PathType) -> bool:
...
+    """Normalize cmd list item to string."""
+    if isinstance(item, bytes):
+        enc = sys.getfilesystemencoding() or "utf-8"
+        return item.decode(enc, "surrogateescape")
+    if isinstance(item, os.PathLike):
+        return os.fspath(item)
+    return item
+
+def quote_sh_list(cmd: List[Union[str, bytes, os.PathLike]]) -> str:
```
pdgendt: Similar

```suggestion
def quote_sh_list(cmd: list[str | bytes | os.PathLike]) -> str:
```

pdgendt (changes requested):

thorsten-klein:
Can you please add a small test to showcase what exactly this PR fixes?
If I understand correctly it is a cosmetic change that makes sure that the printed command on stdout is valid?

src/west/util.py
yvesll: `|` operator was added in python3.10. I saw west in pypi supports 3.9, but in repo it's https://github.com/zephyrproject-rtos/west/blob/76890ef68124369c9b3ccd1ef968e47ab5f3183f/pyproject.toml#L11
which one is correct?

pdgendt: The current _released_ version on pypi supports 3.9, but this has been dropped in `main`, see https://github.com/zephyrproject-rtos/west/pull/839

yvesll:
> Can you please add a small test to showcase what exactly this PR fixes? If I understand correctly it is a cosmetic change that makes sure that the printed command on stdout is valid?

If developer use subprocess api in west extension and the cmd fails with a `CalledProcessError`, the `WestApp` class will catch and then print the cmd
https://github.com/zephyrproject-rtos/west/blob/76890ef68124369c9b3ccd1ef968e47ab5f3183f/src/west/app/main.py#L671

subprocess cmd list accepts bytes, str and path like objects but `quote_sh_list` only accepts str so it may raise another exception in this error handling block.
A real case is https://github.com/zephyrproject-rtos/zephyr/pull/100565

src/west/util.py
```diff
@@ -6,7 +6,9 @@
 
 import os
 import shlex
+import sys
 import textwrap
+from collections.abc import Sequence
```
pdgendt: Use `Iterable` instead as typing hint for `for in` loops

```suggestion
from collections.abc import Iterable
```

src/west/util.py
```diff
@@ -39,9 +41,19 @@ def escapes_directory(path: PathType, directory: PathType) -> bool:
...
+    """Normalize cmd list item to string."""
+    if isinstance(item, bytes):
+        enc = sys.getfilesystemencoding() or "utf-8"
+        return item.decode(enc, "surrogateescape")
+    if isinstance(item, os.PathLike):
+        return os.fspath(item)
+    return item
+
+
+def quote_sh_list(cmd: Iterable[PathType | bytes]) -> str:
```
pdgendt: Even though `PathType` has `str` in it's union collection, I'd rather have it explicit in the argument here. Otherwise it appears like you _should_ pass paths, which isn't the case.

```suggestion
def quote_sh_list(cmd: Iterable[str | bytes | PathType]) -> str:
```

src/west/util.py
```diff
@@ -39,9 +41,19 @@ def escapes_directory(path: PathType, directory: PathType) -> bool:
     return ret
 
 
-def quote_sh_list(cmd: list[str]) -> str:
-    '''Transform a command from list into shell string form.'''
-    return ' '.join(shlex.quote(s) for s in cmd)
+def norm_sh_item(item: PathType | bytes) -> str:
```
pdgendt: Putting it here makes it part of the public API, which I'm not sure is desired. I'd move it inside the `quote_sh_list` function  instead.

marc-hb (changes requested):
https://docs.python.org/3/library/subprocess.html is long and complicated but it's really not clear that it accepts `args` _bytes_... Can you please share a simple, `west`-free, subprocess example that demonstrates that subprocess accepts bytes?

Back to `west`, I agree with @thorsten-klein : this PR needs at least a very simple and clear test case; not something that depends on another fix that is not merged yet.

Even if subprocess really accepts bytes, that does not mean `quote_sh_list()` should accept bytes too. Unicode is a playground for security issues and having it near subprocess makes it even more dangerous. So we should keep things as simple as possible and  functions that accept both encoded and decoded data at the same time seem too error-prone to be kept secure over time. 


> `f'command exited with status {cpe.returncode}: {quote_sh_list(cpe.cmd)}', fatal=True `

So we can make this both simpler and better:
```
f'command exited with status {cpe.returncode}: {cpe.cmd}', fatal=True
```

`shlex.quote()` is cosmetic which is nice on Linux but it's not 100% reliable and can obscure what happens. When something fails, it's much more important to be accurate with `{cpe.cmd}` than to be "beautiful".

From https://docs.python.org/3/library/shlex.html#shlex.quote

> Warning The shlex module is only designed for Unix shells.
The [quote()](https://docs.python.org/3/library/shlex.html#shlex.quote) function is not guaranteed to be correct on non-POSIX compliant shells or shells from other operating systems such as Windows. Executing commands quoted by this module on such shells can open up the possibility of a command injection vulnerability.

yvesll:
The PR was originally to solve following issue:
```python
from pathlib import Path
from west.commands import WestCommand

class SubprocessFail(WestCommand):
    def __init__(self):
        super().__init__(
                'subprocess_fail',
                'Test subprocess fail',
                'Test subprocess fail')

        return parser_adder.add_parser(self.name,
                help=self.help,
                description = self.description)

    def do_run(self, args, ignored):
        self.check_call(["ls", Path("/nonexistent")])
```
Then it will fail in exit:
```shell
  File "/home/yves/Repos/zephyrproject/.venv/bin/west", line 7, in <module>
    sys.exit(main())
             ^^^^^^
  File "/home/yves/Repos/zephyrproject/.venv/lib/python3.12/site-packages/west/app/main.py", line 1199, in main
    app.run(argv or sys.argv[1:])
  File "/home/yves/Repos/zephyrproject/.venv/lib/python3.12/site-packages/west/app/main.py", line 278, in run
  File "/home/yves/Repos/zephyrproject/.venv/lib/python3.12/site-packages/west/app/main.py", line 584, in run_command
    self.run_extension(args.command, argv)
  File "/home/yves/Repos/zephyrproject/.venv/lib/python3.12/site-packages/west/app/main.py", line 739, in run_extension
    self.cmd.run(args, unknown, self.topdir, manifest=self.manifest,
  File "/home/yves/Repos/zephyrproject/.venv/lib/python3.12/site-packages/west/commands.py", line 200, in run
    self.do_run(args, unknown)
  File "/home/yves/Repos/zephyrproject/zephyr/scripts/west_commands/subprocess_fail.py", line 17, in do_run
    self.check_call(["ls", Path("/nonexistent")])
  File "/home/yves/Repos/zephyrproject/.venv/lib/python3.12/site-packages/west/commands.py", line 315, in check_call
    self._log_subproc(args, **kwargs)
  File "/home/yves/Repos/zephyrproject/.venv/lib/python3.12/site-packages/west/commands.py", line 303, in _log_subproc
    self.dbg(f"running '{quote_sh_list(args)}' in "
                         ^^^^^^^^^^^^^^^^^^^
  File "/home/yves/Repos/zephyrproject/.venv/lib/python3.12/site-packages/west/util.py", line 44, in quote_sh_list
    return ' '.join(shlex.quote(s) for s in cmd)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/yves/Repos/zephyrproject/.venv/lib/python3.12/site-packages/west/util.py", line 44, in <genexpr>
    return ' '.join(shlex.quote(s) for s in cmd)
                    ^^^^^^^^^^^^^^
  File "/usr/lib/python3.12/shlex.py", line 327, in quote
    if _find_unsafe(s) is None:
       ^^^^^^^^^^^^^^^
TypeError: expected string or bytes-like object, got 'PosixPath'
```

I agree with that we can ignore byte sequence because it is uncommon and not allowed under Windows. But I am not sure whether print a command without quote it is a correct way.
```shell
# Without quote
ls: cannot access '/nonexistent': No such file or directory
FATAL ERROR: command exited with status 2: ['ls', PosixPath('/nonexistent')]
# With quote
ls: cannot access '/nonexistent': No such file or directory
FATAL ERROR: command exited with status 2: ls /nonexistent
```
IMO, the second one is more friendly to user to reproduce it. User can get more accurate log with `west -vvv`
```shell
running 'ls /nonexistent' in /home/yves/Repos/zephyrproject/zephyr
ls: cannot access '/nonexistent': No such file or directory
FATAL ERROR: command exited with status 2: ls /nonexistent
=== Traceback (enabled by -vvv):
  File "/home/yves/Repos/west/src/west/app/main.py", line 633, in run_command
    self.run_extension(args.command, argv)
  File "/home/yves/Repos/west/src/west/app/main.py", line 791, in run_extension
    self.cmd.run(args, unknown, self.topdir, manifest=self.manifest, config=self.config)
  File "/home/yves/Repos/west/src/west/commands.py", line 214, in run
    self.do_run(args, unknown)
  File "/home/yves/Repos/zephyrproject/zephyr/scripts/west_commands/subprocess_fail.py", line 17, in do_run
    self.check_call(["ls", Path("/nonexistent")])
  File "/home/yves/Repos/west/src/west/commands.py", line 333, in check_call
    subprocess.check_call(args, **kwargs)
  File "/usr/lib/python3.12/subprocess.py", line 413, in check_call
    raise CalledProcessError(retcode, cmd)
subprocess.CalledProcessError: Command '['ls', PosixPath('/nonexistent')]' returned non-zero exit status 2.
```

pdgendt (approved):

marc-hb:
The code change looks good to me, thanks!  The commit message in https://github.com/zephyrproject-rtos/west/commit/2b3445f91cca37b5e8732b5c44254f71425dc6d9 still does not make sense to me:

> subprocess accepts bytes, PathLike obj and str but shlex.quote only
> accept str type. So need convert PathLike object to str.
> Ignore byte sequence because it is uncommon and not allowed under
> Windows.

Where did you see that subprocess accepts bytes?

Do you realize the error message below is NOT from subprocess?

```
>>> import shlex, pathlib
>>> shlex.quote(pathlib.Path("."))
  File "<python-input-1>", line 1, in <module>
    shlex.quote(pathlib.Path("."))
    ~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^
  File "/usr/lib64/python3.13/shlex.py", line 327, in quote
    if _find_unsafe(s) is None:
       ~~~~~~~~~~~~^^^
TypeError: expected string or bytes-like object, got 'PosixPath'

```


----------

(A bit off-topic in this PR)

> IMO, the second one is more friendly to user to reproduce it.

Yes it's more user friendly; except when... it's WRONG:

> Warning The shlex module is only designed for Unix shells.
The [quote()](https://docs.python.org/3/library/shlex.html#shlex.quote) function is not guaranteed to be correct on non-POSIX compliant shells or shells from other operating systems such as Windows. Executing commands quoted by this module on such shells can open up the possibility of a command injection vulnerability.

Incorrect is always less user-friendly than correct: much more time-consuming. Also, joining the list of arguments loses debug information.

yvesll:
> Do you realize the error message below is NOT from subprocess?

I see, the `quote_sh_list` was usually used to quote `args` passed to subprocess, so I just mentioned it.

> Incorrect is always less user-friendly than correct: much more time-consuming. Also, joining the list of arguments loses debug information.

For `debug`, user needs to specify verbose parameters. Otherwise the default console logs look a little duplication.

marc-hb:
So do you plan to change the commit message?

yvesll:
> So do you plan to change the commit message?

No, it's clear.

marc-hb:
It's clear but most of the comments about bytes seem incorrect.

yvesll:
What's do you think about this:
```shell
subprocess args can accept str, PathLike, or bytes (POSIX only), but shlex.quote() only works with str. 
For safety and cross-platform compatibility, convert PathLike to str and ignore bytes since they are rare and unsupported on Windows.
```

marc-hb:
> Where did you see that subprocess accepts bytes?

Still waiting for the answer to that question.

As far as I can see, subprocess accepts strings and `class PathLike`

yvesll:
> > Where did you see that subprocess accepts bytes?
> 
> 
> 
> Still waiting for the answer to that question.
> 
> 
> 
> As far as I can see, subprocess accepts strings and `class PathLike`

https://docs.python.org/3/library/subprocess.html
image

marc-hb:
tl;dr: please drop the reference to Windows in the commit message.


> For safety and cross-platform compatibility, convert PathLike to str and ignore bytes since they are rare and unsupported on Windows.

- "Rare" - agreed. Or, better: "discouraged", on every Operating System.

- "Unsupported on Windows": no, it's not that simple. I did find Windows-specific issues but only when "abusing" the subprocess API a bit. For instance, these encoded bytes work fine across all operating systems, including Windows:

```python
import os, subprocess
subprocess.check_call([os.fsencode('curl'),  os.fsencode('--help')], shell=False)
```


> Changed in version 3.8: args parameter accepts a path-like object if shell is False and a sequence containing bytes and path-like objects on Windows.

Yes I saw that. But it's only a changelog, it's the not part of actual API description. Funny how a changelog refers to a non-existing documentation? I did a fair amount of research and reading and even some testing and I think I solved that mystery.

(BTW avoid screenshots when not necessary https://docs.zephyrproject.org/latest/develop/getting_started/index.html#use-copy-paste)

That changelog was about a bug fix for a Windows-specific bug https://github.com/python/cpython/pull/5914/files

I think the most relevant and interesting document is PEP 519. Long story short, the purpose of PEP 519 is to encourage people to use PathLike objects and to discourage people from using encoded strings (bytes).

https://peps.python.org/pep-0519/#os

> Another way to view this is as a hierarchy of file system path representations (highest- to lowest-level): path → str → bytes. The functions and classes under discussion can all accept objects on the same level of the hierarchy, but they vary in whether they promote or demote objects to another level.

However, Python could not get rid of bytes for backwards compatibility reasons:

https://peps.python.org/pep-0519/#have-fspath-only-return-strings

> Much of the discussion that led to this PEP revolved around whether __fspath__() should be polymorphic and return bytes as well as str or only return str. The general sentiment for this view was that bytes are difficult to work with due to their inherent lack of information about their encoding [...] In the end, it was decided that using bytes to represent paths is simply not going to go away and thus they should be supported to some degree. The hope is that people will gravitate towards path objects like pathlib and that will move people away from operating directly with bytes.


So, I think this is why bytes are never directly mentioned in the API description; only indirectly as in the changelog, or as some PathLike implementation detail.

So the reason why we don't want to support encoded bytes in this quote function is because no extension should ever use bytes on any operating system. Not because misusing the subprocess API on Windows can sometimes fail in some obscure and barely legal use cases.


Some testing:

```python
 # Python 3.12. shell=False is the (sane) default
 import os,subprocess

 # Bad `shell` usage that works only Windows! Because winapi.CreateProcess() is a pretty bad API
 subprocess.check_call('curl --help', shell=False)

 # But once encoded, this fails on all operating systems as it should - even Windows. Go figure.
 subprocess.check_call(os.fsencode('curl --help'), shell=False)
       hp, ht, pid, tid = _winapi.CreateProcess(executable, args,
                       ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
       FileNotFoundError: [WinError 2] The system cannot find the file specified

 # bytes that works everywhere across all operating systems including Windows:
 subprocess.check_call([os.fsencode('curl'),  os.fsencode('--help')], shell=False)

 # Encoded combination that indeed fails only on Windows - no clear reason why.
 # But there is no reason to encode when using `shell=True` anyway.
 subprocess.check_call(os.fsencode('curl --help'), shell=True)
 TypeError: bytes args is not allowed on Windows
```

yvesll:
> tl;dr: please drop the reference to Windows in the commit message.

Make sense! Thanks for the detailed references.

marc-hb (approved):

PR #904, opened 2026-01-12
project: Replace pykwalify with jsonschema

### LINKED ISSUE #807: Deprecate `pyKwalify` for YAML validation
The [pyKwalify](https://pypi.org/project/pykwalify/) project hasn't received updates since 2020 and the highest supported Python version is 3.9.

As an alternative, I propose moving to [JSON schema](https://json-schema.org/)

- An active project
- [YAML as a superset of JSON](https://json-schema-everywhere.github.io/yaml)
- Clear [specifications](https://json-schema.org/specification)
- Backed by large companies

pdgendt:
- Convert pykwalify schema with jsonschema
- Remove some manual testing in Python code with schema validation
- Make name property required in west extension commends (would have raised a `KeyError` before if it was missing)

Fixes https://github.com/zephyrproject-rtos/west/issues/807


marc-hb (commented):
Thanks!
Can this catch quote strings that look like numbers in YAML files and force them to quote them?

Just curious because we just saw two instances of this back to back in Zephyr.

I didn't look at the PR yet sorry.

src/west/west-commands-schema.yaml
```diff
@@ -0,0 +1,40 @@
...
+title: West Commands
+description: |
+  A JSON schema for validation of west commands YAML files. This allows
+  projects managed by west to provide additional commands.
+
+  By convention, this file is named west-commands.yml.
+
+  --------------------------------------------------------------
+
+  DO NOT CHANGE THIS FILE WITHOUT UPDATING THE DOCUMENTATION!
```
thorsten-klein, line 12: Did you update the documentation (since `name` is now required)?

pdgendt:
> Thanks! Can this catch quote strings that look like numbers in YAML files and force them to quote them?
> 
> Just curious because we just saw two instances of this back to back in Zephyr.
> 
> I didn't look at the PR yet sorry.

Not really, no. The version property, for example, currently allows both string and number, to support something like:

```yaml
version: 0.8
```

Is this a problem? I think we would break a lot of users if we take this away?

EDIT: Users could already be forced by setting the type to `string`, no?

src/west/west-commands-schema.yaml
pdgendt: It "was" already required, as it would throw an error if users did not provide it.

But we probably should bump the manifest version (and docs), so that users could keep compatibility with versions that depend on `pykwalify`. Will update.

marc-hb:
>  The version property, for example, currently allows both string and number, to support something like: version: 0.8
> Is this a problem?

It depends where. I didn't mean "everywhere always", sorry for the confusion.

It's not a problem at all where the code is ready and expects both. But it has recently been a problem in Zephyr in two places (https://github.com/zephyrproject-rtos/zephyr/pull/101642 and https://github.com/zephyrproject-rtos/zephyr/pull/101888#discussion_r2684391058) where the code expects a string and the user (or file generator) entered _a **string** that unfortunately looks like a valid number_. 

>  I think we would break a lot of users if we take this away?

Again, it depends where. This new parser can also save users time by "failing fast" and giving them a clear error message that they must quote the strings that look like numbers, instead of some obscure stack trace much later.

marc-hb:
So let's pretend for instance that:

- A user uses a `path` that is also a valid integer. If it's not quoted then `yaml.load()` with automagically return that integer instead of a string. Very debatable `yaml` design choice but it is what it is.
- The `west` code "naively" expects a string and a string only. Why would it expect anything else? It's typed as a `string` in the manifest.

This is exactly what just happened in Zephyr and it could happen anywhere in west too.

In that case, it would be a better experience that `jsonschema` "fails fast" and pinpoints the problematic integer with a useful quoting suggestion rather than `west` blowing up much later with a stack trace difficult to make sense of.

Can `jsonschema` do that? It should considering how many people on the Internet complain about this common `yaml.load()` problem... Maybe pkwalify already did that? I don't know sorry.

src/west/manifest-schema.yaml
```diff
@@ -0,0 +1,235 @@
...
+
+type: object
+additionalProperties: false
+properties:
+  # The "version" key is optional. It's the minimum version manifest
+  # schema that is required to correctly parse the manifest data.
+  version:
+    oneOf:
+      - type: string
+      - type: number
```
marc-hb, line 28: `0.15.1` is not a valid number so it's a bit strange to allow (floating?) numbers. This looks like a hack to work around the funky `yaml.load()` behavior.

BTW `0.8` is not a floating number either in some locales where `0,8` is instead.

uv.lock
```diff
@@ -452,69 +456,139 @@ wheels = [
...
-    { url = "https://files.pythonhosted.org/packages/e7/cd/150fdb96b8fab27fe08d8a59fe67554568727981806e6bc2677a16081ec7/ruamel_yaml_clib-0.2.14-cp314-cp314-win32.whl", hash = "sha256:9b4104bf43ca0cd4e6f738cb86326a3b2f6eef00f417bd1e7efb7bdffe74c539", size = 102394, upload-time = "2025-11-14T21:57:36.703Z" },
-    { url = "https://files.pythonhosted.org/packages/bd/e6/a3fa40084558c7e1dc9546385f22a93949c890a8b2e445b2ba43935f51da/ruamel_yaml_clib-0.2.14-cp314-cp314-win_amd64.whl", hash = "sha256:13997d7d354a9890ea1ec5937a219817464e5cc344805b37671562a401ca3008", size = 122673, upload-time = "2025-11-14T21:57:38.177Z" },
+    { url = "https://files.pythonhosted.org/packages/06/0c/0c411a0ec64ccb6d104dcabe0e713e05e153a9a2c3c2bd2b32ce412166fe/rpds_py-0.30.0-cp310-cp310-macosx_10_12_x86_64.whl", hash = "sha256:679ae98e00c0e8d68a7fda324e16b90fd5260945b45d3b824c892cec9eea3288", size = 370490, upload-time = "2025-11-30T20:21:33.256Z" },
```
marc-hb, line 478: Could `wheels` be upgraded in a separate, routine PR to make the dependency changes reviewable?

src/west/manifest-schema.yaml
pdgendt: Note, that this change matches the `pykwalify` validation for

```yaml
  version:
    required: false
    type: text
```

From the [docs](http://pykwalify.readthedocs.io/en/unstable/validation-rules.html):

> - text
>   - Validates if value is str or number

pdgendt:
> So let's pretend for instance that:
> 
> * A user uses a `path` that is also a valid integer. If it's not quoted then `yaml.load()` with automagically return that integer instead of a string. Very debatable `yaml` design choice but it is what it is.
> * The `west` code "naively" expects a string and a string only. Why would it expect anything else? It's typed as a `string` in the manifest.
> 
> This is exactly what just happened in Zephyr and it could happen anywhere in west too.
> 
> In that case, it would be a better experience that `jsonschema` "fails fast" and pinpoints the problematic integer with a useful quoting suggestion rather than `west` blowing up much later with a stack trace difficult to make sense of.
> 
> Can `jsonschema` do that? It should considering how many people on the Internet complain about this common `yaml.load()` problem... Maybe pkwalify already did that? I don't know sorry.

Both `pykwalify` and `jsonschema` would fail fast if a property needs to be of a string type. If you could point to a particular property or an issue, I could maybe try to improve. But it's outside the scope of this PR.

The same can be said for the `revision` string property, if by coincidence the sha is a number like `123456890`, the schema validation would fail and quotes would be needed.

uv.lock
pdgendt: It's automated when packages are added/removed, where `uv` needs to resolve dependencies. I can't move to a separate PR, but I've split the commits as much as possible.

src/west/manifest-schema.yaml
kartben: https://github.com/zephyrproject-rtos/zephyr/blob/main/scripts%2Fschemas%2Fboard-schema.yaml -- you can prob implement similar validation as we have for board revisions? But starting with something that's functionally equivalent to the status quo with pykwalify is also fine by me

pdgendt: I'd prefer the status quo, and focus this PR on changing the tooling + removing manual checks.

marc-hb:
> Both pykwalify and jsonschema would fail fast if a property needs to be of a string type. If you could point to a particular property or an issue, I could maybe try to improve. But it's outside the scope of this PR.

Thanks! A change about a particular property would be outside the scope of this issue, agreed. I was looking for examples to illustrate my point. I agree this PR should not make a schema change at the same time.

On the other hand, gaining or losing that sort of validation capability _in general_ would definitely be in the scope of this PR! That's what I've been wondering the whole time. I understand you tested that and made sure user feedback is (still) good when jsonschema catches and reports a string that looks like a (floating?!?) number?

pdgendt:
> On the other hand, gaining or losing that sort of validation capability _in general_ would definitely be in the scope of this PR! That's what I've been wondering the whole time. I understand you tested that and made sure user feedback is (still) good when jsonschema catches and reports a string that looks like a (floating?!?) number?

Which property are you referring to? If you refer to the `version`, do you require more testing than that already present in `tests/test_manifest.py`?

marc-hb:
> Which property are you referring to?

Again, I'm just looking for examples, I'm not interested in any particular property. Apologies for the too deep digression on `version`.

This is about testing `jsonschema` itself, not about testing `west`.

> If you refer to the version, do you require more testing than that already present in tests/test_manifest.py?

I wasn't focused on "version". I took a quick look at `tests/test_manifest.py` and I didn't find anything testing `jsonschema` (resp. `pkwalify`) itself. And that's OK, we can hopefully just trust jsonschema. Trust... but verify :-)

Also, I'm interested in the _interactive_ feedback when validation failed. It's not realistic to expect automated testing to guarantee a "user experience" (We're lucky enough when there is any test coverage at all for error handling...)

So, I just went ahead and tested this myself; before and after this PR. As you predicted, the experience is OK in both cases. Neither pkwalify nor jsonschema suggests quoting numbers but they are close enough:

```diff
 zephyr]$ git diff
diff --git a/west.yml b/west.yml
index a11238717de4..4a53b99fa445 100644
--- a/west.yml
+++ b/west.yml
@@ -16,10 +16,10 @@
 
 manifest:
   defaults:
-    remote: upstream
+    remote: 12345
 
   remotes:
-    - name: upstream
+    - name: 12345
       url-base: https://github.com/zephyrproject-rtos
```

With `pkwalify`:

```
west update

FATAL ERROR: Malformed manifest file: zephyr/west.yml 
  Schema file: src/west/manifest-schema.yml
  Hint: Schema validation failed:
 - Value '12345' is not of type 'str'. Path: '/defaults/remote'.
 - Value '12345' is not of type 'str'. Path: '/remotes/0/name'.
```

with `jsonschema`:

```
west update

FATAL ERROR: Malformed manifest file: zephyr/west.yml 
  Schema file: src/west/manifest-schema.yaml
  Hint: 12345 is not of type 'string'
```

As expected, `remote: "12345"` works.

Note `jsonschema` does not give any location hint. Acceptable regression?

uv.lock
marc-hb: > It's automated when packages are added/removed, where uv needs to resolve dependencies.

OK but how come switching from pkwalify to jsonschema changes the `wheels` version?

> I can't move to a separate PR, but I've split the commits as much as possible.

Good enough.

pdgendt:
> Note `jsonschema` does not give any location hint. Acceptable regression?

Updated the error message so it now provides more details:

```shell
$ west update
FATAL ERROR: Malformed manifest file: /home/pdgendt/zephyrproject/zephyr/west.yml
  Schema file: /home/pdgendt/west/src/west/manifest-schema.yaml
  Hint: 12345 is not of type 'string'

Failed validating 'type' in schema['properties']['defaults']['properties']['remote']:
    {'type': 'string'}

On instance['defaults']['remote']:
    12345
```

src/west/manifest-schema.yaml
```diff
@@ -0,0 +1,229 @@
...
+
+        # Path to check out, relative to the west installation root.
+        # The default is the name.
+        path:
+          type: string
+        # Submodules to checkout within the project
+        submodules:
+          oneOf:
+            - type: boolean
+            - type: array
```
marc-hb, line 103: Nice but that does not seem to match your "status quo" intention, does it?

Move to a separate commit?

src/west/manifest-schema.yaml
```diff
@@ -0,0 +1,229 @@
...
+    $ref: "#/$defs/groups"
+
+  # The "self" key specifies values for the project containing the manifest
+  # file (the "manifest repository").
+  self:
+    type: object
+    additionalProperties: false
+    properties:
+      path:
+        type: ["string", "null"]
```
marc-hb, line 147: Why "null"? Does this mean empty?

src/west/manifest-schema.yaml
```diff
@@ -0,0 +1,229 @@
...
+      # Additional manifest data to import from the manifest
+      # repository.
+      import:
+        oneOf:
+          # No boolean import here
+          - type: string
+          - type: array
+            items:
+              type: string
+          - $ref: "#/$defs/import-object"
```
marc-hb, line 161: As above, a bit cryptic, add a comment?

src/west/manifest-schema.yaml
```diff
@@ -0,0 +1,229 @@
...
+            - $ref: "#/$defs/import-object"
+        groups:
+          $ref: "#/$defs/groups"
+        # Arbitrary project user data.
+        userdata: {}
+
+  # If present, a list of project groups to enable and disable. Prefix
+  # a group name with "-" to disable it; prefix with "+" to enable it.
+  group-filter:
+    $ref: "#/$defs/groups"
```
marc-hb, line 138: Is this just a textual replacement or something smarter? I find the syntax quite cryptic, can you add a one-line comment?

src/west/manifest-schema.yaml
```diff
@@ -0,0 +1,229 @@
...
+            items:
+              type: string
+      path-blacklist:
+        oneOf:
+          - type: string
+          - type: array
+            items:
+              type: string
+
+  groups:
```
marc-hb, line 224: Does this have to be a the bottom? It was at the top with pkwalify and moving it makes the side-by-side (and: tedious) comparison significantly more time-consuming.

src/west/manifest-schema.yaml
```diff
@@ -0,0 +1,229 @@
...
+          - type: string
+          - type: array
+            items:
+              type: string
+          - $ref: "#/$defs/import-object"
+      # Arbitrary user data for manifest repository
+      userdata: {}
+
+$defs:
+  import-object:
```
marc-hb, line 166: Was this in pkwalify? I could not find it.

marc-hb (commented):
I don't have any alternative to suggest and the interwebs don't seem to have any either but I find it sad to discover that jsonscheme seems to introduce the very first, 3rd party, mandatory binary dependency for west, correct? This looks like a transparency and security regression to me. The 100+ lines of unreviewable checksums and git pollution in uv.lock is the icing on the cake :-(

src/west/west-commands-schema.yaml
marc-hb: > But in order for users to target a version without the tooling change,
> we need to bump the schema version.

I don't understand this sorry. Can you elaborate the use case?

If the schema is 100% compatible, why would the schema version be bumped?

```
MAINTAINERS.rst:   Decide if west.manifest.SCHEMA_VERSION needs an update:
MAINTAINERS.rst-
MAINTAINERS.rst:   - SCHEMA_VERSION should be updated to X.Y if release vX.Y will have manifest
MAINTAINERS.rst-     syntax changes that earlier versions of west cannot parse.
MAINTAINERS.rst-
MAINTAINERS.rst:   - SCHEMA_VERSION should *not* be changed for west vX.Y if the manifest
MAINTAINERS.rst-     syntax is fully compatible with what west vX.(Y-1) can handle.
```

kartben: > > But in order for users to target a version without the tooling change,
> > we need to bump the schema version.
> 
> I don't understand this sorry. Can you elaborate the use case?
> 
> If the schema is 100% compatible, why would the schema version be bumped?
> 
> ```
> MAINTAINERS.rst:   Decide if west.manifest.SCHEMA_VERSION needs an update:
> MAINTAINERS.rst-
> MAINTAINERS.rst:   - SCHEMA_VERSION should be updated to X.Y if release vX.Y will have manifest
> MAINTAINERS.rst-     syntax changes that earlier versions of west cannot parse.
> MAINTAINERS.rst-
> MAINTAINERS.rst:   - SCHEMA_VERSION should *not* be changed for west vX.Y if the manifest
> MAINTAINERS.rst-     syntax is fully compatible with what west vX.(Y-1) can handle.
> ```

The schema file has an entirely different syntax so... Yeah :)
It's the _semantics_ that's meant to remain identical.

pdgendt: The _manifest syntax_ is the same though? I get Marc's point, but I think it would be beneficial to have different (minor) versions in this case? Not sure how to proceed here.

pdgendt:
> jsonscheme seems to introduce the very first, 3rd party, mandatory binary dependency for west, correct? This looks like a transparency and security regression to me.

Is this a deal-breaker?

src/west/west-commands-schema.yaml
kartben: FWIW in Zephyr I ended up picking a different name for the new manifest (one was .yml the other .yaml) - would doing something similar help?

pdgendt: I did the same here :)

marc-hb:
> Is this a deal-breaker?

No. Do we have a choice anyway? As you found, pkwalify seems orphaned and there is no alternative, is there?

marc-hb:
Actually, could we make validation optional?

Something like:

```python
   try:
      import jsonschema
      try:
        jsonschema.validate()
      ...
   expect ImportError
       self.wrn("f{manifest} not validated, UNSUPPORTED do not report bugs")
```

I wouldn't suggest making this optional if west were used by Zephyr only. But `west` wants to be more generic and universal and maybe run on more "exotic" systems that Zephyr development would not be compatible with.

src/west/west-commands-schema.yaml
marc-hb: I submitted some SCHEMA_VERSION clarifications in PR #909. I hope they explain why I think this should not be bumped.

> FWIW in Zephyr I ended up picking a different name for the new manifest (one was .yml the other .yaml) 

You meant the _schema_.y[a]ml, right? Not the west.yml manifest which should hopefully not be affected.

If I got this correctly, then I don't see how a mere rename of the schema file affects the SCHEMA_VERSION. It would be a very different story if pykwalify and jsonschema were not 100% compatible with each other! But I don't think the mere name of the schema file matters here in any case.

bjarki-andreasen (approved):
We need this sooner than later, zephyr is currently failing to `west packages pip --install` because its missing jsonschema

pdgendt:
> We need this sooner than later, zephyr is currently failing to `west packages pip --install` because its missing jsonschema

The issue is larger, the `west packages` Zephyr extension relies on whatever https://github.com/zephyrproject-rtos/zephyr/blob/0bb29433349fa8d7c472eddc3e462123a87e5138/scripts/zephyr_module.py has as dependencies. It ~~can't~~ shouldn't rely on either `pykwalify` or `jsonschema` to be present.

marc-hb:
> We need this sooner than later, zephyr is currently failing to west packages pip --install because its missing jsonschema

I don't think this would have ~helped~ been enough because `west` and `zephyr` are separate from each other and you don't always have to upgrade `west` when you upgrade Zephyr. Sometimes yes but not systematically and not unnecessarily. They don't have to be "in sync". In other words, it must be possible to combine a cutting-edge Zephyr that has a newer `jsonschema` dependency with a slightly older `west` that does not require `jsonschema` yet.

Anyway https://github.com/zephyrproject-rtos/zephyr/pull/103671 made neither  `pkwalify` nor `jsonschema`  a hard `zephyr_module.py` requirement which is even better.

----

> Actually, could we make validation optional?
> Something like...

... which now looks like a lot like https://github.com/zephyrproject-rtos/zephyr/commit/a0a71dabf8401ab4d :-) Can we have the same here?

PR #905, opened 2026-01-14
west: fix crash when extension command fails before cmd init

vishwas-droid:
While working with west extension commands, I ran into a failure mode where an extension error occurs before the command context (`self.cmd`) is fully initialized.

I spent time tracing the extension loading path and verifying how errors are surfaced at different verbosity levels. In this early-failure case, west attempted to report the error via `self.cmd.err()`, which resulted in an AttributeError and obscured the original cause of the failure.

This change adjusts the error-handling logic to explicitly handle the case where `self.cmd` is not yet available. When a command context exists, the existing behavior and verbosity-controlled traceback output are preserved. When it does not, the error is safely reported via stderr, ensuring the original failure is visible instead of being masked by a secondary exception.

The change is intentionally minimal and localized, and I verified it by exercising the failure path and confirming normal west behavior (including `west --help`) remains unaffected.

### DIFF
--- a/src/west/app/main.py
+++ b/src/west/app/main.py
@@ -462,24 +462,28 @@ class WestApp:
         }
 
     def handle_extension_command_error(self, ece):
+
+
         if self.cmd is not None:
-            msg = f"extension command \"{self.cmd.name}\" couldn't be run"
+            msg = f'extension command "{self.cmd.name}" could not be run'
         else:
-            self.cmd = self.builtins['help']
             msg = "could not load extension command(s)"
 
         if ece.hint:
-            msg += '\n  Hint: ' + ece.hint
+            msg += "\n  Hint: " + ece.hint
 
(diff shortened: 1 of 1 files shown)


pdgendt:
I assume this is already fixed in https://github.com/zephyrproject-rtos/west/pull/894, but not yet released.

vishwas-droid:
@pdgendt > I assume this is already fixed in #894, but not yet released.

Thanks for pointing that out.

I wasn’t aware that this had already been addressed in #894 when I started looking into this. My intent here was to fix the crash I encountered on the released version, and to make sure the failure path is handled cleanly when an extension command fails before cmd initialization.

If this is already covered by #894 and will be resolved by the next release, I’m happy to close this PR to avoid duplication.

Please let me know if you’d prefer this to be closed, or if there’s any value in keeping it around as a backport or reference.

Thanks!

pdgendt:
I'm going to close this for now, as it's already fixed.

PR #906, opened 2026-01-17
project: die if the --manifest-file argument is not relative

marc-hb:
This was found while testing the --topdir pull request #854 in progress. It's valid with or without it.


pdgendt (approved):
A test to catch this would be nice.

carlescufi (approved):

Issue #910, opened 2026-01-21
Can not add project alongside .west folder


chirambaht:
I am having trouble with a project of mine. I was using west v1.2.0, and when I upgraded to v1.3.0 (up to v1.5.0 as well), I found that I could not run west update. Here is my project structure: 
```
project
├── app
│   └── west.yml
├── support_repo_1
├── support_repo_2
├── support_repo_3
└── .west
    └── config
```
I would want to clone a repo to the project folder but west wont allow me to have something like this in my manifest:
```
    - name: support_repo
    path: .
    revision: support_repo_4
    url: git@github.com:support_4.git
```
Is there something that I am missing in all of this. I was following the release notes which reference this MR: #754  and this Issue: #102

pdgendt:
Not sure I follow, but a project can't be cloned to the workspace directory, if that is what you're trying?

chirambaht:
we run the following command to init our project: 
```
west init -m application_repo_url project
```
This will create the folder structure that I presented in the original issue ticket. 
@pdgendt

marc-hb:
Can you share an error message and more detailed reproduction steps? Even better: a sample manifest that reproduces.

BTW:
- https://github.com/zephyrproject-rtos/zephyr/pull/84337 / 
https://github.com/zephyrproject-rtos/zephyr/commit/7d40091fbfedfc51596447212b408a06cefd702b

chirambaht:
Here is the error I am receiving running west v1.5.0: 
```
Traceback (most recent call last):
  File "/opt/python/venv/bin/west", line 7, in <module>
    sys.exit(main())
             ^^^^^^
  File "/opt/python/venv/lib/python3.12/site-packages/west/app/main.py", line 1199, in main
    app.run(argv or sys.argv[1:])
  File "/opt/python/venv/lib/python3.12/site-packages/west/app/main.py", line 269, in run
    self.load_manifest()
  File "/opt/python/venv/lib/python3.12/site-packages/west/app/main.py", line 288, in load_manifest
    self.manifest = Manifest.from_topdir(topdir=self.topdir,
                    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/python/venv/lib/python3.12/site-packages/west/manifest.py", line 1238, in from_topdir
    return Manifest(topdir=topdir, config=config,
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
[...]

PR #911, opened 2026-01-22
manifest: restore "path: ." ability but with a warning

marc-hb:
3 commits. Main commit:

manifest: restore "path: ." ability but with a warning

Restore the ability to have a project at the top level but with an
"unsupported" warning printed every time.

Fixes v1.3 commit https://github.com/zephyrproject-rtos/west/commit/2b2d3f2685e688974690c66dfeb1c8d878c071fd ("manifest: Do not allow projects inside
the west directory") which accidentally broke it with an pretty obscure
backtrace:
```
File "python/site-packages/west/manifest.py", line 2460, in _load_project
    if Path(ret_norm).parts[0] == util.WEST_DIR:
       ~~~~~~~~~~~~~~~~~~~~^^^
IndexError: tuple index out of range
```
Fixes issue number https://github.com/zephyrproject-rtos/west/issues/910

This is STILL not supported!

### DIFF
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -2523,19 +2523,21 @@ class Manifest:
         if ret_norm[0] in '/\\' or os.path.isabs(ret_norm):
             self._malformed(
                 f'project "{ret.name}" has absolute path '
-                f'{ret.path}; this must be relative to the '
+                f'"{ret.path}"; this must be relative to the '
                 f'workspace topdir' + (f' ({self.topdir})' if self.topdir else '')
             )
 
         if ret_norm.startswith('..'):
             self._malformed(
-                f'project "{name}" path {ret.path} '
-                f'normalizes to {ret_norm}, which escapes '
+                f'project "{name}" path "{ret.path}" '
+                f'normalizes to "{ret_norm}", which escapes '
                 f'the workspace topdir'
             )
 
         if Path(ret_norm).parts[0] == util.WEST_DIR:
-            self._malformed(f'project "{name}" path {ret.path} is in the {util.WEST_DIR} directory')
+            self._malformed(
+                f'project "{name}" path "{ret.path}" is in the {util.WEST_DIR} directory'
+            )
 
         return ret
 
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -2534,7 +2534,19 @@ class Manifest:
                 f'the workspace topdir'
             )
 
-        if Path(ret_norm).parts[0] == util.WEST_DIR:
+        _path = Path(ret_norm)
+
+        # Unsupported, see https://github.com/zephyrproject-rtos/zephyr/commit/7d40091fbfedfc
+        if _path == Path('.'):
+            _logger.warning("Making the topdir a git clone is not supported; do not report bugs.")
+            return ret
+
+        # Can any other Path have zero part? Just in case, provide a decent error message
+        # instead of "index out of range" (see #910)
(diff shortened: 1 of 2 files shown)


chirambaht:
Tested this and for the most part it works well, thanks! Only found this when running it on `west update`
```
(.venv) user@6f98e55bdaba:~/project$ west update
west.manifest: WARNING: Making the topdir a git clone is not supported; do not report bugs.
west.manifest: WARNING: Making the topdir a git clone is not supported; do not report bugs.
```
I suppose you wouldn't want the warning to print twice.

src/west/manifest.py
```diff
@@ -2523,19 +2523,31 @@ def _load_project(self, pd: dict, url_bases: dict[str, str], defaults: _defaults
...
             )
 
-        if Path(ret_norm).parts[0] == util.WEST_DIR:
-            self._malformed(f'project "{name}" path {ret.path} is in the {util.WEST_DIR} directory')
+        _path = Path(ret_norm)
+
+        # Unsupported, see https://github.com/zephyrproject-rtos/zephyr/commit/7d40091fbfedfc
+        if _path == Path('.'):
+            _logger.warning("Making the topdir a git clone is not supported; do not report bugs.")
+            return ret
```
pdgendt, line 2542: We don't allow the manifest project to be in the `topdir` directory, I would argue that we shouldn't allow "regular" projects to do it either.

marc-hb: > We don't allow the manifest project to be in the topdir directory,

Is that actually blocked or just discouraged?

As far as I understand,
https://github.com/zephyrproject-rtos/west/commit/2b2d3f2685e688974690c66dfeb1c8d878c071fd says "it is discouraged for the topdir to be a git repo". Anything else on that topic?

Note the manifest does not have to be in git at all: that is documented and works fine.

marc-hb (commented):
> I suppose you wouldn't want the warning to print twice.

Ideally not, but it does not really matter because the workspace is not supposed to be a git repo.

It may not be simple to print this only once. If you find a very simple way with barely additional code then please share.

src/west/manifest.py
```diff
@@ -2523,19 +2523,31 @@ def _load_project(self, pd: dict, url_bases: dict[str, str], defaults: _defaults
...
+
+        # Unexpected
+        if len(_path.parts) < 1:
+            self._malformed(f'project "{name}" path "{ret.path}" has no part?! ({_path})')
+        elif _path.parts[0] == util.WEST_DIR:
+            self._malformed(
+                f'project "{name}" path "{ret.path}" is in the {util.WEST_DIR} directory'
+            )
```
pdgendt, line 2552: Nit, no need for `elif`, `self._malformed` will exit. Can `len(_path.parts) < 1` occur?

```suggestion
        if len(_path.parts) < 1:
            self._malformed(f'project "{name}" path "{ret.path}" has no part?! ({_path})')
        
        if _path.parts[0] == util.WEST_DIR:
            self._malformed(
                f'project "{name}" path "{ret.path}" is in the {util.WEST_DIR} directory'
            )
```

pdgendt (approved):

src/west/manifest.py
marc-hb: > Nit, no need for elif, self._malformed will exit. 

Done.

> Can len(_path.parts) < 1 occur?

I could not find another one besides `Path('.)` but better safe than sorry? I added a comment explaining.

carlescufi (approved):

PR #912, opened 2026-01-24
test_config.py: rename "config" to "wconfig"

marc-hb:
Every object and their dog is named something_config_something, so the bare "config" is a really bad name. Rename it to "wconfig" which is still very short to type but much less ambiguous.


marc-hb:
Any other better name wanted. Anything but "config"!

pdgendt (approved):

PR #913, opened 2026-01-29
feat: add --topdir commandline option

sillydan1:
## Description

This removes the enforcement of "supported topologies" (https://docs.zephyrproject.org/latest/develop/west/workspaces.html#topologies-supported) and enables users to have whatever kind of topology they want.

This also enables the following workflow:
```bash
# Clone the project using regular git, just like every other project
# in the world.
git clone git@github.com:foo/my-project

# Build the project using west.
west --topdir=./build init --local .
cd build
west build <target> <etc>
```

This solution also sidesteps the odd restriction that the topdir should be located in `..`, which can result in polluting random locations and makes some CI workflows difficult.

## Known Issues

For some reason, having `manifest.path` set as an absolute path produces a warning. If any maintainers can explain why this is the case, that would be much appreciated.
Due to this, the unit test suite is failing.

### DIFF
--- a/src/west/app/main.py
+++ b/src/west/app/main.py
@@ -560,6 +560,13 @@ class WestApp:
         )
 
         parser.add_argument(
+            '-t',
+            '--topdir',
+            default=None,
+            help='''Override west topdir.''',
+        )
+
+        parser.add_argument(
             '-q',
             '--quiet',
             default=0,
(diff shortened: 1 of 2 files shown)


pdgendt:
Not sure if you noticed https://github.com/zephyrproject-rtos/west/pull/854

sillydan1:
> Not sure if you noticed #854

Thanks! I will wait for that PR to be merged then. In the meantime, I have adopted the solution described in https://github.com/zephyrproject-rtos/west/issues/774#issuecomment-3369009108

Issue #908, opened 2026-01-20
test_import_loop_detection_self()  fails since cpython 3.13.8 RecursionError change


marc-hb:
(Since Python 3.13.8 commit https://github.com/python/cpython/commit/ebccd1de88d,) `test_import_loop_detection_self()` fails when trying to run it with pytest-xdist 3.8 - even with a single process (`-n 1`)

`complaint = "unsupported operand type(s) for +: 'int' and 'Reader'"`

Reproduced with today's commit ed868f778e25.

This is worth noting because this is currently the only test failing with pytest-xdist. EDIT: `xdist` was only the messenger, merely offsetting the stack. See #919 for reproduction without `xdist`

I checked out the commit 16e2250e45d88f26 where this test was introduced + various others and that test was always incompatible with xdist 3.8.0 (warning: don't forget export `WEST_SKIP_SLOW_TESTS=0`)

To reproduce:

```diff
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -45,6 +45,7 @@ test = [
     "coverage~=7.10",
     "pytest>=8.4,<10.0",
     "pytest-cov~=7.0",
+    "pytest-xdist", # not needed after all, see PR #919 instead
 ]
 types = [
     "mypy~=1.18",
```
EDIT: newer and much better stack trace in later comment below https://github.com/zephyrproject-rtos/west/issues/908#issuecomment-3888725488
<details>
```
uv run poe test  -k import_loop_detect -n 1 # test passes without "-n 1"

  <very long stack, most likely due to the import recursion being tested>

src/west/manifest.py:2300: in _import_path_from_self
    self._import_pathobj_from_self(pathobj_abs, pathobj)
src/west/manifest.py:2327: in _import_pathobj_from_self
    Manifest(topdir=self.topdir, internal_import_ctx=child_ctx)
src/west/manifest.py:1620: in __init__
    self._load_validated()
src/west/manifest.py:2089: in _load_validated
    self._load_self(manifest_data)
src/west/manifest.py:2212: in _load_self
    self._import_from_self(imp)
src/west/manifest.py:2281: in _import_from_self
    self._import_path_from_self(imp)
src/west/manifest.py:2300: in _import_path_from_self
    self._import_pathobj_from_self(pathobj_abs, pathobj)
src/west/manifest.py:2327: in _import_pathobj_from_self
    Manifest(topdir=self.topdir, internal_import_ctx=child_ctx)
[...]

PR #918, opened 2026-02-12
manifest: give _ManifestImportDepth exception a __str__()

marc-hb:
Stop inheriting the superclass string which was the very generic and misleading "Could not import foo.yml, is it present in your manifest repository?"

Found while debugging xdist issue #908.

### DIFF
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -739,7 +739,8 @@ class ManifestVersionError(Exception):
 
 class _ManifestImportDepth(ManifestImportFailed):
     # A hack to signal to main.py what happened.
-    pass
+    def __str__(self):
+        return f'Import of {self.imp} failed: import level too deep'
 
 
 #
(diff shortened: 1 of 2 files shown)


pdgendt (commented):
Can you add a test case for this error?

Also the `self.imp` could contain a larger yaml blob, if this were to be printed, it might not be very readable?

marc-hb:
> Can you add a test case for this error?

AFAIK there is only one "infinite loop" test case and I don't think it prints this exception but I'll give it a try. I experienced this confusion while using a debugger and adding transient log statements.

> Also the self.imp could contain a larger yaml blob, if this were to be printed, it might not be very readable?

Could it? Either way this merely overrides the superclass function which already prints `self.imp` so this seems a bit off-topic.

pdgendt (approved):

marc-hb:
> Can you add a test case for this error?

This was actually easy thanks to `pytest.raises`, thanks for pushing me.

> Also the self.imp could contain a larger yaml blob, if this were to be printed, it might not be very readable?

This is potential issue with the super class but not with this one right now: the constructor is always invoked with a path.

Issue #903, opened 2026-01-08
Clean way to import a west.yml's west commands but no submodule


johnhatten2:
Context :
I have a zephyr module containing drivers and unit tests. This module is normally used by some main application, but here, I need a pipeline (native_sim) to run my unit tests. The only submodule I need is Zephyr with none of Zephyr's submodules. But I also need Zephyr's west commands, which are imported at the same time as Zephyr's west.yml is parsed for submodules. 

Attempt 1
Importing Zephyr with no import, I tried pointing directly at the right location. I tried a lot of variations by moving and renaming zephyr or simply copy pasting its script folder elsewhere, but could never make it work
<img width="616" height="150" alt="Image" src="https://github.com/user-attachments/assets/66f508d2-91ac-4edc-8c2b-a54cd1d32be6" />

Attempt 2
Instead, I tried addind an empty import in the hope west will ignore submodules but import the commands. Didn't work either but I feel like this could be solved in a one liner.
<img width="348" height="131" alt="Image" src="https://github.com/user-attachments/assets/dfd2bba0-49ea-462a-94c4-20c29532caa6" />

I tried this hoping there were some kind of magic word that would allow me todo what I want and it worked! But turns out you can put any string that doesn't map to an existing submodule here and it will work. Looks sketch.
<img width="623" height="133" alt="Image" src="https://github.com/user-attachments/assets/659a2470-02e9-4a17-8d2a-fee722a528fd" />

I consider the import to work when I can run "west boards" and the board list prints properly

marc-hb:
Hi @johnhatten2 !

This looks interesting but I'm afraid you posted it in the wrong place.

- It's not obvious but [`west`](https://github.com/zephyrproject-rtos/west/) tries to be as independent of Zephyr as possible and your description seems to refer to Zephyr a lot? If you can reproduce the problem without Zephyr then by all means please share that step-by-step reproduction here in a new west issue here.
 
- Even if you had filed this in https://github.com/zephyrproject-rtos/zephyr/issues, GitHub issues are focused on solving just one specific issue: either one specific feature or one specific bug. All the trial and error and long way you got there can be interesting but that shouldn't come first in the description. If you click on the green "New" button at https://github.com/zephyrproject-rtos/zephyr/issues you will get templates demonstrating this.

[...]

PR #920, opened 2026-02-16
fix: Allow west commands to be imported from a project subdirectory if manifest is located in subdirectory

### LINKED ISSUE #725: Imported `west-commands` don't resolve when module manifest isn't in module root
When `manifest` isn't in the root of a module, then custom `west-commands` don't resolve when imported from another Zephyr workspace. 

Example: https://github.com/zmkfirmware/zmk

The manifest file is located at `/app/west.yml`. The manifest adds custom commands from `/app/scripts/west-commands.yml`:
```yaml
# ZMK manifest
manifest:
  # ...
  self:
```

When `zmk` is the topdir of the workspace, both the path to `west-commands.yml` and the paths to any script defined therein are interpreted relative to `/app`. 

However, when `zmk` is imported from another Zephyr workspaces as follows, the paths won't resolve correctly:

```yaml
# Main manifest of Zephyr workspace
manifest:
  projects:
    - name: zmk
      # ...
      import: app/west.yml
```

In this case, both the `west-commands` from the imported ZMK manifest as well as the paths to any script defined therein are interpreted as relative to `/zmk` (rather than `/zmk/app`), resulting in none of the commands being found.

---

Note that overwriting `west-commands` from the topdir manifest with `west-commands: app/scripts/west-commands.yml` resolves the first problem of `/zmk/app/scripts/west-commands.yml` being found, but still interprets the paths to any script therein relative to `/zmk`.

nmunnich:
Fixes #725

Note that I relied on the test to tell me if the fix is working, I did not test it with a proper repository setup. Let me know if this is desired/this project is one where manual testing of such kind is necessary. 

I'm unfamiliar with west release cycles - if this is accepted, could someone let me know when I could expect to see a new version of west released, so that we can start recommending setups relying on this to our users?

### DIFF
--- a/src/west/commands.py
+++ b/src/west/commands.py
@@ -701,13 +701,18 @@ def _ext_specs(project):
         except pykwalify.errors.SchemaError as e:
             raise ExtensionCommandError from e
 
+        # Resolve west command extensions relative to the manifest root for
+        # import-derived west-commands entries, otherwise project root.
+        mfst_dir = project._west_commands_manifest_dirs.get(cmd)
+        base_dir = os.path.join(project.abspath, mfst_dir) if mfst_dir else project.abspath
+
         for commands_desc in commands_spec['west-commands']:
-            ret.extend(_ext_specs_from_desc(project, commands_desc))
+            ret.extend(_ext_specs_from_desc(project, commands_desc, base_dir))
     return ret
 
 
-def _ext_specs_from_desc(project, commands_desc):
-    py_file = os.path.join(project.abspath, commands_desc['file'])
+def _ext_specs_from_desc(project, commands_desc, base_dir):
+    py_file = os.path.join(base_dir, commands_desc['file'])
 
     # Verify the YAML's python file doesn't escape the project directory.
     if escapes_directory(py_file, project.abspath):
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -76,6 +76,7 @@ SCHEMA_VERSION = '1.2'
 # The value of a west-commands as passed around during manifest
 # resolution. It can become a list due to resolving imports, even
 # though it's just a str in each individual file right now.
+# TODO: WestCommands should be promoted to a Python class, see issue #959
 WestCommandsType = str | list[str]
 
 # Type for the importer callback passed to the manifest constructor.
@@ -907,6 +908,9 @@ class Project:
         self.groups: GroupsType = groups or []
         self.userdata: Any = userdata
 
+        # Internal helpers
+        self._west_commands_manifest_dirs: dict[str, str] = dict()
+
     @property
     def path(self) -> str:
         return self._path
@@ -1266,6 +1270,7 @@ class ManifestProject(Project):
 
         # Extension commands.
         self.west_commands = _west_commands_list(west_commands)
+        self._west_commands_manifest_dirs: dict[str, str] = dict()
 
     @property
     def abspath(self) -> str | None:
@@ -2723,6 +2728,12 @@ class Manifest:
         west_commands_to_merge = [
             (mfst_dir / cmd).as_posix() for cmd in submanifest._ctx.manifest_west_commands
         ]
+
+        # Keep track of which imported manifest directory each adjusted
+        # west-commands entry came from, so command Python files can be
+        # resolved relative to that manifest root later in commands.py.
+        for adjusted_cmd in west_commands_to_merge:
+            project._west_commands_manifest_dirs.setdefault(adjusted_cmd, str(mfst_dir))
 
         project.west_commands = _west_commands_merge(project.west_commands, west_commands_to_merge)
 
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -2637,13 +2637,19 @@ class Manifest:
 
         _logger.debug(f'resolving import {path} for {project}')
         imported = self._import_content_from_project(project, path)
-        if imported is None:
-            # This can happen if self._ctx.project_importer returns None.
-            # It means there's nothing to do.
-            return
 
-        for data in imported:
-            self._import_data_from_project(project, data, None)
+        match imported:
+            case None:
+                # This can happen if self._ctx.project_importer returns None.
+                # It means there's nothing to do.
+                return
+            case str():
+                self._import_data_from_project(project, imported, path, directory_import=False)
+            case list():
+                for data in imported:
+                    self._import_data_from_project(project, data, path, directory_import=True)
+            case _:
+                raise AssertionError(f'imported content has unexpected type {type(imported)}')
 
         _logger.debug(f'done resolving import {path} for {project}')
 
@@ -2653,25 +2659,40 @@ class Manifest:
         _logger.debug(f'resolving import {imap} for {project}')
 
         imported = self._import_content_from_project(project, imap.file)
-        if imported is None:
-            return
-
-        for data in imported:
-            self._import_data_from_project(project, data, imap)
+        match imported:
+            case None:
+                return
+            case str():
+                self._import_data_from_project(project, imported, imap, directory_import=False)
+            case list():
+                for data in imported:
+                    self._import_data_from_project(project, data, imap, directory_import=True)
+            case _:
+                raise AssertionError(f'imported content has unexpected type {type(imported)}')
 
         _logger.debug(f'done resolving import {imap} for {project}')
 
     def _import_data_from_project(
-        self, project: Project, data: Any, imap: _import_map | None
+        self,
+        project: Project,
+        data: Any,
+        imap_or_mfpath: _import_map | str,
+        directory_import: bool,
     ) -> None:
         # Destructively add the imported data into our 'projects' map.
-
-        if imap is not None:
-            imap_filter = _compose_imap_filters(self._ctx.imap_filter, _imap_filter(imap))
-            imap_path_prefix = imap.path_prefix
-        else:
-            imap_filter = self._ctx.imap_filter
-            imap_path_prefix = '.'
+        match imap_or_mfpath:
+            case _import_map():
+                imap_filter = _compose_imap_filters(
+                    self._ctx.imap_filter, _imap_filter(imap_or_mfpath)
+                )
+                imap_path_prefix = imap_or_mfpath.path_prefix
+                mfst_path = imap_or_mfpath.file
+            case str():
+                imap_filter = self._ctx.imap_filter
+                imap_path_prefix = '.'
+                mfst_path = imap_or_mfpath
+            case _:
+                raise AssertionError(f'imap_or_mfpath has unexpected type {type(imap_or_mfpath)}')
 
         child_ctx = self._ctx._replace(
             imap_filter=imap_filter,
@@ -2689,38 +2710,41 @@ class Manifest:
         try:
             submanifest = Manifest(topdir=self.topdir, internal_import_ctx=child_ctx)
         except RecursionError as e:
-            raise _ManifestImportDepth(None, imap.file if imap else None) from e
+            raise _ManifestImportDepth(None, mfst_path) from e
 
         # Patch up any extension commands in the imported data
         # by allocating them to the project.
-        project.west_commands = _west_commands_merge(
-            project.west_commands, submanifest._ctx.manifest_west_commands
-        )
+
+        # If the manifest was imported from a project subdirectory
+        # (manifest_path is a relative path within the project),
+        # we need to adjust the west_commands paths to be relative
+        # to the project root, not to the manifest subdirectory.
+        mfst_dir = Path(mfst_path) if directory_import else Path(mfst_path).parent
+        west_commands_to_merge = [
+            (mfst_dir / cmd).as_posix() for cmd in submanifest._ctx.manifest_west_commands
+        ]
+
+        project.west_commands = _west_commands_merge(project.west_commands, west_commands_to_merge)
 
     def _import_content_from_project(self, project: Project, path: str) -> ImportedContentType:
         if not (self._ctx.import_flags & ImportFlag.FORCE_PROJECTS) and project.is_cloned():
             try:
-                content = _manifest_content_at(project, path, Manifest.encoding)
+                return _manifest_content_at(project, path, Manifest.encoding)
             except MalformedManifest as mm:
                 self._malformed(mm.args[0])
             except FileNotFoundError:
                 # We may need to fetch a new manifest-rev, e.g. if
                 # revision is a branch that didn't used to have a
                 # manifest, but now does.
-                content = self._ctx.project_importer(project, path)
+                return self._ctx.project_importer(project, path)
             except subprocess.CalledProcessError:
                 # We may need a new manifest-rev, e.g. if revision is
                 # a SHA we don't have yet.
-                content = self._ctx.project_importer(project, path)
+                return self._ctx.project_importer(project, path)
         else:
             # We need to clone this project, or we were specifically
             # asked to use the importer.
-            content = self._ctx.project_importer(project, path)
-
-        if isinstance(content, str):
-            content = [content]
-
-        return content
+            return self._ctx.project_importer(project, path)
 
     def _load_imap(self, imp: dict, src: str) -> _import_map:
         # Convert a parsed self or project import value from YAML into
--- a/tests/test_extension_commands.py
+++ b/tests/test_extension_commands.py
@@ -6,6 +6,7 @@ import subprocess
 import textwrap
 from pathlib import Path
 
+import pytest
 import yaml
 from conftest import GIT, WINDOWS, add_commit, cmd, cmd_raises, yaml_editor
 
@@ -341,6 +342,85 @@ def test_call_imported_project_submanifest_commands_from_project_subdirectory(re
 
     ext_output = cmd('imported-command-from-subdir', cwd=workspace)
     assert 'imported command from subdir works' in ext_output
+
+
+@pytest.mark.skipif(
+    not WINDOWS, reason="non-posix path separators don't work reliably on non-Windows"
+)
+def test_call_imported_project_submanifest_commands_from_project_subdirectory_windows_paths(
+    repos_tmpdir,
+):
+    # Same as the forward-slash test above, but use Windows-style
+    # separators in the manifest file paths.
+    manifest_path = repos_tmpdir / 'repos' / 'zephyr'
+    net_tools_path = repos_tmpdir / 'repos' / 'net-tools'
+
+    MF_SUB_WEST_YML = textwrap.dedent(
+        '''\
+        manifest:
+          self:
+            west-commands: scripts\\west-commands-from-subdir-windows.yml
+        '''
+    )
+    MF_SUB_COMMANDS_YML = textwrap.dedent(
+        '''\
+        west-commands:
+          - file: test\\west-commands\\subdir_command_windows.py
+            commands:
+              - name: imported-command-from-subdir-windows
+                class: ImportedCommandFromProjectSubdirWindows
+                help: imported command from subdir with windows paths help
+        '''
+    )
+    MF_SUB_WEST_PY = textwrap.dedent(
+        '''\
+        from west.commands import WestCommand
+
+        class ImportedCommandFromProjectSubdirWindows(WestCommand):
+            def __init__(self):
+                super().__init__(
+                    'imported-command-from-subdir-windows',
+                    'imported command from subdir with windows paths help',
+                    'imported command from subdir with windows paths description',
+                )
+
+            def do_add_parser(self, parser_adder):
+                return parser_adder.add_parser(self.name)
+
+            def do_run(self, args, unknown):
+                print('imported command from subdir with windows paths works')
+        '''
+    )
+
+    add_commit(
+        net_tools_path,
+        'add imported submanifest extension command with windows paths',
+        files={
+            r'mf_subdir/west.yml': MF_SUB_WEST_YML,
+            r'mf_subdir/scripts/west-commands-from-subdir-windows.yml': MF_SUB_COMMANDS_YML,
+            r'mf_subdir/test/west-commands/subdir_command_windows.py': MF_SUB_WEST_PY,
+        },
+    )
+
+    with yaml_editor(manifest_path / 'west.yml') as mf:
+        net_tools_project = [p for p in mf['manifest']['projects'] if p['name'] == 'net-tools'][0]
+        net_tools_project['import'] = r'mf_subdir/west.yml'
+    subprocess.check_call([
+        GIT,
+        '-C',
+        str(manifest_path),
+        'commit',
+        '-m',
+        'import mf_subdir\\west.yml',
+        'west.yml',
+    ])
+
+    workspace = repos_tmpdir / 'workspace'
+    cmd(['init', '-m', str(manifest_path), str(workspace)])
+    cmd('update', cwd=workspace)
+
+    ext_output = cmd('imported-command-from-subdir-windows', cwd=workspace)
+    assert 'imported command from subdir with windows paths works' in ext_output
 
 
 def test_extension_special_chars(west_update_tmpdir):
--- a/tests/test_extension_commands.py
+++ b/tests/test_extension_commands.py
@@ -266,6 +266,83 @@ def test_extension_command_multiple_commands_same_file(west_update_tmpdir):
     assert 'second command' in ext_output
 
 
+def test_call_imported_project_submanifest_commands_from_project_subdirectory(repos_tmpdir):
+    # If net-tools imports mf_subdir/west.yml and that manifest declares
+    # self: west-commands: scripts/west-commands.yml, then file paths in that
+    # west-commands YAML are resolved relative to the imported manifest root.
+    # The same paths must work whether a project is imported or initialized directly;
+    # importing must never break anything.
+    manifest_path = repos_tmpdir / 'repos' / 'zephyr'
+    net_tools_path = repos_tmpdir / 'repos' / 'net-tools'
+
+    MF_SUB_WEST_YML = textwrap.dedent(
+        '''\
+        manifest:
+          self:
+            west-commands: scripts/west-commands-from-subdir.yml
+        '''
+    )
+    MF_SUB_COMMANDS_YML = textwrap.dedent(
+        '''\
+        west-commands:
+          - file: test/west-commands/subdir_command.py
+            commands:
+              - name: imported-command-from-subdir
+                class: ImportedCommandFromProjectSubdir
+                help: imported extension help
+        '''
+    )
+    MF_SUB_WEST_PY = textwrap.dedent(
+        '''\
+        from west.commands import WestCommand
+
+        class ImportedCommandFromProjectSubdir(WestCommand):
+            def __init__(self):
+                super().__init__(
+                    'imported-command-from-subdir',
+                    'imported command from subdir help',
+                    'imported command from subdir description',
+                )
+
+            def do_add_parser(self, parser_adder):
+                return parser_adder.add_parser(self.name)
+
+            def do_run(self, args, unknown):
+                print('imported command from subdir works')
+        '''
+    )
+
+    add_commit(
+        net_tools_path,
+        'add imported submanifest extension command',
+        files={
+            'mf_subdir/west.yml': MF_SUB_WEST_YML,
+            'mf_subdir/scripts/west-commands-from-subdir.yml': MF_SUB_COMMANDS_YML,
+            'mf_subdir/test/west-commands/subdir_command.py': MF_SUB_WEST_PY,
+        },
+    )
+
+    with yaml_editor(manifest_path / 'west.yml') as mf:
+        net_tools_project = [p for p in mf['manifest']['projects'] if p['name'] == 'net-tools'][0]
+        net_tools_project['import'] = 'mf_subdir/west.yml'
+    subprocess.check_call([
+        GIT,
+        '-C',
+        str(manifest_path),
+        'commit',
+        '-m',
+        'import mf_subdir/west.yml',
+        'west.yml',
+    ])
+
+    workspace = repos_tmpdir / 'workspace'
+    cmd(['init', '-m', str(manifest_path), str(workspace)])
+    cmd('update', cwd=workspace)
+
+    ext_output = cmd('imported-command-from-subdir', cwd=workspace)
+    assert 'imported command from subdir works' in ext_output
+
+
 def test_extension_special_chars(west_update_tmpdir):
     # Detect any unexpected changes in the way we've been handling backslashes and other
     # special characters. Changes in how we handle such edge cases may or may not be desired
--- a/tests/test_manifest.py
+++ b/tests/test_manifest.py
@@ -2116,6 +2116,101 @@ def test_import_project_submanifest_commands_both(manifest_repo):
     assert p1.west_commands == expected
 
 
+def test_import_project_submanifest_commands_from_project_subdirectory(manifest_repo):
+    # When a manifest is imported from a project subdirectory (e.g., mf_subdir/west.yml),
+    # and that manifest defines west-commands, the paths should be
+    # resolved relative to the manifest subdirectory.
+    # This test covers both string and import-map forms, for file and directory imports.
+
+    p1 = manifest_repo / '..' / 'p1'
+    create_repo(p1)
+    # West reads the manifest from the 'manifest-rev' branch of the project, so we need to
+    # create it and add the manifest there.
+    create_branch(p1, 'manifest-rev', checkout=True)
+    add_commit(
+        p1,
+        'add mf_subdir/west.yml with west-commands',
+        files={
+            'mf_subdir/west.yml': '''\
+                                manifest:
+                                  projects:
+                                  - name: p2
+                                    url: url-placeholder2
+                                  self:
+                                    west-commands: p2subdir/west-commands.yml
+                                ''',
+        },
+    )
+    # Checkout master (which is empty) to ensure the west.yml file is not present in the
+    # file system, and that the west-commands are resolved from the 'manifest-rev' branch
+    # using git, not the currently checked out version.
+    checkout_branch(p1, 'master')
+
+    # Case A: import as a string path to the submanifest file.
+    with open(manifest_repo / 'west.yml', 'w') as f:
+        f.write('''\
+        manifest:
+          projects:
+          - name: p1
+            url: url-placeholder
+            import: mf_subdir/west.yml
+        ''')
+
+    # The west_commands path should be 'mf_subdir/p2subdir/west-commands.yml',
+    # not 'west-commands.yml', to be resolved correctly
+    # relative to the project root. See issue #725.
+    p1_proj = MF().get_projects(['p1'])[0]
+    expected = ['mf_subdir/p2subdir/west-commands.yml']
+    assert p1_proj.west_commands == expected
+
+    # Case B: import using an import-map whose 'file' is a file path.
+    with open(manifest_repo / 'west.yml', 'w') as f:
+        f.write('''\
+      manifest:
+        projects:
+        - name: p1
+          url: url-placeholder
+          import:
+            file: mf_subdir/west.yml
+      ''')
+
+    # Reload and check the west_commands were resolved the same way.
+    p1_proj = MF().get_projects(['p1'])[0]
+    expected = ['mf_subdir/p2subdir/west-commands.yml']
+    assert p1_proj.west_commands == expected
+
+    # Case C: import as a string path to the submanifest directory.
+    with open(manifest_repo / 'west.yml', 'w') as f:
+        f.write('''\
+      manifest:
+        projects:
+        - name: p1
+          url: url-placeholder
+          import: mf_subdir
+      ''')
+
+    # Reload and check the west_commands were resolved the same way.
+    p1_proj = MF().get_projects(['p1'])[0]
+    expected = ['mf_subdir/p2subdir/west-commands.yml']
+    assert p1_proj.west_commands == expected
+
+    # Case D: import using an import-map whose 'file' is a directory.
+    with open(manifest_repo / 'west.yml', 'w') as f:
+        f.write('''\
+      manifest:
+        projects:
+        - name: p1
+          url: url-placeholder
+          import:
+            file: mf_subdir
+      ''')
+
+    # Reload and check the west_commands were resolved the same way.
+    p1_proj = MF().get_projects(['p1'])[0]
+    expected = ['mf_subdir/p2subdir/west-commands.yml']
+    assert p1_proj.west_commands == expected
+
+
 def test_import_map_error_handling():
     # Make sure we handle expected errors when loading import:
     # values that are maps.


tests/test_manifest.py
```diff
@@ -2095,6 +2095,48 @@ def test_import_project_submanifest_commands_both(manifest_repo):
...
+        },
+    )
+    checkout_branch(p1, 'master')
+
+    p1_proj = MF().get_projects(['p1'])[0]
+    # The west_commands path should be 'app/west-commands.yml',
+    # not 'west-commands.yml', to be resolved correctly
+    # relative to the project root.
+    expected = ['app/west-commands.yml']
+    assert p1_proj.west_commands == expected
```
marc-hb, line 2211: Nit: the `expected` variable does not seem necessary.

tests/test_manifest.py
```diff
@@ -2095,6 +2095,48 @@ def test_import_project_submanifest_commands_both(manifest_repo):
...
+    # and that manifest defines west-commands, the paths should be
+    # resolved relative to the manifest subdirectory.
+    # This tests _import_path_from_project with a string path.
+
+    with open(manifest_repo / 'west.yml', 'w') as f:
+        f.write('''\
+        manifest:
+          projects:
+          - name: p1
+            url: p1-url
```
marc-hb:
```suggestion
            url: url-placeholder
```
makes it clearer it's unused here.

tests/test_manifest.py
```diff
@@ -2095,6 +2095,48 @@ def test_import_project_submanifest_commands_both(manifest_repo):
...
+            import: app/west.yml
```
marc-hb: Please correct me but this line looks the key thing being tested so I think it deserves a bit more "fluff". If nothing else to make it stand out.
```suggestion
            # Non-standard location in a subdirectory (not just "true")
            import: mf_subdir/west.yml
```

tests/test_manifest.py
```diff
@@ -2095,6 +2095,48 @@ def test_import_project_submanifest_commands_both(manifest_repo):
...
+    create_branch(p1, 'manifest-rev', checkout=True)
+    add_commit(
+        p1,
+        'add app/west.yml with west-commands',
+        files={
+            'app/west.yml': '''\
+                                manifest:
+                                  projects:
+                                  - name: p2
+                                    url: p2-url
```
marc-hb:
```suggestion
                                    url: url-placeholder2
```

tests/test_manifest.py
```diff
@@ -2095,6 +2095,48 @@ def test_import_project_submanifest_commands_both(manifest_repo):
...
+                                  self:
+                                    west-commands: west-commands.yml
```
marc-hb: How about another subdirectory here too just to stress the code a bit more and catch even more bugs? Who knows what the future has in store.
```suggestion
                                    west-commands: p2subdir/west-commands.yml
```

tests/test_manifest.py
```diff
@@ -2095,6 +2095,48 @@ def test_import_project_submanifest_commands_both(manifest_repo):
...
+                                ''',
+        },
+    )
+    checkout_branch(p1, 'master')
+
+    p1_proj = MF().get_projects(['p1'])[0]
+    # The west_commands path should be 'app/west-commands.yml',
+    # not 'west-commands.yml', to be resolved correctly
+    # relative to the project root.
```
marc-hb:
```suggestion
    # relative to the project root. See issue #12345
```

src/west/manifest.py
marc-hb, line 1: Can you please:
- fit the commit subject in ~80 characters as usual
- add a short commit message? At least: "fixes bug #12345 where X when Y"

src/west/manifest.py
```diff
@@ -2678,8 +2678,22 @@ def _import_data_from_project(
...
+        # If the manifest was imported from a project subdirectory
+        # (manifest_path is a relative path within the project),
+        # we need to adjust the west_commands paths to be relative
+        # to the project root, not to the manifest subdirectory.
+        west_commands_to_merge = submanifest._ctx.manifest_west_commands
+        if manifest_path and manifest_path != _WEST_YML:
+            manifest_dir = os.path.dirname(manifest_path)
+            if manifest_dir:
+                west_commands_to_merge = [
+                    os.path.join(manifest_dir, cmd).replace('\\', '/')
```
marc-hb: Please avoid `os.path` and switch to the newer `Pathlib` instead. You should never have to handle slashes and backslashes manually.

src/west/manifest.py
```diff
@@ -2678,8 +2678,22 @@ def _import_data_from_project(
 
         # Patch up any extension commands in the imported data
         # by allocating them to the project.
+        
+        # If the manifest was imported from a project subdirectory
+        # (manifest_path is a relative path within the project),
+        # we need to adjust the west_commands paths to be relative
+        # to the project root, not to the manifest subdirectory.
+        west_commands_to_merge = submanifest._ctx.manifest_west_commands
+        if manifest_path and manifest_path != _WEST_YML:
```
marc-hb: Is this `if` really needed? Prefixing with `.` should be a no-op and will make the code simpler.

(the comment can stay the same)

nmunnich:
Should pass the linter this time (sorry!)

src/west/manifest.py
```diff
@@ -2642,12 +2642,12 @@ def _import_map_from_project(self, project: Project, imp: dict) -> None:
...
 
         for data in imported:
-            self._import_data_from_project(project, data, imap)
+            self._import_data_from_project(project, data, imap, imap.file)
 
         _logger.debug(f'done resolving import {imap} for {project}')
 
     def _import_data_from_project(
-        self, project: Project, data: Any, imap: _import_map | None
+        self, project: Project, data: Any, imap: _import_map | None, manifest_path: str | None = None
```
marc-hb: Adding a new argument signature seems overkill considering the internal function is used in only two places in the same file and that at least one seems to be always `None` now (mutually exclusive). A single parameter `imap_or_path` for both should be enough?

```python
    match imap_or_path:
        case _import_map():
             path = imap_or_path.file
             imap_filter = _compose_imap_filters(self._ctx.imap_filter, _imap_filter(imap_or_path))
        case str() | Path():
             path = imap_or_path
             ...
        case None:
             # still needed?
        case _:
             raise ...

```

src/west/manifest.py
```diff
@@ -2642,12 +2642,13 @@ def _import_map_from_project(self, project: Project, imp: dict) -> None:
...
         for data in imported:
-            self._import_data_from_project(project, data, imap)
+            self._import_data_from_project(project, data, imap, imap.file)
 
         _logger.debug(f'done resolving import {imap} for {project}')
 
     def _import_data_from_project(
-        self, project: Project, data: Any, imap: _import_map | None
+        self, project: Project, data: Any,
+        imap: _import_map | None, manifest_path: str | None = None
```
marc-hb: Adding a new argument signature seems overkill considering the internal function is used in only two places in the same file and that at least one seems to be always None now (mutually exclusive). A single parameter imap_or_path for both should be enough?
```python
    match imap_or_path:
        case _import_map():
             path = imap_or_path.file
             imap_filter = _compose_imap_filters(self._ctx.imap_filter, _imap_filter(imap_or_path))
        case str() | Path():
             path = imap_or_path
             ...
        case None:
             # still needed?
        case _:
             raise ...
```

src/west/manifest.py
```diff
@@ -2678,8 +2679,22 @@ def _import_data_from_project(
...
+        # If the manifest was imported from a project subdirectory
+        # (manifest_path is a relative path within the project),
+        # we need to adjust the west_commands paths to be relative
+        # to the project root, not to the manifest subdirectory.
+        west_commands_to_merge = submanifest._ctx.manifest_west_commands
+        if manifest_path and manifest_path != _WEST_YML:
+            manifest_dir = os.path.dirname(manifest_path)
+            if manifest_dir:
+                west_commands_to_merge = [
+                    os.path.join(manifest_dir, cmd).replace('\\', '/')
```
marc-hb: Avoid os.path and switch to the newer Pathlib instead. You should never have to handle slashes and backslashes manually.

marc-hb (commented):
Thank you so much for taking this! See comments below.

Sorry if some comments appear twice, I had to make them again because you force pushed in the mean time. Please keep force-pushing, it's just Github support for force-pushes is pretty bad.

src/west/manifest.py
nmunnich: Hope this works.

src/west/manifest.py
nmunnich: Was mostly aiming to avoid some unnecessary os.path calls before I realised that manifests are sometimes not called west.yml.

tests/test_manifest.py
nmunnich: Renamed accordingly, but I'm personally not a fan of the comment because it ends up being a yaml comment at that point and that feels hacky/wrong to me. Could add a comment below/above if you feel it's necessary?

tests/test_manifest.py
nmunnich: I agree, but there's an existing pattern that I was following. See `test_import_project_submanifest_commands` and `test_import_project_submanifest_commands_both`, which could also have said variable removed. I'd prefer not to deviate here.

nmunnich (commented):
Thanks for the quick review! (I hope the linter is happy with me this time, `uv run poe all` doesn't seem to check whatever that last test is)

src/west/manifest.py
```diff
@@ -2647,16 +2647,21 @@ def _import_map_from_project(self, project: Project, imp: dict) -> None:
...
+                imap_filter = _compose_imap_filters(self._ctx.imap_filter,
+                                                    _imap_filter(imap_or_path))
+                imap_path_prefix = imap_or_path.path_prefix
+                path = imap_or_path.file
+            case str():
+                imap_filter = self._ctx.imap_filter
+                imap_path_prefix = '.'
+                path = imap_or_path
+            case _:
+                raise AssertionError(f'imap_or_path has unexpected type {type(imap_or_path)}')
```
marc-hb:
```suggestion
                raise AssertionError(f'imap_or_path {imap_or_path} has unexpected type {type(imap_or_path)}')
```
You can never be too verbose when something that unexpected and catastrophic happens.

src/west/manifest.py
```diff
@@ -2647,16 +2647,21 @@ def _import_map_from_project(self, project: Project, imp: dict) -> None:
...
-            imap_path_prefix = imap.path_prefix
-        else:
-            imap_filter = self._ctx.imap_filter
-            imap_path_prefix = '.'
+        match imap_or_path:
+            case _import_map():
+                imap_filter = _compose_imap_filters(self._ctx.imap_filter,
+                                                    _imap_filter(imap_or_path))
+                imap_path_prefix = imap_or_path.path_prefix
+                path = imap_or_path.file
```
marc-hb: `path` is just too vague sorry; it can hardly be more vague.
```suggestion
                mnft_path = imap_or_path.file
```

src/west/manifest.py
```diff
@@ -2674,12 +2679,24 @@ def _import_data_from_project(
...
 
         # Patch up any extension commands in the imported data
         # by allocating them to the project.
+
+        # If the manifest was imported from a project subdirectory
+        # (manifest_path is a relative path within the project),
+        # we need to adjust the west_commands paths to be relative
+        # to the project root, not to the manifest subdirectory.
+        west_commands_to_merge = submanifest._ctx.manifest_west_commands
+        manifest_dir = PurePosixPath(path).parent
```
marc-hb: Why `Pure` and why `Posix`?

src/west/manifest.py
```diff
@@ -2674,12 +2679,24 @@ def _import_data_from_project(
...
+        if manifest_dir != PurePosixPath('.'):
```
marc-hb: Please don't optimize this, at least not before someone actually profiles it and complains it makes a significant performance difference.

src/west/manifest.py
```diff
@@ -2674,12 +2679,24 @@ def _import_data_from_project(
...
+            west_commands_to_merge = [
+                (manifest_dir / cmd).as_posix() for cmd in west_commands_to_merge
```
marc-hb:
```suggestion
                str(manifest_dir / cmd) for cmd in submanifest._ctx.manifest_west_commands
```

Or even better: keep them as `Path` if it works, cause why not? More types is better.

tests/test_manifest.py
marc-hb: You are probably overestimating the consistency in our test code.

How about this: which variant gives the more user-friendly error message when the test fails? Let's go with that one.

src/west/manifest.py
```diff
@@ -2674,12 +2679,24 @@ def _import_data_from_project(
...
+            raise _ManifestImportDepth(None, path) from e
 
         # Patch up any extension commands in the imported data
         # by allocating them to the project.
+
+        # If the manifest was imported from a project subdirectory
+        # (manifest_path is a relative path within the project),
+        # we need to adjust the west_commands paths to be relative
+        # to the project root, not to the manifest subdirectory.
+        west_commands_to_merge = submanifest._ctx.manifest_west_commands
```
marc-hb: I think you can drop this line. There is a balance: you don't want to "inline" too many variables and compose 6, 7 or 8 function calls all at once because good (variable) names act as "documentation" and help understand intermediate states and the code. But when you have something like `west_commands_xxxx = sthing.sthing_west_commands`, that intermediate variable does not help the reader tell what it is.

tests/test_manifest.py
marc-hb: This is just test data so having comments there really does not shock me. But that comment was not important, that information can easily be found nearby.

src/west/manifest.py
marc-hb: Wait: converting this to a string could trigger another exception. Ignore.

src/west/manifest.py
```diff
@@ -2647,16 +2647,21 @@ def _import_map_from_project(self, project: Project, imp: dict) -> None:
         _logger.debug(f'done resolving import {imap} for {project}')
 
     def _import_data_from_project(
-        self, project: Project, data: Any, imap: _import_map | None
+        self, project: Project, data: Any, imap_or_path: _import_map | str
```
marc-hb:
```suggestion
        self, project: Project, data: Any, imap_or_mfpath: _import_map | str
```

marc-hb:
> (I hope the linter is happy with me this time, ...

Still no luck.

Whenever a GitHub workflow fails, always save the log in a comment:

https://github.com/zephyrproject-rtos/west/actions/runs/22082397164/job/63814166855?pr=920

That's anOTHER way GitHub sucks with force-pushes: it becomes very hard/impossible to find older runs.

> ... uv run poe all doesn't seem to check whatever that last test is)

Strange, works for me:

```
uv run poe all
...
Poe => ruff format
1 file reformatted, 21 files left unchanged
```
Also, faster:

```shell
uv run poe # lists all
uv run poe format
```

tests/test_manifest.py
nmunnich: I think I lean *very* very slightly towards `expected`.

src/west/manifest.py
nmunnich: Overall much simpler now with no if's whatsoever.

Re: Posix, PurePosixPath, etc:
For better or worse (almost certainly worse most of the time) I'm doing this development on Windows. Testing *outside* of WSL, the overall path needs to be
* converted to a string
* converted to a posixpath

in order for tests to succeed. I used posixpath everywhere because I liked using the same type of path across multiple lines - now that it's condensed I lean towards `(Path()).as_posix()` (Alternative would be `str(PurePosixPath())`.

nmunnich (commented):
Ah, the formatter *does* run, but for whatever reason it thinks it needs to format 22 files (with no difference made to them) on my machine. I've kept that formatting on ``manifest.py``, maybe now it'll be happy? Just another quirk of Windows I suppose.

tests/test_manifest.py
marc-hb: So I just made such a test fail (for another reason) and there's basically no significant output difference when it fails hence no user-friendliness difference.

There is not a lot of consistency either: some tests use an `expected` name and others not. It makes sense to use a variable when the expected result is long and would make the assert line too long and that is the pattern I observed. But the data is short enough in this test.

There is no value in the name either because the "expectation" is pretty obvious when the last line of a test is an `assert`.

All this being said: this is really not important so I'll leave it to you.

nmunnich:
Hmmm. I'm not sold on adding tests to test that the exceptions are indeed raised. Don't think the AssertionError can ever be raised anyway, it's just good practice to have it there.

src/west/manifest.py
marc-hb: I think I understand what's going on, thanks. Give me a moment to think about it.

marc-hb (commented):
> Hmmm. I'm not sold on adding tests to test that the exceptions are indeed raised. Don't think the AssertionError can ever be raised anyway, it's just good practice to have it there.

I agree. Error-handling is notoriously lacking coverage across the entire industry (one of the reasons software is so insecure), but I don't think we need test coverage for `raise InternalBug("this should never happen")`.

@pdgendt is big on test coverage and rightly so, let's wait until next week when he is back. Every PR needs 2 approvals anyway.

src/west/manifest.py
marc-hb: Please try to:
- Replace `.as_posix()` with `str(manifest_dir / cmd)` as already suggested above
- Replace  `assert p1_proj.west_commands == expected` with `for ... zip` loop similar to the one I just submitted in #921 

If the test(s) still fail (how?) then try cherry-pick my #921 on top.

BTW thanks for testing on Windows! Much appreciated :-)

nmunnich: Replaced and tests pass now.

marc-hb (dismissed):

src/west/manifest.py
```diff
@@ -2674,13 +2680,21 @@ def _import_data_from_project(
...
         # by allocating them to the project.
-        project.west_commands = _west_commands_merge(
-            project.west_commands, submanifest._ctx.manifest_west_commands
-        )
+
+        # If the manifest was imported from a project subdirectory
+        # (manifest_path is a relative path within the project),
+        # we need to adjust the west_commands paths to be relative
+        # to the project root, not to the manifest subdirectory.
+        mfst_dir = Path(mfst_path).parent
```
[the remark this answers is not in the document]
pdgendt: @marc-hb maybe related to #921 ?

tests/test_manifest.py
```diff
@@ -2095,6 +2095,49 @@ def test_import_project_submanifest_commands_both(manifest_repo):
...
+    # not 'west-commands.yml', to be resolved correctly
+    # relative to the project root. See issue #725.
+    expected = ['mf_subdir/p2subdir/west-commands.yml']
+    for a, e in zip(p1_proj.west_commands, expected, strict=True):
+        assert PurePath(a) == PurePath(e)
+
+
 def test_import_map_error_handling():
     # Make sure we handle expected errors when loading import:
     # values that are maps.
```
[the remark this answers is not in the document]
marc-hb: That's an interesting suggestion. But that's also a MASSIVE duplication of the other test code for very little additional coverage :-( @nmunnich maybe you could easily combine both tests (yours and the copilot ~duplicate) into a single test like this:

```yaml
        manifest:
          projects:
           
          - name: pA
            url: url-placeholderA
            import: mf_subdirA/west.yml

          - name: pB
            url: url-placeholderB
            import:
               file: mf_subdirB/west.yml
```
Then you can re-use the identical `add_commit()` twice in both A and B.

src/west/manifest.py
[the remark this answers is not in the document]
marc-hb: Yes, related to #921 (which I just demoted to draft). 

On one hand we should stick to Pathlib abstractions, avoid hardcoding forward slashes and not use Unix paths objects internally when running on Windows. This is not just about backslashes, there are other differences.

On the other hand we don't want `west manifest --resolve` to convert forward slashes to backslashes on Windows. EDIT: or do we? See #928

I need again more time to think about it and do more testing, sorry :-(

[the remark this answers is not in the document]
marc-hb: This is a much bigger can of worms than I expected. I wish I could "unsee" it but I can't :-(

For instance, on Windows:
```
west list -f '{path}'      # forward slashes
west list -f '{abspath}'   # backslashes 
west list -f '{posixpath}' # forward slash
west list topdir           # absolute and forward since #375
```
https://docs.zephyrproject.org/latest/develop/west/west-apis.html#west.manifest.Project

Also relevant:
- #411 and more
- #375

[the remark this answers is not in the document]
nmunnich: Personally, I would argue that resolving that inconsistency is out of scope here, and I would prefer not to block waiting for a fix/resolution. Would you be alright with taking the PurePosixPath suggestion and going forward with that here?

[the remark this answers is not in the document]
marc-hb: I'm not sure it's a pre-existing inconsistency. This could be all by design - which this PR may or may not be consistent with.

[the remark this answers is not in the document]
marc-hb: So, there is some pre-existing inconsistency. But some of it is right in the middle of this PR. Example:

Input:
```
    - name: hal_espressif
      path: modules///hal\\\espressif
      west-commands: west\\\west-commands.yml
```

`west manifest --resolve` output:
```
    - name: hal_espressif
      path: modules/hal/espressif              # normalized to POSIX
      west-commands: west\\\west-commands.yml  # unchanged
```
We need to _decide_ at the very least whether this pre-existing behavior is correct or not before merging this PR. I mean we can't modify code without even _knowing_ whether the modified code is buggy or not.

Note this is only the output of `west manifest --resolve`. How does the actual import behave? On Windows and Linux?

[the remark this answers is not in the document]
nmunnich: I've gone through the pain of doing a bunch of testing of actual imports in PowerShell. This may be simpler than we think:
 - No matter what `path` is set to, it always resolves in `west manifest --resolve` with `/`. So `path: modules/\//\/\/\\\\/\//\./\zmk\/\/\/\.\\\/\test` turned into `modules/zmk/test`.
 - No matter what `west-commands` is set to, it remained the same in the west manifest. In addition, regardless of how many extra `\` and `/` combinations I put, my dummy west command was being picked up and executed correctly. 

I've now modified the code somewhat, so that the west-commands will always resolve with posix paths, for consistency with `path`. I noticed a comment on L2468 that seems relevant to the conversation, and took the approach used there.

I did notice another bug while doing my testing:
``` title="manifest A"
manifest:
  projects:
    - name: p1
      url: placeholder
      revision: west-test
      import: app/west.yml
      west-commands: app/scripts/west-commands.yml
```

``` title="west-commands.yml"
west-commands:
  - file: test/west-commands/west.py
    commands:
      - name: west
        class: West
        help: Print hello world to the console
```

When resolving this, while the `west-commands.yml` file is located correctly and the commands are visible in `west help`, trying to actually execute the command doesn't work because west is trying to resolve it as `<module-path>/test/west-commands/west.py` rather than `<module-path>/app/test/west-commands/west.py`. My current thinking is that we need to add `manifestdir` as a property to `Project`, so that we can refer to it on L691 of `commands.py`. That feels like a rather extensive change though, would you have any alternative idea?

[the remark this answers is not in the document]
marc-hb: I've been working on adding some 2 "backslash" tests related to this (#928). I have this one ready but I want to finish a similar one for `west-commands`:
https://github.com/marc-hb/west/commits/backslash-tests/

I should be able to get the last one ready on Friday. I would like these "pure" tests to be merged before any change (feature or bug fix) in `west` itself.

Just a heads-up, sorry for not answering your comment yet - will do that on Friday too.

src/west/manifest.py
```diff
@@ -2497,7 +2496,7 @@ def _load_project(self, pd: dict, url_bases: dict[str, str], defaults: _defaults
             path=path,
             submodules=self._load_submodules(pd.get('submodules'), f'project {name}'),
             clone_depth=pd.get('clone-depth'),
-            west_commands=pd.get('west-commands'),
+            west_commands=Path(pd.get('west-commands')).as_posix() if pd.get('west-commands') else None,
```
marc-hb: This normalizes `west-commands:` values for projects but not for `self` which seems to come from somewhere else :-(

marc-hb: If we go for `west-commands` normalization, then maybe just change the type to a `Path` across the board which can help catch inconsistencies.

src/west/manifest.py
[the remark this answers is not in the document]
marc-hb: > I would like these "pure" tests to be merged before any change (feature or bug fix) in west itself.

- #928

marc-hb:
As discussed previously, this PR breaks my new tests submitted in #928. It's probably OK and these tests in #928 can likely be adjusted but it needs to be discussed because it's still a technically backwards-incompatible change. The easiest way to discuss and to record that behavior change is to merge #928 to fill the test coverage gap :-)

marc-hb:
One more thought: this needs to be split into at least 2 commits, maybe 2 PRs:
1. the west-commands Path normalization
2. the prefix bug fix

nmunnich:
Just an fyi, I think it is unlikely that I'll find the time to continue on this until next month. If it's still open in April I'll have another go, but I'm encouraging anyone who has the time to continue this sooner.

marc-hb:
> One more thought: this needs to be split into at least 2 commits, maybe 2 PRs:
>
> 1. the west-commands Path normalization
> 2. the prefix bug fix

I can do 1. after #928 is merged. I can probably rebase 2. as well, should be trivial enough (famous last words)

nmunnich:
Rebased onto main. I also added a test and a fix for the west-commands not being imported correctly after the manifests are imported.

src/west/manifest.py
```diff
@@ -2689,13 +2697,27 @@ def _import_data_from_project(
...
         # by allocating them to the project.
-        project.west_commands = _west_commands_merge(
-            project.west_commands, submanifest._ctx.manifest_west_commands
-        )
+
+        # If the manifest was imported from a project subdirectory
+        # (manifest_path is a relative path within the project),
+        # we need to adjust the west_commands paths to be relative
+        # to the project root, not to the manifest subdirectory.
+        mfst_dir = Path(mfst_path).parent if _is_yml(mfst_path) else Path(mfst_path)
```
marc-hb, line 2709: I don't think you can use `_is_yml()` here. When looking at other places where `_is_yml()` is already used, you can see it is used only to collect all `submanifests/*.{yml,yaml}` files. `_is_yml()` is not used to decide whether `submanifests` is a file or a directory. 

In the same places where `_is_yml()` is used, you will find the code that decides whether `submanifests` is a file or a directory. That code seems consistent with the documentation (but please check).

I'm afraid this could require a git query in some cases.

BTW this call to `_is_yml()` seems new, I can't find it in earlier revision https://github.com/zephyrproject-rtos/west/commit/51c2f44ad4461c90f3991fb481e21b79fd10d2da. What changed?

tests/test_manifest.py
marc-hb, line 1: Please squash fixup commits:
https://docs.zephyrproject.org/latest/contribute/contributor_expectations.html#pr-requirements

Zephyr is not your typical Github project :-)

src/west/manifest.py
marc-hb, line 1: Can you please add the test(s) into a separate commit after the fix so it's easier to test with and without the fix? Sorry to ask this but my head hurts every time I'm trying to understand what works and what does not and when. It's good practice generally speaking.

src/west/commands.py
```diff
@@ -701,13 +701,18 @@ def _ext_specs(project):
         except pykwalify.errors.SchemaError as e:
             raise ExtensionCommandError from e
 
+        # Resolve command Python files relative to the manifest root for
```
marc-hb, line 704:
```suggestion
        # Resolve west command extensions relative to the manifest root for
```

tests/test_manifest.py
```diff
@@ -2116,6 +2116,67 @@ def test_import_project_submanifest_commands_both(manifest_repo):
...
+                                manifest:
+                                  projects:
+                                  - name: p2
+                                    url: url-placeholder2
+                                  self:
+                                    west-commands: p2subdir/west-commands.yml
+                                ''',
+        },
+    )
+    checkout_branch(p1, 'master')
```
marc-hb, line 2151: Is this "hiding" `manifest-rev` to make sure the import sticks to it anyway and does not look at the checkout instead? If I'm correct, that's subtle west behavior and worth a one-line comment. If misunderstood this, then maybe this is worth another comment :-)

src/west/manifest.py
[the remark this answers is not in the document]
marc-hb: Sorry I'm paying attention to this only now:

> I did notice another bug while doing my testing:
```
projects:
    - name: p1
      import: app/west.yml
      west-commands: app/scripts/west-commands.yml
```
First, I'm really not a fan of such "shortcuts + import" combinations BTW. This is adding `west-commands` from another git repo directly _and_ "shortcutting" the  `import app/west.yml` which could also add the same `west-commands.yml` file a second time! Does it get added a second time in your test case? I don't think you said. If yes, is that second time an error? A warning? I hope it's documented somewhere but I'm not holding my breath. Also, why adding the same `west-commands.yml` file twice?

If `app/west.yml` is not adding `west-commands.yml` second time then why is it not in `app/west.yml`? Why force every `app/west.yml` importer to manually add extensions?

Don't get me wrong: I don't see anything wrong with having the `west-commands:` shortcut ALONE without the `import`. But if a project becomes big and bold enough to have its own `west.yml` manifest, then shouldn't that submanifest import be a the single module "entry point" and take care of implementation details like the location of `west-commands` too? Or am I missing some real-world use case here?

All this being said, the ship has sailed: this has been allowed so the behavior should be specified. But: can we deal with this separately from #920? (which has got surprisingly more complicated than I expected already). If yes separately, could you please file a separate bug? Thx in advance.

cc:
- #903

tests/test_extension_commands.py
```diff
@@ -266,6 +266,77 @@ def do_run(self, args, unknown):
...
+    # self: west-commands: scripts/west-commands.yml, then file paths in that
+    # west-commands YAML are resolved relative to the imported manifest root.
+    manifest_path = repos_tmpdir / 'repos' / 'zephyr'
+    net_tools_path = repos_tmpdir / 'repos' / 'net-tools'
+
+    add_commit(
+        net_tools_path,
+        'add imported submanifest extension command',
+        files={
+            'mf_subdir/west.yml': textwrap.dedent(
```
marc-hb, line 280: Local constants would be more readable and more flexible:

```suggestion
        files={
            'mf_subdir/west.yml': MF_SUB_WEST_YML,
            'mf_subdir/scripts/west-commands.yml' MF_SUB_WEST_CMDS,
```

tests/test_extension_commands.py
```diff
@@ -266,6 +266,77 @@ def do_run(self, args, unknown):
     assert 'second command' in ext_output
 
 
+def test_import_project_submanifest_commands_from_project_subdirectory(repos_tmpdir):
+    # If net-tools imports mf_subdir/west.yml and that manifest declares
+    # self: west-commands: scripts/west-commands.yml, then file paths in that
+    # west-commands YAML are resolved relative to the imported manifest root.
```
marc-hb, line 272:
```suggestion
    # west-commands YAML are resolved relative to the imported manifest root.
    # ... because the same paths must work when imported and not: importing
    # must never break anything.
```
Just trying to demonstrate the importance of this test (and fix)

tests/test_extension_commands.py
```diff
@@ -266,6 +266,77 @@ def do_run(self, args, unknown):
...
+                        print('imported extension works')
+                '''
+            ),
+        },
+    )
+
+    with yaml_editor(manifest_path / 'west.yml') as mf:
+        net_tools_project = [p for p in mf['manifest']['projects'] if p['name'] == 'net-tools'][0]
+        net_tools_project['import'] = 'mf_subdir/west.yml'
+    subprocess.check_call([GIT, '-C', str(manifest_path), 'add', 'west.yml'])
```
marc-hb, line 322: I think you can pass the filename as an argument to the next `git commit` and drop this.

GIT is _insanely_ slow on WINE :-) Pretty slow on Windows too.

tests/test_extension_commands.py
```diff
@@ -266,6 +266,77 @@ def do_run(self, args, unknown):
...
+        '-C',
+        str(manifest_path),
+        'commit',
+        '-m',
+        'import mf_subdir/west.yml',
+    ])
+
+    workspace = repos_tmpdir / 'workspace'
+    cmd(['init', '-m', str(manifest_path), str(workspace)])
+    cmd('update', cwd=workspace)
```
marc-hb, line 334: Can you start from `west_init_tmpdir` instead of `repos_tmpdir` and avoid this? Sorry if not possible.

tests/test_manifest.py
```diff
@@ -2116,6 +2116,67 @@ def test_import_project_submanifest_commands_both(manifest_repo):
     assert p1.west_commands == expected
 
 
+def test_import_project_submanifest_commands_from_project_subdirectory(manifest_repo):
```
marc-hb, line 2119: As of 6f600e37be95cf3167801838940339a0561924bb, you have two tests with the same name 6f600e37be95cf3167801838940339a0561924bb in two different files. Can you please find different names? We don't need a lot of difference but we need the ability to run a single one at a time from `pytest -k`.

marc-hb (commented):
This PR opened a big backslash / PosixPath can of worms that kept both of us busy for a while in this comment: https://github.com/zephyrproject-rtos/west/pull/920#discussion_r2830240711

But I don't see any corresponding test as of the latest version 6f600e37be95cf3167801838940339a0561924bb.

Please:
- use special characters like I did in #928
- add some naive `assert(west manifest --resolve)` at the end of the relevant tests.

If this test is "naive" then I'm afraid it will fail either on Linux or on Windows but cannot pass on both at the same time. Then we can discuss whether we make the test Windows-specific like I did in #928, or whether we change `west` behavior to make the output the same on both. I will work on either option.

src/west/manifest.py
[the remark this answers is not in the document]
nmunnich: In this case the bug I was trying to describe is that which is shown by the newest test and fixed in this PR in the last commit. Writing it this way was simply a way to avoid writing the app/west.yml where the script gets imported down, I _think_ but can't recall completely that I simply didn't import the commands in app/west.yml when testing this. 

The interaction I care about is that of the newest test, and I agree that any interaction with multiple import locations of the same westcommands file should be delt with separately from this.

tests/test_manifest.py
nmunnich: I have no idea about this one - I was simply following the same format used in many other tests in this file, see e.g. L2110.

tests/test_extension_commands.py
nmunnich: Does not seem to be possible. I think it's initialised too early, but have not dug deeper.

nmunnich:
> This PR opened a big backslash / PosixPath can of worms that kept both of us busy for a while in this comment: [#920 (comment)](https://github.com/zephyrproject-rtos/west/pull/920#discussion_r2830240711)
> 
> But I don't see any corresponding test as of the latest version [6f600e3](https://github.com/zephyrproject-rtos/west/commit/6f600e37be95cf3167801838940339a0561924bb).
> 
> Please:
> 
>     * use special characters like I did in #928
> 
>     * add some naive `assert(west manifest --resolve)` at the end of the relevant tests.
> 
> 
> If this test is "naive" then I'm afraid it will fail either on Linux or on Windows but cannot pass on both at the same time. Then we can discuss whether we make the test Windows-specific like I did in #928, or whether we change `west` behavior to make the output the same on both. I will work on either option.

I'm quite uncertain what you mean here. My understanding of the situation is that I happened to be developing on Windows, and that ended up finding an inconsistency in the way West has been handling filepaths. I still don't really consider it in the scope of this PR to test or fix this inconsistency - my goal here is to fix the bug described by the original issue.

tests/test_manifest.py
marc-hb: Mmmm... I think you should be able to have some understanding of your code you're adding, including test code. I'm afraid "I just copied this from this other test" does not sound good enough. This is especially true for imports which is a very (too?) complex and subtle feature. This PR is about imports and this is an import question so this question is definitely on-topic.

This line is a git command in a test for a meta-git tool, it's not some obscure test framework detail.

Note I'm NOT asking you to make any change here now, I'm only trying to understand it myself.

marc-hb (dismissed):
I just tried to get another look at this, really sorry for the wait.

I see the more recent commits that you  added for the issue that you found and described in this comment: https://github.com/zephyrproject-rtos/west/pull/920#discussion_r2855625436

But the sequence of test additions and fixes in the git log didn't completely make sense to me. So, I tried to run the tests on each commit separately and I found that only the last commit https://github.com/zephyrproject-rtos/west/commits/e4d0866ee45b0076218417084be48b33b0b304ed is passing all the tests. Most other commits fail like this:

 ```
 .
 .
 .
 .
 FAILED tests/test_project.py::test_update_some_with_imports - AssertionError: imap_or_mfpath has unexpected type <class 'NoneType'>
FAILED tests/test_project.py::test_update_recovery - AssertionError: imap_or_mfpath has unexpected type <class 'NoneType'>
FAILED tests/test_project.py::test_import_project_release_dir - AssertionError: imap_or_mfpath has unexpected type <class 'NoneType'>
FAILED tests/test_project.py::test_import_project_release_fork - AssertionError: imap_or_mfpath has unexpected type <class 'NoneType'>
FAILED tests/test_project.py::test_import_project_release - AssertionError: imap_or_mfpath has unexpected type <class 'NoneType'>
FAILED tests/test_project.py::test_import_project_rolling - AssertionError: imap_or_mfpath has unexpected type <class 'NoneType'>
=========== 23 failed, 272 passed, 1 xfailed in 27.29s ==============================
```

Every git commit must pass all its tests.

"Test-driven" development and writing tests first is great but for git bisectability and other history reasons, the order should be flipped before submission and features should come first and tests second in the end. Or all in the same commit if small enough, that's less important as long as all commits pass.

This being said, flipping the order is probably not enough in this case because a lot of the failing tests are untouched by this PR? No commit should knowingly introduce bugs fixed later in the same PR:

https://docs.zephyrproject.org/latest/contribute/contributor_expectations.html#pr-requirements

As described in this page, commits should be as small as possible... but not smaller.

**Comment by marc-hb** [comment]
> My understanding of the situation is that I happened to be developing on Windows, and that ended up finding an inconsistency in the way West has been handling filepaths. I still don't really consider it in the scope of this PR to test or fix this inconsistency - my goal here is to fix the bug described by the original issue.

I could be convinced that these are truly independent of each other, but my current impression is that the Windows/Unix inconsistencie(s) when handling this path were so far hidden by the bug that this PR is fixing.

Car analogy: as long as the engine wasn't even starting, no one really cared that Windows drove on the left and Unix on the right. Cause they weren't even driving at all in the first place. Now does that mean you can fix and start the engine and still not care about the inconsistency? I would say "no" but again maybe I'm simplifying things too much and maybe my analogy is wrong.

File: tests/test_manifest.py (hunk 24 above)
Comment (nmunnich) [reply tests/test_manifest.py:2147]: Fair enough. I dug into it somewhat and your initial thinking was correct. I've added a comment here for future investigators to discover, but have not added the comment in the other places where this appears.

File: src/west/manifest.py (hunk 22 above)
Comment (nmunnich) [reply src/west/manifest.py]: Previously we were testing only the cases where we import from a file. But import blocks can also be directories, so I added this to account for that scenario.

Good callout that `_is_yml()` shouldn't be used for this though. I've added some plumbing which should pass through whether the manifest file came from a directory or a direct file properly, and I've adjusted the test in the first commit to showcase all four scenarios.

**Comment by nmunnich** [comment]
Very sorry about the messy commits in the previous force-push. I took care to be a bit more rigorous this time, since I have more time available again. In the last one I made a silly mistake and had one crucial line which should have been in the second commit in the fourth commit.... Each commit should have all tests passing.

Regarding our inconsistency, you are in the end correct that it does seem to affect this after all. My understanding was that we have:

- the west file itself
- the output of `west manifest --resolve`, i.e. what west sees
- the ability of west to find and run the files found by the manifest

The previous discussion appeared when we were discussing west turning 1 into 2, where both Linux and Windows seem to behave perfectly fine regardless of what type of paths we choose, and hence I thought we were safe here. Going from 2 to 3, we see an issue appearing in Linux where it cannot interpret the windows style paths in the resolved manifest correctly.

I've added a windows-only test in the last commit which shows this. If the test is run on Linux it fails, on windows it passes.

Personally I was considering the parsing of 1 into 2 not normalising paths into posix-style to be a bug until #928. I would still consider that to be the easiest and most "proper" fix, especially since other paths in west are normalised in that manner. At the same time I feel I don't have enough stakes in the project or a weighty enough voice to get an agreement on that, hence I was really hoping we could just ignore it :(

**Comment by nmunnich** [comment]
Tried running `uv run poe all` inside of a Python3 vscode devcontainer to hopefully fix whatever metadata/encoding issue ruff was complaining about. The diff of the last forcepush looks very awkward, the contents are completely identical....

File: src/west/manifest.py [hunk 30]
```diff
@@ -2717,10 +2737,12 @@ def _import_content_from_project(self, project: Project, path: str) -> ImportedC
             # asked to use the importer.
             content = self._ctx.project_importer(project, path)
 
+        import_target_is_dir = isinstance(content, list)
```
marc-hb: So this is "saving" whether it was a list or not in a new boolean to tell the caller, while still converting single elements into a list in the next block... both dropping the list information and then restoring it in a boolean...

Please don't do this. If the `_import_content_from_project()` caller needs that information, then just drop the `content = [ content]` and let the callers deal with the type difference directly: it is precisely the new information they now want! AFAICT there are only two places where this function it called and they are next to each other.

Funny enough, `_import_content_from_project()` already claims to be returning `ImportedContentType = str | list(str)`! But doesn't yet?

src/west/manifest.py
```diff
@@ -83,6 +83,8 @@
 ImportedContentType = str | list[str] | None
 ImporterType = Callable[['Project', str], ImportedContentType]
 
+_ImportedProjectData = tuple[list[str] | None, bool]
```
marc-hb: I find this confusing, see below.

src/west/manifest.py
```diff
@@ -2652,26 +2654,36 @@ def _import_map_from_project(self, project: Project, imp: dict) -> None:
 
         _logger.debug(f'resolving import {imap} for {project}')
 
-        imported = self._import_content_from_project(project, imap.file)
+        imported, import_target_is_dir = self._import_content_from_project(project, imap.file)
```
marc-hb: If we passed an import map, then the target cannot be a directory, correct?

If correct then please add an `assert` here. Not so much to catch bugs (although that won't hurt) but to help the reader understand the flow between these functions better.

src/west/manifest.py
```diff
@@ -2652,26 +2654,36 @@ def _import_map_from_project(self, project: Project, imp: dict) -> None:
...
 
         _logger.debug(f'done resolving import {imap} for {project}')
 
     def _import_data_from_project(
-        self, project: Project, data: Any, imap: _import_map | None
+        self,
+        project: Project,
+        data: Any,
+        imap_or_mfpath: _import_map | str,
+        import_target_is_dir: bool,
```
marc-hb: imap_or_mfpath has 2 possibilities and import_target_is_dir is a boolean so it looks like we have 2x2 possibilities. But in reality, there are only 3, correct?
If correct, is there a way to have a single parameter for these 3 possibilities instead? And then the three corresponding case in the match below. I think these 3 cases would correspond to the documentation and make the code so much easier to read.
If needed, maybe add a new enum / discriminant and maybe extend `ImportedContentType` and/or `WestCommandType`, not sure which. In any case prefer import code that is more similar to the import documentation.

src/west/commands.py
```diff
@@ -701,13 +701,18 @@ def _ext_specs(project):
         except pykwalify.errors.SchemaError as e:
             raise ExtensionCommandError from e
 
+        # Resolve west command extensions relative to the manifest root for
```
marc-hb, line 704:
```suggestion
        # Resolve west command extensions relative to their respective manifest root for
```

src/west/manifest.py
```diff
@@ -904,6 +904,7 @@ def __init__(
         self.clone_depth = clone_depth
         self.path = os.fspath(path or name)
         self.west_commands = _west_commands_list(west_commands)
+        self._west_commands_manifest_dirs: dict[str, str] = dict()
```
marc-hb: This is not a new constructor parameter, please don't insert it in the middle of the list of such parameters.

(The other, evil `ManifestProject` class is not organized as well)

marc-hb: Actually, would it be simpler and clearer to just promote `WestCommandsType` to some very basic but actual Python class?

tests/test_extension_commands.py
```diff
@@ -343,6 +343,84 @@ def do_run(self, args, unknown):
     assert 'imported extension works' in ext_output
 
 
+def test_call_imported_project_submanifest_commands_from_project_subdirectory_windows_paths(
+    repos_tmpdir,
+):
+    if not WINDOWS:
+        return
```
marc-hb: skip instead

src/west/commands.py
marc-hb: Is this fixing something that was not described in #725, or did you split the fix for #725 into two commits? As of the latest push https://github.com/zephyrproject-rtos/west/commits/daf584f9544b25433029b27f3c0b0a7d854929f5, the commit messages are empty so this does not help :-( 

If that use case is not described in #725 yet then can you please add it there? Or file a new bug entirely if you prefer. Should be much easier than reading the test code.

marc-hb (commented):
Finally getting to this, sorry for the delay.

PS: this PR has also changed quite significantly since the very first push. That also explains why it's been open for a while :-)

src/west/manifest.py
nmunnich: See https://github.com/zephyrproject-rtos/west/pull/920/changes/277b14becbb400a8a91a450d1b89e236f20a6f95#r3363910096, docs seem to state we have four possibilities.

src/west/manifest.py
nmunnich: According to the documentation, it can still be a directory. 
["file: Optional. The name of the manifest file or directory to import. This defaults to west.yml if not present."](https://docs.zephyrproject.org/latest/develop/west/manifest.html#option-3-mapping)
Manifest could have something like
```
import:
  file: my_directory
```

src/west/commands.py
nmunnich: I'm not really sure where to draw the line here, to be honest. The title of #725 is "Imported west-commands don't resolve when module manifest isn't in module root". 
- If we interpret "resolve" here as "they are picked up by `west manifest --resolve`, then this is fixing something not described in #725. 
- If we interpret it as "The west command won't run", then it isn't described in #725 and then this is a fix to a followup bug which is only visible once #725 was fixed in the first commit.

src/west/manifest.py
nmunnich: I did consider this, and gave it a brief try, but it felt like a much more invasive refactoring than I was comfortable doing, and that would _further_ explode the size of this PR (which as you said has evolved quite a bit). It most likely should have been a class from the very start tbh - `_west_commands_list` feels like a constructor, `_west_commands_maybe_delist` feels like a string representation, `_west_commands_merge` would be an elegant `__iadd__`....

I've moved the _west_commands_manifest_dirs down a bit.

src/west/manifest.py
```diff
@@ -2653,25 +2659,40 @@ def _import_map_from_project(self, project: Project, imp: dict) -> None:
...
-        for data in imported:
-            self._import_data_from_project(project, data, imap)
+        match imported:
+            case None:
+                return
+            case str():
+                self._import_data_from_project(project, imported, imap, False)
+            case list():
+                for data in imported:
+                    self._import_data_from_project(project, data, imap, True)
```
marc-hb:
```suggestion
                    self._import_data_from_project(project, data, imap, directory_import=True)
```
I'm sorry to nitpick but the number of combination in this area of the code gives me a headache every time and I feel like a small changes like this go a very long way to mitigate the pain. To be clear: I find this code hard to read _before_ this PR.

src/west/manifest.py
```diff
@@ -2653,25 +2659,40 @@ def _import_map_from_project(self, project: Project, imp: dict) -> None:
...
 
         _logger.debug(f'done resolving import {imap} for {project}')
 
     def _import_data_from_project(
-        self, project: Project, data: Any, imap: _import_map | None
+        self,
+        project: Project,
+        data: Any,
+        imap_or_mfpath: _import_map | str,
+        import_target_is_dir: bool,
```
marc-hb: See above
```suggestion
        *,
        directory_import: bool,
```

tests/test_extension_commands.py
```diff
@@ -266,6 +266,83 @@ def do_run(self, args, unknown):
...
+        '''\
+        manifest:
+          self:
+            west-commands: scripts/west-commands.yml
+        '''
+    )
+    MF_SUB_COMMANDS_YML = textwrap.dedent(
+        '''\
+        west-commands:
+          - file: test/west-commands/west.py
```
marc-hb:
```suggestion
          - file: test/west-commands/ext-subdir-725.py
```

Please use less generic/more random names to
- make it very clear these are just samples and not following any sort of naming convention or API
- massively help `git grep` and other code searches. Especially in case of a test failure but not just.
- avoid collisions. You never know: test code can get copied, extended, etc.

It does not matter much what the name is in this case as long as it's not super generic like `west.py` (and TBH: a bit confusing for `west.py`)

tests/test_extension_commands.py
```diff
@@ -266,6 +266,83 @@ def do_run(self, args, unknown):
...
+            commands:
+              - name: imported-extension
```
marc-hb:
```suggestion
              - name: subdir-cmd-715-1
```
See above.

tests/test_extension_commands.py
```diff
@@ -266,6 +266,83 @@ def do_run(self, args, unknown):
...
+                class: ImportedExtension
+                help: imported extension help
+        '''
+    )
+    MF_SUB_WEST_PY = textwrap.dedent(
+        '''\
+        from west.commands import WestCommand
+
+        class ImportedExtension(WestCommand):
```
marc-hb:
```suggestion
        class SubdirExt725(WestCommand):
```
or anything less generic, see above.

tests/test_extension_commands.py
```diff
@@ -266,6 +266,83 @@ def do_run(self, args, unknown):
...
+                    'imported-extension',
+                    'imported extension help',
+                    'imported extension description',
+                )
+
+            def do_add_parser(self, parser_adder):
+                return parser_adder.add_parser(self.name)
+
+            def do_run(self, args, unknown):
+                print('imported extension works')
```
marc-hb:
```suggestion
                print('hello from subDir Ext 725')
```

src/west/manifest.py
```diff
@@ -2724,6 +2728,12 @@ def _import_data_from_project(
             (mfst_dir / cmd).as_posix() for cmd in submanifest._ctx.manifest_west_commands
         ]
 
+        # Keep track of which imported manifest directory each adjusted
+        # west-commands entry came from, so command Python files can be
+        # resolved relative to that manifest root later in commands.py.
```
marc-hb, line 2734: Please add some `TODO: WestCommands should be promoted to a Python class, see issue #nnn`

src/west/manifest.py
marc-hb: > but it felt like a much more invasive refactoring than I was comfortable doing, and that would further explode the size of this PR

Fair enough; I was afraid of that :-(

More generally speaking, I feel sorry for the "trap" you/we seemed to have walked into. This should have been a simple fix but the complexity of this code (and of the feature itself TBH) really got in the way.


>  It most likely should have been a class from the very start tbh - _west_commands_list feels like a constructor, _west_commands_maybe_delist feels like a string representation, _west_commands_merge would be an elegant __iadd__....

Great observations, would you mind quickly dumping those in a new west enhancement issue? No need to write a novel: just link back to this PR instead! That will provide lots of reading already :-) Then just add a TODO in the code linking to that issue.

marc-hb (commented):
Only a few style and comment changes requested, especially in test code which I hadn't really looked at before. Everything else looks good.

Thank you so much for not giving up, much appreciated.

marc-hb:
Forgot to say: I haven't looked much at the 3rd commit that adds the new backslash test that I requested, but you have "suffered" long enough, this is only test code and it looks like a great starting point for me to take over and finally deal with this portability issue. Thanks again.

nmunnich:
Should be everything addressed now, I hope. Thanks for being happy to take over the portability stuff.

marc-hb (approved):

nmunnich:
Copilot comments addressed.

nmunnich:
Actually, one more thing I just noticed. In this PR I use `match` statements, which I find helps with reasoning and understanding. The `pyproject.toml` has Python 3.10 as the minimum version, which is the version where `match` was introduced, but pypi seems to suggest that the minimum Python for west is 3.9. 
I hope pypi is just outdated and that needs to be fixed?

marc-hb (approved):
https://pypi.org/project/west/1.5.0/ supports Python 3.9 because 3.9 was still supported when 1.5.0 was released. But 3.9 is not supported anymore now and the next west release is already not compatible with 3.9 anymore. If you look at the CI checks, you'll see that all compatible Python versions get tested.

src/west/manifest.py
```diff
@@ -2653,25 +2664,40 @@ def _import_map_from_project(self, project: Project, imp: dict) -> None:
...
+        match imported:
+            case None:
+                return
+            case str():
+                self._import_data_from_project(project, imported, imap, directory_import=False)
+            case list():
+                for data in imported:
+                    self._import_data_from_project(project, data, imap, directory_import=True)
+            case _:
+                raise AssertionError(f'imported content has unexpected type {type(imported)}')
```
pdgendt, line 2676: The return type of `_import_content_from_project` leaks importer behavior, this is fine for the internal handling, but not for a user-supplied `project_importer` callback.

We shouldn't infer file-vs-directory from the response shape and instead carry it as ground truth from the source that knows:

- For `_manifest_content_at:` the answer is `ptype == 'tree'`. Return it.
- For the user-supplied `project_importer`: keep `directory_import=False` as the default when going through the callback

```python
    def _import_content_from_project(
        self, project, path
    ) -> tuple[ImportedContentType, bool]:  # (content, is_directory)
        # ...
```

pdgendt (commented):
Sorry for the long overdue feedback, the open issues can go to follow-up PRs, just one comment I think we should incorporate in this PR.

src/west/manifest.py
marc-hb: I'm afraid I asked earlier the opposite change in https://github.com/zephyrproject-rtos/west/pull/920#discussion_r3363831967...

I just found this in the (long) pydoc of `Manifest.__init__()`

```
        The *importer* callback's return value should be a string containing
        manifest data, or a list of strings if ``file`` is a directory
        containing YAML files. A return value of None will cause the import
        to be ignored.
```
Also:

```python
# Type for the importer callback passed to the manifest constructor.
# (ImportedContentType is just an alias for what it gives back.)
ImportedContentType = str | list[str] | None
ImporterType = ...
```

So a boolean seems redundant

pdgendt (approved):

PR #921, opened 2026-02-17
tests: stop hardcoding a forward slash in 'scripts/west-commands.yml'

marc-hb:
Use backwards slashes on Windows to catch any future hardcoding mistake.

Minor oversight found while discussing #920

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -31,8 +31,19 @@ GIT = shutil.which('git')
 # name.
 GIT_INIT_HAS_BRANCH = False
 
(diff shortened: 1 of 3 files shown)


marc-hb:
Following some feedback from @pdgendt , to be replaced by more specific tests like this:
https://github.com/marc-hb/west/tree/backslash-tests

marc-hb:
> Following some feedback from @pdgendt , to be replaced by more specific tests like this:
https://github.com/marc-hb/west/tree/backslash-tests

Submitted #928 instead.
- #928

PR #923, opened 2026-02-20
Add new WINE.md help file

marc-hb:
Slow and not 100% reliable but still very useful to quickly test Pathlib and other issues without juggling with multiple systems.

This helped me a lot with #928

cc:
- @jhol (spotted in https://github.com/msys2/MINGW-packages/pull/28122)

### DIFF
--- a/WINE.md
+++ b/WINE.md
@@ -1,0 +1,99 @@
+https://www.winehq.org/ provides a way to run Windows binaries on Linux.
+
+Q: Why would a west developer need WINE considering west is portable Python code?
+A: To quickly put themselves in the shoes of Windows users and troubleshoot
+   Windows-specific issues without:
+
+- A Windows licence
+- A different computer to run it, or the inconvenience of dual-booting
+- The inconvenience of synchronizing source files across computers
+
+Warning: WINE is not 100% identical to Windows and WINE is _not_ able to replicate all
+issues happening on Windows. But enough issues can be reproduced, including Path-related
+issues.
+
+This is a quick and dirty recipe to run west tests using WINE.  This process is not
+officially supported and these instructions can "bitrot" and may have fallen apart by the
+time you read them: you have been warned. Please submit fixes or better ways if you know
+them. If humanly possible, keep this working over ssh. That is: without any GUI.
+
+Do NOT replace Windows by WINE to validate releases!
+
+WINE
+----
+
+First you need to install WINE. Try the following commands and your Linux distribution will
+likely tell you what packages are missing:
+
+```
+wine   --help
+wine64 --help
+wine32 --help
+```
+At least one of these must print WINE's help message before you proceed.
+
+Some distributions let you install both at the same time. To find which one the "wine"
+command currently points at:
+```
+wine cmd
+dir c:
+```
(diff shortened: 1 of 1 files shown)


marc-hb:
I tried `uv` but it's not necessary now and it fails like this:

```
cd west
rm -rf .venv # do not share `uv` with WINE
wenv uv version # OK
wenv uv format # OK
wenv uv run poe # FAIL

   Building docopt==0.6.2
  × Failed to build `west @ file:///Z:/home/marc/west`
  ├─▶ The build backend returned an error
  ╰─▶ Call to `setuptools.build_meta.build_editable` failed (exit code: 1)

      [stderr]
      Fatal Python error: Failed to import encodings module
      Python runtime state: core initialized
      ImportError: bad magic number in 'encodings': b'\xcb\r\r\n'

      Current thread 0x0000015c (most recent call first):
        <no Python frame>

      hint: This usually indicates a problem with the package or the build environment.

wenv pip -v install encodings

ERROR: Could not find a version that satisfies the requirement encodings (from versions: none)
ERROR: No matching distribution found for encodings

wenv python -c 'import encodings; print(encodings)'
<module 'encodings' from 'Z:\\tmp\\wenv-d574c107\\Lib\\encodings\\__init__.pyc'>
```

pdgendt (commented):
Hate to whine about this (pun intended), but do we need a "quick and dirty recipe" for something that clearly isn't well supported?

marc-hb:
Good question!

- Because many people would not be aware this possibility even exists. See this new file as a (quiet) "advertisement" in a corner.
- Because (other) Un\*x users suspecting the possibility might exist would likely waste a couple hours of trial and error exactly like I did before finding and aligning the planets as indicated in this new, small text file. Or, more likely, they would give up after less than 30 min and fetch and dust off some Windows system from their attic instead[*]. I didn't give up because I didn't have access to Windows - necessity is the mother of invention!
- Even when this doc gets out of date, it will still be a much better starting point than nothing at all.


[*] ... and then they would watch that Windows system downloading and applying updates for hours and hours, but at least they could do something on their main, Linux system in the meantime. I digress.

kartben:
It looks to me as if this should just be a wiki page maybe?

marc-hb:
That definitely looks like wiki material! On the other hand, starting a new `west` wiki just for this page feels overkill and that page would feel alone and likely not seen much. Now you have to start somewhere and a first page is by definition alone for some time - but I'm afraid that one would be alone for a very long time in a "west wiki".

Also, wikis tend to be for users and especially new users whereas this is 100% a _development_ trick.

I could move it (and the MAINTAINERS.rst file) to a new `doc/` subdirectory to reduce distraction?

nmunnich:
As someone more on the periphery of west, I'd support a new `doc/` subdirectory. I'd suggest that the "hacking on west" section of the readme get its own file in that case, and link to said file from the main readme.

pdgendt:
> As someone more on the periphery of west, I'd support a new `doc/` subdirectory. I'd suggest that the "hacking on west" section of the readme get its own file in that case, and link to said file from the main readme.

I need to get back to https://github.com/zephyrproject-rtos/west/pull/860, where this might be moved to.

PR #925, opened 2026-02-20
version.py: add comment with workaround to run pytest from scratch.

marc-hb:
`importlib.metadata.version("west")` requires a west package. After a surprisingly long time, I realized there's none when running pytest from scratch. That's because (un)installations seem to always leave one behind. So this "accidentally" tends to work. As a workaround, recommend the very well tried and tested tactic of... installing - and uninstalling.

cc: @thorsten-klein

### DIFF
--- a/src/west/version.py
+++ b/src/west/version.py
@@ -7,6 +7,14 @@
 
 import importlib.metadata
 
+# The package metadata required can be missing when running pytest directly.  One
+# possible workaround is:
+#
+#   cd west && pipx install -e . && pipx uninstall west
+#
+# This cycle leaves behind a `west/src/west.egg-info/PKG-INFO` file which is enough.
+# `pip install --break-system-packages ...` can achieve the same result but with more
+# disruption than pipx.
 __version__ = importlib.metadata.version("west")
 #
 # MAINTAINERS:


pdgendt:
> `importlib.metadata.version("west")` requires a west package. After a surprisingly long time, I realized there's none when running pytest from scratch.

How exactly do you get into this situation?

marc-hb:
> How exactly do you get into this situation?

Easy:
```
pip uninstall west
pipx uninstall west
cd west && git clean -fdx # DANGEROUS! Alternative: rm -rf west/*egg*
pytest ...
```

marc-hb:
> I realized there's none **when running pytest from scratch**. 

Emphasis in case you missed this.

`uv ... pytest ...` does not have that problem because `uv` "installs" first. `uv` is great but it adds a layer of isolation which makes it hard to bisect `cpython` or `ruamel` (#908), or to run tests with WINE (#923), or debug issues in general.

`uv` isolation is either a blessing or a curse depending on what you're doing.

"All problems in computer science can be solved by another level of indirection."

"Except for too many layers of indirection!"

thorsten-klein:
What about maintaining the version in the python code and setuptools will read it from there?

```
 pyproject.toml      | 5 ++++-
 src/west/version.py | 4 +---
 2 files changed, 5 insertions(+), 4 deletions(-)

diff --git a/pyproject.toml b/pyproject.toml
index a4b12b5..c12af43 100644
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "west"
-version = "1.5.99"
+dynamic = ["version"]
 authors = [{name = "Zephyr Project", email = "devel@lists.zephyrproject.org"}]
 description = "Zephyr RTOS Project meta-tool"
 classifiers = [
@@ -62,6 +62,9 @@ package-dir = {"" = "src"}
 zip-safe = false
 include-package-data = true
 
+[tool.setuptools.dynamic]
+version = {attr = "west.version.__version__"}
+
 [tool.setuptools.packages.find]
 where = ["src"]
 namespaces = false
diff --git a/src/west/version.py b/src/west/version.py
index b6ddd08..6043d6d 100644
--- a/src/west/version.py
+++ b/src/west/version.py
@@ -5,9 +5,7 @@
 # This is the Python 3 version of option 3 in:
 # https://packaging.python.org/guides/single-sourcing-package-version/#single-sourcing-the-version
 
-import importlib.metadata
-
-__version__ = importlib.metadata.version("west")
+__version__ = "1.5.99"
 #
 # MAINTAINERS:
 #
```

marc-hb:
> What about maintaining the version in the python code and setuptools will read it from there?

I'm really not a packaging expert but LGTM. In fact, having the version in a file called `version.py`  looks even better: ever "closer" to the versioned source itself and "package-technology-independent". This unsurprisingly solves the issue of running pytest without a package!

What is the "best practice" for this? Assuming there is one...

pdgendt:
> What is the "best practice" for this? Assuming there is one...

Clicking on the  link in the version.py file leads to https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html#dynamic-metadata which describes the changes from @thorsten-klein 

It would be a cleaner solution definitely.

PR #928, opened 2026-03-03
Add test coverage for special characters in manifest and west commands

marc-hb:
Pure tests changes.

Prompted by the #920 fix.

cc: @nmunnich

### DIFF
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -14,6 +14,7 @@ import uuid
 from pathlib import Path, PurePath
 
 import pytest
+import yaml
 
 from west.app import main
 
@@ -68,6 +69,19 @@ WINDOWS = platform.system() == 'Windows'
 #
 # Contextmanager
 #
+
+
+@contextlib.contextmanager
+def yaml_editor(yaml_f: str | Path):
+    # Fail fast if not writable
+    with open(yaml_f, 'r+') as f:
+        pass
+    with open(yaml_f) as f:
+        mf = yaml.safe_load(f)
+    yield mf
+    # Overwrite file
+    with open(yaml_f, 'w') as f:
+        yaml.safe_dump(mf, f, sort_keys=False)
 
 
 @contextlib.contextmanager
--- a/tests/conftest.py
+++ b/tests/conftest.py
@@ -39,25 +39,26 @@ manifest:
     remote: test-local
 
   remotes:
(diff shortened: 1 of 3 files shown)


pdgendt:
There's a CI issue, also non of your commits has the `Signed-off-by`, I was under the impression that was required in this repo.

marc-hb:
I will add the missing signed off, thanks! Hopefully everything else looks good :-D

The coverage tool is complaining about a lack of source code for the extension test.py file located... in /tmp/pytest/ ?!? That does not make sense to me, could you take a look? This file isn't even new, I'm merely moving it...

pdgendt:
> I will add the missing signed off, thanks! Hopefully everything else looks good :-D

👍 

> The coverage tool is complaining about a lack of source code for the extension test.py file located... in /tmp/pytest/ ?!? That does not make sense to me, could you take a look? This file isn't even new, I'm merely moving it...

Files created/moved during testing should be omitted from coverage, see:

https://github.com/zephyrproject-rtos/west/blob/5f5931fee70b26f654666a962dddb96ee8c8cdb9/pyproject.toml#L76-L81

marc-hb:
`coverage` configuration was a bit of a mess. Simplified in #929.
EDIT: #929 should be merged first, then removed from this.
EDIT: done.

pdgendt (approved):

marc-hb:
@nmunnich PTAL

nmunnich:
Code changes look fine to me. Did you want this specially tested on Windows natively?

marc-hb:
Thx @nmunnich for your time, much appreciated. I tested this thoroughly with WINE (see #923) and based on other WINE experience I'm pretty confident it does catch Windows-specific paths issues. Also, GitHub CI runs Windows. All this being said, it would be even better if you could spent 3 more minutes running these tests on Window cause you can never have too much coverage! You can only waste time and/or resources testing the wrong priorities but this one should be really quick. Up to you.

Also, if it looks OK then please feel free to hit the "Approve" button :-)

nmunnich (approved):
Tests pass on Windows native.

marc-hb:
Note to self: this #928 does not involve any `import`. It may seem like it because `west-commands:` in the top-level manifest points (directly!) to a file located in a west project/other git repo. But that is not an `import`.

PR #929, opened 2026-03-04
pyproject.toml: simplify coverage filter

marc-hb:
Drop the "omit" rules and use a simpler "source = src/" to declare
what we are interested in. This is the first recommendation at
https://coverage.readthedocs.io/en/latest/source.html#source-execution

Stop using the `pytest --cov-config=pyproject.toml` option as this
file format seems silently ignored there.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,19 +66,13 @@ include-package-data = true
 where = ["src"]
 namespaces = false
 
+# Note this section is only for interactive `coverage run ...` use and ignored by "gh-test"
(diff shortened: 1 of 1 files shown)


marc-hb:
This works with `coverage run /usr/bin/pytest` but not with `pytest --cov-report=html --cov-config=pyproject.toml --cov=west` :-(

 `--cov-config=` seems to do nothing at all

pdgendt (approved):
Nice improvement, thanks!

carlescufi (approved):

PR #930, opened 2026-03-04
west: add --color option to control output colorization

### LINKED ISSUE #651: Add west --color=always and west config color=always
Git has an (somewhat inconsistent) set of "color = always" options to override the automated terminal detection. Demo of one of the (too many) options:

```
git diff                            # colored           :-)
git diff                | less      # not colored       :-(
git diff --color=always | less  -R  # colored again!    :-(
```

It would nice for `west` to offer the same feature.

While rarer, `color=never` is sometimes needed too. For instance some CI engines try hard to pretend they are terminals to capture colors but then can fail to display them for some reason. 

cc:
- https://github.com/zephyrproject-rtos/west/pull/551#issuecomment-1439462843 - mysterious colorama issues

TharakaUJ:
Add a global --color option with values 'always', 'never', and 'auto' to override automatic terminal detection.

The command-line option takes precedence over configuration settings. When provided, it overrides the value returned by WestCommand.color_ui.

fixes #651

### DIFF
--- a/src/west/app/main.py
+++ b/src/west/app/main.py
@@ -93,6 +93,7 @@ class EarlyArgs(NamedTuple):
     version: bool  # True if -V was given
     zephyr_base: str | None  # -z argument value
     verbosity: int  # 0 if not given, otherwise counts
+    color: str | None  # --color argument value ('always', 'never', 'auto')
     command_name: str | None
 
     # Other arguments are appended here.
@@ -106,16 +107,18 @@ def parse_early_args(argv: list[str]) -> EarlyArgs:
     version = False
     zephyr_base = None
     verbosity = 0
+    color = None
     command_name = None
     unexpected_arguments = []
 
     expecting_zephyr_base = False
+    expecting_color = False
 
     def consume_more_args(rest):
         # Handle the 'Vv' portion of 'west -hVv'.
 
-        nonlocal help, version, zephyr_base, verbosity
-        nonlocal expecting_zephyr_base
+        nonlocal help, version, zephyr_base, verbosity, color
+        nonlocal expecting_zephyr_base, expecting_color
 
         if not rest:
             return
@@ -145,6 +148,13 @@ def parse_early_args(argv: list[str]) -> EarlyArgs:
     for arg in argv:
         if expecting_zephyr_base:
             zephyr_base = arg
+            expecting_zephyr_base = False
+        elif expecting_color:
+            if arg in ('always', 'never', 'auto'):
+                color = arg
+            else:
+                unexpected_arguments.append(f'--color={arg}')
+            expecting_color = False
         elif arg.startswith('-h'):
             help = True
             consume_more_args(arg[2:])
@@ -170,13 +180,23 @@ def parse_early_args(argv: list[str]) -> EarlyArgs:
                 zephyr_base = arg[3:]
             else:
                 zephyr_base = arg[2:]
+        elif arg == '--color':
+            expecting_color = True
+        elif arg.startswith('--color='):
+            color_val = arg[8:]
+            if color_val in ('always', 'never', 'auto'):
+                color = color_val
+            else:
+                unexpected_arguments.append(arg)
         elif arg.startswith('-'):
             unexpected_arguments.append(arg)
         else:
             command_name = arg
             break
 
-    return EarlyArgs(help, version, zephyr_base, verbosity, command_name, unexpected_arguments)
+    return EarlyArgs(
+        help, version, zephyr_base, verbosity, color, command_name, unexpected_arguments
+    )
 
 
 class LogFormatter(logging.Formatter):
@@ -222,6 +242,7 @@ class WestApp:
         self.subparser_gen = None  # an add_subparsers() return value
         self.cmd = None  # west.commands.WestCommand, eventually
         self.queued_io = []  # I/O hooks we want self.cmd to do
+        self.color = None  # 'always', 'never', 'auto', or None
 
         for group, classes in BUILTIN_COMMAND_GROUPS.items():
             lst = [cls() for cls in classes]
@@ -253,9 +274,8 @@ class WestApp:
         # Use verbosity to determine west API log levels
         self.setup_west_logging(early_args.verbosity)
 
-        # Makes ANSI color escapes work on Windows, and strips them when
-        # stdout/stderr isn't a terminal
-        colorama.init()
+        # Store color mode for later use
+        self.color = early_args.color
 
         # See if we're in a workspace. It's fine if we're not.
         # Note that this falls back on searching from ZEPHYR_BASE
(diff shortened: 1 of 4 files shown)


TharakaUJ:
I initially considered adding support for a `west config color always` style configuration. However, since the configuration system expects section.key format (e.g., color.ui), I decided not to extend it in that direction.

Also for now 'auto' value just initialize `colorama`  as default (existing behaviour).

Let me know if any changes to be made.
Thank you.

bencefr:
This change would be very welcome with regards to environments like VS Code where executing child processes like `west build` as tasks we cannot cleanly provide a tty attached to get colored output. While environment variables like `FORCE_COLOR=1  CLICOLOR_FORCE=1  TERM=xterm-256color` does help with some underlying tools, they don't help with west's colorama.

marc-hb:
Thanks @TharakaUJ ! Can you please run `uv run poe format` and force push? Please also look at the other CI check. Do NOT spend too much time with cosmetics before review (in case the code has to change), but `uv run poe format` takes approximately zero time to run! Maybe the other CI check is very quick to fix too; I haven't looked at it yet.

TharakaUJ:
Thanks! I ran `uv run poe format` and fixed the lint issues. Force pushed the updated commit.

pdgendt:
@TharakaUJ thanks for the PR, I still need to do a full review, but I wonder if we could test this somehow. Codecov coverage shows that most of the changes isn't even executed during testing.

TharakaUJ:
> @TharakaUJ thanks for the PR, I still need to do a full review, but I wonder if we could test this somehow. Codecov coverage shows that most of the changes isn't even executed during testing.

ok. I will look into adding test cases. Thank you!

pdgendt:
Make sure to run the formatter on the newly added tests, thanks!

src/west/app/main.py
```diff
@@ -739,12 +769,45 @@ def setup_west_logging(self, verbosity):
...
+            # Handle both string values (always/never/auto) and
+            # boolean values (true/false) for backward compatibility
+            if config_value in ('always', 'never', 'auto'):
+                color_mode = config_value
+            elif config_value in ('true', 'True', '1'):
+                color_mode = 'always'
+            elif config_value in ('false', 'False', '0'):
+                color_mode = 'never'
+            else:
+                color_mode = 'auto'
```
pdgendt: Nit, having `'always'`, `'never'` and `'auto'` handled in a separate if is confusing.

```suggestion
            if config_value in ('always', 'true', 'True', '1'):
                color_mode = 'always'
            elif config_value in ('never', 'false', 'False', '0'):
                color_mode = 'never'
            else:
                color_mode = 'auto'
```

pdgendt (approved):

TharakaUJ:
Thanks, applied the suggestion.

marc-hb:
@TharakaUJ please squash and force-push, the zephyrproject cares about git history and does not accept "fixup" commits in it (like https://github.com/zephyrproject-rtos/west/pull/930/changes/147920c23a7a9912faab6ea49ddf9fa67384a2c4)

- https://docs.zephyrproject.org/latest/contribute/contributor_expectations.html#pr-requirements
- https://blog.buenzli.dev/announcing-development-on-flirt/#the-pr-workflow
- https://github.com/zephyrproject-rtos/zephyr/issues/39194
- https://zachholman.com/posts/git-commit-history/ "Utter Disregard for Git Commit History"

TharakaUJ:
Thank you. Squashed the fixup commit.

src/west/app/main.py
```diff
@@ -741,12 +769,43 @@ def setup_west_logging(self, verbosity):
...
         logger.addHandler(LogHandler())
 
+    def init_colorama(self):
+        color_mode = self.color_mode
+
+        if color_mode is None and self.config:
+            config_value = self.config.get('color.ui', 'auto')
+            # Handle both string values (always/never/auto) and
+            # boolean values (true/false) for backward compatibility
+            if config_value in ('always', 'true', 'True', '1'):
```
marc-hb: I don't think we can do that because `color.ui=True` (or 1, etc.) had no effect since forever. Now with this PR it becomes `always`. That could seriously mess up some CI logs and others.

I think it's OK for `True` and `always` to have different meanings.

src/west/app/main.py
```diff
@@ -741,12 +769,43 @@ def setup_west_logging(self, verbosity):
...
+                color_mode = 'never'
+            else:
+                color_mode = 'auto'
+
+        if color_mode == 'always':
+            colorama.init(strip=False)
+        elif color_mode == 'never':
+            colorama.init(strip=True)
+        else:
+            # 'auto' or None: use colorama's default behavior
```
marc-hb:
```suggestion
            # 'auto' or None or "true": defer to colorama's terminal detection
```

src/west/app/main.py
```diff
@@ -741,12 +769,43 @@ def setup_west_logging(self, verbosity):
 
         logger.addHandler(LogHandler())
 
+    def init_colorama(self):
+        color_mode = self.color_mode
+
+        if color_mode is None and self.config:
+            config_value = self.config.get('color.ui', 'auto')
```
marc-hb: We don't want to access `self.config.get('color.ui',...` in two different places in the code, this seems too complicated and too high-maintenance. You can read the config value at a different time and place if there's a good reason for that, but don't add a new place while keeping the old place. 

`grep.color` and other quirks add enough complexity already :-(

marc-hb: Please keep using `getboolean()` and catch the `ValueError` when it's `always` or `never` rather than duplicating the special values from `configparser`.

marc-hb (changes requested):
Sorry I really don't understand why we need this new `cmd._color_override` field and new override function. It should be possible to keep all the logic centralized in a single accessor like `color_ui()` rather than scattered in different places like this, why not?

BTW you can probably remove some of the `if self.color_ui` checks in the code to simplify it now that this PR uses `colorama.init(strip=True)` (some other places will likely still need that check).

Last but not least: make sure you test thoroughly on both Linux and Windows. I started to do some manual testing that I will summarize soon in #651 (EDIT: done) and I found some existing issues even before this PR. Fixing them is out of scope for this PR but we have to be sure this #930 is not making anything worse. Also, #651 provides a good list of manual tests.

TharakaUJ:
Thank you the detailed feedback. I will look into necessary changes.

TharakaUJ:
I made some changes and force pushed. I thought a early review might helpful to stay on needed track.
also, unfortunately I won't be able to test this on windows until Monday.

TharakaUJ:
> Sorry I really don't understand why we need this new `cmd._color_override` field and new override function. It should be possible to keep all the logic centralized in a single accessor like `color_ui()` rather than scattered in different places like this, why not?

Got rid of most of the _color_override logic. But still could not centralize all the logic, mainly because I'm not sure whether calling `calorama.init()` in `westCommand` would initialize it more than once. I will look into it.

marc-hb:
> Got rid of most of the _color_override logic. 

Thanks and thanks for the heads-up.

> But still could not centralize all the logic, mainly because I'm not sure whether calling calorama.init() in westCommand would initialize it more than once. I will look into it.

Why would we need `colorama.init()` is WestCommands? Thanks to your new `colorama.init(strip=True)`, we should be able to have LESS color-related code in `WestCommands`. If not in this PR, at least in the future. But for sure no additional color-related code in `WestCommands`, or why? Colorization is global.

Ideally, the only color-related code in WestCommands should be decisions like "how important is this log message?" and never "What does the user/configuration/terminal want?".

OK, now I just realized we haven't looked at colorization from a west API perspective... no idea what is the colorization status there :-(

PR #936, opened 2026-03-26
Fix "west init" failure to detect reinitialization in some cases

marc-hb:
Fixes #935, see commit messages and #935 for details.

cc: @thorsten-klein

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -306,6 +306,13 @@ below.
         #
         # https://docs.python.org/3/library/pathlib.html#pathlib.PurePath.parent
         manifest_dir = Path(args.directory or os.getcwd()).resolve()
+
+        try:
+            already = util.west_topdir(manifest_dir, fall_back=False)
+            self.die_already(already)
+        except util.WestNotFound:
+            pass
+
         manifest_filename = args.manifest_file or 'west.yml'
         manifest_file = manifest_dir / manifest_filename
         topdir = manifest_dir.parent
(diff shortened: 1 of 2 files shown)


src/west/app/project.py
```diff
@@ -410,6 +417,15 @@ def bootstrap(self, args) -> Path:
...
         manifest_abspath = topdir / manifest_path
 
+        # We already looked for old .west/ pollution earlier but only at the highest
+        # topdir/.west level because we didn't know the manifest_path yet. Now that we know
+        # "manifest_path", look again for any existing, clashing .west/ located at a lower
+        # level somewhere between the new topdir and the new manifest_path. This check will be
+        # especially useful once new --topdir feature is available, see
+        # https://github.com/zephyrproject-rtos/west/issues/774
+        already = util.west_topdir(manifest_abspath, fall_back=False)
+        topdir.samefile(already) or self.die_already(already)
```
pdgendt: Nit: prefer to use an actual `if` statement

```suggestion
        if not topdir.samefile(already):
            self.die_already(already)
```

pdgendt (approved):

src/west/app/project.py
marc-hb: Done. I like how `check() or die()` reads like plain English but I admit it's not a common idiom outside some languages.

carlescufi (approved):

carlescufi:
@thorsten-klein FYI

Issue #941, opened 2026-04-08
Remove the comma seperator when west update failed for multiple projects


yvesll:
When west update fails for multiple projects, the error message lists projects separated by commas:
```bash
ERROR: update failed for projects: project1, project2, project3
```
But west update accepts multiple projects separated by spaces. As a result, the error output is not copy‑paste friendly for retrying the failed projects.

It would be more user-friendly if the output matched the input format of west update.

marc-hb:
This output is not designed to be re-used as input and `west update` is not expected to fail regularly. When `west update` fails, the user is expected to spend time scrolling up and looking at the logs to find what went wrong. The user is not expected to blindly retry the failed repos with a copy/paste as fast as possible.

Even in the "unsupported" case where you are regularly using a regularly flaky server and/or network (!?!), you can re-run your previous `west update` command unchanged and it should be very quick on repos that are already up-to-date.

Dropping the commas would make the list looking more like a sentence and slower to parse by humans.

yvesll:
Thanks for the explanation. Let me clarify a concrete use case from my workflow.

In my case, `west update` failures are often caused by uncommitted local changes in a small subset of repos. I already know these changes are disposable, so my typical flow is:

1. Identify the failed projects from the error output.
2. Use `west forall` to clean/reset only those repos.
3. Re-run `west update` on the failed projects only.

This avoids re-walking all projects again, which can be slow in large workspaces (140+ projects) and unnecessary when the failure scope is already known. The failed-project list is not used blindly, but after reviewing logs and understanding the failure reason.

I agree this is a very small QoL request, and it’s perfectly fine if it’s not something the project wants to support.

marc-hb:
Thanks for the details on the use case, I think I understand better now.

I think there should be a more generic, safer and generally better way to address it: something like a `west compare --format {name}` command that prints the names and only the names of repos with changes in them. Would that work for you? I can approximate it with these inconvenient but already existing commands: `west -q forall -c 'git status -s | grep -q . && printf "${WEST_PROJECT_NAME}\n"'` or `west compare | awk '/^===/ { print $2 }'`. Please give those a try, see if they help?

- Generally speaking, I think command output should be either meant for humans or meant for machines. Trying to make compatible with both makes it good for neither IMHO. This is why `west list` has a `--format` option already.
- I would much prefer a predictable and reliable command like `west compare --format {name}` rather than a "try it and see what failed" approach. That seems very flaky.

Thoughts?

[...]

PR #946, opened 2026-04-22
project: compare: add -f/--format for machine-readable output

yvesll:
Mirror "west list --format" on "west compare": when --format is given, print one line per dirty project using the format string and skip the banner/status output. Supports the same keys as "west list" ({name}, {url}, {path}, {abspath}, {posixpath}, {revision}, {sha}, {groups}), including ManifestProject special-casing.

Unlike default compare output (documented as unstable), --format output is a stable interface suitable for scripting, e.g.:

    west update $(west compare --format '{name}')

https://github.com/zephyrproject-rtos/west/issues/941

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -863,13 +863,38 @@ class Compare(_ProjectCommand):
             The command also prints output for the manifest repository if it
             has nonempty status.
 
-            The output is meant to be human-readable, and may change. It is
-            not a stable interface to write scripts against. This command
-            requires git 2.22 or later.'''),
+            The default output is meant to be human-readable, and may change.
+            It is not a stable interface to write scripts against. If you need
+            machine-readable output (e.g. to feed back into "west update" or
+            "west forall"), pass --format; see FORMAT STRINGS below.
+
+            This command requires git 2.22 or later.'''),
         )
 
     def do_add_parser(self, parser_adder):
-        parser = self._parser(parser_adder, epilog=ACTIVE_CLONED_PROJECTS_HELP)
+        parser = self._parser(
+            parser_adder,
+            epilog=f'''\
+{ACTIVE_CLONED_PROJECTS_HELP}
+
+FORMAT STRINGS
+--------------
+
+When --format is given, "west compare" prints one line per project
+with changes, rendered using the given Python format string. The
+supported keys are the same ones documented in "west list --help".
+Unlike the default output, --format output is a stable interface
+suitable for scripting, for example:
+
+POSIX:
+
+    west compare --format '{{name}}' | while read -r name; do west update "$name"; done
+
+PowerShell:
+
+    west compare --format '{{name}}' | ForEach-Object {{ west update $_ }}
+''',
+        )
         parser.add_argument(
             'projects',
             metavar='PROJECT',
@@ -900,6 +925,13 @@ class Compare(_ProjectCommand):
                     or any compare.ignore-branches configuration
                     option''',
         )
+        parser.add_argument(
+            '-f',
+            '--format',
+            help='''if given, print one line per project with changes
+                    using this format string (stable, machine-readable
+                    output); see FORMAT STRINGS below''',
+        )
         return parser
 
     def do_run(self, args, ignored):
@@ -915,6 +947,13 @@ class Compare(_ProjectCommand):
 
         failed = []
         printed_output = False
+
+        def compare_project(project, fmt):
+            if fmt is None:
+                self.compare(project)
+            else:
+                self._format_project(project, fmt)
+
         for project in self._cloned_projects(args, only_active=not args.all):
             if isinstance(project, ManifestProject):
                 # West doesn't track the relationship between the manifest
@@ -922,7 +961,7 @@ class Compare(_ProjectCommand):
                 # in printing output for comparisons that makes sense.
                 if self._has_nonempty_status(project):
                     try:
-                        self.compare(project)
+                        compare_project(project, args.format)
                         printed_output = True
                     except subprocess.CalledProcessError:
                         failed.append(project)
@@ -960,7 +999,7 @@ class Compare(_ProjectCommand):
                 ):
                     continue
 
-                self.compare(project)
+                compare_project(project, args.format)
                 printed_output = True
             except subprocess.CalledProcessError:
                 failed.append(project)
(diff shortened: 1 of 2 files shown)


src/west/app/project.py
```diff
@@ -904,14 +942,21 @@ def do_run(self, args, ignored):
 
         failed = []
         printed_output = False
+
+        def compare_project(project):
```
marc-hb: The least "global" variable use, the better
```suggestion
        def compare_project(project, fmt):
            if fmt is None
```

src/west/app/project.py
```diff
@@ -963,6 +1008,42 @@ def compare(self, project):
         self.print_rev_info(project)
         self.print_status(project)
 
+    def compare_format(self, project, fmt):
```
```suggestion
    def format_project(self, project, fmt):
```
This function is not related to `compare` any more. In fact it's mostly a duplicate of code in `List.do_run()` that does the same thing with only small differences. Please avoid copy/paste/diverge and extract the code in `List.do_run()` into a new `format_project()` function in a preliminary, zero-functional change commit. Then you can use that new function in `west compare` in a second commit.

I can see some minor differences like for instance `args.manifest_path_from_yaml`: these can be passed as arguments to the new `format_project()` function if needed.

src/west/app/project.py
```diff
@@ -852,13 +852,43 @@ def __init__(self):
...
+- name: project name in the manifest
+- url: full remote URL as specified by the manifest
+- path: the relative path to the project from the top level,
+  as specified in the manifest where applicable
+- abspath: absolute and normalized path to the project
+- posixpath: like abspath, but in posix style, that is, with '/'
+  as the separator character instead of '\\'
+- revision: project's revision as it appears in the manifest
+- sha: project's HEAD revision as a SHA
+- groups: project groups, as a comma-separated list
```
marc-hb: This is mostly a copy of `west list -h`. Better to refer to it instead, mentioning minor usage differences if needed.

src/west/app/project.py
```diff
@@ -852,13 +852,43 @@ def __init__(self):
...
+            epilog=f'''\
+{ACTIVE_CLONED_PROJECTS_HELP}
+
+FORMAT STRINGS
+--------------
+
+When --format is given, "west compare" prints one line per project with
+changes. Arguments to the format string are accessed by name. Unlike the
+default output, --format output is a stable interface suitable for scripting
+(for example: "west update $(west compare --format '{{name}}')").
```
marc-hb: Special characters in project names probably don't work anyway, but just to be on the safer side and to demonstrate safe shell practices:

```suggestion
(for example: west compare --format '{{name}}' | while read -r ; do west update "$REPLY"; done ).
```

Yes this is slightly slower but it's just an example.

tests/test_project.py
```diff
@@ -628,6 +628,107 @@ def test_compare(config_tmpdir, west_init_tmpdir):
...
+
+    # --exit-code must still fire when --format produced any output.
+    with pytest.raises(SystemExit):
+        cmd('compare --exit-code --format {name}')
+
+    # Cleanup and confirm empty output again.
+    os.unlink(foo)
+    assert cmd('compare --format {name}') == ''
+
+    # Now exercise a real (non-manifest) project.
```
marc-hb: This test is very long, can you please split it into 2 or 3 different tests? Also, they can now run in parallel with pytest-xdist and `pytest -n auto -k compare`

tests/test_project.py
```diff
@@ -628,6 +628,107 @@ def test_compare(config_tmpdir, west_init_tmpdir):
     assert 'mybranch' in cmd('compare --no-ignore-branches')
 
 
+def test_compare_format(config_tmpdir, west_init_tmpdir):
```
marc-hb: Thanks for the thorough test coverage!

src/west/app/project.py
yvesll: Added a share function to format_project

src/west/app/project.py
yvesll: updated

src/west/app/project.py
yvesll: Thanks for the info. I have added cmd demo for POSIX and windows when name contains special characters.

src/west/app/project.py
yvesll: Removed and point to `west list -h`

tests/test_project.py
yvesll: Split into 3 test functions.

src/west/app/project.py
```diff
@@ -863,13 +863,47 @@ def __init__(self):
...
+default output, --format output is a stable interface suitable for
+scripting.
+
+For typical manifests where project names are plain identifiers
+(letters, digits, '_', '-', '.'), command substitution works in
+both POSIX shells and PowerShell:
+
+    west update $(west compare --format '{{name}}')
+
+If names may contain characters that a POSIX shell would interpret
```
marc-hb: Please just give one, safe example for unix and one for powershell and don't talk about special characters. These --help texts tend to be too long already, we don't want them to slowly turn into shell scripting guides. There are plenty of such guides on the internet already. To be honest, we could even do with zero example because the purpose of `--format` is already explained in plenty of detail and `--format` is also a very common thing to do but OK, one brief example is always nice.

marc-hb (commented):
Please trim down the `--help`, otherwise everything looks great! Thank you.

src/west/app/project.py
yvesll: Trimmed, thanks.

src/west/app/project.py
```diff
@@ -149,6 +149,86 @@ def has_output():
 
         return has_output()
 
+    def format_project(self, project, fmt, *, manifest_path_from_yaml=False):
```
pdgendt: There's a naming convention used here, that "private" functions are prefixed with an `_` underscore. We should probably apply that here?

src/west/app/project.py
```diff
@@ -900,6 +925,14 @@ def do_add_parser(self, parser_adder):
                     or any compare.ignore-branches configuration
                     option''',
         )
+        parser.add_argument(
+            '-f',
+            '--format',
+            default=None,
```
pdgendt: This already is the default `default`, omit.

pdgendt (commented):
Thanks! Looks good otherwise.

yvesll:
> Thanks! Looks good otherwise.

Thanks for the review! All resolved.

pdgendt (approved):

marc-hb (approved):
Great work, thank you!

PR #953, opened 2026-05-22
support `west init --topdir`

### LINKED ISSUE #774: `west init` gives wrong config `path` for a manifest file in nested directories

thorsten-klein:
Fixes https://github.com/zephyrproject-rtos/west/issues/774 (Successor of https://github.com/zephyrproject-rtos/west/pull/775 EDIT: and #854)

This PR introduces a new --topdir option for west init, allowing users to explicitly specify the west workspace directory (topdir) during initialization.

This is particularly useful when using --local, as the current behavior always sets the parent of the given directory as the topdir. In addition, west is currently unable to determine both manifest.path and manifest.file from a single directory argument as the path can be interpreted in multiple ways (e.g. it can be split at various places).

With the new --topdir argument, users can now directly control where the west workspace is created.

Parameterized tests for west init --topdir are added as well.

Moreover, the help texts and descriptions are (hopefully) simplified.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -24,10 +24,8 @@ from urllib.parse import urlparse
 from west import util
 from west.commands import CommandError, Verbosity, WestCommand
 from west.configuration import Configuration
-from west.manifest import MANIFEST_REV_BRANCH as MANIFEST_REV
-from west.manifest import QUAL_MANIFEST_REV_BRANCH as QUAL_MANIFEST_REV
-from west.manifest import QUAL_REFS_WEST as QUAL_REFS
 from west.manifest import (
+    _WEST_YML,
     ImportFlag,
     Manifest,
     ManifestImportFailed,
@@ -35,6 +33,9 @@ from west.manifest import (
     Submodule,
     _manifest_content_at,
 )
+from west.manifest import MANIFEST_REV_BRANCH as MANIFEST_REV
+from west.manifest import QUAL_MANIFEST_REV_BRANCH as QUAL_MANIFEST_REV
+from west.manifest import QUAL_REFS_WEST as QUAL_REFS
 from west.manifest import is_group as is_project_group
 from west.util import expand_path
 
@@ -236,19 +237,82 @@ class Init(_ProjectCommand):
             'init',
             'create a west workspace',
             f'''\
-Creates a west workspace.
+Initialize a west workspace (topdir) from a west manifest (`west.yml`) by
+creating a `.west` directory in the topdir and a local configuration file
+`.west/config`.
+The West manifest can come from either a git repository (that will be cloned
+during workspace initialization) or from an already existing local directory.
 
-With -l, creates a workspace around an existing local repository;
-without -l, creates a workspace by cloning a manifest repository
-by URL.
+West only reads the manifest data from the manifest tree.
+In detail, this is the manifest file and any files it imports. Everything else
+in that tree is not used by west at all (including git history or branches).
 
-With -m, clones the repository at that URL and uses it as the
-manifest repository. If --mr is not given, the remote's default
-branch will be used, if it exists.
+Common arguments
+----------------
+In a nutshell:
 
-With neither, -m {MANIFEST_URL_DEFAULT} is assumed.
+    topdir/manifest.path/manifest.file
 
-Warning: 'west init' renames and/or deletes temporary files inside the
+Each component can have subdirectories, even the last one despite its name
+`manifest.file`.
+
+-t / --topdir WORKSPACE_DIR
+    The workspace directory; `.west` is created inside it. Defaults are
+    described per mode below.
+
+--mp / --manifest-path SUBPATH
+    Path, relative to the workspace, where the manifest tree
+    lives within the workspace (sets the `manifest.path` config option).
+    In remote mode: where the manifest repository is cloned.
+    In local mode: a newer and clearer alternative to the positional argument.
+
+--mf / --manifest-file FILE
+    Path of the manifest file inside the manifest directory (sets the
+    `manifest.file` config option); relative to `manifest.path` and may
+    include subdirectories. Defaults to '{_WEST_YML}'.
+
+
+1. Using a Manifest Repository (default)
+----------------------------------------
+West clones the given repository (provided via `-m / --manifest-url`).
+Note, that the repository must contain a west manifest.
+
+If no `-m / --manifest-url` is provided, west uses the Zephyr URL by default:
+  {MANIFEST_URL_DEFAULT}.
+
+The topdir (where `.west` is created) is determined as follows:
+- argument `-t / --topdir` if provided
+- otherwise: the positional argument `directory` if provided (deprecated;
+  prefer `-t/--topdir`)
+- otherwise: the current working directory
+
+The positional argument `directory` and `-t / --topdir` cannot be combined
+in cloning mode; use `--manifest-path SUBPATH` instead to control where
+within the workspace the manifest repository is cloned (defaults to the
+manifest's `self: path:` if declared, otherwise the manifest URL's basename,
+e.g. <topdir>/zephyr).
+
+With `--mr`, the revision (branch, tag, or sha) of the repository can be
+specified that will be used. It defaults to the repository's default branch.
+
+2. Using a Local Manifest
+-------------------------
+If `-l / --local` is given, west initializes a workspace from an already
+present manifest tree.
+
+The topdir (where `.west` is created) is determined as follows:
+- argument `-t / --topdir` if provided
+- otherwise (legacy form): `manifest_directory`'s parent, or the current
+  working directory if neither the positional nor `--manifest-path` is given
+
+The manifest tree location can be specified either as the positional
+argument `manifest_directory` (a path to the manifest directory) or via the
+newer `--manifest-path` (a subdirectory of the workspace). The two forms
+cannot be combined.
+
+Known Issues
+------------
+'west init' renames and/or deletes temporary files inside the
 workspace being created. This fails on some filesystems when some
 development tool or any other program is trying to read/index these
 temporary files at the same time. For instance, it is required to stop
@@ -273,9 +337,12 @@ below.
         parser = self._parser(
             parser_adder,
             usage='''
+  init with a repository:
+  %(prog)s [-m URL] [--mr MF_REVISION] [--mf FILE] [--mp SUBPATH]
+                [-o=GIT_CLONE_OPTION] [-t WORKSPACE_DIR] [directory]
 
-  %(prog)s [-m URL] [--mr REVISION] [--mf FILE] [-o=GIT_CLONE_OPTION] [directory]
-  %(prog)s -l [--mf FILE] directory
+  init with an existing local manifest:
+  %(prog)s -l [-t WORKSPACE_DIR] [--mf FILE] [--mp SUBPATH] [manifest_directory]
 ''',
         )
 
@@ -284,6 +351,7 @@ below.
         parser.add_argument(
             '-m',
             '--manifest-url',
+            metavar='URL',
             help='''manifest repository URL to clone;
                     cannot be combined with -l''',
         )
@@ -292,6 +360,7 @@ below.
             '--clone-opt',
             action='append',
             default=[],
+            metavar='GIT_CLONE_OPTION',
             help='''additional option to pass to 'git clone'
                     (e.g. '-o=--depth=1'); may be given more than once;
                     cannot be combined with -l''',
@@ -300,21 +369,53 @@ below.
             '--mr',
             '--manifest-rev',
             dest='manifest_rev',
+            metavar='MF_REVISION',
             help='''manifest repository branch or tag name
                     to check out first; cannot be combined with -l''',
-        )
-        parser.add_argument(
-            '--mf', '--manifest-file', dest='manifest_file', help='manifest file name to use'
         )
         parser.add_argument(
             '-l',
             '--local',
             action='store_true',
-            help='''use "directory" as an existing local
-                    manifest repository instead of cloning one from
-                    MANIFEST_URL; .west is created next to "directory"
-                    in this case, and manifest.path points at
-                    "directory"''',
+            help='''initialize workspace from an already existing local
+                    manifest instead of cloning a git repository;
+                    cannot be combined with -m''',
+        )
+        parser.add_argument(
+            '--mf',
+            '--manifest-file',
+            dest='manifest_file',
+            metavar='FILE',
+            help=f'''manifest file name (sets the manifest.file config
+                     option); the relative path of the manifest file
+                     within the manifest directory.
+                     Defaults to {_WEST_YML}.''',
+        )
+        parser.add_argument(
+            '--mp',
+            '--manifest-path',
+            dest='manifest_path',
+            metavar='SUBPATH',
+            help='''path of the manifest repository within the workspace
+                    (sets the manifest.path config option); SUBPATH must be
+                    relative and is interpreted relative to the workspace
+                    root (-t/--topdir, the positional directory in remote
+                    mode, or cwd, in that order); in remote mode this is
+                    where the manifest repository is cloned and defaults to
+                    the manifest's "self: path:" if declared, otherwise the
+                    manifest URL's basename (e.g. "zephyr"); in local mode
+                    this is an alternative to the positional argument and
+                    cannot be combined with it''',
+        )
+        parser.add_argument(
+            '-t',
+            '--topdir',
+            dest='topdir',
+            metavar='WORKSPACE_DIR',
+            help='''the workspace directory where .west directory
+                    will be created (WORKSPACE_DIR/.west);
+                    the workspace directory may already exist
+                    and WORKSPACE_DIR/.west must not exist''',
         )
         parser.add_argument(
             '--rename-delay',
@@ -330,9 +431,13 @@ below.
             'directory',
             nargs='?',
             default=None,
-            help='''with -l, the path to the local manifest repository;
-            without it, the directory to create the workspace in (defaulting
-            to the current working directory in this case)''',
+            metavar='directory',
+            help='''With --local: the path to a local directory which contains
+                    a west manifest (west.yml); cannot be combined with
+                    --manifest-path.
+                    Without --local (remote mode): the workspace directory
+                    where .west will be created. This form is deprecated;
+                    prefer -t/--topdir. It cannot be combined with -t.''',
         )
 
         return parser
@@ -364,6 +469,9 @@ below.
         if args.manifest_file is not None and Path(args.manifest_file).is_absolute():
             self.die(f'--mf {args.manifest_file} argument is not relative')
 
+        if args.manifest_path is not None and Path(args.manifest_path).is_absolute():
+            self.die(f'--mp {args.manifest_path} argument is not relative')
+
         self.die_if_no_git()
 
         if args.local:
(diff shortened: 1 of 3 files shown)


pdgendt:
This PR has been long overdue but I'm a bit on the fence. I wanted to throw a last brain dump before going with this proposal.

- A concern is that `directory` gains a third meaning when `-t` and positional are both set in remote mode: "where to clone the manifest within the workspace."
- What `--topdir` brings to the table is fine, but it disconnects the local and remote options further, giving more permutations of arguments.

I think we should avoid the third meaning mentioned earlier, and add an explicit `--manifest-path SUBPATH` argument to make it unambiguous.

What this looks like at the call site:
```
  # Local mode
  west init -l                          # cwd is the manifest
  west init -l my-manifest              # ./my-manifest is the manifest
  west init -l -t /ws my-manifest       # explicit workspace override

  # Remote mode
  west init                             # default URL, workspace = cwd (removed by follow-up PR)
  west init -m URL                      # workspace = cwd
  west init -m URL my-workspace         # workspace = ./my-workspace (deprecated form; emit warning?)
  west init -m URL -t /ws               # workspace = /ws (canonical)

  # Special case: clone manifest into a subpath of the workspace
  west init -m URL -t /ws --manifest-path zephyr/repos/manifest
```

Pros:
  - Eliminates positional directory's third meaning. For the current proposal, `west init -m URL -t WORKSPACE DIR` means "DIR is where the manifest is cloned within the workspace." This part has the highest cognitive load.
  - Shrinks the validation surface. You now have to validate "manifest is inside topdir" because the positional can be a sibling. Under this rule, `--manifest-path` is always a relative subpath of `-t`, validated as a path-clean check; no "is X relative to Y" geometry needed.
  - Cuts the test matrix. Some ~13 valid permutations are mostly the cross-product of overloaded meanings. The unified rule produces ~6: each mode × {default topdir / explicit--t} × {default manifest path / explicit---manifest-path}.
  - Smaller help text. Reads "positional = manifest with -l, workspace without; -t overrides workspace; --manifest-path overrides clone subpath." Three rules, all single-meaning.

Example for `argparse`:
```python
    parser.add_argument(
        '--mp',
        '--manifest-path',
        dest='manifest_path',
        metavar='SUBPATH',
        help='''with -m: the path within the workspace where the
                manifest repository is cloned (i.e. sets the
                manifest.path config option);
                SUBPATH must be relative and is interpreted relative
                to the workspace root (--topdir, the positional
                directory, or cwd, in that order);
                defaults to the manifest URL's basename''',
  )
```

  - `manifest.path` config key it writes is the same one the resolved manifest already uses everywhere -> no new schema.
  - Doesn't take the place of `--mf`: `--manifest-path` is the directory, `--mf` stays the filename within it. Together they cover the same surface as `--mf zephyr/west.yml` would in the current PR, with cleaner semantics (each flag answers exactly one question).
  - The "defaults to the manifest URL's basename" line preserves the existing behaviour from `bootstrap()` where `manifest_path = PurePath(urlparse(manifest_url).path).name` is the fallback — so users who don't pass `--manifest-path` get exactly what they get today.

EDIT: we could allow `--manifest-path SUBPATH` with local mode as an alternative to the positional argument; it would have the same meaning in any mode (set manifest.path).

thorsten-klein:
Sounds valid to me.

> west init -l -t /ws my-manifest

Any reason why not using `-l <local dir>`?
This would result in `west init -l my-manifest -t /ws`, which feels much more convenient to me as it is analog to remote init `-r url`.

pdgendt:
> Sounds valid to me.
> 
> > west init -l -t /ws my-manifest
> 
> Any reason why not using `-l <local dir>`? This would result in `west init -l my-manifest -t /ws`, which feels much more convenient to me as it is analog to remote init `-r url`.

I think the `--local`/`-l` should be a simple "enable local mode", where the _positional argument_ would change behavior. That is what we currently also have (you could be doing `west init my_repo -l` currently).

I very much like the idea of having both,
`--manifest-file FILENAME` -> `manifest.file` config setting
`--manifest-path SUBPATH` -> `manifest.path` config setting

The positional `directory` argument in local mode just becomes optional, and has the same meaning as manifest path.

src/west/app/project.py
```diff
@@ -236,19 +237,58 @@ def __init__(self):
...
-by URL.
-
-With -m, clones the repository at that URL and uses it as the
-manifest repository. If --mr is not given, the remote's default
-branch will be used, if it exists.
+Initialize a west workspace (topdir) from a west manifest (`west.yml`) by
+creating a `.west` directory in the topdir and a local configuration file
+`.west/config`.
+The West manifest can come from either a git repository (that will be cloned
+during workspace initialization) or from an already existing local directory.
```
marc-hb, line 244: I would like to add something like:
> In either case, `west` only cares about what manifest-related data in the manifest repository and ignores the rest. Among others, `west` ignores version control of the manifest repository (a `west` manifest never version controls itself)

Better phrasing welcome.

marc-hb (commented):
> I think we should avoid the third meaning mentioned earlier, and add an explicit --manifest-path SUBPATH argument to make it unambiguous.

Like all great solutions, this looks obvious with hindsight.

We discussed this issue in https://github.com/zephyrproject-rtos/west/pull/854#discussion_r2688877590 but didn't think of this - too obsessed with "backwards-compatibility".

Backwards compatibility is very important and I think one of the main reasons this new `--topdir` is taking so long: we really don't want to get the new UI wrong and get stuck with it. But I forgot there that we do NOT need some "fake" backwards-compatibility when people use the new `--topdir` parameter: cause no one has ever used --topdir before. So we can perfectly say "the confusing positional argument is not allowed anymore when you use --topdir" which indeed removes a big chunk of the combinatorial explosion of parameters. Lovely! This is the other UI mistake I made in  https://github.com/zephyrproject-rtos/west/pull/854#discussion_r2688877590: trying to keep the command-line options  independent of each other and all combinations valid. We absolutely don't need all combinations to be valid, quite the opposite. We only don't want to break existing usage (yet). "All input is evil until proven otherwise" - especially command line input ;-)


BTW please tell confirm I understood this above correctly @pdgendt, otherwise the yet other headache caused by this new feature will really require me to see a doctor :-D


> EDIT: we could allow --manifest-path SUBPATH with local mode as an alternative to the positional argument; it would have the same meaning in any mode (set manifest.path).

Agreed!

---------

I don't think we have yet clarified what relative parameters are relative to, which was hopefully the last UI question left open in https://github.com/zephyrproject-rtos/west/pull/854#discussion_r2700543305. But the --manifest-path topic has higher "precedence".

marc-hb:
> I think the --local/-l should be a simple "enable local mode", where the positional argument would change behavior. That is what we currently also have (you could be doing west init my_repo -l currently).

Agreed. We can make the brand new `--topdir` incompatible with the positional argument because it's extremely frequent to have combination of inputs incompatible with each other. For instance: -m and -l are already incompatible with each other. Granted, the incompatibility in this last example is much more obvious but still not major UI surprise when hitting some invalid combination; this happens all the time.

On the other hand, having a _variable_ meaning and/or syntax for the  `--local` option would be extremely unusual and confusing.  I'm not surprised that when looking at `west init --local $local_dir`, people sometimes wrongly and unfortunately assume that `$local_dir` is for `--local`. Yet this has never been true and changing that (either across west versions or based on other parameters) would **increase** that confusion massively. It's simply too late now. `--local` should not be changed.

BTW, banning the positional argument as soon as `--topdir` is used reduces the "attack surface" of this `--local $local_dir` interpretation mistake. Cause you can't do it anymore.



At this point I can't resist duplicating the quote that I already made in [#854 ](https://github.com/zephyrproject-rtos/west/pull/854#discussion_r2830476621)

 https://interrupt.memfault.com/blog/practical_zephyr_west

> Notice that we no longer specify the current directory in the call to west init using “.”. In fact, the directory - optionally passed as the last argument to west init - is interpreted differently by West when using the --local flag or the -m arguments:
> 
> With --local, the directory specifies the path to the local manifest repository.
> Without the --local flag, the directory refers to the topdir and thus the folder in which to create the workspace (defaulting to the current working directory in this case).
> Awkward, but this approach is probably used due to legacy reasons. If no --local flag is used and no repository is specified using -m, West defaults to using the Zephyr repository.

See: that positional argument has always been evil :-)

> I very much like the idea of having both,
> --manifest-file FILENAME -> manifest.file config setting
> --manifest-path SUBPATH -> manifest.path config setting

Yes, this is crystal-clear - unlike other bits of the `west init` UI.

marc-hb:
@thorsten-klein do you prefer to make the UI change suggested above before my conflicting #954? #954 and the conflicts are both very small.

src/west/app/project.py
```diff
@@ -236,19 +237,72 @@ def __init__(self):
...
+The topdir (where `.west` is created) is determined as follows:
+- argument `-t / --topdir` if provided
+- otherwise: the positional argument `directory` if provided (deprecated;
+  prefer `-t/--topdir`)
+- otherwise: the current working directory
+
+The positional argument `directory` and `-t / --topdir` cannot be combined
+in remote mode; use `--manifest-path SUBPATH` instead to control where
+within the workspace the manifest repository is cloned (defaults to the
+manifest URL's basename, e.g. <topdir>/zephyr).
```
pdgendt: I think `self.path` in yaml has precedence over the URL's basename.

src/west/app/project.py
```diff
@@ -236,19 +237,72 @@ def __init__(self):
...
+
+With `--mr`, the revision (branch, tag, or sha) of the repository can be
+specified that will be used. It defaults to the repository's default branch.
+
+2. Using a Local Manifest
+-------------------------
+If `-l / --local` is given, west initializes a workspace from an already
+existing `west.yml`. Its location can be specified either as the positional
```
pdgendt: Don't use `west.yml` in the description, can be configured.

src/west/app/project.py
```diff
@@ -300,21 +359,52 @@ def do_add_parser(self, parser_adder):
...
             '--local',
             action='store_true',
-            help='''use "directory" as an existing local
-                    manifest repository instead of cloning one from
-                    MANIFEST_URL; .west is created next to "directory"
-                    in this case, and manifest.path points at
-                    "directory"''',
+            help='''initialize workspace from an already existing local
+                    manifest instead of cloning a manifest;
+                    cannot be combined with -m''',
```
pdgendt, line 382:
```suggestion
            help='''initialize workspace from an already existing local
                    manifest instead of cloning a manifest repository;
                    cannot be combined with -m''',
```

src/west/app/project.py
```diff
@@ -300,21 +359,52 @@ def do_add_parser(self, parser_adder):
...
+            '--manifest-path',
+            dest='manifest_path',
+            metavar='SUBPATH',
+            help='''path of the manifest repository within the workspace
+                    (sets the manifest.path config option); SUBPATH must be
+                    relative and is interpreted relative to the workspace
+                    root (-t/--topdir, the positional directory in remote
+                    mode, or cwd, in that order); in remote mode this is
+                    where the manifest repository is cloned and defaults to
+                    the manifest URL's basename (e.g. "zephyr"); in local
```
pdgendt: Ditto comment, `self.path` before URL basename.

pdgendt (commented):
I think this is very good.

pdgendt:
The CI fails on formatting `test_project.py`.

pdgendt:
Windows CI isn't happy

pdgendt:
Squashed the last 2 commits (was a fixup commit)

marc-hb:
User interface design looks great to me. I haven't reviewed the code yet. Before people do, should you squash the 1st and 3rd commit? I can't see any value in keeping them separate, is there? At this point in time, I feel like 2 commits should be enough. Commit 1 = new feature, Commit 2 = new tests. Unless I'm missing a "natural" split between two subfeatures?

I mean the 3rd commit looks like a "Fixup" to me. A very big one but still a fixup.

pdgendt:
> User interface design looks great to me. I haven't reviewed the code yet. Before people do, should you squash the 1st and 3rd commit? I can't see any value in keeping them separate, is there? At this point in time, I feel like 2 commits should be enough. Commit 1 = new feature, Commit 2 = new tests. Unless I'm missing a "natural" split between two subfeatures?
> 
> I mean the 3rd commit looks like a "Fixup" to me. A very big one but still a fixup.

Sure, squashed to feature commits and tests, please review.

pdgendt:
Applied Copilot's valid suggestions.

src/west/app/project.py
```diff
@@ -236,19 +237,73 @@ def __init__(self):
...
-With neither, -m {MANIFEST_URL_DEFAULT} is assumed.
-
-Warning: 'west init' renames and/or deletes temporary files inside the
+Initialize a west workspace (topdir) from a west manifest (`west.yml`) by
+creating a `.west` directory in the topdir and a local configuration file
+`.west/config`.
+The West manifest can come from either a git repository (that will be cloned
+during workspace initialization) or from an already existing local directory.
+
+West only reads the manifest data from the manifest repository or directory.
```
```suggestion
West only reads the manifest data from the manifest tree.
```
... cause west forgets it was a repository immediately after cloning - if cloning at all.

Sorry to insist on this but it has been an FAQ. Understandable confusion probably fueled by:
- the fact that `west init` without `-l` does clone the manifest initially. That gives a wrong impression.
- terminology like this one.

src/west/app/project.py
```diff
@@ -236,19 +237,73 @@ def __init__(self):
...
+history or branches).
+
+Common arguments
+----------------
+-t / --topdir WORKSPACE_DIR
+    The workspace directory; `.west` is created inside it. Defaults are
+    described per mode below.
+
+--mp / --manifest-path SUBPATH
+    Path, relative to the workspace, where the manifest repository
```
```suggestion
    Path, relative to the workspace, where the manifest tree
```

src/west/app/project.py
```diff
@@ -236,19 +237,73 @@ def __init__(self):
...
+    lives within the workspace (sets the `manifest.path` config option).
+    In remote mode: where the manifest repository is cloned.
+    In local mode: an alternative to the positional argument.
```
```suggestion
    In local mode: a newer and clearer alternative to the positional argument.
```

src/west/app/project.py
```diff
@@ -236,19 +237,73 @@ def __init__(self):
...
+
+--mf / --manifest-file FILE
+    Path of the manifest file inside the manifest directory (sets the
+    `manifest.file` config option). Defaults to '{_WEST_YML}'.
```
```suggestion
    `manifest.file` config option). Relative to the `manifest.path` parent. Defaults to '{_WEST_YML}'. Can include subdirectories.
```
Not obvious for something named "file".

src/west/app/project.py
```diff
@@ -236,19 +237,73 @@ def __init__(self):
...
+The West manifest can come from either a git repository (that will be cloned
+during workspace initialization) or from an already existing local directory.
+
+West only reads the manifest data from the manifest repository or directory.
+In detail, this is the manifest file and any files it imports. Everything else
+in that repository or directory is not used by west at all (including git
+history or branches).
+
+Common arguments
+----------------
```
marc-hb, line 251:
```suggestion
----------------
In a nutshell:

    topdir/manifest.path/manifest.file

Each component can have subdirectories, even the last one despite its name `manifest.file`.
```

src/west/app/project.py
```diff
@@ -236,19 +237,73 @@ def __init__(self):
...
+  {MANIFEST_URL_DEFAULT}.
+
+The topdir (where `.west` is created) is determined as follows:
+- argument `-t / --topdir` if provided
+- otherwise: the positional argument `directory` if provided (deprecated;
+  prefer `-t/--topdir`)
+- otherwise: the current working directory
+
+The positional argument `directory` and `-t / --topdir` cannot be combined
+in remote mode; use `--manifest-path SUBPATH` instead to control where
```
```suggestion
in cloning mode; use `--manifest-path SUBPATH` instead to control where
```

src/west/app/project.py
```diff
@@ -236,19 +237,73 @@ def __init__(self):
...
+manifest's `self: path:` if declared, otherwise the manifest URL's basename,
+e.g. <topdir>/zephyr).
+
+With `--mr`, the revision (branch, tag, or sha) of the repository can be
+specified that will be used. It defaults to the repository's default branch.
+
+2. Using a Local Manifest
+-------------------------
+If `-l / --local` is given, west initializes a workspace from an already
+existing manifest. Its location can be specified either as the positional
```
```suggestion
present manifest tree. Its location can be specified either as the positional
```

src/west/app/project.py
```diff
@@ -236,19 +237,73 @@ def __init__(self):
...
+argument `manifest_directory` (a path to the manifest directory) or via
```
```suggestion
argument `manifest_directory` (a path to the manifest directory) or via the newer
```

src/west/app/project.py
```diff
@@ -236,19 +237,73 @@ def __init__(self):
...
+`--manifest-path` (a subpath of the workspace). The two forms cannot be
```
```suggestion
`--manifest-path` (a subdirectory of the workspace). The two forms cannot be
```

src/west/app/project.py
```diff
@@ -236,19 +237,73 @@ def __init__(self):
...
+combined.
+
+The topdir (where `.west` is created) is determined as follows:
+- argument `-t / --topdir` if provided
+- otherwise (legacy form): `manifest_directory`'s parent, or the current
+  working directory if neither the positional nor `--manifest-path` is given
```
marc-hb, line 306: Can you please move this topdir section first for "top-down" consistency with other parts of the doc?

src/west/app/project.py
```diff
@@ -300,21 +360,53 @@ def do_add_parser(self, parser_adder):
             '--mr',
             '--manifest-rev',
             dest='manifest_rev',
+            metavar='REVISION',
```
```suggestion
            dest='manifest_rev',
            metavar='MF_REVISION',
```

src/west/app/project.py
```diff
@@ -300,21 +360,53 @@ def do_add_parser(self, parser_adder):
...
             '-l',
             '--local',
             action='store_true',
-            help='''use "directory" as an existing local
-                    manifest repository instead of cloning one from
-                    MANIFEST_URL; .west is created next to "directory"
-                    in this case, and manifest.path points at
-                    "directory"''',
+            help='''initialize workspace from an already existing local
+                    manifest instead of cloning a manifest repository;
```
```suggestion
                    manifest instead of cloning a git repository;
```

src/west/app/project.py
```diff
@@ -380,29 +479,47 @@ def local(self, args) -> Path:
...
-        except util.WestNotFound:
-            pass
+        # Resolve the manifest directory. With --manifest-path, the manifest
+        # location is a subpath of the workspace; without it, the positional
+        # argument (or cwd) is the manifest directory directly. Note we call
+        # resolve() so that 'west init -l .' works as expected -- without it,
+        # Path('.').parent is Path('.') rather than Path('..').
+        if args.manifest_path is not None:
+            topdir = Path(args.topdir or os.getcwd()).resolve()
+            manifest_dir = (topdir / args.manifest_path).resolve()
```
```suggestion
           abs_mf_path = (topdir / args.manifest_path).resolve()
```
"path" is just as vague as "dir", BUT it is consistent with `manifest.path`.

I normally don't like minor, cosmetic changes adding noise to an already big commit. But this particular commit already touches pretty much every line in this function, so this is a great opportunity to make this more consistent and clearer.

Moreover, this big change is all about absolute and relative paths.

src/west/app/project.py
```diff
@@ -380,29 +479,47 @@ def local(self, args) -> Path:
...
+        else:
+            manifest_dir = Path(args.directory or os.getcwd()).resolve()
+            topdir = Path(args.topdir or manifest_dir.parent).resolve()
 
-        manifest_filename = args.manifest_file or 'west.yml'
+        manifest_filename = args.manifest_file or _WEST_YML
```
```suggestion
        manifest_file = args.manifest_file or _WEST_YML
        abs_manifest_file = manifest_dir / manifest_file
```
See rationale above.

src/west/app/project.py
```diff
@@ -413,8 +530,34 @@ def local(self, args) -> Path:
         return topdir
 
     def bootstrap(self, args) -> Path:
-        topdir = Path(abspath(args.directory or os.getcwd()))
-        self.banner('Initializing in', topdir)
+        # Determine the workspace topdir. The positional <directory> in
```
```suggestion
        # Determine the workspace/topdir. The positional <directory> in
```

src/west/app/project.py
```diff
@@ -413,8 +530,34 @@ def local(self, args) -> Path:
...
+        if args.topdir:
+            topdir = Path(args.topdir)
+        elif args.directory:
+            self.wrn(
+                "the positional <directory> argument in remote mode is "
+                "deprecated; use -t/--topdir instead"
+            )
+            topdir = Path(args.directory)
+        else:
+            topdir = Path()
```
```suggestion
            topdir = Path('.')
```

src/west/app/project.py
```diff
@@ -486,14 +629,21 @@ def bootstrap(self, args) -> Path:
         manifest = Manifest.from_data(
             temp_manifest.read_text(encoding=Manifest.encoding), import_flags=ImportFlag.IGNORE
         )
-        if manifest.yaml_path:
-            manifest_path = manifest.yaml_path
+
+        if args.manifest_path:
+            # User-supplied --manifest-path fully overrides the default
+            # (the manifest's "self: path:" or the URL basename).
+            manifest_path = Path(args.manifest_path)
```
marc-hb, line 645: An actual override of `self.path` should print a warning. Some wrote `self.path` for a reason. I have experience with one actual manifest using `self.path` and this override would have broken it.

marc-hb (commented):
I have only style comments left, the UI and logic LGTM.

I have only skimmed the test code but it looks really well designed and easy to read and maintain, so more than good enough.

pdgendt:
> I have only style comments left, the UI and logic LGTM.
> 
> I have only skimmed the test code but it looks really well designed and easy to read and maintain, so more than good enough.

Applied the suggestions, thanks!

pdgendt (approved):

marc-hb (commented):
I am testing this and it seems to break `--local` in some strange ways:

```
west init -l --mp zephyr # fails
FATAL ERROR: -l cannot be combined with -m, -o or --mr

west init -l     zephyr
# still works OK

west init -l     notfound # misleading error message
FATAL ERROR: -l cannot be combined with -m, -o or --mr

west init -l  ~/tmp/zephyr/ # fails
FATAL ERROR: -l cannot be combined with -m, -o or --mr
```

Any idea?

marc-hb (approved):

PR #954, opened 2026-05-24
project: stop defaulting `west init` to upstream Zephyr

marc-hb:
Stop defaulting `west init` to `west init -m
https://github.com/zephyrproject-rtos/zephyr/`. The argument must be passed explicitly.

Thanks to argparse, a bare `west init` now errors like this:

```
usage:

  west init [-m URL] [--mr REVISION] [--mf FILE] [-o=GIT_CLONE_OPTION] [directory]
  west init -l [--mf FILE] directory

west init: error: one of the arguments -m/--manifest-url -l/--local is required
```

This is part of the wider effort to make west independent from Zephyr: https://github.com/zephyrproject-rtos/west/issues/246

This should also help nudge people towards creating their own manifests, which will mitigate Zephyr's "tragedy of the commons" where many people want their stuff to be present the default upstream manifest, while also complaining that the default is way too big and slow (currently around 10 Gigabytes)

- https://github.com/zephyrproject-rtos/zephyr/issues/91061
- https://github.com/zephyrproject-rtos/zephyr/issues/108783#issuecomment-4526275800

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -29,10 +29,12 @@ By default the manifest file is named ``west.yml``.
 You use ``west init`` to set up this directory, then ``west update`` to fetch
 and/or update the repositories named in the manifest.
 
-By default, west uses `upstream Zephyr's manifest file
-<https://github.com/zephyrproject-rtos/zephyr/blob/main/west.yml>`_, but west
-doesn't care if the manifest repository is zephyr or not. You can and are
-encouraged to make your own manifest repositories to meet your needs.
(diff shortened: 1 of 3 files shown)


marc-hb:
Kept as draft because it git-conflicts with #953 which should go first. But the discussion can start.

pdgendt (approved):

fabiobaltieri (approved):

henrikbrixandersen (approved):

Issue #963, opened 2026-06-08
`west init` on Windows fails  when `shutil.move` falls back to copy+rmtree (stdlib `rmtree` doesn't strip read-only bit on git pack files)


swoboda1337:
## Summary

When `west init`'s `shutil.move(tempdir, manifest_abspath)` at `src/west/app/project.py:396` falls back to its `copytree + rmtree` path, the `rmtree(src)` step fails with `PermissionError: [WinError 5]` on the freshly-cloned `manifest-tmp/.git/objects/pack/*.idx` / `*.pack` files. Root cause is a stdlib defect: `shutil.rmtree` does not strip the read-only attribute before `os.unlink`, and git writes pack files read-only.

This is **deterministic** — every git clone produces read-only pack files, so every fallback-path execution hits it. Distinct from #684 (closed as user error, but the traceback was the same bug) and #558 / #756 (intermittent NTFS rename collisions, partially mitigated).

Additionally, the surrounding `except shutil.Error` handler at line 397 does not catch `PermissionError` (which is `OSError`, not `shutil.Error`), so the failure escapes `self.die()` and surfaces as a raw traceback.

## Reproduction

Windows 11, Python 3.13, west 1.5.0:

```
python -m west init -m https://github.com/<any-repo>.git C:\path\to\new\workspace
```

Any condition that prevents the initial `os.rename` inside `shutil.move` from succeeding (AV/indexer briefly holding a handle, etc.) routes execution into the copy+rmtree fallback, and the rmtree then crashes.

## Traceback

```
File ".../west/app/project.py", line 396, in bootstrap
  shutil.move(os.fspath(tempdir), os.fspath(manifest_abspath))
File ".../Python313/Lib/shutil.py", line 874, in move
  rmtree(src)
File ".../Python313/Lib/shutil.py", line 790, in rmtree
  return _rmtree_unsafe(path, onexc)
File ".../Python313/Lib/shutil.py", line 629, in _rmtree_unsafe
  onexc(os.unlink, fullname, err)
File ".../Python313/Lib/shutil.py", line 625, in _rmtree_unsafe
  os.unlink(fullname)
PermissionError: [WinError 5] Access is denied:
  '...\manifest-tmp\.git\objects\pack\pack-<sha>.idx'
```

## Root cause

1. Git writes pack `.idx` / `.pack` files with the read-only attribute set (intentional — they're immutable once written).
[...]

PR #964, opened 2026-06-08
project: use os.rename() instead of shutil.move() for cloned manifest

marc-hb:
As reported in #963 (thanks!) and seen in the source of https://github.com/python/cpython/blob/main/Lib/shutil.py, `shutil.move()` tries an (atomic) `os.rename()` first, then falls back on a copy+delete in case of an OSError raised by `os.rename()`. The source and destination being on different filesystems is one of the (too many?) conditions caught by this OSError and how `shutil.move()` claims to "optimize" moves on the same filesystem thanks to `os.rename()`. Maybe there isn't another, simple and reliable way to check whether source and destionation are on the same filesystem and perform this optimization?

In this particular `project.py` use case however, we _know_ that both are on the same filesystem because we purposely located them in the same directory so we can benefit from very fast and atomic renames guaranteed by practically all filesystems in the universe. So, let's simplify `shutil.move()` down to `os.rename()` in this particular case. Quoting the same `Lib/shutil.py#move()` source:

    os.rename() is preferably used if the source and destination are on the
    same filesystem. [...]
    It's recommended to use os.rename() if atomic move is strictly required.

But wait... if we know we are always on the same filesystem, then why does this even matter? os.rename() should always succeed inside shutil.move() and this minor cleanup should never make any visible difference, right? To answer that we need to go back to #963, NTFS and antivirus scanners which when combined together can break even the simplest things like renaming a directory. While this simplification is far from making west compatible with such utterly broken systems, it does remove a lot of complexity and headaches when such a failure happens and it recommends the existing --rename-delay workaround.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -516,18 +516,16 @@ below.
 
         self.dbg('moving', tempdir, 'to', manifest_abspath, level=Verbosity.DBG_EXTREME)
 
-        # As shutil.move() is used to relocate tempdir, if manifest_abspath
-        # is an existing directory, tmpdir will be moved _inside_ it, instead
-        # of _to_ that path - this must be avoided. If manifest_abspath exists
-        # but is not a directory, then semantics depend on os.rename(), so
-        # avoid that too...
+        # If manifest_abspath exists but is not a directory, then semantics
+        # depend on os.rename(), so avoid that
         if manifest_abspath.exists():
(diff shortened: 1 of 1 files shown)


swoboda1337 (approved):
Yeah look good to me.

pdgendt (approved):

kartben (approved):

PR #965, opened 2026-06-09
tests: manifest: Add parse/yaml roundtrip test for nested west-commands

pdgendt:
Proof-of-concept for a test where resolved west commands in a project's subdirectory does not survive a roundtrip to yaml.

WIP - fix needed

PR #962, opened 2026-06-08
check manifest --resolve in _subdir_special_chars() test

marc-hb:
3 commits. Main one:

Extend test that was just added by commit https://github.com/zephyrproject-rtos/west/commit/1152ac24137490b4c8020f73470b9961a87a8e84 ("tests: extension-commands: Add test checking \\ windows-style variants for west-commands")

This exposes new issue https://github.com/zephyrproject-rtos/west/issues/961 where the behavior is not consistent, not even on Windows alone.

Signed-off-by: Marc Herbert <marc.herbert@gmail.com>

Issue #961, opened 2026-06-08
manifest --resolve normalizes west-commands: "sometimes"


marc-hb:
Depending on various conditions, `manifest-resolve` sometimes outputs special characters exactly as entered like this:
```
    west-commands:
    - mf_subdir/scripts/\winsubA\\\west-commands-from-subdir-windows.yml
```
but in other situations it normalizes paths like this instead:
```
    west-commands:
    - mf_subdir/scripts/winsubA/west-commands-from-subdir-windows.yml
```

At first I thought it would behave one way on Windows and the other way on Unix. It does, but it is unfortunately worse. I just discovered while extending test coverage (#962)  for the very recent ca6bb644b068 fix for #725 that the two different behaviors can be observed _on Windows alone_. I think that's really not OK.

This was already discussed in:
- https://github.com/zephyrproject-rtos/west/pull/920#discussion_r2879475020
- https://github.com/zephyrproject-rtos/west/pull/920#discussion_r2855625436

... and in probably a few other places in #920 but I didn't think it would be inconsistent **even on the same OS**.

PS: in a private chat @pdgendt also suspects that `mf_subdir` might be "silently" added. Should it be?

cc:
- #965
- #966

marc-hb:
From @nmunnich  in https://github.com/zephyrproject-rtos/west/pull/920#issuecomment-4522505716

> Regarding our inconsistency, you are in the end correct that it does seem to affect this after all. My understanding was that we have:
> 
> - the west file itself
> - the output of west manifest --resolve, i.e. what west sees
> - the ability of west to find and run the files found by the manifest
>
> The previous discussion appeared when we were discussing west turning 1 into 2, where both Linux and Windows seem to behave perfectly fine regardless of what type of paths we choose, and hence I thought we were safe here. Going from 2 to 3, we see an issue appearing in Linux where it cannot interpret the windows style paths in the resolved manifest correctly.

So the ability of find and run files is OK, which is good. On other hand I found that your #920 seems to introduce inconsistent `.as_posix()` normalization... on Windows alone. Can you please look into this?

> I've added a windows-only test in the last commit which shows this. If the test is run on Linux it fails, on windows it passes.

I've ported this test to Unix in #962, please review.

marc-hb:
Setting priority "high" until we have at the very least a clear decision on what the behavior SHOULD be in both the short and long terms. A bug is an unexpected behavior. It's not possible to assess the severity of a bug when we don't even know what the expected behavior is!

[...]

PR #966, opened 2026-06-10
manifest: model west-commands as per-entry records

### LINKED ISSUE #959: Improvement: WestCommandsType should be a class
See https://github.com/zephyrproject-rtos/west/pull/920#discussion_r3367818364

> Actually, would it be simpler and clearer to just promote WestCommandsType to some very basic but actual Python class?

> I did consider this, and gave it a brief try, but it felt like a much more invasive refactoring than I was comfortable doing, and that would further explode the size of this #920 PR (which as you said has evolved quite a bit). It most likely should have been a class from the very start tbh - _west_commands_list feels like a constructor, _west_commands_maybe_delist feels like a string representation, _west_commands_merge would be an elegant __iadd__....

Make sure to include the _west_commands_manifest_dirs in the class somehow - it could be that a subclass of dict would be most suitable for WestCommandsType.

pdgendt:
Replace the `WestCommandsType` alias and the parallel `_west_commands_manifest_dirs` dict with a frozen `WestCommandsEntry` dataclass carrying (`path`, `base_dir`). `Project.west_commands_entries` is the new canonical attribute; the old `west_commands` becomes a deprecated read-only property. Schema accepts `str` / `list[str]` / `list[{file, base-dir}]` so resolved manifests roundtrip through `as_yaml()` without losing the `base_dir` needed to resolve `file:` paths inside specs imported from a project subdirectory.

Fixes https://github.com/zephyrproject-rtos/west/issues/959
Supersedes https://github.com/zephyrproject-rtos/west/pull/965

### DIFF
--- a/src/west/commands.py
+++ b/src/west/commands.py
@@ -662,7 +662,7 @@ def extension_commands(config: Configuration, manifest: Manifest | None = None):
 
     specs = OrderedDict()
     for project in manifest.projects:
-        if project.west_commands:
+        if project.west_commands_entries:
             specs[project.path] = _ext_specs(project)
     return specs
 
@@ -673,10 +673,11 @@ def _ext_specs(project):
 
     ret = []
 
-    for cmd in project.west_commands:
+    for entry in project.west_commands_entries:
+        cmd = entry.path
         spec_file = os.path.join(project.abspath, cmd)
 
-        # Verify project.west_commands isn't trying a directory traversal
+        # Verify the entry's path isn't trying a directory traversal
         # outside of the project.
         if escapes_directory(spec_file, project.abspath):
             raise ExtensionCommandError(
@@ -702,10 +703,9 @@ def _ext_specs(project):
         except pykwalify.errors.SchemaError as e:
             raise ExtensionCommandError from e
 
-        # Resolve west command extensions relative to the manifest root for
-        # import-derived west-commands entries, otherwise project root.
-        mfst_dir = project._west_commands_manifest_dirs.get(cmd)
-        base_dir = os.path.join(project.abspath, mfst_dir) if mfst_dir else project.abspath
+        # We must resolve the paths to west commands relative to the
+        # west manifest's path from the project root.
+        base_dir = os.path.join(project.abspath, entry.base_dir)
 
         for commands_desc in commands_spec['west-commands']:
             ret.extend(_ext_specs_from_desc(project, commands_desc, base_dir))
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -59,7 +59,7 @@ QUAL_REFS_WEST = 'refs/west/'
 #: v1.0.x, so that users can say "I want schema version 1" instead of
 #: having to keep using '0.13', which was the previous version this
 #: changed.)
-SCHEMA_VERSION = '1.2'
+SCHEMA_VERSION = '1.6'
 # MAINTAINERS:
 #
 # - Make sure to update _VALID_SCHEMA_VERS if you change this.
@@ -216,6 +216,7 @@ _VALID_SCHEMA_VERS = [
     '0.12',
     '0.13',
     '1.0',
+    '1.2',
     SCHEMA_VERSION,
 ]
 
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -15,8 +15,10 @@ import re
 import shlex
 import subprocess
 import sys
+import warnings
 from collections import deque
 from collections.abc import Callable, Iterable
+from dataclasses import dataclass
 from pathlib import Path, PurePosixPath
 from typing import TYPE_CHECKING, Any, NamedTuple, NoReturn
 
@@ -73,11 +75,25 @@ SCHEMA_VERSION = '1.6'
 # Internal helpers
 #
 
-# The value of a west-commands as passed around during manifest
-# resolution. It can become a list due to resolving imports, even
-# though it's just a str in each individual file right now.
-# TODO: WestCommands should be promoted to a Python class, see issue #959
-WestCommandsType = str | list[str]
+
+@dataclass(frozen=True)
+class WestCommandsEntry:
+    '''A single west-commands spec declared in a West project.
+
+    ``path`` is the spec YAML location, relative to the project root.
+
+    ``base_dir`` is the directory that the spec's ``file:`` entries
+    (paths to the Python files implementing the commands) are resolved
+    against, also relative to the project root. The common case is an
+    empty ``base_dir``, meaning the project root; a non-empty
+    ``base_dir`` is set when the entry was promoted from a submanifest
+    imported from a project subdirectory (so the Python files inside
+    the spec are relative to that subdirectory, not the project root).
+    '''
+
+    path: str
+    base_dir: str = ''
+
 
 # Type for the importer callback passed to the manifest constructor.
 # (ImportedContentType is just an alias for what it gives back.)
@@ -232,36 +248,93 @@ def _load(data: str) -> Any:
         raise MalformedManifest(data) from e
 
 
-def _west_commands_list(west_commands: WestCommandsType | None) -> list[str]:
-    # Convert the raw data from a manifest file to a list of
-    # west_commands locations. (If it's already a list, make a
-    # defensive copy.)
+def _parse_west_commands(raw: Any) -> list[WestCommandsEntry]:
+    # Normalize raw west-commands YAML data into a list of
+    # WestCommandsEntry. Accepts:
+    #
+    #   - None or empty                    -> []
+    #   - str                              -> [WestCommandsEntry(str)]
+    #   - list of (str | dict | Entry)     -> mixed list, dicts must
+    #                                         have a 'file' key and may
+    #                                         have a 'base-dir' key
+    #   - already-parsed list[Entry]       -> defensive copy
+    #
+    # Raises MalformedManifest on any other shape.
 
-    if west_commands is None:
+    if not raw:
         return []
-    elif isinstance(west_commands, str):
-        return [west_commands]
-    else:
-        return list(west_commands)
+    if isinstance(raw, str):
+        return [WestCommandsEntry(path=raw)]
+    if not isinstance(raw, list):
+        raise MalformedManifest(
+            f'invalid west-commands value {raw!r}: '
+            'expected a string, a list of strings, or a list of '
+            "{file, base-dir} maps"
+        )
+
+    ret: list[WestCommandsEntry] = []
+    for item in raw:
+        match item:
+            case WestCommandsEntry():
+                ret.append(item)
+            case str():
+                ret.append(WestCommandsEntry(path=item))
+            case dict():
+                file = item.get('file')
+                if not isinstance(file, str) or not file:
+                    raise MalformedManifest(
+                        f'invalid west-commands entry {item!r}: missing or non-string \'file\' key'
+                    )
+                base_dir = item.get('base-dir', '')
+                if not isinstance(base_dir, str):
+                    raise MalformedManifest(
+                        f'invalid west-commands entry {item!r}: \'base-dir\' must be a string'
+                    )
+                extras = set(item) - {'file', 'base-dir'}
+                if extras:
+                    raise MalformedManifest(
+                        f'invalid west-commands entry {item!r}: unexpected keys {sorted(extras)}'
+                    )
+                ret.append(WestCommandsEntry(path=file, base_dir=base_dir))
+            case _:
+                raise MalformedManifest(
+                    f'invalid west-commands element {item!r}: '
+                    'expected a string or a {file, base-dir} map'
+                )
+    return ret
 
 
-def _west_commands_maybe_delist(west_commands: list[str]) -> WestCommandsType:
-    # Convert a west_commands list to a string if there's
-    # just one element, otherwise return the list itself.
+def _emit_west_commands(entries: list[WestCommandsEntry]) -> str | list:
+    # Serialize a list of WestCommandsEntry into the simplest YAML form
+    # that preserves info:
+    #   - all base_dir == ''  ->  single str (1 entry) or list[str]
+    #   - any non-empty       ->  list of (str | {file, base-dir})
 
-    if len(west_commands) == 1:
-        return west_commands[0]
-    else:
-        return west_commands
+    if all(not e.base_dir for e in entries):
+        if len(entries) == 1:
+            return entries[0].path
+        return [e.path for e in entries]
+    return [e.path if not e.base_dir else {'file': e.path, 'base-dir': e.base_dir} for e in entries]
 
 
-def _west_commands_merge(wc1: list[str], wc2: list[str]) -> list[str]:
-    # Merge two west_commands lists, filtering out duplicates.
+def _merge_west_commands(
+    wc1: list[WestCommandsEntry], wc2: list[WestCommandsEntry]
+) -> list[WestCommandsEntry]:
+    # Merge two west_commands lists, dropping any entry from wc2
+    # whose path is already in wc1 (first wins).
 
-    if wc1 and wc2:
-        return wc1 + [wc for wc in wc2 if wc not in wc1]
-    else:
-        return wc1 or wc2
+    if not wc1:
+        return list(wc2)
+    if not wc2:
+        return list(wc1)
+    paths = {e.path for e in wc1}
+    kept = []
+    for e in wc2:
+        if e.path in paths:
+            _logger.debug('west-commands: keeping first %s, dropping duplicate', e.path)
+        else:
+            kept.append(e)
+    return list(wc1) + kept
 
 
 # Manifest import handling
@@ -411,11 +484,11 @@ class _import_ctx(NamedTuple):
     # the import tree in precedence order.
     group_filter_q: deque
 
-    # The list of west command names provided by the manifest
+    # The list of west command entries provided by the manifest
     # repository itself. This is mutable state in the same way
     # 'projects' is. Manifests which are imported earlier get
     # higher precedence here as usual.
-    manifest_west_commands: list[str]
+    manifest_west_commands_entries: list[WestCommandsEntry]
 
     # The current restrictions on which projects the importing
     # manifest is interested in.
@@ -824,8 +897,13 @@ class Project:
     - ``clone_depth``: clone depth to fetch when first cloning the
       project, or ``None`` (the revision should not be a SHA
       if this is used)
-    - ``west_commands``: list of YAML files where extension commands in
-      the project are declared
+    - ``west_commands_entries``: list of `WestCommandsEntry` records
+      describing the YAML files where extension commands in the project
+      are declared, and the base directory each spec's ``file:`` entries
+      are resolved against
+    - ``west_commands`` (**deprecated**): list of the YAML file paths
+      from ``west_commands_entries``. Loses per-entry ``base_dir`` info;
+      emits ``DeprecationWarning`` on read.
     - ``topdir``: the top level directory of the west workspace
       the project is part of, or ``None``
     - ``remote_name``: the name of the remote which should be set up
@@ -846,7 +924,7 @@ class Project:
             f'Project("{self.name}", "{self.url}", '
             f'revision="{self.revision}", path={self.path!r}, '
             f'clone_depth={self.clone_depth}, '
-            f'west_commands={self.west_commands}, '
+            f'west_commands_entries={self.west_commands_entries}, '
             f'topdir={self.topdir!r}, '
             f'groups={self.groups!r}, '
             f'userdata={self.userdata!r})'
@@ -865,7 +943,7 @@ class Project:
         path: PathType | None = None,
         submodules: SubmodulesType = False,
         clone_depth: int | None = None,
-        west_commands: WestCommandsType | None = None,
+        west_commands: Any = None,
         topdir: PathType | None = None,
         remote_name: str | None = None,
         groups: GroupsType | None = None,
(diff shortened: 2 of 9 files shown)


marc-hb:
I think the tests in #962 should got first because they record/"document" the existing behavior, as it was just changed. Then, this PR can adjust the tests if needed which will again document any behavior change (or none if none)

marc-hb (commented):
I'm afraid "entry" sounds pretty vague.


We have a massive and self-inflicted name collision between:

- The `west-commands:` field in the west.yml manifest
- The "west-commands.yml" file(s) it points at
- The OTHER `west-commands:` header line in those files!
- The "west commands" descriptors in those files. "Descriptor" is an exception: theonly specific name around here? I wish we had more of these.
- The "west commands" .py files that actually implement the commands
- The mypy `WestCommandsType`
- The WestCommand python class
- A new "entry" class added by this PR
- project.west_commands
- others?


EVERYTHING is a "West command"! Unless it's an "extension", which does not help at all.

This mass collision starts in the user interface and documentation at https://docs.zephyrproject.org/latest/develop/west/extensions.html and unsurprisingly gets worse in the code. That relevant code has already given me many headaches and this one PR will obviously increase the count. I would like those collisions to ideally stop or at least start be mitigated.

Naming is hard, but this PR is definitely the right time to formalize some better terminology. I'm not implying a mass search/replace, at least not yet. But at least finalizing some "correct" names to accelerate communication, bug tracking and reviews.

To be fair, some of the above are unique enough not to need a specific name. For instance, there is only one Python file (per command), so a "west command python file" is specific enough. Not concise but good enough. On the other hand, the stuff in YAML files definitely needs more specific names. Even more so considering it's the user interface, even more so considering we have multiple YAML files.

Let me think about some terminology, just a heads-up.

src/west/manifest.py
```diff
@@ -231,36 +248,86 @@ def _load(data: str) -> Any:
...
+    if not isinstance(raw, list):
+        raise MalformedManifest(
+            f'invalid west-commands value {raw!r}: '
+            'expected a string, a list of strings, or a list of '
+            "{file, base-dir} maps"
+        )
 
+    ret: list[WestCommandsEntry] = []
+    for item in raw:
+        if isinstance(item, WestCommandsEntry):
```
marc-hb: In general, prefer the newer `case` and pattern matching. But don't go and  change this before we close the high-level discussions.

src/west/manifest.py
```diff
@@ -73,11 +75,25 @@
...
 
-# The value of a west-commands as passed around during manifest
-# resolution. It can become a list due to resolving imports, even
-# though it's just a str in each individual file right now.
-# TODO: WestCommands should be promoted to a Python class, see issue #959
-WestCommandsType = str | list[str]
+
+@dataclass(frozen=True)
+class WestCommandsEntry:
+    '''A single west-commands spec declared in a project.
```
marc-hb: The term "project" is also problematic :-(

For `west`, a project (not my choice...) is a git repo:
https://docs.zephyrproject.org/latest/develop/west/manifest.html#projects

But in most IDEs and even Zephyr sometimes, a "project" can be the workspace:

- https://code.visualstudio.com/docs/editing/userinterface#_basic-layout
- https://docs.zephyrproject.org/latest/develop/getting_started/index.html#get-zephyr-and-install-python-dependencies 
> In the following instructions the name "zephyrproject" is used for the workspace,...

EDIT: one possible reason for this confusion is: many development "projects" are single-repo. That is: they don't use west or anything like it.

I think in the case the fix is easy: use either "workspace topdir" or "git top dir".

src/west/manifest.py
```diff
@@ -231,36 +248,86 @@ def _load(data: str) -> Any:
...
 
-    if wc1 and wc2:
-        return wc1 + [wc for wc in wc2 if wc not in wc1]
-    else:
-        return wc1 or wc2
+def _merge_west_commands(
+    wc1: list[WestCommandsEntry], wc2: list[WestCommandsEntry]
+) -> list[WestCommandsEntry]:
+    # Merge two west_commands lists, dropping any entry from wc2
+    # whose path is already in wc1 (first wins).
```
marc-hb, line 324: Such an override should print at least a DEBUG.

src/west/manifest.py
```diff
@@ -902,14 +974,28 @@ def __init__(
...
-        self._west_commands_manifest_dirs: dict[str, str] = dict()
+    @property
+    def west_commands(self) -> list[str]:
+        '''Deprecated: paths from `west_commands_entries`.
+
+        Use `west_commands_entries` instead — that exposes the per-entry
+        ``base_dir`` needed to resolve Python files inside the spec when
+        the entry was promoted from a subdirectory submanifest.
+        '''
+        warnings.warn(
```
marc-hb, line 995: Why not just fail? This wasn't part of any documented API AFAICT

marc-hb (commented):
I'm confused for now, sorry. I don't understand why this requires a schema change. Maybe I need  more time to better understand the problem statement.

I think I understand the value of having some `base_dir` in the internal implementation but I don't understand yet why this has to be exposed in the (changed) user interface. Why can't `base_dir` simply be added as a prefix in the output of `west manifest --resolve`?

Adding that `base_dir` does not seem to add any value to the user interface, it looks like a pure hack to make the output of `west manifest --resolve` work again.

Also, do we really need to add yet another level of lists? We already have two. That bit really looks like a new feature, not a bug fix.

---------------

> Let me think about some terminology, just a heads-up.

- west-commands.yml is a DESCRIPTOR LIST  (e.g: `cmd_descs_list` or longer). LIST sounds vague but I think we have no other, top-level list in this context and we can always extend it to "(command) descriptor list" in any context where it could be ambiguous. Hopefully, making a difference between the list, its file and the pointer to it from the manifest is obvious enough from the context when needed and they don't need different names.

- Each DESCRIPTOR has a single pointer to one *command* PYTHON FILE.

- Each PYTHON FILE has a list of command CLASSES. No collision there.

src/west/manifest.py
pdgendt: I think _for West_, it's clear what a project is? Maybe should just name it like that.

```suggestion
    '''A single west-commands spec declared in a West project.
```

src/west/manifest.py
pdgendt: Sure

src/west/manifest.py
pdgendt: Sure, the existing code also did not print though.

src/west/manifest.py
pdgendt: I was under the impression that all fields that do not start with an underscore are considered public? Anyway we don't need to break users of this.

pdgendt (commented):
> I'm confused for now, sorry. I don't understand why this requires a schema change. Maybe I need more time to better understand the problem statement.
> 
> I think I understand the value of having some `base_dir` in the internal implementation but I don't understand yet why this has to be exposed in the (changed) user interface. Why can't `base_dir` simply be added as a prefix in the output of `west manifest --resolve`?

The problem is that `west manifest --resolve` outputs a single yaml file, aggregated from the imported ones. The `west-commands.yml` descriptor files stay in their projects untouched.
We can't simply add the prefix, that conflates two genuinely different directories:

- `path` = location of the `west-commands.yml` spec file, relative to project root (e.g. `mf_subdir/scripts/winsubA/west-commands.yml`)
- `base_dir` = the root that the `file:` entries inside that spec resolve against (e.g. `mf_subdir`)

The Python file resolves as `base_dir / desc['file']`, i.e. relative to `mf_subdir`, not relative to the spec file's own directory `mf_subdir/scripts/winsubA/`. So you can't fold `base_dir` into the `path` prefix; they point at different places.

> Adding that `base_dir` does not seem to add any value to the user interface, it looks like a pure hack to make the output of `west manifest --resolve` work again.

True, `base-dir` is meaningless for a hand-authored manifest, but it _is_ needed for a resolved manifest to work. I wouldn't call it a hack, but it is an internal detail surfacing in the user-facing schema.

> Also, do we really need to add yet another level of lists? We already have two.

There's no level added? Currently we have the following "levels":

1. the manifest `west-commands:` field -> pointer(s) to spec file(s). Was `str | list[str]`.
2. inside each `west-commands.yml`, the top-level `west-commands:` key -> list of descriptors.
3. inside each descriptor, `commands:` -> list of command classes.

This PR changes level 1 only, and it widens the element type of that existing list from `str` to `str | {file, base-dir}`.

> That bit really looks like a new feature, not a bug fix.

We're not adding new features? Users can't express anything new. It is genuinely a bug fix IMO.

> > Let me think about some terminology, just a heads-up.
> 
> * west-commands.yml is a DESCRIPTOR LIST  (e.g: `cmd_descs_list` or longer). LIST sounds vague but I think we have no other, top-level list in this context and we can always extend it to "(command) descriptor list" in any context where it could be ambiguous. Hopefully, making a difference between the list, its file and the pointer to it from the manifest is obvious enough from the context when needed and they don't need different names.
> * Each DESCRIPTOR has a single pointer to one commands PYTHON FILE.
> * Each PYTHON FILE has a list of command CLASSES. No collision there.

I like the idea of using **descriptor**.

src/west/manifest.py
marc-hb: That is a typical Python convention but in west I believe the supported API is only the functions that have a pydoc - which I believe is leveraged to produce https://docs.zephyrproject.org/latest/develop/west/west-apis.html

nmunnich:
> EVERYTHING is a "West command"! Unless it's an "extension", which does not help at all.

To come back to this a bit, I very much agree that this is a massive source of confusion and likely a reason why the related bits are hard to reason about. I propose "West Commands Index" be used for the west-commands.yml file in which west-command-python-files are listed along with the west extension commands that they define. We can then call "WestCommandsEntry" "WestCommandsIndexPath" or something, we can refer to those defined within a west manifest as its west commands indices...

I think picking a proper name for that particular file actually resolves a lot of the unclarity simply by providing a good mental reference point from which other naming can branch off of.

marc-hb:
> The problem is that west manifest --resolve outputs a single yaml file, aggregated from the imported ones. The west-commands.yml descriptor files stay in their projects untouched.

Great summary, thanks. I've been thinking about this... "inconsistency" a little bit (not as much as I wanted to yet), and maybe `west manifest --resolve` should aggregate not just the manifests but the "commands index" files too, producing not just one but two files. The former would point at the latter. Could be something like `workspace/.west/all-resolved-commands.yml` by default, with the possibility to name it differently with a command line option. That would certainly avoid a ~backwards~-incompatible change to solve a problem that should be solved, but that was discovered more or less by "chance" and that no user seems to have been complaining about so far. I mean how many people use west extensions through the output of `west manifest --resolve`? Apparently and unsurprisingly very few.

---------------

> To come back to this a bit, I very much agree that this [lack of names] is a massive source of confusion

Thanks for concurring, that means a lot from the person who did 96ad4f32e29e0508c and friends :-)

pdgendt:
> > The problem is that west manifest --resolve outputs a single yaml file, aggregated from the imported ones. The west-commands.yml descriptor files stay in their projects untouched.
> 
> Great summary, thanks. I've been thinking about this... "inconsistency" a little bit (not as much as I wanted to yet), and maybe `west manifest --resolve` should aggregate not just the manifests but the "commands index" files too, producing not just one but two files. The former would point at the latter. Could be something like `workspace/.west/all-resolved-commands.yml` by default, with the possibility to name it differently with a command line option. That would certainly avoid a backwards-incompatible change to solve a problem that should be solved, but that was discovered more or less by "chance" and that no user seems to have been complaining about so far. I mean how many people use west extensions through the output of `west manifest --resolve`? Apparently and unsurprisingly very few.

Not particularly thrilled about this. `west manifest --freeze/--resolve` simply output to `stdout`, there's no "file" concept to start with.

marc-hb:
> Not particularly thrilled about this. west manifest --freeze/--resolve simply output to stdout, there's no "file" concept to start with.

Not thrilled about this either but I think it's a small enough egg to break.

I think we should look at what are the practical use cases for `west manifest --resolve`?

In an informative, read-only, "pure SBOM" use case, west extensions don't matter. This is probably the most common use case and why we never heard of this bug before and just discovered it "by chance".

So in this particular context we must focus on the more "active" use case where people actually try to feed the resolved manifest back to `west`, for instance to reproduce a build. If they use the resolved manifest only to "bootstrap" the initial download and then pivot to the "real" manifest (while of course making sure nothing changes in git), then we don't care about commands in the "resolved" manifest either. Note they can (and should) re-produce the resolved manifest after pivoting and compare as a safety check.

So I think that leaves only the case of people actually trying to use the resolved manifest to both:
1. bootstrap the download
2. not carry the `resolved-commands.yml` file.
3. not pivot to the real one and actually use the former to `west build` etc.

So yes: the (small minority of?) people bent on doing all that would have to change 2. and copy one extra file, which is still not the end of the world.

Unless I missed some use cases?

> Could be something like workspace/.west/all-resolved-commands.yml by default, with the possibility to name it differently with a command line option. 

Maybe we could have a middle-ground with a smaller `workspace/.west/all-painful-subdir-commands.yml` (better name wanted) containing ONLY the extensions that have "subdir issues"? Even less pretty (granted) but now we've hopefully reduced the number of users who actually need to carry this file to close to zero.

> True, `base-dir` is meaningless for a hand-authored manifest, but it is needed for a resolved manifest to work. I wouldn't call it a hack, but it is an internal detail surfacing in the user-facing schema.

So `base-dir` not the prettiest thing either... pick your poison?

marc-hb:
> True, base-dir is meaningless for a hand-authored manifest, but it is needed for a resolved manifest to work. I wouldn't call it a hack, but it is an internal detail surfacing in the user-facing schema.

If we must make a ~backwards~-incompatible schema change, then how about offering the possibility of _embedding_ west commands descriptors inside manifest files? Now that would be "real" feature offered to users. One less file!

This could be used by `west manifest --resolve` for either all commands, or only for the ones having "subdir issues".

-------------

> ... where people actually try to feed the resolved manifest back to west, for instance to reproduce a build.

Interestingly, NONE of the solutions mentioned so far achieves "perfect" reproducibility for the simple reason that:

> The problem is that west manifest --resolve outputs a single yaml file, aggregated from the imported ones.

_That_ will always be a major difference no matter what. It's the "original sin" and why we are stuck between a rock and a hard place. 

> west manifest --freeze/--resolve "simply" outputs to stdout

That "simplicity" (quotes mine) is deceiving.

nmunnich:
> That would certainly avoid a backwards-incompatible change to solve a problem that should be solved

 I don't see the backwards incompatible change that you're seeing. From what I can see, the new schema should be backwards compatible with existing schemas?
 
 Personally I would be in favor of adding base-dir to the schema as a solution. It feels like the alternative suggested approach would be much more complex, and has a much higher chance of followup issues appearing. I think a schema change is necessary anyway to allow round trips, because we (as far as I can tell) currently do not support lists of west-command-index. Consider:
 
 ```
 manifest:
  self:
   import:
    - a/west.yml
    - b/west.yml
 ``` 
where both of those import a west-command-index via `self`. While slightly more meaningful for a hand-authored manifest than `base-dir`, I would say in practice they are likely equally meaningless.

marc-hb:
> I don't see the backwards incompatible change that you're seeing. From what I can see, the new schema should be backwards compatible with existing schemas?

My bad, I meant "incompatible schema change" and I added "backwards" by force of habit. 

>  I think a schema change is necessary anyway to allow round trips, because we (as far as I can tell) currently do not support lists of west-command-index. Consider: ...

How would some generated `aggregated-commands.yml` not support this example?

marc-hb:
I just remembered that `rsync` v2.6.7 already solved a similar "base_dir" issue with a small but powerful change in how it treats source arguments when using its `--relative` option. https://download.samba.org/pub/rsync/rsync.1#opt--relative

Imagine you want to copy `subdir/file.yml` to `destination/` in a single step / single command. That is, you want `destination/subdir/file.yml` to be created.  `destination/subdir/` does not exist yet.

Before rsync v2.6.7, none of these did it:
- `rsync base_dir/subdir/file.yml destination` copies to `destination/file.yml`
- `rsync --relative base_dir/subdir/file.yml destination/` copies to `destination/base_dir/subdir/file.yml` - wrong again.

You could of course always `pushd base_dir/ && rsync subdir/file.yml ../destination` but that is tedious and error-prone, and it becomes impractical when you want to copy from a number of `base_dir1/`, `base_dir2/`, ... scattered in different places.

So, rsync v2.6.7 added the following "dot separator":

- `rsync --relative base_dir/./subdir/file.yml destination/` 

This creates `destination/subdir/file.yml` at last.

Would such a "dot separator" work in this case too? I think we can assume that no one used any dot separator in their `west-commands` before (why would they have?!), meaning a new, incompatible schema and version bump is not required!  I also think this dot separator syntax would be a "goldilocks" amount of schema complexity for "an internal detail surfacing in the user-facing schema".

EDIT: any other separator should work too. Maybe a separator that filesystems do NOT parse successfully would be more explicit - and then it would _absolutely_ not require any schema version bump. I suspect `rsync` used a dot for CLI auto-complete reasons.

nmunnich:
> How would some generated aggregated-commands.yml not support this example?

I suppose it could, it just feels to me like it would become even more of a hacky solution, especially considering that there might be some edge case where someone _does_ want west-commands-index to be a list.

> Would such a "dot separator" work in this case too?

To me this feels like it would be an even hackier way of doing things. Especially if we wanted to make a separator that filesystems do not parse successfully, then to my eyes this wouldn't be much different to having
```west-commands: "subdir/file.yml base_dir: base_dir```
which I feel would be adding a hack to pretend that the schema wasn't changed.

If we were starting from scratch, the correct way of handling it would likely be to just state that the west-command-python-file paths specified in a west-commands-index file must be relative to the project root, rather than relative to the west manifest. Perhaps we can do something where we deprecate the current path handling and set the schema type as `str | list[str]`, and accept that the round-trip cannot happen when the west-commands-index file uses a deprecated setup?

File: src/west/manifest.py (hunk 4 above)
Comment (pdgendt) [reply src/west/manifest.py:995]: `west_commands` is documented public API, though. In the Zephyr docs (doc/develop/west/west-apis.rst) the Project autoclass documents it explicitly, including a version directive:

```rst
.. autoclass:: west.manifest.Project

   .. (note: attributes are part of the class docstring)
   ...
   .. versionchanged:: 0.8.0
      The *west_commands* attribute is now always a list. In previous
      releases, it could be a string or ``None``.
```

It's rendered on https://docs.zephyrproject.org/latest/develop/west/west-apis.html (attributes come from the Project class docstring, which autoclass picks up).

We should make it deprecated IMO.

src/west/commands.py
```diff
@@ -701,10 +702,13 @@ def _ext_specs(project):
...
-        # import-derived west-commands entries, otherwise project root.
-        mfst_dir = project._west_commands_manifest_dirs.get(cmd)
-        base_dir = os.path.join(project.abspath, mfst_dir) if mfst_dir else project.abspath
+        # Each entry's `file:` paths inside the spec resolve against
+        # entry.base_dir (a subdirectory of the project root when the
+        # entry was promoted from a subdirectory submanifest) or the
+        # project root when base_dir is empty.
+        base_dir = (
+            os.path.join(project.abspath, entry.base_dir) if entry.base_dir else project.abspath
+        )
```
nmunnich:
```suggestion
        # We must resolve the paths to west commands relative to the west manifest's path from the project root
        base_dir = os.path.join(project.abspath, entry.base_dir)
```
Since entry.base_dir defaults to '', there is no need to have this if statement. I also think the comment is not quite suited to be here, since it's discussing the details of the manifest while this file is focused around running the commands. I do think a comment of _some_ kind is helpful here, though. Not entirely happy with my suggestion, but you get the idea

src/west/manifest.py
```diff
@@ -2710,32 +2800,33 @@ def _import_data_from_project(
...
+        # entries keep an empty base_dir.
         mfst_dir = Path(mfst_path) if directory_import else Path(mfst_path).parent
-        west_commands_to_merge = [
-            (mfst_dir / cmd).as_posix() for cmd in submanifest._ctx.manifest_west_commands
+        base_dir = '' if mfst_dir == Path('.') else str(mfst_dir)
+        adjusted = [
+            WestCommandsEntry(
+                path=(mfst_dir / e.path).as_posix(),
+                base_dir=base_dir,
+            )
```
nmunnich: How does this interact if the manifest that we are importing from already has a base_dir defined? I've not thought about this in much detail, but I'm pretty sure this ends up being an awkward scenario.

nmunnich (commented):
I suppose the discussion about the overall approach has stalled. Are we going to continue with this approach then?

Left some comments to progress the approach if yes.

src/west/manifest.py
pdgendt: Updated to compose an existing `base_dir` with prefixing our own if a nested manifest was imported, instead of discarding it.

nmunnich: I'm not certain that is the correct way to handle it, but I may be wrong. Could you add a test testing this case specifically, to make sure that the command is called correctly?

tests/test_extension_commands.py
```diff
@@ -279,14 +279,22 @@ def test_call_imported_project_submanifest_commands_from_project_subdirectory(re
...
 
     MF_SUB_WEST_YML = textwrap.dedent(
         '''\
         manifest:
           self:
-            west-commands: scripts/west-commands-from-subdir.yml
+            west-commands:
+            - scripts/west-commands-from-subdir.yml
+            - file: nested/west-commands-nested.yml
+              base-dir: nested
```
nmunnich, line 297: This test tests whether the base-dir is resolved correctly, which is very nice to have in general.

What I meant in my earlier comment, though, was specifically the case where
- Manifest B is in a subdirectory and has a west-commands
- Manifest A is in a subdirectory, imports manifest B, and the manifest was constructed using `west manifest --resolve`, so it contains the new `base_dir`
- Our project manifest imports manifest A

Then we'd end up with our "double base_dir" case, where we currently compose the two base_dir, but my hunch is that if we already have a base_dir set we need to use the existing base_dir only rather than compose, hence why I want this case tested specifically.

pdgendt: I don't think that's an issue? See the test in the last commit, is that the scenario you describe?

nmunnich: I spent a bit of time trying to recreate my test scenario manually and wasn't able to find a problem, so I guess it works as desired. With that resolved, all of my open points are done and I think this is good to go.

Zzz210s:
Sorry for the noise, and for arriving late to this: I opened #1006 before noticing this PR exists. It is a much smaller alternative for the same issue #959 - a `WestCommands` list subclass that carries the manifest directories, with no schema change and no change to the public `west_commands` attribute.

A maintainer should pick the direction and I am happy to close mine; this PR is the one that actually solves the `as_yaml()` round-trip problem, which mine does not. If it goes forward instead, I would be glad to help: rebasing it on current `main`, or adding tests around the `manifest --resolve` round-trips where #961 / #962 showed the sharp edges.

PR #970, opened 2026-07-02
West-rs

pdgendt:
## West, ported to Rust

This is a proof-of-concept rewrite of `west` as a standalone Rust binary, with a thin Python layer for extension-command dispatch (`west build`, `west flash`, and the rest of the Zephyr ecosystem's commands still run unmodified).

> [!NOTE]
> This port was created with AI assistance.

The built-in commands (`init`, `update`, `list`, `manifest`, `config`, `diff`, `status`, `compare`, `forall`, `grep`, …) are implemented natively — no Python on the hot path. Manifest parsing, config handling, and the VCS layer all live in a shared `west-core` crate; the same code backs both the CLI binary and the PyO3 bindings that extensions import.

Packaging is a single maturin wheel: `pip install` puts the `west` binary on your `PATH` and installs the `_west_native` module extensions link against.

### Trying it out

Building the wheel compiles the Rust binary, so **a Rust toolchain must be installed** (`rustup` / stable `cargo`).

```console
$ cd zephyrproject
$ pip install https://github.com/zephyrproject-rtos/west/archive/refs/pull/970/head.zip  # needs a Rust compiler installed
$ west config migrate   # config is now in TOML (your old files are kept)
$ west update           # …or any other command
```

`west config migrate` reads your existing v1 config and writes the TOML equivalent at the same scope; the original files are left in place, so you can roll back by reinstalling upstream `west`.

Feedback welcome — this is a PoC, so rough edges and missing niceties are expected.

---

### 🚀 Performance & scale

- **Parallel by default.** `update`, `diff`, `status`, `grep`, `forall`, and `compare` run projects concurrently (rayon; `update.jobs` defaults to `min(num_cpus, 8)`).
- **No interpreter startup for built-ins.** v1 paid Python startup + import cost on every invocation. The Rust binary starts instantly.

### 📦 Distribution & dependencies

- **Zero runtime Python dependencies.** v1 pulled in `colorama`, `pyyaml`, `pykwalify`, and `packaging`. v2's `pyproject.toml` has `dependencies = []` — parsing, validation, and coloring all live in the Rust core. Fewer supply-chain surfaces, no version conflicts with the user's environment.
- **Installable without any Python environment.** `cargo install west` gives a single binary. The `pip install` path still works and additionally ships the extension bindings.

### ✨ UX improvements

- **Clean stdout/stderr split**. Data on stdout, chrome (banners, progress, summaries) on stderr, so `west diff > out.patch` and `west grep PATTERN > matches` give machine-readable output.
- **Progress bars** for parallel `update`, and clean pipe handling — `west … | head` exits quietly instead of erroring.
- **Format flexibility.** Manifests and west-commands files can be YAML, TOML, or JSON; config is TOML.

### ⚖️ Trade-offs

- Building from source needs a Rust toolchain (until prebuilt wheels are published on CI).
- v2 is a deliberate breaking-change line; config moves to TOML, a few long-deprecated Python APIs are dropped. `west config migrate` converts the config automatically and leaves the old files in place.
- A couple of parity gaps remain (e.g. `manifest.version` isn't validated yet, no published JSON Schema). _tracked, not forgotten._


henrikbrixandersen:
> This is a proof-of-concept rewrite of west as a standalone Rust binary, with a thin Python layer for extension-command dispatch (west build, west flash, and the rest of the Zephyr ecosystem's commands still run unmodified).

I couldn't find the arguments for doing this rewrite PoC in the AI-generated wall of text. Please summarize.

pdgendt:
> > This is a proof-of-concept rewrite of west as a standalone Rust binary, with a thin Python layer for extension-command dispatch (west build, west flash, and the rest of the Zephyr ecosystem's commands still run unmodified).
> 
> I couldn't find the arguments for doing this rewrite PoC in the AI-generated wall of text. Please summarize.

Apologies! I have update the PR description.

M1cha:
> Packaging is a single maturin wheel: pip install puts the west binary on your PATH and installs the _west_native module extensions link against.

While that works and there are probably other packages doing the same thing, that's a misuse of pip IMO.

> Why a Rust port of west?

While I like Rust, I don't think any of these reasons make sense:

- `Parallel by default.` I think only `west update` profits from that. There have been multiple attempts at implementing this and none of them failed due to technical reasons. It was always stuff like "I don't want to invest time into supporting Windows" or "do we really want to rely on this random dependency?"
- `No interpreter startup for built-ins.` Why is that important? Zephyr heavily relies on Python. The startup time of west builtins is the least of our concerns.
- `Zero runtime Python dependencies.` Why is that important? Zephyr heavily relies on Python and already needs tons of packages anyway.
- `UX improvements`: These either are already supported by west or could easily be added to the Python version.

Also, what about west plugins? These are a very important part of west. Today, any west project can register new west commands without the need for compilation.

marc-hb:
I'm learning Rust and I love it but I share @M1cha's opinion. Rust is specifically designed to solve (C/C++) problems that `west` does not have because it's not in the same problem space. Python is far from perfect but it's been pretty good for `west` and there has been a lot of knowledge accumulated over the years. West dependencies are very minimal and universal and have not been that big of a burden. Users most likely need them already anyway for other Python software.

More importantly, pretty much everyone knows at least a bit of Python and can help with "drive-by" contributions and testing - no toolchain or compilation required. Just do a quick edit and done. Much, much fewer people know Rust and recompiling is never fun.

I agree the parallelism is lacking in west, but I don't find that enough to tip the scale.

PS: not sure why the config file format needs to change, its content has always been very basic.

marc-hb:
BTW I can imagine why you started incrementally from the existing git history, but I still don't think a _published_ Rust branch should be based on existing history. There is just too much difference for such a base to be useful. This should be a branch with no shared history or an entirely different repo.

pdgendt:
> > Packaging is a single maturin wheel: pip install puts the west binary on your PATH and installs the _west_native module extensions link against.
> 
> While that works and there are probably other packages doing the same thing, that's a misuse of pip IMO.

It's a convenience, but also a natural upgrade path. The west extensions are written in Python, allowing west to be installed in the same Python environment is pretty handy.

The Rust port actually enables the opposite: no Python required. We can package West natively for different platforms.

> > Why a Rust port of west?
> 
> While I like Rust, I don't think any of these reasons make sense:
> 
> * `Parallel by default.` I think only `west update` profits from that. There have been multiple attempts at implementing this and none of them failed due to technical reasons. It was always stuff like "I don't want to invest time into supporting Windows" or "do we really want to rely on this random dependency?"

I even created one: https://github.com/zephyrproject-rtos/west/pull/755. There are multiple built-in commands written to be parallel from the start, but `west update` was my focus indeed. Similar to the interpreter startup-time, it starts adding up, if you ran many west commands locally, or automated in CI/CD workflows.

> * `No interpreter startup for built-ins.` Why is that important? Zephyr heavily relies on Python. The startup time of west builtins is the least of our concerns.

`West != Zephyr`. Zephyr would be the main user of west, but West is standalone. I think one of the reasons that people don't adopt West, is because it is written in Python. And if you start adding up the startup time of each west command, I do think it makes sense.
Why do extension commands need to be written in Python? In Zephyr it's because it is easily ported to different platforms, for downstream users this probably isn't a concern. Why not allow shell scripts instead?

> * `Zero runtime Python dependencies.` Why is that important? Zephyr heavily relies on Python and already needs tons of packages anyway.

Again `West != Zephyr`, but with the uprise of many supply-chain attacks, this is a compelling argument for any developer. IMO, Zephyr has _too many_ Python dependencies.

> * `UX improvements`: These either are already supported by west or could easily be added to the Python version.

Maybe, but a v1 version would need to stay backwards compatible, this would be a time to have breaking changes.

> Also, what about west plugins? These are a very important part of west. Today, any west project can register new west commands without the need for compilation.

And that would still be the case for v2? The West v2 package would be distributed as a pre-built wheel, for all (popular) platforms. The same is true for any package with native components (for example `jsonschema`).

> BTW I can imagine why you started incrementally from the existing git history, but I still don't think a _published_ Rust branch should be based on existing history. There is just too much difference for such a base to be useful. This should be a branch with no shared history or an entirely different repo.

Sure, I wanted to keep the tests's history as close as possible, and created a PR for some more visibility.

marc-hb:
> Why do extension commands need to be written in Python? 

Good question, which I finally filed in:
- #974

pdgendt:
> PS: not sure why the config file format needs to change, its content has always been very basic.

Not having any type information in the configuration file is a bad design IMO. The "everything is text" requires additional verification and parsing for no reason, it's especially cumbersome for lists.
We can't even handle the "append" `west config -a` command, if the key/value doesn't exist yet.

marc-hb:
> I even created one: https://github.com/zephyrproject-rtos/west/pull/755. There are multiple built-in commands written to be parallel from the start, but west update was my focus indeed. Similar to the interpreter startup-time, it starts adding up, if you ran many west commands locally, or automated in CI/CD workflows.

The time in "west update" and similar is not spent in west itself much, but in fetching repos one by one.


> I think one of the reasons that people don't adopt West, is because it is written in Python...

?! Python is one of the most popular languages by far, so familiarity is clearly not an issue here. Rust would clearly make that worse, at least for now.

> ... and if you start adding up the startup time of each west command, I do think it makes sense.

Ah! So what you really meant is: "... because west is slow?" That makes more sense to me. That's a possibility but I honestly doubt it. You shouldn't be invoking west many times when you build or when you do anything for that matter. If you need some west data in your build (or in anything), then get that data once and save it, don't keep invoking west. Same as any other data: never compute the same thing repeatedly. If west is in someone's bottleneck, I'm pretty sure they are doing something wrong. Or very niche.



> We can't even handle the "append" west config -a command, if the key/value doesn't exist yet.

I think this was by (debatable) design and it does not seem related to the config file format or anything else here, is it?

> Not having any type information in the configuration file is a bad design IMO. The "everything is text" requires additional verification and parsing for no reason, it's especially cumbersome for lists.

TOML is a good upgrade compared to INI but this complete rewrite looks a like a big enough change already :-)

marc-hb:
> > This should be a branch with no shared history or an entirely different repo.

I think this should be an entirely new "west-rs" project - because it very much is. To make the switch easy, "west-rs" can be configured so "cargo install west-rs" adds the command "west" in the PATH. Then both versions can coexist and compete. Competition is good! Confusion and "takeovers" are not, please don't make that same "chardet" mistake: https://lwn.net/Articles/1061534/

Maybe "west-rs" will be a great success and the Python project will be dead in a few years? That would be fine as long as it was clear what was what and users can always switch and compare easily.

henrikbrixandersen:
> think this should be an entirely new "west-rs" project - because it very much is. To make the switch easy, "west-rs" can be configured so "cargo install west-rs" adds the command "west" in the PATH. Then both versions can coexist and compete.

I think this would be the best approach too.

PR #976, opened 2026-07-23
Support executable extension commands

### LINKED ISSUE #974: Support "light-weight" extensions in any language
_Originally posted by @marc-hb in https://github.com/zephyrproject-rtos/zephyr/issues/112680#issuecomment-4963973189_

> There is no technical reason for this [`west dt`] to be a west command. @pdgendt agrees that this could be a script
> ... 
> @pdgendt would prefer to have it as something similar to west twister, where the script is in scripts/ and the west command is just an entry point to a script with all the logic that lives in scripts/

I feel like people write west extensions for two distinct reasons:
1. access to the west Python APIs
2. "Dynamic PATH convenience" (better name wanted). That is: not having to locate workspace-specific commands and not having to tweak their PATH, forgetting to tweak it again and run the same command but from a different and unexpected workspace.

I suspect it's not the first time some code wants 2. without caring about 1. 

I've been thinking it could be nice to have some "light weight" extension ability in west. I mean the ability for any workspace to point `west random_thing` at `workspace/somewhere/random_thing.exe` where the latter **does not have to be in Python** and does not care about any west API. Maybe `west` could pass just a couple values in some JSON file, starting with the workspace location. Those few values would a much looser "API" that requires much less attention to backwards compatibility and west versions.

pdgendt:
West extension commands currently must be Python `WestCommand` subclasses, coupling every extension to the west API even when it only needs west's ability to locate and run a workspace-relative command.

This adds a second kind of extension command: one backed by an executable in any language. It reuses the existing `file:` key and adds a per-command `exec:` selector:

```yaml
west-commands:
  # WestCommand extension: the file is imported and this subclass is
  # instantiated, with full access to the west API.
  - file: scripts/west_commands.py
    commands:
      - name: my-api-cmd
        class: MyApiCommand

  # Executable extension run through an interpreter.
  - file: scripts/gen_report.py
    commands:
      - name: gen-report
        exec: python3        # runs: python3 scripts/gen_report.py <args...>

  # Executable extension run directly (compiled binary or +x script).
  - file: tools/provision
    commands:
      - name: provision
        exec: true           # runs: tools/provision <args...>
```

- `exec:` may be `true` (run the file directly), a string, or a list (interpreter + args).
- `class` and `exec` are mutually exclusive; commands without `exec` keep the current Python behavior, so existing extensions are unaffected.
- West forwards all arguments (including `-h/--help`) verbatim and passes workspace context via `WEST_*` env vars (`WEST`, `WEST_TOPDIR`, `WEST_MANIFEST_PATH`, `WEST_PROJECT_PATH`, `WEST_VERBOSE`, and `WEST_COMMAND` for dispatch when one file is registered under several names).
- The interpreter form makes interpreted scripts work on Windows too, where a bare script can't be executed directly.

Fixes: https://github.com/zephyrproject-rtos/west/issues/974

### DIFF
--- a/src/west/commands.py
+++ b/src/west/commands.py
@@ -617,6 +617,108 @@ WARNING: in file {self.py_file},
         return cmd
 
 
+class _ExecExtensionCommand(WestCommand):
+    # An extension command backed by an executable.
+    #
+    # Unlike class-based extensions, which are Python WestCommand subclasses
+    # with access to the full west API, an executable extension is any
+    # executable (in any language) named by the 'file' key of a west-commands
+    # file, selected with the 'exec' key of a command. West does not parse its
+    # arguments: everything after the command name (including -h/--help) is
+    # forwarded verbatim, and workspace context is passed via WEST_*
+    # environment variables. The invoked command name is passed as
+    # WEST_COMMAND so that a single executable registered under several names
+    # can dispatch on the subcommand.
+    #
+    # If 'interpreter' is empty, the file is run directly; otherwise it is run
+    # as arguments to 'interpreter' (e.g. ['python3']).
+
+    def __init__(
+        self,
+        name: str,
+        help: str,
+        exec_file: str,
+        interpreter: list[str],
+        project_path: str,
+    ):
+        super().__init__(
+            name,
+            # 'help' is ignored for extensions, see
+            # https://github.com/zephyrproject-rtos/west/issues/927
+            help='',
+            description=help,
+            accepts_unknown_args=True,
+            requires_workspace=True,
+        )
+        self._exec_file = exec_file
+        self._interpreter = interpreter
+        self._project_path = project_path
+
+    def do_add_parser(self, parser_adder):
+        # add_help=False so that -h/--help is forwarded to the executable
+        # instead of being intercepted by west's argument parser.
+        return parser_adder.add_parser(
+            self.name,
+            help=self.help,
+            description=self.description,
+            add_help=False,
+        )
+
+    def do_run(self, args: argparse.Namespace, unknown: list[str]):
+        # The backing file must exist however it is run, whether directly or
+        # as an argument to an interpreter. Check this first on every platform,
+        # since os.access(..., X_OK) below returns False for a missing file
+        # too, which would otherwise be reported as the misleading "not
+        # executable".
+        if not os.path.exists(self._exec_file):
+            self.die(f'extension command executable "{self._exec_file}" not found')
+
+        # When run directly (no interpreter), the file itself must also be
+        # executable. This X_OK check is only meaningful on POSIX; on Windows
+        # it is not, so let subprocess report any such failure instead.
+        if not self._interpreter and os.name == 'posix':
+            if not os.access(self._exec_file, os.X_OK):
+                self.die(f'extension command executable "{self._exec_file}" is not executable')
+
+        env = os.environ.copy()
+        # The context "API" for executable extensions. Keep this small
+        # and stable; it is a much looser contract than the Python API.
+        env['WEST'] = sys.argv[0]
+        # The command name as invoked by the user, so a single executable
+        # registered under multiple names can dispatch on the subcommand.
+        env['WEST_COMMAND'] = self.name
+        assert self.topdir is not None  # requires_workspace is True
+        env['WEST_TOPDIR'] = self.topdir
+        if self.has_manifest and self.manifest.abspath:
+            env['WEST_MANIFEST_PATH'] = self.manifest.abspath
+        env['WEST_PROJECT_PATH'] = self._project_path
+        env['WEST_VERBOSE'] = str(int(self.verbosity))
+
+        argv = [*self._interpreter, self._exec_file, *unknown]
+        self._log_subproc(argv)
+        try:
+            proc = subprocess.run(argv, env=env)
+        except OSError as e:
+            self.die(f'could not run extension command "{self.name}": {e}')
+
+        if proc.returncode:
+            raise CommandError(proc.returncode)
+
+
+@dataclass
+class _ExecExtFactory:
+    exec_file: str
+    interpreter: list[str]
+    name: str
+    help: str
+    project_path: str
+
+    def __call__(self):
+        return _ExecExtensionCommand(
+            self.name, self.help, self.exec_file, self.interpreter, self.project_path
+        )
+
+
 @dataclass
 class WestExtCommandSpec:
     # An object which allows instantiating a west extension.
@@ -712,26 +814,54 @@ def _ext_specs(project):
 
 
 def _ext_specs_from_desc(project, commands_desc, base_dir):
-    py_file = os.path.join(base_dir, commands_desc['file'])
+    file = os.path.join(base_dir, commands_desc['file'])
 
-    # Verify the YAML's python file doesn't escape the project directory.
-    if escapes_directory(py_file, project.abspath):
+    # Verify the YAML's file doesn't escape the project directory.
+    if escapes_directory(file, project.abspath):
         raise ExtensionCommandError(
-            hint=f'extension command python file "{commands_desc["file"]}" '
+            hint=f'extension command file "{commands_desc["file"]}" '
             f'escapes project path {project.path}'
         )
 
-    # Create the command thunks.
+    # Create the command thunks. Each command is either a class-based
+    # extension, selected with 'class', or an executable extension, selected
+    # with 'exec'.
     thunks = []
     for command_desc in commands_desc['commands']:
         name = command_desc['name']
-        attr = command_desc.get('class', name)
         help = command_desc.get('help', f'(no help provided; try "west {name} -h")')
-        factory = _ExtFactory(py_file, name, attr)
+
+        if 'exec' in command_desc:
+            if 'class' in command_desc:
+                raise ExtensionCommandError(
+                    hint=f"west command '{name}' sets both 'class' and 'exec'"
+                )
+            interpreter = _exec_interpreter(name, command_desc['exec'])
+            factory = _ExecExtFactory(file, interpreter, name, help, project.path)
+        else:
+            attr = command_desc.get('class', name)
+            factory = _ExtFactory(file, name, attr)
+
         thunks.append(WestExtCommandSpec(name, project, help, factory))
 
     # Return the thunks for this project.
     return thunks
+
+
+def _exec_interpreter(name, value):
+    # Turn the 'exec' value of a command into the interpreter argument list
+    # to prefix the file with. 'true' means "run the file directly" (empty
+    # list); a string or list of strings names the interpreter.
+    if value is True:
+        return []
+    if isinstance(value, str):
+        return [value]
+    if isinstance(value, list) and all(isinstance(v, str) for v in value):
+        return list(value)
+    raise ExtensionCommandError(
+        hint=f"west command '{name}' has an invalid 'exec' value; "
+        "expected true, a string, or a list of strings"
+    )
 
 
 def _commands_module_from_file(file):
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -59,7 +59,7 @@ QUAL_REFS_WEST = 'refs/west/'
 #: v1.0.x, so that users can say "I want schema version 1" instead of
 #: having to keep using '0.13', which was the previous version this
 #: changed.)
-SCHEMA_VERSION = '1.2'
+SCHEMA_VERSION = '1.6'
 # MAINTAINERS:
 #
 # - Make sure to update _VALID_SCHEMA_VERS if you change this.
@@ -216,6 +216,7 @@ _VALID_SCHEMA_VERS = [
     '0.12',
     '0.13',
     '1.0',
+    '1.2',
     SCHEMA_VERSION,
 ]
 
--- a/src/west/west-commands-schema.yml
+++ b/src/west/west-commands-schema.yml
@@ -18,6 +18,9 @@ mapping:
     sequence:
       - type: map
         mapping:
+          # The file backing the commands below. For "class" commands it is
+          # imported as a Python module; for "exec" commands it is run as an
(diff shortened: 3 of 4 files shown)


src/west/commands.py
```diff
@@ -617,6 +617,108 @@ def __call__(self):
...
+        exec_file: str,
+        interpreter: list[str],
+        project_path: str,
+    ):
+        super().__init__(
+            name,
+            # 'help' is ignored for extensions, see
+            # https://github.com/zephyrproject-rtos/west/issues/927
+            help='',
+            description=help,
```
marc-hb, line 649: Considering the previous arg, can we call this one something else than `help`?

src/west/commands.py
```diff
@@ -617,6 +617,108 @@ def __call__(self):
...
+
+        argv = [*self._interpreter, self._exec_file, *unknown]
+        self._log_subproc(argv)
+        try:
+            proc = subprocess.run(argv, env=env)
+        except OSError as e:
+            self.die(f'could not run extension command "{self.name}": {e}')
+
+        if proc.returncode:
+            raise CommandError(proc.returncode)
```
marc-hb, line 705: Is that not losing the error messages?

marc-hb (commented):
Apologies for the belated review.

I'm really not a fan of the user interface. It tries to shoehorn new concepts into old fields which seems to preserve some consistency... while still being very different and inconsistent. Can we have brand new fields instead?
```yaml
# WestCommand extension: the file is imported and this subclass is
  # instantiated, with full access to the west API.
  - file: scripts/west_commands.py
    commands:
      - name: my-api-cmd
        class: MyApiCommand
      - name: my-api-cm2
        class: MyApiCommand2

  # Executable extension run through an interpreter.
  - file: scripts/gen_report.py
    commands:
      - name: gen-report
        exec: python3        # runs: python3 scripts/gen_report.py <args...>

  # Executable extension run directly (compiled binary or +x   script).
  - file: tools/provision
    commands:
      - name: provision
        exec: true           # runs: tools/provision <args...>
```

The purpose of the original file+commands design is to allow multiple commands in a single Python file. The new stuff will never have multiple "commands" in a single "file" yet it keeps the same structure and names.

In the original design, a `file` contains `commands`, the former is at a higher level. But in the new interpreter case, the interpreter is found at a lower level than the script. It looks backwards.

I'm really not a fan of "loosely typed" fields that can be either a boolean or a file.

The exec true/interpreter difference exists only for Windows. It would be nice to make the simpler Unix case... simpler and less verbose.

Let me think and try to suggest some alternative.

marc-hb:
> Let me think and try to suggest some alternative.

As this new feature is basically `subprocess.run()`, how about:

```yaml

  # Executable extension (run through an interpreter or not)
  - name: gen-report
    subprocess:
      - args: [python3, scripts/gen_report.py] # additional args are appended
        shell: false # by default
        timeout: 0 # by default

  # WestCommand extension: the file is imported and this subclass is
  # instantiated, with full access to the west API.
  - file: scripts/west_commands.py
    commands:
      - name: my-api-cmd
        class: MyApiCommand
```

Note: I am _not_ suggesting that west passes a `**kwwargs` to subprocess "as is". Only a carefully selected list.

marc-hb:
> West forwards all arguments (including -h/--help) verbatim 

Yes.

> and passes workspace context via WEST_* env vars (WEST, WEST_TOPDIR, WEST_MANIFEST_PATH, WEST_PROJECT_PATH, WEST_VERBOSE

Can we avoid (evil) environment variables and pass some temporary `west_cmdname.json` file instead?

https://nibblestew.blogspot.com/2021/03/never-use-environment-variables-for.html

marc-hb:
We can drop this from 1.6 if that's too much change.

pdgendt:
> Apologies for the belated review.
> 
> I'm really not a fan of the user interface. It tries to shoehorn new concepts into old fields which seems to preserve some consistency... while still being very different and inconsistent. Can we have brand new fields instead?

This is a first proposal, of course.

> The purpose of the original file+commands design is to allow multiple commands in a single Python file. The new stuff will never have multiple "commands" in a single "file" yet it keeps the same structure and names.

Why not? You can still have shared functionality, even in a shell script, that's why I've set the actual West command name in the `WEST_COMMAND` environment variable.

> In the original design, a `file` contains `commands`, the former is at a higher level. But in the new interpreter case, the interpreter is found at a lower level than the script. It looks backwards.
> 
> I'm really not a fan of "loosely typed" fields that can be either a boolean or a file.
> 
> The exec true/interpreter difference exists only for Windows. It would be nice to make the simpler Unix case... simpler and less verbose.

It doesn't only serve a purpose on Windows. One useful scenario that comes to mind, is a node javascript file, and the interpreter being `node`.

> Let me think and try to suggest some alternative.

Sure

> Can we avoid (evil) environment variables and pass some temporary `west_cmdname.json` file instead?
> 
> [nibblestew.blogspot.com/2021/03/never-use-environment-variables-for.html](https://nibblestew.blogspot.com/2021/03/never-use-environment-variables-for.html)

Maybe, but occasionally, environment variables are useful. No strong opinion here.

> Note: I am _not_ suggesting that west passes a `**kwwargs` to subprocess "as is". Only a carefully selected list.

I'd guess that's the first request users would make; "How can I pass my additional arguments?" Why shouldn't we? Is there a security risk? I don't see the problem here.

marc-hb:
> Why not? You can still have shared functionality, even in a shell script, that's why I've set the actual West command name in the WEST_COMMAND environment variable.

Thanks, I missed that.

I still don't think `file:` should be at the top though. I think the command name should be at the top because the command name is the user interface. 

```yaml
  # Executable extension (run through an interpreter or not)
  - name: gen-report
    subprocess:
      - args: [python3, scripts/gen_report.py] # + CLI args

  - name: gen-stats
    subprocess:
      - args: [python3, scripts/gen_report.py, stats] # + CLI args
```
I suspect the `file:` was put at the top because it made the implementation easier: `import` every `file:` and done. No need to check for duplicates. But there is no duplicate import issue in this new use case. Also, isn't `python3` the executable "file:" and `scripts/gen_report.py` just an argument? Confusing.

> It doesn't only serve a purpose on Windows. One useful scenario that comes to mind, is a node javascript file, and the interpreter being node.

Good point. What I meant is: it should be very simple and short to run just one executable on Unix without requiring too many fields. That's a common case motivating this entire feature and we should "make simple things easy and more complex things possible".

marc-hb:
> Maybe, but occasionally, environment variables are useful.

There are "useful" as a last fallback when there is no good alternative available. Exactly like global variables.

> > Note: I am not suggesting that west passes a **kwwargs to subprocess "as is". Only a carefully selected list.

> I'd guess that's the first request users would make; "How can I pass my additional arguments?" Why shouldn't we? Is there a security risk? I don't see the problem here.

Because some parameters require the code that actually invokes `subprocess` to behave differently. For instance `capture_output=` or `check=`. Unless west "parses" them and behaves accordingly? But now you have the problem of documenting which arguments are merely passed through versus those that _also_ change the caller's (west) behavior. Confusing.

Because we don't know yet how popular this new feature can get. We have some clearly identified candidate use cases but that's all. So not worth spending a lof cycles designing "advanced" usage yet - YAGNI.

`Popen` is a complex interface with a lot of pitfalls, I think it would be very risky to just pass an entire `kwargs` without control. And yes that also feels like it could increase the "attack surface".

Last but not least, different people can be using different Python versions with different set of supported parameters. What happens when some parameters are not supported? Garbage in, garbage out?

pdgendt:
_Removed this from v1.6.0 so we can continue with a release already._

thorsten-klein:
I’m not a big fan of exposing implementation details in the interface, such as `subprocess`.

For simplicity, I would expect the most straightforward interface to look like this:

```yaml
- name: my-command
  run: [python3, scripts/gen_report.py, stats]
```

If users prefer, they could also use an expanded YAML list:

```yaml
- name: other-command
  run:
    - python3
    - scripts/other.py
    - path/to/some/config.json
```

This keeps the interface simple and focused on what the user wants to execute, without exposing how it is implemented internally.

marc-hb:
> I’m not a big fan of exposing implementation details in the interface, such as subprocess.

When looking at the code, you can see `subprocess` is not really an "implementation detail" but basically the only thing performed. Hence the idea to also pass-through other parameters like `shell=...`, `timeout=...` and maybe others.

What the use cases in #974 request is to never make this feature more complex than a simple `subprocess` pass-through. People who need something more complex should write a Python extension.

All this being said, I still like your name `run`! I find it simple and intuitive, thanks. And it's close enough to `subprocess.run()`.

```
- name: my-command
  run: [python3, scripts/gen_report.py, stats]
```

pdgendt:
> For simplicity, I would expect the most straightforward interface to look like this:
> 
> ```yaml
> - name: my-command
>   run: [python3, scripts/gen_report.py, stats]
> ```

The one issue here is, again, path resolution. If any of the arguments is a path relative to the west project, intuitively you'd change the `cwd` before running the command. But then forwarded arguments from the west call silently resolve against the project directory.

Alternative is to expand variables in the `run` arguments, and provide a small set that expands (`${WEST_PROJECT_PATH}`, `${WEST_TOPDIR}`)

```yaml
- name: gen-report
  run: [python3, ${WEST_PROJECT_PATH}/scripts/gen_report.py, --config, ${WEST_PROJECT_PATH}/cfg/report.yml]
```

marc-hb:
> The one issue here is, again, path resolution. If any of the arguments is a path relative to the west project, intuitively you'd change the cwd before running the command. But then forwarded arguments from the west call silently resolve against the project directory.

Yes, this relative paths topic is critical and we get only one chance to get it right - future corrections would break everything. Also, thanks kinda the main point of #974. Thanks for highlighting this.

I'm afraid the only "deterministic" choice is to anchor relative paths to the _root of the git checkout where the .yml file defines the command_. I believe anything else would work or fail "sometimes". west topdir does not work because the git repo can be moved around. Maybe less frequently but the .yml file can also be moved inside its own repo.

To keep the implementation simple, I think that anchoring should be performed by simply _changing the current directory to that git checkout root_. Unlike "manually" resolving relative paths, that requires near zero code. But more importantly, it avoids trying (and failing) to guess whether `stats` is a path or not:
```yaml
- name: my-command
  run: [python3, scripts/gen_report.py, stats] # is "stats" a path or not?
```

> Alternative is to expand variables in the run arguments, and provide a small set that expands (${WEST_PROJECT_PATH}, ${WEST_TOPDIR})

That could be interesting feature but:
- I think these should be optional and not required. So we must still define what happens when none of these is used anyway.
- It can be added later after we gather some real-world usage and users feedback (YAGNI)

> But then forwarded arguments from the west call silently resolve against the project directory.

Yes, this is the (only?) price to pay. But that problem is trivially avoided by using absolute arguments and it also gives _consistency_ by handling CLI arguments the same as arguments hard-coded inside the .yaml file.

Maybe we could also have a `${WEST_PREVIOUS_CWD}` (or rather: the corresponding JSON field) for the (rare?) command that really wants to access paths relative to where the extension was invoked from?

marc-hb:
> and passes workspace context via WEST_* env vars (WEST, WEST_TOPDIR, WEST_MANIFEST_PATH, WEST_PROJECT_PATH, WEST_VERBOSE

> Can we avoid (evil) environment variables and pass some temporary `west_cmdname.json` file instead?
> https://nibblestew.blogspot.com/2021/03/never-use-environment-variables-for.html

There is another way JSON is superior to environment variables: the difference between a missing variable and a blank variable is very clear.  `${UNSET_VAR}` and `${BLANK_VAR}` expand the same by default on Unix https://mywiki.wooledge.org/BashFAQ/112

To be fair, temporary files also have some issues:
- Where to create them. https://docs.python.org/3/library/tempfile.html can hopefully help.
- How to find them. Maybe ONE environment variable WEST_EXTENSION_PARAMS can do that. Not great but much better than a growing collection of variables.
- Cleaning them up. Even on failure. Or not on failure, because they are useful when investigating failures? Or yes on failure, because too many failures clog the temporary directory? Never easy.

Pushing this idea one step further, maybe we could have the entire JSON data in a single WEST_EXTENSION_PARAMS environment variable.

thorsten-klein:
I completely agree with @pdgendt on this.

As a user, I would expect the command to run with my current working directory as its working directory. This becomes especially important when passing arguments to the command, for example:

```
west mycommand ./config.json
```

In this case, `./config.json` should naturally be resolved relative to the directory from which I invoked `west`. Anything else would be rather surprising and unintuitive.

I think the simple solution is exactly what @pdgendt suggested: define a small set of well-known anchors that can be used when a path needs to be resolved. If no anchor is specified and a relative path is used, it should be interpreted relative to the current working directory from which the command was invoked.

As @pdgendt already suggested, I also see the following anchors as useful:

- `{west_topdir}` → path containing `.west`
- `{manifest_dir}` / `{project_dir}` → path containing `west.yml`
- `{this_dir}` → path containing the `west_commands.yml` file
- `{cwd}` → current working directory

That gives users explicit control when they need paths relative to a specific location, while keeping the default behavior intuitive and consistent with how command-line tools generally work.

marc-hb:
> west mycommand ./config.json
>
> In this case, ./config.json should naturally be resolved relative to the directory from which I invoked west. Anything else would be rather surprising and unintuitive.

Yes, but that's just one sample use case. Not every command will have local paths argument on the command line. Changing current directory might be more useful for other commands.

> If no anchor is specified and a relative path is used, it should be interpreted relative to the current working directory from which the command was invoked.

It's dangerous to have relative paths _instead the .yaml file_ that depend on where the command is run from. That is too non-deterministic, a security risk and confusing like the above. Even more confusing than your example because the confusion then comes from different places. So this means the `{west_...}` anchors become practically mandatory if you do not change current directory.

The fundamental issue is: for relative paths convenience, two "current directories" are needed but we can have only one at a time.

Unless... there is a `change_cwd` boolean flag in the .yaml file that lets each command decides for itself what it thinks is best! It would require near-zero code and complexity and let command writers decide what choice is best for each command.

PR #865, opened 2025-10-09
repo: format Python files with ruff and enforce in CI

pdgendt:
Moving forward I would like the Python formatting to be enforced in CI, this ensures there's little to no discussion needed on how code should be formatted in PRs.

In Zephyr, the `ruff` formatter is enforced in CI for new files. As there are a lot of existing Python files there, it made sense to not  format in one go (350+ excluded files ATM).
In West there aren't a lot of files (16 need formatting changes), so that's why I propose to apply this once.

`--preview` is enabled to have Github Action annotations.

_Can we just bite the bullet here, please?_

EDIT cc:
- https://github.com/zephyrproject-rtos/west/pull/768 # unrelated PR where a related discussion happened
- https://github.com/zephyrproject-rtos/zephyr/pull/81070

PR #977, opened 2026-07-24
ruff: Bump version and drop preview mode

pdgendt:
Ruff 0.16 broadens the default rule set that extend-select builds on, fix issues before the actual bump.
See individual commits for details.

Preview mode was only enabled to get GitHub annotation output from the formatter, which is stable as of Ruff 0.16. Drop it and reformat the files that relied on the preview-only hug-parens style.

Sadly this does format some (mostly test) files, if not desired, the preview mode could be kept.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -19,6 +19,7 @@ from functools import partial
 from os.path import abspath, basename, relpath
 from pathlib import Path, PurePath
 from time import perf_counter
+from typing import ClassVar
 from urllib.parse import urlparse
 
 from west import util
@@ -2503,17 +2504,17 @@ class Grep(_ProjectCommand):
     #
     # - the 'failed' handling below for proper exit codes
     # - color handling works as expected
-    TOOLS = ['git-grep', 'ripgrep', 'grep']
+    TOOLS: ClassVar[list[str]] = ['git-grep', 'ripgrep', 'grep']
 
     DEFAULT_TOOL = 'git-grep'
 
-    DEFAULT_TOOL_ARGS = {
+    DEFAULT_TOOL_ARGS: ClassVar[dict[str, list[str]]] = {
         'git-grep': [],
         'ripgrep': [],
         'grep': ['--recursive'],
     }
 
-    DEFAULT_TOOL_EXECUTABLES = {
+    DEFAULT_TOOL_EXECUTABLES: ClassVar[dict[str, list[str]]] = {
         # git-grep is intentionally omitted: use self._git
         'ripgrep': ['rg', 'ripgrep'],
         'grep': ['grep'],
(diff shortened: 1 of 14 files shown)


kartben (approved):

marc-hb (commented):
Afraid this is exactly the kind of conflicting git churn I predicted  in #865. While there are some actual code changes, a lot of it is minor and pointless whitespace changes not even making any actual style difference.

Could some of this be avoided by changing some settings? Ruff should provide such "backwards-compatibility" settings.

pdgendt:
> Afraid this is exactly the kind of conflicting git churn I predicted in #865. While there are some actual code changes, a lot of it is minor and pointless whitespace changes not even making any actual style difference.
> 
> Could some of this be avoided by changing some settings? Ruff should provide such "backwards-compatibility" settings.

I already stated that in the PR description:

> Sadly this does format some (mostly test) files, if not desired, the preview mode could be kept.

marc-hb:
I wish I had noticed this "preview" mode (new to me) when it was introduced. It's really good to finally turn it off and get to a more stable place.

So are you saying that there is no "backwards-compatibility" setting for any of the whitespace changes? Because 1) they were all a side-effect of this preview mode? 2) so ruff did not bother provide any switch for them? (it could have anyway)

If I understood all this correctly then: LGTM

pdgendt:
> I wish I had noticed this "preview" mode (new to me) when it was introduced. It's really good to finally turn it off and get to a more stable place.
> 
> So are you saying that there is no "backwards-compatibility" setting for any of the whitespace changes? Because 1) they were all a side-effect of this preview mode? 2) so ruff did not bother provide any switch for them? (it could have anyway)

Yes, I only found out too after it was merged, there was [already a request](https://github.com/astral-sh/ruff/issues/19537#issuecomment-3116606623) to make it opt-in/opt-out, but the authors don't want to have that.

> If I understood all this correctly then: LGTM

:+1:

src/west/app/project.py
```diff
@@ -704,7 +704,7 @@ def create(self, directory: Path, exist_ok: bool = True) -> None:
             self.die(f'Cannot initialize in {directory}: permission denied')
         except FileExistsError:
             self.die(f'Cannot initialize in {directory}: it already exists')
-        except Exception as e:
+        except Exception as e:  # noqa: BLE001 intentional catch-all for a clean error message
```
marc-hb, line 707: Finally! pylint had W0718 for ages.

Can't wait to have this one enabled in https://github.com/zephyrproject-rtos/zephyr :-) because this is a powerful way to hide bugs. I digress.

marc-hb (approved):

src/west/app/project.py
marc-hb: > Finally! pylint had W0718 for ages. Can't wait to have this one enabled in https://github.com/zephyrproject-rtos/zephyr 


Literally less than 5 minutes later:

https://github.com/zephyrproject-rtos/zephyr/pull/114736/changes#diff-c0c813b8b08b01a31a1cd7886ecaf7846a82bcf6cf3cf4ee37c0570592a33e49R393

PR #978, opened 2026-07-24
ci: fix and harden the Linux distro test workflow

pdgendt:
Fixes the `fedora:rawhide` failure in the distro test workflow and corrects a deeper issue: the workflow was not actually testing each distro's Python.

- Use each distro's own Python. `uv` defaults to python-preference=managed and silently downloads its own standalone CPython, bypassing the distro interpreter on every matrix entry. Set `UV_NO_MANAGED_PYTHON` and `UV_PYTHON_DOWNLOADS=never` so uv uses the system Python — what this workflow is meant to test.
- Disable `fedora:rawhide`. It ships a prerelease Python (3.15) that west doesn't support and that lacks dev-dependency wheels; on the distro's own interpreter it only exercises an unsupported config. `fedora:latest` keeps Fedora covered on a released Python.
- Coherent run steps. Report the interpreter `uv` actually resolves (via `uv run`, after sync) instead of the bare `python3`, and pass `--frozen` to the `poe run` for parity with `uv sync` and the main test workflow.
- Refresh local-run docs. Replace the stale `tox` instructions with the current `uv + poe` commands.

### DIFF
--- a/.github/workflows/test-distros.yml
+++ b/.github/workflows/test-distros.yml
@@ -53,7 +53,7 @@ jobs:
           - "debian:stable"
           - "debian:testing"
           - "fedora:latest"
-          - "fedora:rawhide"
+          # - "fedora:rawhide"
           - "opensuse/tumbleweed:latest"
           - "ubuntu:devel"
     steps:
--- a/.github/workflows/test-distros.yml
+++ b/.github/workflows/test-distros.yml
@@ -79,14 +79,14 @@ jobs:
         with:
           persist-credentials: false
 
-      - name: Display Python version
-        run: python3 -c "import sys; print(sys.version); print(sys.platform)"
-
       - name: Install uv
         run: pip3 install uv --break-system-packages
 
       - name: Install Python dependencies
         run: uv sync --frozen
 
+      - name: Display Python version
+        run: uv run --frozen python -c "import sys; print(sys.version); print(sys.platform)"
+
       - name: Run tasks
-        run: uv run poe all
+        run: uv run --frozen poe all
--- a/.github/workflows/test-distros.yml
+++ b/.github/workflows/test-distros.yml
@@ -17,8 +17,8 @@ on:
 # 1. Start a docker container with (read-only) access to the source files
 # $ docker run -it -v $(pwd):/west:ro <container>
 #
-# 2. Install, depending on the OS, git, python-pip and curl, for example in Arch linux:
-# $ pacman -Syu --noconfirm git python-pip curl
+# 2. Install, depending on the OS, git, python and curl, for example in Arch linux:
+# $ pacman -Syu --noconfirm git python curl
 #
 # 3. Clone the repository to a temporary directory
 # $ git config --global --add safe.directory /west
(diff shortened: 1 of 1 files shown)


kartben (dismissed):

.github/workflows/test-distros.yml
```diff
@@ -22,13 +22,15 @@ on:
 #
 # 3. Clone the repository to a temporary directory
 # $ git config --global --add safe.directory /west
-# $ git clone -q /west /west-tmp
+# $ git clone -q /west /west-tmp && cd /west-tmp
 #
-# 4. Install tox
-# $ pip3 install tox --break-system-packages
+# 4. Install uv
+# $ pip3 install uv --break-system-packages
```
marc-hb: pipx should work in this case and it does not require breaking system packages.

.github/workflows/test-distros.yml
```diff
@@ -22,13 +22,15 @@ on:
 #
 # 3. Clone the repository to a temporary directory
 # $ git config --global --add safe.directory /west
-# $ git clone -q /west /west-tmp
+# $ git clone -q /west /west-tmp && cd /west-tmp
 #
-# 4. Install tox
-# $ pip3 install tox --break-system-packages
+# 4. Install uv
```
marc-hb:
```suggestion
# 4. Install uv. You can also try the version provided by your operating system.
```

Some distros ship `uv` now (e.g. https://packages.fedoraproject.org/pkgs/uv/uv/) so that's another option worth mentioning for not "breaking system packages"

marc-hb (commented):
I'm not sure I agree with the last commit "ci: only run the test suite on the distro matrix". I think it is useful to check that linters can run on any distro.

If some linter version ever causes some trouble in future, then we should 1) be aware 2) exclude just those linter versions in a non-distro specific way.

> "format" even rewrites files in place

Running "something ALL" should never rewrite files in place - CI or not. Rewriting files in place when you have uncommitted work in progress is a disaster, so it should happen only for very clearly identified commands like "format".

.github/workflows/test-distros.yml
pdgendt: But that is another dependency? The multi-distro-test was created to stay as close to the Python/venv provided by the OS. Not sure I agree here.

.github/workflows/test-distros.yml
pdgendt: We could, but I'd only go there if _all_ distros ship it.

marc-hb: > We could, but I'd only go there if all distros ship it.

I don't see why. This is just a suggestion in a comment.

There will always be some distros that don't include `uv` - there are literally hundreds of them and no way to tell.

(just remembered about https://repology.org/project/uv/versions)

.github/workflows/test-distros.yml
marc-hb: > But that is another dependency? 

Yes but that's much better than "breaking system packages" and `pipx` is really THE way to avoid this issue. These instructions are only a convenience for west developers (About 1.5 people? :D) , this is not a "getting started guide".

If you really want no dependency then:

```
curl -LsSf https://astral.sh/uv/install.sh | sh
```

I'm only half-joking: there should really be a link to:
https://docs.astral.sh/uv/getting-started/installation/

marc-hb (commented):
> I'm not sure I agree with the last commit "ci: only run the test suite on the distro matrix". I think it is useful to check that linters can run on any distro.

BTW this last commit looks pretty independent from the rest. Keep it for later to get the rest merged faster?

marc-hb (approved):

kartben (approved):

Issue #967, opened 2026-06-23
Add a config file for west to pick the zephyr workspace path from


matthesa-drg:
**Description**
For many development or CI setups, it would be very helpful for west to be able to configure the zephyr workspace path from a config file instead of environment variables only. This is especially useful, when the zephyr project is not located within the workspace folder.

> ```
> project/
>   <project source code>
> project_ws/
>   <workspace code>
> ```

For west to find the workspace folder now, you need to configure the `ZEPHYR_BASE` environment variable, which is sometimes cumbersome or not possible.

**Proposed Solution**
Introduce a config file like `.westenv`, which is used to configure the `ZEPHYR_BASE` variable.

marc-hb:
Just like `git` and many other tools, **the current directory** is what matters the most to west. This is what allows switching between different workspaces quickly and reliably. ZEPHYR_BASE is a legacy hack. 

Keep in mind west is not Zephyr-specific. `west build` and others are not part of west, they are only part of Zephyr.

If your application is outside the west workspace (https://docs.zephyrproject.org/latest/develop/application/index.html#zephyr-freestanding-app), then west does not care about it. Zephyr and `west build` do, but west alone does not.

`find_package(Zephyr)`and the CMake package registry are probably what you want, please confirm:
https://docs.zephyrproject.org/latest/build/zephyr_cmake_package.html#zephyr-cmake-package

> Introduce a config file like .westenv, which is used to configure the ZEPHYR_BASE variable.

How would west switch between west workspaces then?


EDIT: https://docs.zephyrproject.org/latest/develop/env_vars.html#env-vars-zephyrrc could also be useful.

matthesa-drg:
@marc-hb Sorry for the late reply.

Maybe I am missing something about west here, but if we assume the legacy hack `ZEPHYR_BASE` (any my proposed feature, which builds up on this) is gone, how would I tell west where to find the workspace? 
Looking at my example folder structure again, I would like to be able to execute project-specific west commands within the "project" folder. But that's currently not possible as it doesn't detect any workspace. This is only possible when I use west from within the "project_ws" folder.
Is this the intended way of working with west and thus no solution exists for my problem?

pdgendt:
[...]

PR #979, opened 2026-07-27
project: Add support for .westenv file (#967)

matthesa-drg:
This commit adds support for a .westenv file to configure the zephyr workspace. Fixes ticket #967.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -13,3 +13,4 @@ htmlcov/
 .dir-locals.el
 .venv/
 junit.xml
+.idea/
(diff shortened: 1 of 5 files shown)


pdgendt (changes requested):
We're stripping all Zephyr specific handling from the West repository, this would re-introduce such a mechanism.

marc-hb:
> Fixes ticket https://github.com/zephyrproject-rtos/west/issues/967.

We explained in 967 that this is not a way forward.

PR #982, opened 2026-07-29
project: simplify status output detection

### LINKED ISSUE #981: _has_nonempty_status() is racy
`_has_nonempty_status()` added by commit eff0e1ba2a2 is racing with git. If git is slow (e.g.: on Windows/NTFS), then it may wrongfully report an empty `git status`.

```python
        def has_output():
            # 'git status --porcelain' prints nothing if there
            # are no notable changes, so any output at all
            # means we should run 'git status' on the project.
            stdout, stderr = None, None
            try:
                stdout, stderr = popen.communicate(timeout=0.1). # <==== racing with git
            except subprocess.TimeoutExpired:
                pass
```

As of west 1.5/1.6, this function is used in `west status` and `west compare`.

cc: @yvesll , @mbolivar

pdgendt:
`_has_nonempty_status()` drove 'git status --porcelain' through a manual `Popen` with a repeated `communicate(timeout=0.1)` loop plus `popen.kill()`, added by commit eff0e1ba2a2d ("west status: only print output for projects with nonempty status") to bail out as soon as git printed any output.

That optimization never optimized anything: `communicate()` does not return already-available data, it blocks until the child closes its pipes (EOF), which normally means git has already exited. The early-abort could therefore never fire before the command finished, and the surrounding `poll()`/`communicate()` machinery only added complexity.

Run 'git status --porcelain' to completion via the existing `project.git()` helper and check whether it produced any output. The behaviour is unchanged; the output is bounded and the command is fast. `CalledProcessError` from real git failures is already handled by every caller.

Fixes: eff0e1ba2a2d ("west status: only print output for projects with nonempty status")
Fixes https://github.com/zephyrproject-rtos/west/issues/981

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -120,36 +120,13 @@ class _ProjectCommand(WestCommand):
         )
 
     def _has_nonempty_status(self, project):
-        # Check if the project has any status output to print. We
-        # manually use Popen in order to try to exit as quickly as
-        # possible if 'git status' prints anything.
-
-        popen = subprocess.Popen(
-            ['git', 'status', '--porcelain'],
-            stdout=subprocess.PIPE,
-            stderr=subprocess.PIPE,
-            cwd=project.abspath,
-        )
-
-        def has_output():
-            # 'git status --porcelain' prints nothing if there
-            # are no notable changes, so any output at all
-            # means we should run 'git status' on the project.
-            stdout, stderr = None, None
-            try:
-                stdout, stderr = popen.communicate(timeout=0.1)
-            except subprocess.TimeoutExpired:
-                pass
-            return stdout or stderr
-
-        while True:
-            if has_output():
-                popen.kill()
-                return True
-            if popen.poll() is not None:
-                break
-
-        return has_output()
+        # Check if the project has any status output to print.
+        #
+        # 'git status --porcelain' prints nothing if there are no
+        # notable changes, so any output at all means we should run
+        # 'git status' on the project.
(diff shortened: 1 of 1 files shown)


marc-hb:
From #981:

> Actually, I missed the retry loop and filed #981 too fast (I saw this timeout while looking at something else). Now I'm not so sure there is a bug.

pdgendt:
> From #981:
> 
> > Actually, I missed the retry loop and filed #981 too fast (I saw this timeout while looking at something else). Now I'm not so sure there is a bug.

I still think the simplification is justified?

> The early-abort could never fire early anyway: communicate() does not return already-available data, it blocks until the child closes its pipes (EOF), which normally means git has already exited.

kartben (approved):

marc-hb (changes requested):
> I still think the simplification is justified?

> The early-abort could never fire early anyway: 

After taking a closer look, I think this is right: it looks this optimization never optimized anything in the first place.

This also means it wasn't racy at all and that this PR is fixing a totally different bug than my bogus #981. Yet Claude's commit message happily claims to fix... both!? As it starts with a copy of my bug. The code removal looks OK to me now, but that commit message cannot be merged as is. Good reminder to never trust AI blindly.

pdgendt:
> The code removal looks OK to me now, but that commit message cannot be merged as is. Good reminder to never trust AI blindly.

Thanks, updated the commit message.

marc-hb (approved):

PR #984, opened 2026-07-30
main: remove special handling of ZEPHYR_BASE

pdgendt:
West is a general-purpose tool and should not rely on Zephyr internals. Drop the `ZEPHYR_BASE` special handling:

- west no longer sets the `ZEPHYR_BASE` environment variable, and the `zephyr.base` / `zephyr.base`-prefer config options are no longer used.
- `west_topdir()` no longer falls back to searching from `ZEPHYR_BASE`.

The public `west_topdir()` `fall_back` argument and the `-z/--zephyr-base` command line option are kept as ignored no-ops for backwards compatibility.

### DIFF
--- a/src/west/app/main.py
+++ b/src/west/app/main.py
@@ -24,7 +24,7 @@ import textwrap
 import traceback
 from collections import OrderedDict
 from io import StringIO
-from pathlib import Path, PurePath
+from pathlib import Path
 from subprocess import CalledProcessError
 from typing import NamedTuple
 
@@ -260,8 +260,6 @@ class WestApp:
         colorama.init()
 
         # See if we're in a workspace. It's fine if we're not.
-        # Note that this falls back on searching from ZEPHYR_BASE
-        # if the current directory isn't inside a west workspace.
         try:
             self.topdir = west_topdir()
         except WestNotFound:
@@ -547,9 +545,9 @@ class WestApp:
             '-z',
             '--zephyr-base',
             default=None,
-            help='''Override the Zephyr base directory. The
-                    default is the manifest project with path
-                    "zephyr".''',
+            help='''Deprecated and ignored. West no longer manages the
+                    Zephyr base directory; this option has no effect and
+                    will be removed in a future release.''',
         )
 
         parser.add_argument(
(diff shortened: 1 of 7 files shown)


src/west/app/main.py
```diff
@@ -779,136 +777,8 @@ def run_extension(self, name, argv):
...
 
-        # HACK: try to set ZEPHYR_BASE.
-        #
-        # Currently required by zephyr extensions like "west build".
-        #
-        # TODO: get rid of this. Instead:
-        #
-        # - support a WEST_DIR environment variable to specify the
-        #   workspace if we're not running under a .west directory
-        #   (controversial)
```
marc-hb, line 790: cc:
- #967

marc-hb:
Looks like great progress towards #246 but also likely to break some eggs... which ones exactly?

Could you summarize which use cases this would break versus not and try to gauge the "popularity" of each?

This change needs to be well advertised in any case.

> west_topdir() no longer falls back to searching from ZEPHYR_BASE.

Based on the discussions in #967, I think we need (at least) `west -C ...` to offer an alternative _before_ removing that feature.

pdgendt:
Most of Zephyr has been ready for this since 2020. The extension commands self-locate and don't need west to export `ZEPHYR_BASE`. There is **one** dependency that is currently masked by west setting `ZEPHYR_BASE`: application `find_package(Zephyr REQUIRED HINTS $ENV{ZEPHYR_BASE})` relies on the CMake user package registry (`west zephyr-export`) once the env var is gone. That's the egg most likely to break.

### What makes this possible (already in Zephyr)

- `scripts/west_commands/zephyr_ext_common.py` derives `ZEPHYR_BASE = os.environ.get('ZEPHYR_BASE', THIS_ZEPHYR)` where `THIS_ZEPHYR` is the repo root relative to the script. Every extension command imports from here, so they keep working with the env var unset. This was added *specifically to enable this removal* https://github.com/zephyrproject-rtos/zephyr/commit/e819fa46cd6
- `west twister` and standalone `scripts/twister` both set/derive `ZEPHYR_BASE` from `__file__` themselves.
- `ZephyrConfig.cmake` resolves the base four ways: `-DZEPHYR_BASE` → `$ENV{ZEPHYR_BASE}` → CMake user package registry → workspace-relative search.

### Breaks / changes behaviour

| Use case | Impact | Fix |
|---|---|---|
| `west build` in a workspace where the CMake package registry was **never** populated and `ZEPHYR_BASE` isn't set otherwise | `find_package(Zephyr)` fails: *"Could not find a package configuration file provided by Zephyr"*. Today west's auto-`ZEPHYR_BASE` hides this. | None, already in the [getting started guide](https://docs.zephyrproject.org/latest/develop/getting_started/index.html#get-zephyr-and-install-python-dependencies) that users should export using `west zephyr-export` |
| Running a west command from **inside the zephyr repo but outside a `.west` workspace**, relying on the `ZEPHYR_BASE`→topdir fallback | `WestNotFound` | This is the #967 discussion |
| `west flash`/`debug` on **Intel Cyclone V** | `intel_cyclonev.py:240,301` do `'--cd=' + os.environ.get('ZEPHYR_BASE')` with no fallback → `TypeError` | Easy fix https://github.com/zephyrproject-rtos/zephyr/pull/115043 |
| Scripts relying on `zephyr.base` / `zephyr.base-prefer` config or `-z/--zephyr-base` | Silently ignored (kept as no-ops) | Breaks, intentionally |

PR #986, opened 2026-07-31
alias: support early args in alias expansion

pdgendt:
An alias whose expansion starts with an early/global option (e.g. `-v`) previously failed: alias expansion assigned `alias.args[0]` as the command name without re-parsing early args, so the option was treated as an unknown command.

- Fix parsing of the `-z` early argument
The early parser never cleared the "expecting a value" flag, so `-z <path>` swallowed the rest of argv and no command name was found; `west -z /p <alias>` crashed with a KeyError.
- Handle long early arguments
`--help` and `--zephyr-base[=…]` are accepted by the top-level parser but were unknown to the early parser, so `west --zephyr-base /p <cmd>` errored on the path and `west --help <alias>` expanded the alias instead of describing it.
- Support early args in alias expansion
 Re-parse the expanded argv on each iteration, so early args coming from an alias (`-v`, `-h`, `-z`, `-V`, …) are handled instead of being taken as the command name.
- Replace the command name by position
The expansion looked the command name up in argv by value, which could splice over an earlier option value, e.g. `west -z <alias> <alias>`.
- Stop expanding aliases when help is requested
The check was done once before the loop, so an alias expanding to `-h` kept expanding and help described the final command instead of the alias.
- Apply verbosity from an alias to west logging
West's own log level is set up before the command line is fully known, so an alias's `-v` only reached the command, not the west API loggers.

See commits for more details.

### DIFF
--- a/src/west/app/main.py
+++ b/src/west/app/main.py
@@ -619,7 +619,10 @@ class WestApp:
                     if arg == early_args.command_name:
                         argv = argv[:i] + alias.args + argv[i + 1 :]
                         break
-                early_args = early_args._replace(command_name=alias.args[0])
+                # Re-parse the expanded argv so early args coming from the
+                # alias itself (e.g. "-v") are handled instead of being
+                # mistaken for the command name.
+                early_args = parse_early_args(argv)
 
         self.handle_early_arg_errors(early_args)
         args, unknown = self.west_parser.parse_known_args(args=argv)
--- a/src/west/app/main.py
+++ b/src/west/app/main.py
@@ -89,9 +89,9 @@ class EarlyArgs(NamedTuple):
     # - setting up log levels from the verbosity level
 
     # Expected arguments:
-    help: bool  # True if -h was given
-    version: bool  # True if -V was given
-    zephyr_base: str | None  # -z argument value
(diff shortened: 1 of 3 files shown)


marc-hb (commented):
I think "xfail" should be reserved for bugs for which we have at least some idea/plan of how they should get fixed, even if that idea is vague and in the very distant future. Also, in verbose mode each xfail should print at least one reference to the corresponding bug.

Can this ever be fixed? I mean fixed without rewriting west in Rust ;-). I'm not sure it can. EDIT: and that's OK, we can live with that alias limitation. aliases are only a convenience feature.

The alternative is to convert this to a "normal" test that asserts the outcome is the bad but expected one - with only a comment mentioning that this outcome sucks instead of "xfail". Such a "normal" test is also more restrictive because it will immediately detect and report any change (always a good thing) whereas an "xfail" test will happily keep xfailing with pretty much any output.

pdgendt:
> Can this ever be fixed? I mean fixed without rewriting west in Rust ;-). I'm not sure it can. EDIT: and that's OK, we can live with that alias limitation. aliases are only a convenience feature.

Converted this PR to fix the limitation rather than to keep it.

PR #988, opened 2026-08-03
Refresh dependencies and rationalise version constraints

pdgendt:
Refreshes `uv.lock` to the latest versions, allows the new majors of `packaging` and `mypy`, and makes the constraint policy consistent:

- **Runtime dependencies** use lower bounds only (`>=`). As a published
  library, west is installed alongside arbitrary other packages, so
  pre-emptive upper caps on common deps like `packaging`/`pyyaml` can make an
  environment unresolvable and force a new west release on every dependency
  major. Cap reactively on a known incompatibility instead.
- **Dev/test dependencies** keep upper bounds (normalised to `~=`). They are
  not installed by consumers, and capping them keeps CI reproducible so an
  unrelated upstream release cannot fail a green PR.

The `packaging` floor is lowered to `>=22.0` rather than kept at the inherited `25.0`: west only uses `packaging.version.parse`, and 22.0 is the release that removed `LegacyVersion`, so `parse()` raises `InvalidVersion` on malformed manifest versions as west expects.

Supersedes https://github.com/zephyrproject-rtos/west/pull/945, https://github.com/zephyrproject-rtos/west/pull/938, https://github.com/zephyrproject-rtos/west/pull/951


marc-hb (approved):
(This may sound surprising but I actually looked at this)

henrikbrixandersen (approved):

PR #989, opened 2026-08-04
app: suggest closest match for unknown commands

jostn:
Add a helpful suggestion when a user enters an unknown command by matching against built-ins, extensions, and aliases.

Example:
  west updat
  west: unknown command "updat"; do you need to run this inside a workspace? Did you mean "update"?

### DIFF
--- a/src/west/app/main.py
+++ b/src/west/app/main.py
@@ -12,6 +12,7 @@ Nothing in here is public API.
 '''
 
 import argparse
+import difflib
 import logging
 import os
 import platform
@@ -719,7 +720,35 @@ class WestApp:
             )
         else:
             extra_help = 'do you need to run this inside a workspace?'
-        self.print_usage_and_exit(f'west: unknown command "{command_name}"; {extra_help}')
+
+        suggestion = self.suggest_command(command_name)
+        suggestion_msg = f' Did you mean "{suggestion}"?' if suggestion else ''
+        self.print_usage_and_exit(
+            f'west: unknown command "{command_name}"; {extra_help}{suggestion_msg}'
+        )
+
+    def suggest_command(self, command_name):
+        candidates = set(self.builtins)
+        if self.extensions:
+            candidates.update(self.extensions)
+        if self.aliases:
+            candidates.update(self.aliases)
+        if not candidates:
+            return None
+
+        lower_to_name = {}
+        for name in candidates:
+            lower_to_name.setdefault(name.lower(), name)
+
+        lowered = command_name.lower()
+        if lowered in lower_to_name and command_name not in candidates:
+            return lower_to_name[lowered]
+
+        matches = difflib.get_close_matches(lowered, lower_to_name.keys(), n=1, cutoff=0.6)
+        if matches:
+            return lower_to_name[matches[0]]
+
+        return None
 
     def print_usage_and_exit(self, message):
         self.west_parser.print_usage(file=sys.stderr)
--- a/tests/test_main.py
+++ b/tests/test_main.py
@@ -3,7 +3,7 @@ import sys
 from pathlib import Path
 
 import pytest
-from conftest import cmd, cmd_subprocess
+from conftest import cmd, cmd_raises, cmd_subprocess
 
 import west.version
 
@@ -48,3 +48,10 @@ def test_module_run(tmp_path, monkeypatch):
     # check that that the sys.path was correctly inserted
     expected_path = Path(__file__).parents[1] / 'src'
     assert actual_path == [f'{expected_path}', 'initial-path']
+
+
+def test_unknown_command_suggests_closest_match():
+    exc_info, _ = cmd_raises('updat', SystemExit)
+
+    assert 'unknown command "updat"' in str(exc_info.value)
+    assert 'Did you mean "update"?' in str(exc_info.value)


src/west/app/main.py
```diff
@@ -719,7 +720,35 @@ def handle_unknown_command(self, command_name):
...
+        )
+
+    def suggest_command(self, command_name):
+        candidates = set(self.builtins)
+        if self.extensions:
+            candidates.update(self.extensions)
+        if self.aliases:
+            candidates.update(self.aliases)
+        if not candidates:
+            return None
```
marc-hb, line 737: It's good to be paranoid but you can drop this: I don't think `self.builtins` is ever going to be empty :-)

src/west/app/main.py
```diff
@@ -719,7 +720,35 @@ def handle_unknown_command(self, command_name):
...
+
+        lower_to_name = {}
```
marc-hb, line 739:
```suggestion
        lowers_to_name = {}
```

src/west/app/main.py
```diff
@@ -719,7 +720,35 @@ def handle_unknown_command(self, command_name):
...
+        for name in candidates:
+            lower_to_name.setdefault(name.lower(), name)
```
marc-hb, line 741: Why `setdefault()`? Candidates are unique because you are using a `set`. Unless some candidates differ only by case, but then I don't see how `setdefault()` would make a difference.

src/west/app/main.py
```diff
@@ -719,7 +720,35 @@ def handle_unknown_command(self, command_name):
...
+
+        lowered = command_name.lower()
+        if lowered in lower_to_name and command_name not in candidates:
```
marc-hb, line 744: I don't understand the `and command_name not in candidates` bit, I think that logic deserves at the very least a comment.

1. Either callers are never supposed to invoke this function when there is an exact, case-sensitive match (which is a waste of time). Then `and command_name not in candidates` achieves nothing. 

2. Or, (future) callers are allowed to be "lazy" for some unknown reason and invoke this when there is an exact, case-sensitive match. Then I don't really understand what happens here. Care to explain?

Note there is a single caller for now and AFAIK it will not call this with an exact match for now.

The 1. or 2. expectation/contract should be in the pydoc.

marc-hb: BTW it should be easy enough to make this function more generic and re-usable for other things than commands in the future (some parameters maybe?), I mean `fix_typo()` instead of `suggest_command()`. For that, move the first part of it that builds the "candidates" list of commands outside the function.

tests/test_main.py
```diff
@@ -48,3 +48,10 @@ def test_module_run(tmp_path, monkeypatch):
     # check that that the sys.path was correctly inserted
     expected_path = Path(__file__).parents[1] / 'src'
     assert actual_path == [f'{expected_path}', 'initial-path']
+
+
+def test_unknown_command_suggests_closest_match():
+    exc_info, _ = cmd_raises('updat', SystemExit)
+
+    assert 'unknown command "updat"' in str(exc_info.value)
+    assert 'Did you mean "update"?' in str(exc_info.value)
```
marc-hb, line 57: Thanks for adding one test! Please add a few more, especially for case (in)sensitivity coverage. EDIT: because case sensitivity is the area where this new function has the most "logic". `difflib` should have its own test coverage, no need to duplicate `difflib` test coverage here.

`@pytest.mark.parametrize` makes that very easy, look for existing examples in this repo.

marc-hb (changes requested):
Thanks for this! I was concerned about adding one extra dependency because `west` tries to have a very small set. Glad to discover that difflib is part of the standard library!

Comments below but before you go and address them, let's first make sure @pdgendt green lights the idea. I'd hate you wasting your time.

src/west/app/main.py
pdgendt: Do we even need the lowercase handling? For `difflib` matching an exact case-insensitive hit scores 1.0 already.

src/west/app/main.py
```diff
@@ -719,7 +720,35 @@ def handle_unknown_command(self, command_name):
             )
         else:
             extra_help = 'do you need to run this inside a workspace?'
-        self.print_usage_and_exit(f'west: unknown command "{command_name}"; {extra_help}')
+
+        suggestion = self.suggest_command(command_name)
+        suggestion_msg = f' Did you mean "{suggestion}"?' if suggestion else ''
+        self.print_usage_and_exit(
+            f'west: unknown command "{command_name}"; {extra_help}{suggestion_msg}'
+        )
```
pdgendt, line 728: Should we replace the `extra_help` with the suggestion? Or at least start with the hint.

pdgendt (commented):
> Comments below but before you go and address them, let's first make sure @pdgendt green lights the idea. I'd hate you wasting your time.

I do wonder what the added benefit is here. For West users, I don't think there's that much of a problem of "finding" the right command, as this will mostly point out a user's typo.

A more common issue, is that a workspace is in a state where extension commands are not being resolved, but again, a suggestion for a built-in or alias does not really help here.

That being said, it's a nice gimmick and maybe there's a case to be made for users that have many cryptic aliases (If the aliases are 3 characters or more, otherwise the threshold of 0.6 will not be met).

I'm not opposed to having this, but more test coverage and the comments from @marc-hb should be addressed.

marc-hb:
> That being said, it's a nice gimmick 

To be clear: I don't think it's a game changer either. But this is surprisingly little code that relies entirely on the standard library and that looks very low maintenance. So why not!

src/west/app/main.py
marc-hb: Interesting... I found nothing about case (in)sensitivity in https://docs.python.org/3/library/difflib.html !?

BTW, if you think case-sensitivity is a simple question, then take a look at these:
- https://www.b-list.org/weblog/2018/nov/26/case/
- https://dev.to/amartyajha/cve-2026-28292-how-a-simple-case-sensitivity-bug-turns-simple-git-into-a-remote-code-execution-4ki3
- https://lore.kernel.org/lkml/CAHk-=wjajMJyoTv2KZdpVRoPn0LFZ94Loci37WLVXmMxDbLOjg@mail.gmail.com/
- etc.

src/west/app/main.py
pdgendt: Next to that, the `command_name` itself should be discarded from the candidates, otherwise a self-referencing alias would print, well itself.

If `alias.foo = foo --bar` is set and `foo` does not exist:
```
west: unknown command "foo"; do you need to run this inside a workspace? Did you mean "foo"?
```

Simply add:
```Python
    candidates.discard(command_name)
```

marc-hb: Good catch @pdgendt . I think this demonstrates the need for a lot more test coverage :-)

pdgendt:
@jostn any feedback, will you update the PR as per the comments?

PR #990, opened 2026-08-06
project: fix TypeError in 'update --stats --keep-descendants'

pdgendt:
The keep-descendants branch of `Update.update()` measures how long the 'git status' invocation takes, but the elapsed time was computed from `perf_counter` without calling it, subtracting a float from the function object. As a result 'west update --stats --keep-descendants' crashed with a `TypeError` whenever a descendant branch was left checked out.
Add the missing call parentheses.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -1841,7 +1841,7 @@ class Update(_ProjectCommand):
                 start = perf_counter()
             project.git('status')
             if take_stats:
-                stats['get current status'] = perf_counter - start
+                stats['get current status'] = perf_counter() - start
         elif try_rebase:
             # Attempt a rebase.
             self.inf(f'west update: rebasing to {MANIFEST_REV} {sha}')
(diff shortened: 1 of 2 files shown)


marc-hb (commented):
Good catch but I'm concerned about long-term test organization and maintainability. Would it be possible to fix the test coverage gap by tweaking an existing test instead? Rather than adding yet another one.

The documentation of the added test "Regression test: 'west update --stats --keep-descendants' used to crash with a TypeError" looks more like a git log and a test description. It should just say "provide test coverage for X" because it does not really matter what exact bug was present in the formerly not covered code.

pdgendt:
> Good catch but I'm concerned about long-term test organization and maintainability. Would it be possible to fix the test coverage gap by tweaking an existing test instead? Rather than adding yet another one.
> 
> The documentation of the added test "Regression test: 'west update --stats --keep-descendants' used to crash with a TypeError" looks more like a git log and a test description. It should just say "provide test coverage for X" because it does not really matter what exact bug was present in the formerly not covered code.

Updated an existing test about how west update treats local branches carrying local commits instead of adding a new one. Also reworded the commit message.

marc-hb (approved):

PR #992, opened 2026-08-07
Feature/add topdir overwrite command

### LINKED ISSUE #967: Add a config file for west to pick the zephyr workspace path from

matthesa-drg:
This PR adds the `-C` argument to west, which functions similar to the [command in git](https://git-scm.com/docs/git#Documentation/git.txt--Cpath)

Partially fixes #967

### DIFF
--- a/src/west/util.py
+++ b/src/west/util.py
@@ -54,7 +54,22 @@ class WestNotFound(RuntimeError):
     '''Neither the current directory nor any parent has a west workspace.'''
 
 
-def west_dir(start: PathType | None = None) -> str:
+_topdir_override: str | None = None
+
+
+def set_topdir_override(path: PathType | None) -> None:
+    '''Force west_topdir() to always return *path* instead of searching.
+
+    Implements the ``west -C`` option: once set, every west_topdir() call
+    in the process (unless it passes ``no_overwrite=True``) returns this
+    directory instead of searching from a start directory, cwd, or
+    ZEPHYR_BASE. Pass None to clear the override.
+    '''
+    global _topdir_override
+    _topdir_override = os.fspath(path) if path is not None else None
+
+
+def west_dir(start: PathType | None = None, no_overwrite: bool = False) -> str:
     '''Returns the absolute path of the workspace's .west directory.
 
     Starts the search from the start directory, and goes to its
@@ -63,14 +78,22 @@ def west_dir(start: PathType | None = None) -> str:
 
     Raises WestNotFound if no .west directory is found.
     '''
-    return os.path.join(west_topdir(start), WEST_DIR)
+    return os.path.join(west_topdir(start, no_overwrite=no_overwrite), WEST_DIR)
 
 
-def west_topdir(start: PathType | None = None, fall_back: bool = True) -> str:
+def west_topdir(
+    start: PathType | None = None, fall_back: bool = True, no_overwrite: bool = False
+) -> str:
     '''
     Like west_dir(), but returns the path to the parent directory of the .west/
     directory instead, where project repositories are stored
     '''
+    if not no_overwrite and _topdir_override is not None:
+        override_dir = Path(_topdir_override)
+        if (override_dir / WEST_DIR).is_dir():
+            return os.fspath(override_dir)
+        raise WestNotFound(f'-C path is not a west workspace: {override_dir}')
+
     cur_dir = Path(start or os.getcwd())
 
     while True:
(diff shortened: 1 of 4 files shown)


.gitignore
```diff
@@ -13,3 +13,4 @@ htmlcov/
 .dir-locals.el
 .venv/
 junit.xml
+.idea/
```
marc-hb: `west/.gitignore` is not the place for personal preferences, please drop this commit. You must configure a global gitignore file in your home directory. After you put this there, it will automatically apply to all git clones at once. Having everyone pollute every git repo with their personal tools and preferences does not scale.

Also, this is unrelated to topdir

marc-hb (commented):
This does not seem enough because `west_topdir()` is used in other places. So these other places now ignore `-C`, is that OK? Probably a case-by-case answer.

I see two possibilities:
- Either `west_topdir()` is kept as the "authoritative" source. Then the new code `-C` code must be moved there. I don't know whether that's possible or a good idea.
- Or, `west_topdir()` is renamed to `west_searchtopdir()` and every call site is carefully inspected to see if it now must take the new `-C` parameter into account or ignore it.

All this while keeping in the back of our heads that it would be nice to also have some sort of global west config topdir fallback in the future.

marc-hb:
> This does not seem enough because west_topdir() is used in other places. So these other places now ignore -C, is that OK? Probably a case-by-case answer.

As a sheer coincidence I just found this 7 years-old, unmerged PR of mine, please take a look:
- #275

Should it be resurrected?

matthesa-drg:
I took a look at your suggestions and figured out that moving the topdir overwrite into `west_topdir()` is probably the better idea (IMO) as this function is responsible for getting the path to the current topdir.
I had to add the `no_overwrite` flag that prevents this as commands like `west init` use the `west_topdir()` function to check whether a folder already contains a workspace. In this case we want it to ignore the overwrite.

I also took a look at the PR you linked and to me it looks like it has a different goal.

src/west/util.py
```diff
@@ -54,7 +54,22 @@ class WestNotFound(RuntimeError):
     '''Neither the current directory nor any parent has a west workspace.'''
 
 
-def west_dir(start: PathType | None = None) -> str:
+_topdir_override: str | None = None
```
pdgendt, line 57: Sorry, but globals are not acceptable IMO. Also see https://github.com/zephyrproject-rtos/west/issues/149

matthesa-drg: Okay. But then it rules out the solution to move this into the `west_topdir` function without changing its signature significantly, right?

src/west/util.py
```diff
@@ -63,14 +78,22 @@ def west_dir(start: PathType | None = None) -> str:
 
     Raises WestNotFound if no .west directory is found.
     '''
-    return os.path.join(west_topdir(start), WEST_DIR)
+    return os.path.join(west_topdir(start, no_overwrite=no_overwrite), WEST_DIR)
 
 
-def west_topdir(start: PathType | None = None, fall_back: bool = True) -> str:
+def west_topdir(
+    start: PathType | None = None, fall_back: bool = True, no_overwrite: bool = False
```
marc-hb: No "negative" booleans and double negations please. Default values do not have to be "False".

```suggestion
    start: PathType | None = None, fall_back: bool = True, overwrite: bool = True
```

src/west/util.py
marc-hb: There is a global `WestApp` object (with a reference to #149 in it :-)) Maybe you could add a field there? Just thinking out loud sorry.

src/west/util.py
```diff
@@ -63,14 +78,22 @@ def west_dir(start: PathType | None = None) -> str:
...
+) -> str:
     '''
     Like west_dir(), but returns the path to the parent directory of the .west/
     directory instead, where project repositories are stored
     '''
+    if not no_overwrite and _topdir_override is not None:
```
marc-hb, line 91:
```suggestion
    if overwrite and _topdir_override is not None:
```
No "negative" booleans and double negations please. Default values do not have to be "False".

Not a big fan of the name "overwrite", too vague. Can you be a bit more specific, for instance:

```suggestion
    if cli_arg_override and _topdir_arg_override is not None:
```
Just a suggestion, hope you can find better.

marc-hb (commented):
Considering how pervasive the "topdir" is across the entire codebase, I've been wondering whether we shouldn't rather run the ENTIRE test suite twice: once without `-C` and another time with `-C`...

PR #993, opened 2026-08-24
project: fix cached update of a project containing nested projects

mniestroj:
The manifest format allows one project's path to contain another project's path: _check_paths_are_unique() only rejects projects that share the exact same path. Zephyr's west.yml relies on this today, with bsim at tools/bsim and the babblesim components under tools/bsim/components/.

Without a cache such a layout updates fine, because init_project() runs 'git init', which happily initializes a repository in a directory that already exists and is not empty. With any cache configured the same code path runs 'git clone' instead, and git clone refuses a non-empty destination:

    === updating bsim (tools/bsim):
    --- bsim: cloning from /cache-zephyrproject/tools/bsim
    fatal: destination path '/builds/.../tools/bsim' already exists and
    is not an empty directory.
    ERROR: update failed for project bsim

Whether this triggers depends on the update order. Manifests are updated in order, so it fails whenever a nested project is listed before the project whose path contains it. In Zephyr's alphabetically sorted manifest every babblesim_* component sorts before bsim, so enabling the babblesim group together with any cache fails deterministically.

Keep 'git clone' as the normal path: cloning from a local cache hardlinks objects, which is what makes caching fast. When the destination is not empty, fall back to the 'git init' sequence and seed the new repository from the cache with a fetch, so its objects still come from the cache rather than from the network. With the default smart fetch strategy a SHA revision present in the cache then needs no network access at all, exactly as for a cloned project.

Fixes #934

Assisted-by: Claude:claude-opus-5

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -1981,6 +1981,71 @@ class Update(_ProjectCommand):
             self.dbg(f'{project.name}: update auto-cache ({cache_dir}) with remote')
             project.git(['remote', 'update', '--prune'], cwd=cache_dir, check=False)
 
+    def init_project_in_place(self, project):
+        # init_project() helper. Create the project's repository at
+        # project.abspath and set up the convenience remote, without
+        # fetching anything.
+
+        init_cmd = ['init', project.abspath]
+        # Silence the very verbose and repetitive init.defaultBranch
+        # warning (10 lines per new git clone). The branch
+        # 'placeholder' will never have any commit so it will never
+        # actually exist.
+        if self.git_version_info >= (2, 28, 0):
+            init_cmd.insert(1, '--initial-branch=init_placeholder')
+
+        project.git(init_cmd, cwd=self.topdir)
+
+        # This remote is added as a convenience for the user.
+        # However, west always fetches project data by URL, not name.
+        # The user is therefore free to change the URL of this remote.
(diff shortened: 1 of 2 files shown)


src/west/app/project.py
```diff
@@ -1981,6 +1981,55 @@ def handle_auto_cache(self, project):
...
+            )
+            if cp.returncode == 0:
+                refspecs.append(cp.stdout.decode('ascii').strip())
+
+        git_dir = Path(project.abspath) / '.git'
+        remove_git_dir = not os.path.lexists(git_dir)
+        seeded = False
+        try:
+            self.init_project_in_place(project)
+            project.git(['-c', 'protocol.version=2', 'fetch', '--tags', '--', cache_dir, *refspecs])
```
pdgendt: Nit; the `protocol.version` argument could use a comment why this is needed, and maybe should be guarded with a git version check (`>= 2.18.0` from 2018?) similar to the `--initial-branch` above.

pdgendt (commented):
Overal LGTM, could you split the change from the tests into separate commits?

src/west/app/project.py
mniestroj: Added a check on `>= 2.18.0` and tested with git `2.15.0` to make sure it works. In the latter case west will fallback to network access instead of doing a fetch from local cache.

PR #994, opened 2026-08-26
project: implement mirroring for "west init/update"

steverae-work:
The concept is to emulate a smart "git insteadof" feature
in order to reduce the rate-limit and networking issues
when accessing upstream repositories.

If the upstream repo has been mirrored to the URL
defined by the "west config url.insteadof" configuration,
then fetch from the mirror "instead of" from upstream
repository.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -50,6 +50,10 @@ class _ProjectCommand(WestCommand):
     # Helper class which contains common code needed by various commands
     # in this file.
 
+    def __init__(self, *args, **kwargs):
+        super().__init__(*args, **kwargs)
+        self._url_insteadof_mappings = None
+
     def _parser(self, parser_adder, **kwargs):
         # Create and return a "standard" parser.
 
@@ -208,17 +212,15 @@ class _ProjectCommand(WestCommand):
 
         self.inf(result, colorize=False)
 
-    def _resolve_mirror_url(self, url, cwd=None):
-        # Shared by Init and Update to rewrite 'url' using the
-        # 'url.insteadof' configuration option, if set. The value must
-        # be a JSON array of strings in the form "remote=mirror". Each mapping is
-        # tried in order, and the first mirror that 'git ls-remote --exit-code'
-        # reports as reachable is returned. Falls back to the original 'url'
-        # if no mapping matches or no mirror is reachable. If 'cwd' is set,
-        # probing runs in that working directory.
+    def _parsed_url_insteadof_mappings(self):
+        if self._url_insteadof_mappings is not None:
+            return self._url_insteadof_mappings
+
+        # Parse and validate this once, since update can resolve many URLs.
         url_insteadof_config = self.config.get('url.insteadof')
         if not url_insteadof_config:
-            return url
+            self._url_insteadof_mappings = ()
+            return self._url_insteadof_mappings
 
         try:
             url_mappings = json.loads(url_insteadof_config)
@@ -234,6 +236,7 @@ class _ProjectCommand(WestCommand):
                 '"remote>mirror" strings'
             )
 
+        parsed_mappings = []
         for mapping in url_mappings:
             if not isinstance(mapping, str):
                 self.die(
@@ -252,6 +255,20 @@ class _ProjectCommand(WestCommand):
                     f'invalid url.insteadof mapping entry {mapping!r}: empty remote or mirror value'
                 )
 
+            parsed_mappings.append((remote, mirror))
+
+        self._url_insteadof_mappings = tuple(parsed_mappings)
+        return self._url_insteadof_mappings
+
+    def _resolve_mirror_url(self, url, cwd=None):
+        # Shared by Init and Update to rewrite 'url' using the
+        # 'url.insteadof' configuration option, if set. The value must
+        # be a JSON array of strings in the form "remote=mirror". Each mapping is
+        # tried in order, and the first mirror that 'git ls-remote --exit-code'
+        # reports as reachable is returned. Falls back to the original 'url'
+        # if no mapping matches or no mirror is reachable. If 'cwd' is set,
+        # probing runs in that working directory.
+        for remote, mirror in self._parsed_url_insteadof_mappings():
             if not url.startswith(remote):
                 continue
 
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -2241,8 +2241,9 @@ class Update(_ProjectCommand):
         # Fetch the workspace repository from the freshly synced auto-cache
         # when auto-cache is enabled; otherwise fetch directly from the remote URL
         # or the url.insteadof mirror URL.
-        fetch_url = self.project_auto_cache(project) or project.url
-        fetch_url = self._resolve_mirror_url(fetch_url, cwd=project.abspath)
+        fetch_url = self.project_auto_cache(project)
+        if not fetch_url:
+            fetch_url = self._resolve_mirror_url(project.url, cwd=project.abspath)
 
         # -f is needed to avoid errors in case multiple remotes are
         # present, at least one of which contains refs that can't be
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -7,6 +7,7 @@
 
 import argparse
 import hashlib
+import json
 import os
 import shlex
 import shutil
@@ -206,6 +207,71 @@ class _ProjectCommand(WestCommand):
             self.die(f'invalid format string {shlex.quote(fmt)}')
 
         self.inf(result, colorize=False)
+
+    def _resolve_mirror_url(self, url, cwd=None):
+        # Shared by Init and Update to rewrite 'url' using the
+        # 'url.insteadof' configuration option, if set. The value must
+        # be a JSON array of strings in the form "remote=mirror". Each mapping is
+        # tried in order, and the first mirror that 'git ls-remote --exit-code'
+        # reports as reachable is returned. Falls back to the original 'url'
+        # if no mapping matches or no mirror is reachable. If 'cwd' is set,
+        # probing runs in that working directory.
+        url_insteadof_config = self.config.get('url.insteadof')
+        if not url_insteadof_config:
+            return url
+
+        try:
+            url_mappings = json.loads(url_insteadof_config)
+        except json.JSONDecodeError:
+            self.die(
+                f'invalid url.insteadof configuration: expected JSON array of "remote=mirror" '
+                f'strings, got {url_insteadof_config!r}'
+            )
+
+        if not isinstance(url_mappings, list):
+            self.die(
+                'invalid url.insteadof configuration: expected JSON array of '
+                '"remote=mirror" strings'
+            )
+
+        for mapping in url_mappings:
+            if not isinstance(mapping, str):
+                self.die(
+                    'invalid url.insteadof configuration: each entry must be a '
+                    '"remote=mirror" string'
+                )
+
+            remote, sep, mirror = mapping.partition('=')
+            if not sep:
+                self.die(f'invalid url.insteadof mapping entry {mapping!r}: missing "=" separator')
+
+            remote = remote.strip()
+            mirror = mirror.strip()
+            if not remote or not mirror:
+                self.die(
+                    f'invalid url.insteadof mapping entry {mapping!r}: empty remote or mirror value'
+                )
+
+            remote = remote.rstrip('/') + '/'
+            mirror = mirror.rstrip('/') + '/'
+            if not url.startswith(remote):
+                continue
+
+            mirror_url = mirror + url.removeprefix(remote)
+            result = subprocess.run(
+                ['git', '-c', 'core.askPass=true', 'ls-remote', '--exit-code', mirror_url],
+                check=False,
+                capture_output=True,
+                text=True,
+                cwd=cwd,
+            )
+            if not result.returncode:
+                self.inf(f"    [info] using mirror: {mirror_url}")
+                return mirror_url
+            else:
+                self.dbg(f'    [debug] mirror was specified, but "{mirror_url}" was not found...')
+
+        return url
 
 
 class Init(_ProjectCommand):
@@ -582,6 +648,7 @@ below.
         shutil.rmtree(tempdir2)
 
         # Clone the manifest repository into a temporary directory.
+        manifest_url = self._resolve_mirror_url(manifest_url)
         try:
             self.small_banner(
                 f'Cloning manifest repository from {manifest_url}'
@@ -1958,8 +2025,9 @@ class Update(_ProjectCommand):
             cache_dir_parent = Path(cache_dir).parent
             cache_dir_parent.mkdir(parents=True, exist_ok=True)
             self.dbg(f'{project.name}: create auto-cache for {project.url} in {cache_dir}')
+            _url = self._resolve_mirror_url(project.url, cwd=cache_dir_parent)
             project.git(
-                ['clone', '--mirror', '--', project.url, os.fspath(cache_dir)], cwd=cache_dir_parent
+                ['clone', '--mirror', '--', _url, os.fspath(cache_dir)], cwd=cache_dir_parent
             )
             self.create_auto_cache_info(project, cache_dir)
         else:
@@ -2154,9 +2222,11 @@ class Update(_ProjectCommand):
         # automatically fetch the auto-cache so that it is up-to-date
         self.handle_auto_cache(project)
 
-        # fetch workspace repository from freshly synced auto-cache if
-        # auto-cache is used. Otherwise fetch directly from remote url.
+        # Fetch the workspace repository from the freshly synced auto-cache
+        # when auto-cache is enabled; otherwise fetch directly from the remote URL
+        # or the url.insteadof mirror URL.
         fetch_url = self.project_auto_cache(project) or project.url
+        fetch_url = self._resolve_mirror_url(fetch_url, cwd=project.abspath)
 
         # -f is needed to avoid errors in case multiple remotes are
         # present, at least one of which contains refs that can't be
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -241,7 +241,7 @@ class _ProjectCommand(WestCommand):
                     '"remote=mirror" string'
                 )
 
-            remote, sep, mirror = mapping.partition('=')
+            remote, sep, mirror = mapping.partition('>')
             if not sep:
                 self.die(f'invalid url.insteadof mapping entry {mapping!r}: missing "=" separator')
 
@@ -252,8 +252,6 @@ class _ProjectCommand(WestCommand):
                     f'invalid url.insteadof mapping entry {mapping!r}: empty remote or mirror value'
                 )
 
-            remote = remote.rstrip('/') + '/'
-            mirror = mirror.rstrip('/') + '/'
             if not url.startswith(remote):
                 continue
 
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -265,10 +265,10 @@ class _ProjectCommand(WestCommand):
                 env={**os.environ, 'GIT_TERMINAL_PROMPT': '0'},
             )
             if not result.returncode:
-                self.inf(f"    [info] using mirror: {mirror_url}")
+                self.inf(f"using mirror: {mirror_url}")
                 return mirror_url
             else:
-                self.dbg(f'    [debug] mirror was specified, but "{mirror_url}" was not found...')
+                self.dbg(f'mirror was specified, but "{mirror_url}" was not found...')
 
         return url
 
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -257,11 +257,12 @@ class _ProjectCommand(WestCommand):
 
             mirror_url = mirror + url.removeprefix(remote)
             result = subprocess.run(
-                ['git', '-c', 'core.askPass=true', 'ls-remote', '--exit-code', mirror_url],
+                ['git', 'ls-remote', '--exit-code', mirror_url],
                 check=False,
                 capture_output=True,
                 text=True,
                 cwd=cwd,
+                env={**os.environ, 'GIT_TERMINAL_PROMPT': '0'},
             )
             if not result.returncode:
                 self.inf(f"    [info] using mirror: {mirror_url}")
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -224,26 +224,26 @@ class _ProjectCommand(WestCommand):
             url_mappings = json.loads(url_insteadof_config)
         except json.JSONDecodeError:
             self.die(
-                f'invalid url.insteadof configuration: expected JSON array of "remote=mirror" '
+                f'invalid url.insteadof configuration: expected JSON array of "remote>mirror" '
                 f'strings, got {url_insteadof_config!r}'
             )
 
         if not isinstance(url_mappings, list):
             self.die(
                 'invalid url.insteadof configuration: expected JSON array of '
-                '"remote=mirror" strings'
+                '"remote>mirror" strings'
             )
 
         for mapping in url_mappings:
             if not isinstance(mapping, str):
                 self.die(
                     'invalid url.insteadof configuration: each entry must be a '
-                    '"remote=mirror" string'
+                    '"remote>mirror" string'
                 )
 
             remote, sep, mirror = mapping.partition('>')
             if not sep:
-                self.die(f'invalid url.insteadof mapping entry {mapping!r}: missing "=" separator')
+                self.die(f'invalid url.insteadof mapping entry {mapping!r}: missing ">" separator')
 
             remote = remote.strip()
             mirror = mirror.strip()
--- a/tests/test_url_insteadof.py
+++ b/tests/test_url_insteadof.py
@@ -278,37 +278,6 @@ def test_url_insteadof_no_prefix_match(repos_tmpdir):
     assert (workspace / 'net-tools' / '.git').check(dir=1)
 
 
-def test_url_insteadof_with_trailing_slashes(repos_tmpdir):
-    """Test that url.insteadof handles trailing slashes correctly."""
-    remotes = repos_tmpdir / 'repos'
-    mirrors = repos_tmpdir / 'mirrors'
-    mirrors.mkdir()
-
-    original_net_tools = remotes / 'net-tools'
-    mirror_net_tools = mirrors / 'net-tools'
-    subprocess.check_call(
-        ['git', 'clone', '--mirror', str(original_net_tools), str(mirror_net_tools)]
-    )
-
-    workspace = repos_tmpdir / 'workspace'
-    manifest = remotes / 'zephyr'
-
-    cmd(['init', '-m', str(manifest), str(workspace)], env={'ZEPHYR_BASE': None})
-
-    # Test with and without trailing slashes (should be normalized)
-    cmd(
-        ['config', 'url.insteadof', json.dumps([mapping_entry(f'{remotes}/', f'{mirrors}/')])],
-        cwd=workspace,
-    )
-    log_config(cwd=workspace)
-
-    output = cmd('update', cwd=workspace)
-
-    assert 'using mirror:' in output
-    normalized_output = output.replace('\\', '/')
-    assert str(mirror_net_tools).replace('\\', '/') in normalized_output
-
-
 def test_url_insteadof_first_working_mirror_wins(repos_tmpdir):
     """Test that first reachable mirror is used when multiple match."""
     remotes = repos_tmpdir / 'repos'
@@ -367,21 +336,6 @@ def test_url_insteadof_empty_config(repos_tmpdir):
     # Should work normally
     assert (workspace / 'net-tools' / '.git').check(dir=1)
     assert 'irror' not in output
-
-
-def test_url_insteadof_invalid_json_fails(repos_tmpdir):
-    """Test that malformed url.insteadof JSON fails the command."""
-    remotes = repos_tmpdir / 'repos'
-    workspace = repos_tmpdir / 'workspace'
-    manifest = remotes / 'zephyr'
-
-    cmd(['init', '-m', str(manifest), str(workspace)], env={'ZEPHYR_BASE': None})
-
-    cmd(['config', 'url.insteadof', '["missing-end"'], cwd=workspace)
-    log_config(cwd=workspace)
-
-    with pytest.raises(SystemExit):
-        cmd('update', cwd=workspace)
 
 
 @pytest.mark.parametrize(
--- a/tests/test_url_insteadof.py
+++ b/tests/test_url_insteadof.py
@@ -62,7 +62,7 @@ def test_url_insteadof_init_manifest_clone(repos_tmpdir):
 
 def test_url_insteadof_prevents_credential_prompt(repos_tmpdir):
     """Test that url.insteadof probe doesn't trigger credential prompts."""
-    # This is a regression test to ensure -c core.askPass=true is used
+    # This is a regression test to ensure GIT_TERMINAL_PROMPT=0 is used.
     remotes = repos_tmpdir / 'repos'
     workspace = repos_tmpdir / 'workspace'
     manifest = remotes / 'zephyr'
--- a/tests/test_url_insteadof.py
+++ b/tests/test_url_insteadof.py
@@ -18,7 +18,7 @@ def log_config(cwd=None):
 
 
 def mapping_entry(remote, mirror):
-    return f'{remote}={mirror}'
+    return f'{remote}>{mirror}'
 
 
 def test_url_insteadof_init_manifest_clone(repos_tmpdir):
@@ -387,14 +387,14 @@ def test_url_insteadof_invalid_json_fails(repos_tmpdir):
 @pytest.mark.parametrize(
     'config_value, expected_error',
     [
-        (json.dumps({'remote1': 'mirror1'}), 'expected JSON array of "remote=mirror" strings'),
+        (json.dumps({'remote1': 'mirror1'}), 'expected JSON array of "remote>mirror" strings'),
         (
             json.dumps([{'remote1': 'mirror1'}]),
-            'each entry must be a "remote=mirror" string',
+            'each entry must be a "remote>mirror" string',
         ),
-        (json.dumps(['remote1-mirror1']), 'missing "=" separator'),
-        (json.dumps(['=mirror1']), 'empty remote or mirror value'),
-        (json.dumps(['remote1=']), 'empty remote or mirror value'),
+        (json.dumps(['remote1=mirror1']), 'missing ">" separator'),
+        (json.dumps(['>mirror1']), 'empty remote or mirror value'),
+        (json.dumps(['remote1>']), 'empty remote or mirror value'),
     ],
 )
 def test_url_insteadof_invalid_mapping_format_fails(repos_tmpdir, config_value, expected_error):
--- a/tests/test_url_insteadof.py
+++ b/tests/test_url_insteadof.py
@@ -1,0 +1,548 @@
+# Copyright (c) 2026, Infineon Technologies
+# SPDX-License-Identifier: Apache-2.0
+
+"""Tests for url.insteadof configuration feature."""
+
+import json
+import subprocess
+
+import pytest
+from conftest import (
+    cmd,
+    cmd_raises,
+)
+
+
+def dump_url_insteadof_config(cwd=None):
+    print(cmd('config -l', cwd=cwd).rstrip())
+
+
+def mapping_entry(remote, mirror):
+    return f'{remote}={mirror}'
+
+
+def test_url_insteadof_init_manifest_clone(repos_tmpdir):
+    """Test that url.insteadof works during 'west init' manifest clone."""
+    remotes = repos_tmpdir / 'repos'
+    mirrors = repos_tmpdir / 'mirrors'
+    mirrors.mkdir()
+
+    # Mirror the manifest repository (zephyr)
+    original_zephyr = remotes / 'zephyr'
+    mirror_zephyr = mirrors / 'zephyr'
+    subprocess.check_call(['git', 'clone', '--mirror', str(original_zephyr), str(mirror_zephyr)])
+
+    workspace = repos_tmpdir / 'workspace'
+
+    # Set url.insteadof in global config before init
+    # (init reads global config even before workspace exists)
+    remotes_url = 'file://' + str(remotes).replace('\\', '/')
+    mirrors_url = 'file://' + str(mirrors).replace('\\', '/')
+    cmd(
+        [
+            'config',
+            '--global',
+            'url.insteadof',
+            json.dumps([mapping_entry(remotes_url, mirrors_url)]),
+        ]
+    )
+    dump_url_insteadof_config()
+
+    try:
+        # Now run init, which should use the mirror for cloning the manifest
+        cmd(['init', '-m', str(original_zephyr), str(workspace)], env={'ZEPHYR_BASE': None})
+
+        # Verify manifest was cloned
+        assert (workspace / 'zephyr' / '.git').check(dir=1)
+        assert (workspace / 'zephyr' / 'west.yml').check(file=1)
+    finally:
+        # Clean up global config
+        cmd(['config', '--global', '-d', 'url.insteadof'])
+
+
+def test_url_insteadof_prevents_credential_prompt(repos_tmpdir):
+    """Test that url.insteadof probe doesn't trigger credential prompts."""
+    # This is a regression test to ensure -c core.askPass=true is used
+    remotes = repos_tmpdir / 'repos'
+    workspace = repos_tmpdir / 'workspace'
+    manifest = remotes / 'zephyr'
+
+    cmd(['init', '-m', str(manifest), str(workspace)], env={'ZEPHYR_BASE': None})
+
+    # Configure mirror pointing to nonexistent HTTPS repo (would normally prompt)
+    cmd(
+        [
+            'config',
+            'url.insteadof',
+            json.dumps([mapping_entry(str(remotes), 'https://example.com/nonexistent')]),
+        ],
+        cwd=workspace,
+    )
+    dump_url_insteadof_config(cwd=workspace)
+
+    # This should complete without hanging on credential prompt
+    # The test will timeout if credential prompt blocks
+    cmd('update', cwd=workspace)
+
+    # Should fall back to original since mirror doesn't exist
+    assert (workspace / 'net-tools' / '.git').check(dir=1)
+
+
+def test_url_insteadof_basic_prefix_match(repos_tmpdir):
+    """Test url.insteadof with a simple prefix replacement."""
+    # Setup: create a mirror of net-tools in a different location
+    remotes = repos_tmpdir / 'repos'
+    mirrors = repos_tmpdir / 'mirrors'
+    mirrors.mkdir()
+
+    # Clone net-tools to mirror location
+    original_net_tools = remotes / 'net-tools'
+    mirror_net_tools = mirrors / 'net-tools'
+    subprocess.check_call(
+        ['git', 'clone', '--mirror', str(original_net_tools), str(mirror_net_tools)]
+    )
+
+    # Create workspace with url.insteadof config
+    workspace = repos_tmpdir / 'workspace'
+    manifest = remotes / 'zephyr'
+
+    # Initialize workspace
+    cmd(['init', '-m', str(manifest), str(workspace)], env={'ZEPHYR_BASE': None})
+
+    # Configure url.insteadof to redirect from remotes to mirrors
+    cmd(
+        ['config', 'url.insteadof', json.dumps([mapping_entry(str(remotes), str(mirrors))])],
+        cwd=workspace,
+    )
+    dump_url_insteadof_config(cwd=workspace)
+
+    # Run update - should use mirror for net-tools
+    output = cmd('update', cwd=workspace)
+
+    # Verify mirror was used (normalize paths for Windows/Unix compatibility)
+    assert 'using mirror:' in output
+    # Normalize both paths to use forward slashes for comparison
+    expected_mirror = str(mirror_net_tools).replace('\\', '/')
+    normalized_output = output.replace('\\', '/')
+    assert expected_mirror in normalized_output
+
+    # Verify project was cloned successfully
+    assert (workspace / 'net-tools' / '.git').check(dir=1)
+
+
+def test_url_insteadof_basic_prefix_match_json_array(repos_tmpdir):
+    """Test url.insteadof with a JSON array of remote=mirror strings."""
+    remotes = repos_tmpdir / 'repos'
+    mirrors = repos_tmpdir / 'mirrors'
+    mirrors.mkdir()
+
+    original_net_tools = remotes / 'net-tools'
+    mirror_net_tools = mirrors / 'net-tools'
+    subprocess.check_call(
+        ['git', 'clone', '--mirror', str(original_net_tools), str(mirror_net_tools)]
+    )
+
+    workspace = repos_tmpdir / 'workspace'
+    manifest = remotes / 'zephyr'
+
+    cmd(['init', '-m', str(manifest), str(workspace)], env={'ZEPHYR_BASE': None})
+
+    cmd(
+        [
+            'config',
+            'url.insteadof',
+            json.dumps(
+                [
+                    mapping_entry('https://example.invalid/', 'https://mirror.invalid/'),
+                    mapping_entry(str(remotes), str(mirrors)),
+                ]
+            ),
+        ],
+        cwd=workspace,
+    )
+    dump_url_insteadof_config(cwd=workspace)
+
+    output = cmd('update', cwd=workspace)
+
+    assert 'using mirror:' in output
+    assert str(mirror_net_tools).replace('\\', '/') in output.replace('\\', '/')
+    assert (workspace / 'net-tools' / '.git').check(dir=1)
+
+
+def test_url_insteadof_nonexistent_mirror_fallback(repos_tmpdir):
+    """Test that nonexistent mirrors fall back to original URL."""
+    remotes = repos_tmpdir / 'repos'
+    workspace = repos_tmpdir / 'workspace'
+    manifest = remotes / 'zephyr'
+
+    # Initialize workspace
+    cmd(['init', '-m', str(manifest), str(workspace)], env={'ZEPHYR_BASE': None})
+
+    # Configure url.insteadof with nonexistent mirror path
+    nonexistent = repos_tmpdir / 'does-not-exist'
+    cmd(
+        ['config', 'url.insteadof', json.dumps([mapping_entry(str(remotes), str(nonexistent))])],
+        cwd=workspace,
+    )
+    dump_url_insteadof_config(cwd=workspace)
+
+    # Run update - should fall back to original URL
+    output = cmd('update', cwd=workspace)
+
+    # Verify mirror was tried but original URL was used
+    assert 'was not found' in output or 'using mirror:' not in output
+
+    # Verify projects were still cloned successfully from original
+    assert (workspace / 'net-tools' / '.git').check(dir=1)
+    assert (workspace / 'subdir' / 'Kconfiglib' / '.git').check(dir=1)
+
+
+def test_url_insteadof_multiple_mappings(repos_tmpdir):
+    """Test url.insteadof with multiple JSON mappings."""
+    remotes = repos_tmpdir / 'repos'
+    mirrors1 = repos_tmpdir / 'mirrors1'
+    mirrors2 = repos_tmpdir / 'mirrors2'
+    mirrors1.mkdir()
+    mirrors2.mkdir()
+
+    # Create two different mirrors for different repos
+    original_net_tools = remotes / 'net-tools'
+    original_kconfiglib = remotes / 'Kconfiglib'
+    mirror1_net_tools = mirrors1 / 'net-tools'
+    mirror2_kconfiglib = mirrors2 / 'Kconfiglib'
+
+    subprocess.check_call(
+        ['git', 'clone', '--mirror', str(original_net_tools), str(mirror1_net_tools)]
+    )
+    subprocess.check_call(
+        ['git', 'clone', '--mirror', str(original_kconfiglib), str(mirror2_kconfiglib)]
+    )
+
+    workspace = repos_tmpdir / 'workspace'
+    manifest = remotes / 'zephyr'
+
+    cmd(['init', '-m', str(manifest), str(workspace)], env={'ZEPHYR_BASE': None})
+
+    # Configure multiple mappings with different repo-specific prefixes
+    # The manifest uses file:// URLs, so we need to mirror those
+    remotes_url = 'file://' + str(remotes).replace('\\', '/')
+    net_tools_url = remotes_url + '/net-tools'
+    kconfiglib_url = remotes_url + '/Kconfiglib'
+    mirror1_net_tools_url = 'file://' + str(mirror1_net_tools).replace('\\', '/')
+    mirror2_kconfiglib_url = 'file://' + str(mirror2_kconfiglib).replace('\\', '/')
+
+    mirror_config = [
+        mapping_entry(net_tools_url, mirror1_net_tools_url),
+        mapping_entry(kconfiglib_url, mirror2_kconfiglib_url),
+    ]
+    cmd(
+        ['config', 'url.insteadof', json.dumps(mirror_config)],
+        cwd=workspace,
+    )
+    dump_url_insteadof_config(cwd=workspace)
+
+    cmd('update', cwd=workspace)
+
+    # Verify that at least the projects were cloned successfully
+    # (mirrors may or may not be used depending on exact URL matching)
+    assert (workspace / 'net-tools' / '.git').check(dir=1)
+    assert (workspace / 'subdir' / 'Kconfiglib' / '.git').check(dir=1)
+
+
+def test_url_insteadof_no_prefix_match(repos_tmpdir):
+    """Test url.insteadof when no mapping matches the URL prefix."""
+    remotes = repos_tmpdir / 'repos'
+    workspace = repos_tmpdir / 'workspace'
+    manifest = remotes / 'zephyr'
+
+    cmd(['init', '-m', str(manifest), str(workspace)], env={'ZEPHYR_BASE': None})
+
+    # Configure url.insteadof with unrelated prefix
+    cmd(
+        [
+            'config',
+            'url.insteadof',
+            json.dumps([mapping_entry('https://github.com/', 'https://mirror.example.com/')]),
+        ],
+        cwd=workspace,
+    )
+    dump_url_insteadof_config(cwd=workspace)
+
+    # Run update - should use original URLs since prefix doesn't match
+    output = cmd('update', cwd=workspace)
+
+    # No mirror messages should appear
+    assert 'using mirror:' not in output
+
+    # Projects should still be cloned from original
+    assert (workspace / 'net-tools' / '.git').check(dir=1)
+
+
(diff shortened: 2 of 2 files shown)


src/west/app/project.py
```diff
@@ -582,6 +582,18 @@ def bootstrap(self, args) -> Path:
         shutil.rmtree(tempdir2)
 
         # Clone the manifest repository into a temporary directory.
+        github_mirror = os.environ.get('GITHUB_MIRROR')
```
steverae-work: @pdgendt   (  this cannot be a "west config" as this is in the "west init" command... )

steverae-work:
I'm confused - the documentation for this is in a different repo:
zephyr/doc/develop/west/built-in.rst
- correct?

src/west/app/project.py
pdgendt: Kind of, the configuration options have 3 scopes: system, global or local. You can't have it in a local/workspace scope that you're initializing.
There wouldn't be an issue for `west update`.

I also have been wanting to add an early argument to west to pass configuration options (similar to [git -c](https://git-scm.com/docs/git#Documentation/git.txt--cnamevalue)), which could be used here. This would also close the gap for a `insteadof`-like feature here.

src/west/app/project.py
```diff
@@ -582,6 +582,18 @@ def bootstrap(self, args) -> Path:
...
+        if github_mirror:
+            if manifest_url.startswith("https://github.com"):
+                mirror_manifest_url = f"{github_mirror}" + manifest_url.removeprefix("https://github.com")
```
pdgendt: Other than being hosted on Github, West has no relation to any git server, nor should it. This hard-coded mechanism is not acceptable IMO.

pdgendt (changes requested):
> I'm confused - the documentation for this is in a different repo: zephyr/doc/develop/west/built-in.rst
> 
> * correct?

Yes, I need to pick up
- https://github.com/zephyrproject-rtos/west/pull/860

steverae-work:
- updated to remove GITHUB_MIRROR env variable
- documents in: https://github.com/zephyrproject-rtos/zephyr/pull/117653

src/west/app/project.py
steverae-work: got it - thanks

src/west/app/project.py
steverae-work: got it - thanks

src/west/app/project.py
```diff
@@ -582,6 +582,23 @@ def bootstrap(self, args) -> Path:
         shutil.rmtree(tempdir2)
 
         # Clone the manifest repository into a temporary directory.
+        url_mirror_config = self.config.get('blobs.url-mirror')
```
pdgendt: "blobs" is a term used in the Zephyr extension `west blobs` and also has no relation to west itself.

By the way, as west uses `git` under the hood, you _can already_ specify an `insteadof` globally:

```shell
$ git config --global url."https://gitlab.intra/silabs-mirror/".insteadOf "https://github.com/"
```

rettichschnidi:
> The concept is to emulate a smart "git insteadof" feature in order to reduce the rate-limit and networking issues when accessing "github.com".
> 
> If the "github.com" repo has been mirrored to the URL defined by the GITHUB_MIRROR environment variable, then fetch from the mirror "instead of" from "github.com".

I fail to see the need for the "smart" part of this PR.

In real world, I'd assume that either one is mirroring _all_ Zephyr repos or _none_, so a simple, catch-all `git insteadOf` would do. And if one wants to mirror only selective repositories, some slightly more scoped `git insteadOf`s would do the trick.

marc-hb (dismissed):
Unfortunately, when there is a single commit message then you need to keep the PR description and commit message synchronized manually. This is due to how the zephyr project uses Github:

https://docs.zephyrproject.org/latest/contribute/guidelines.html#contribution-workflow

> Note that this implementation is not limited to "github.com"
as it supports mirroring from any specified URL.

If this is not github-specific (which is good), then don't mention github in the commit message and don't make it look like it is.

If you search "insteadof" in https://github.com/zephyrproject-rtos/west then you instantly find #615 where this topic was already discussed. Please make sure you absorb all past discussions and of course refer to them in the commit message.

Now about the code itself, I'm pretty amazed that https://github.com/zephyrproject-rtos/west/pull/994/changes/0c3fb727d3f9829d420b07bf3f2b20f89ceed7ea includes 3 nearly identical copies of the same code, duplicated in 3 different places. Isn't this something Claude is supposed to trivially avoid?

src/west/app/project.py
steverae-work: what "west config" value would you suggest which covers both the "west init/update" feature and the "west blobs" feature?

thanks in advance

steverae-work:
> > The concept is to emulate a smart "git insteadof" feature in order to reduce the rate-limit and networking issues when accessing "github.com".
> > If the "github.com" repo has been mirrored to the URL defined by the GITHUB_MIRROR environment variable, then fetch from the mirror "instead of" from "github.com".
> 
> I fail to see the need for the "smart" part of this PR.
> 
> In real world, I'd assume that either one is mirroring _all_ Zephyr repos or _none_, so a simple, catch-all `git insteadOf` would do. And if one wants to mirror only selective repositories, some slightly more scoped `git insteadOf`s would do the trick.

https://github.com/orgs/zephyrproject-rtos/repositories?type=all
currently has 182 repositories
Our company mirror currently mirrors 34 of them,
because there is a cost to maintaining mirrors ( which *also* does hit the github.com rate-limits too )

And even if we did mirror them all, whenever a new repo is added upstream, the "git insteadOf" feature
would break in all of our pipeline jobs, until that repo was mirrored too...

IMO, while possible to create 34 "git insteadOf" configurations, it is not very practical; so to add a "smart" capability ( with a fallback ) to "west" seemed to be a good idea - hence, I created this PR

... thanks for your comment

steverae-work:
> If this is not github-specific (which is good), then don't mention github in the commit message and don't make it look like it is.

The issue _is_ all about the GitHub rate-limiting and networking issues that we are regularly facing - but I was asked to make the code more generic ( which is a good idea )

steverae-work:
> If you search "insteadof" in https://github.com/zephyrproject-rtos/west then you instantly find #615 where this topic was already discussed. Please make sure you absorb all past discussions and of course refer to them in the commit message.

thanks for this

However, I see a discussion that has been going on for years without a solution....

marc-hb:
> The issue is all about the GitHub rate-limiting and networking issues that we are regularly facing - 

Rate-limiting is not specific to Github at all. If you don't rate-limit on the Internet then you die. https://lwn.net/Articles/1008897/ etc.

> but I was asked to make the code more generic ( which is a good idea )

I'm merely asking to keep the commit message and PR description reasonably synchronized with the code. At least for major "upgrades" like this one.

> thanks for this
> However, I see a discussion that has been going on for years without a solution....

Can you solve it?

steverae-work:
> Now about the code itself, I'm pretty amazed that [0c3fb72](https://github.com/zephyrproject-rtos/west/commit/0c3fb727d3f9829d420b07bf3f2b20f89ceed7ea) includes 3 nearly identical copies of the same code, duplicated in 3 different places. Isn't this something Claude is supposed to trivially avoid?

yes - that is true

Originally, there were only two:
- one for the "west init" which used the "subprocess.run()" function, and
- one for the "west update" which used the "project.git()" function

and there was no support for a semi-colon separated list ....
( so each block was ~ 8 lines )

Then, testing revealed and the "handle_auto_cache()" required this code too

Then, there was a request to handle a semi-colon separated list, so the code grew....

However, I chose to prioritize the KISS principle over the DRY principle in this case

I would welcome other comments - thanks

steverae-work:
> I'm merely asking to keep the commit message and PR description reasonably synchronized with the code. At least for major "upgrades" like this one.

OK -- I will work on that

steverae-work:
> > thanks for this
> > However, I see a discussion that has been going on for years without a solution....
> 
> Can you solve it?

I have no idea
it seems like it is really a different issue that this one
( but I can look at it more carefully in the coming weeks .... )

marc-hb:
> However, I chose to prioritize the KISS principle over the DRY principle in this case

I really don't see how duplicating the same code three times _in the same file_ helps with KISS, sorry. I agree that "a little copying is (sometimes) better than a little dependency" but that does not seem to apply here.

steverae-work:
OK
- I have updated the git commit message to drop the references to "github.com"
- I still need to change the "west config blobs.url-mirror" to something more appropriate

steverae-work:
OK
- removed the duplicated code
- I still need to change the "west config blobs.url-mirror" to something more appropriate

steverae-work:
> * I still need to change the "west config blobs.url-mirror" to something more appropriate

AI suggests:

A [blobs.*](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20Visual%20Studio%20Code/e4c7e7b1d6/resources/app/out/vs/code/electron-browser/workbench/workbench.html) namespace implies blob-specific scope, but this option affects [west init](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20Visual%20Studio%20Code/e4c7e7b1d6/resources/app/out/vs/code/electron-browser/workbench/workbench.html)/west update (core, not blob-specific) too. A better name would live in a namespace that reflects its true purpose — generic URL rewriting for any git fetch, similar to git's own url.<base>.insteadOf concept.

Suggestion: [url.mirror](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20Visual%20Studio%20Code/e4c7e7b1d6/resources/app/out/vs/code/electron-browser/workbench/workbench.html) (or [url.mirrors](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20Visual%20Studio%20Code/e4c7e7b1d6/resources/app/out/vs/code/electron-browser/workbench/workbench.html) if you want to signal it's a list)

Rationale:

- No existing [url.*](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20Visual%20Studio%20Code/e4c7e7b1d6/resources/app/out/vs/code/electron-browser/workbench/workbench.html) section in west config today, so it doesn't collide with [manifest.*](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20Visual%20Studio%20Code/e4c7e7b1d6/resources/app/out/vs/code/electron-browser/workbench/workbench.html), [update.*](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20Visual%20Studio%20Code/e4c7e7b1d6/resources/app/out/vs/code/electron-browser/workbench/workbench.html), [build.*](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20Visual%20Studio%20Code/e4c7e7b1d6/resources/app/out/vs/code/electron-browser/workbench/workbench.html), grep.*, [zephyr.*](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20Visual%20Studio%20Code/e4c7e7b1d6/resources/app/out/vs/code/electron-browser/workbench/workbench.html).
- It's short, and mirrors the naming/spirit of Git's own [url.<base>.insteadOf](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20Visual%20Studio%20Code/e4c7e7b1d6/resources/app/out/vs/code/electron-browser/workbench/workbench.html), which anyone familiar with Git config will recognize immediately.
- It doesn't imply blob-only or init/update-only scope — accurately reflects that it's consulted by [Init](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20Visual%20Studio%20Code/e4c7e7b1d6/resources/app/out/vs/code/electron-browser/workbench/workbench.html), [Update](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20Visual%20Studio%20Code/e4c7e7b1d6/resources/app/out/vs/code/electron-browser/workbench/workbench.html), and [west blobs](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20Visual%20Studio%20Code/e4c7e7b1d6/resources/app/out/vs/code/electron-browser/workbench/workbench.html) alike.


... I think I like ```url.mirrors``` the best

@pdgendt - what do you think?

src/west/app/project.py
```diff
@@ -207,6 +207,38 @@ def delay(func, project):
...
+        # Falls back to the original 'url' if no mirror matches or none
+        # are reachable.
+        url_mirror_config = self.config.get('blobs.url-mirror')
+        if not url_mirror_config:
+            return url
+
+        for url_mirror in url_mirror_config.split(';'):
+            src, dst = url_mirror.split('=')
+            src = src.rstrip('/') + '/'
+            dst = dst.rstrip('/') + '/'
```
marc-hb: Can this leverage Python's built-in support for JSON instead of manually parsing?
```
[url]
insteadof = { "remote1" : "mirror1", "remote2" : "mirror2" }
```

src/west/app/project.py
```diff
@@ -207,6 +207,38 @@ def delay(func, project):
...
+            if not url.startswith(src):
+                continue
+
+            mirror_url = dst + url.removeprefix(src)
+            result = ls_remote(
+                ['ls-remote', '--exit-code', mirror_url.replace("https://", "https://:@")]
```
marc-hb: I don't see why any of this feature should be hardcoded to https.

marc-hb (commented):
I just spent a fair amount of time (re-)reading #615 and a good few more.

I think this should simply mimic git's `insteadOf`. The implementation would be much smaller and simpler while being more generic at the same time and support not just mirroring but other use cases like overriding (for branches).

I really don't like the "fallback" behavior. It's too non-deterministic and it's really hard to get right. For instance, it only considers `git ls-remote` failures right now, not sure why.

In any case, this needs some basic test coverage. While significantly different, #869 has some tests that could be used as a starting point.

marc-hb:
> The issue is all about the GitHub rate-limiting and networking issues that we are regularly facing - but I was asked to make the code more generic ( which is a good idea )

NOT questioning the value of this PR but... you know that `west` also supports caching, right?
```
west update --help

CACHING
-------
Projects are typically initialized by fetching from remote URLs.
However, they can also be cloned from local caches on the file system.
Three types of caches are supported, and they may be used together.
```

thorsten-klein:
For our use case, the approach from this PR does not scale because the mirror URL needs to be inherited by downstream projects. With the `west` config approach, each user would have to configure the option separately. The same applies to Git’s `insteadOf` feature. If I understand correctly it also does not work if manifests use `url:`, but only if they use `remote:`

This is why we have been using the approach from #869 for the past nine months, and it has been working well for us. We specify the mirrors in our category manifest, and all downstream projects inherit the URLs automatically. There is no need for downstream users to configure anything in Git or `west` themselves.

I’m therefore considering reopening #869, especially since it seems that more users need this functionality now. At this point, #869 still seems like the better approach to me.

marc-hb:
> With the west config approach, each user would have to configure the option separately. [...] There is no need for downstream users to configure anything in Git or west themselves.

I'm sorry but this really looks like a non-issue to me and certainly not one worth messing with manifest imports. Long reply in
- https://github.com/zephyrproject-rtos/west/pull/995#pullrequestreview-5061791601

marc-hb:
> If I understand correctly it also does not work if manifests use url:, but only if they use `remote:`

It could be nice to also support bare `url:` if (and only if) doable with limited complexity but that does not feel like a important use case either. Breaking down a manifest `url` seems like a very small effort/requirement to enable mirroring.

src/west/app/project.py
steverae-work: sorry - it is not hard-coded to "https" ...
but if "https" _is_ used, then it rewrites the URL to include ":@" so that the "git ls-remote" command does not prompt for a user name and password

src/west/app/project.py
steverae-work: do I understand this request correctly?

you are suggesting to change:
```
west config --global     url.mirrors   "remote1=mirror1;remote2=mirror2"
```
to
```
west config --global     url.insteadof   '{ "remote1" : "mirror1", "remote2" : "mirror2" }'
```
... so that it can be processed as JSON rather than parsing the string?

( I do not see any other references to "json" in this repo - is this a new thing? )

steverae-work:
In this latest commit:
- used "url.mirrors" for the "west config" command ( since there were no objections )
- used "-c core.askPass=true" ( instead of the hack ) to prevent "git ls-remote" prompting for username and password when the repo does not exist
- fixed issue which broken ```--name-cache``` and ```--path-cache```

src/west/app/project.py
steverae-work: however, this hack was reworked in the current commit...

steverae-work:
this commit fixes the "poe gh-format" issues

steverae-work:
these changes:
- add pytests ( as created by GitHub Copilot:claude-sonnet-4.5 )
- fix the poe format in that "tests/test_url_mirrors.py" file

src/west/app/project.py
marc-hb: > west config --global     url.insteadof   '{ "remote1" : "mirror1", "remote2" : "mirror2" }'

Yes exactly that.

I was also surprised that JSON seems never used yet in this repo. It's standard, universal, available or even built-in most languages (including Python) and it's coming with #904 anyway.

steverae-work:
this change:
- fixes the poe lint issues in the "tests/test_url_mirrors.py" file

src/west/app/project.py
marc-hb: The parsing and network accesses should be separated, so parsing problems "fail fast" before attempting any network access (and some minutes-long TCP timeout). I mean invalid input should cause `west` to abort ASAP.

Re-parsing every time for every url is a waste; it should be done either lazily or just once at startup.

src/west/app/project.py
```diff
@@ -207,6 +207,36 @@ def delay(func, project):
...
+        url_mirror_config = self.config.get('url.mirrors')
+        if not url_mirror_config:
+            return url
+
+        for url_mirror in url_mirror_config.split(';'):
+            src, dst = url_mirror.split('=')
+            src = src.rstrip('/') + '/'
+            dst = dst.rstrip('/') + '/'
+            if not url.startswith(src):
+                continue
```
marc-hb, line 273: git insteadof uses (only) the longest match, which seems very useful.

src/west/app/project.py
```diff
@@ -207,6 +207,36 @@ def delay(func, project):
...
+
+            mirror_url = dst + url.removeprefix(src)
+            result = ls_remote(['-c', 'core.askPass=true', 'ls-remote', '--exit-code', mirror_url])
+            if not result.returncode:
+                self.inf(f"    [info] using mirror: {mirror_url}")
+                return mirror_url
+            else:
+                self.dbg(f'    [debug] mirror was specified, but "{mirror_url}" was not found...')
```
```suggestion
                self.wrn(f'{mirror_url} failed: {result}')
```
It's expected to have configured mirrors unreachable sometimes. It's not expected to have them unreachable all the time.

src/west/app/project.py
```diff
@@ -207,6 +207,36 @@ def delay(func, project):
...
+            src, dst = url_mirror.split('=')
+            src = src.rstrip('/') + '/'
+            dst = dst.rstrip('/') + '/'
+            if not url.startswith(src):
+                continue
+
+            mirror_url = dst + url.removeprefix(src)
+            result = ls_remote(['-c', 'core.askPass=true', 'ls-remote', '--exit-code', mirror_url])
+            if not result.returncode:
+                self.inf(f"    [info] using mirror: {mirror_url}")
```
marc-hb: Did you ever look at these printed messages?
```suggestion
                self.inf(f"using mirror: {mirror_url}")
```

src/west/app/project.py
```diff
@@ -207,6 +207,36 @@ def delay(func, project):
 
         self.inf(result, colorize=False)
 
+    def _resolve_mirror_url(self, url, ls_remote):
```
marc-hb: I don't understand why a lambda argument and a fairly large number of variable git options are needed to merely test whether some remote locations are valid and up. Can you explain?

src/west/app/project.py
```diff
@@ -2157,6 +2203,12 @@ def fetch(self, project, stats, take_stats):
         # fetch workspace repository from freshly synced auto-cache if
         # auto-cache is used. Otherwise fetch directly from remote url.
         fetch_url = self.project_auto_cache(project) or project.url
+        fetch_url = self._resolve_mirror_url(
```
marc-hb: Why would you use a mirror when you have a cache?

marc-hb (commented):
I'm sorry to ask but did you carefully review every line written by Claude?

src/west/app/project.py
steverae-work: all caches need to be seeded the first time;
this provides for seeding from a mirror

src/west/app/project.py
steverae-work: because some calls use "subprocess.run()" and others use "project.git()" - which take different options

src/west/app/project.py
steverae-work: I did, and to me, these looked good in the logs

```
=== Initializing in /builds/RaeSteven/test_c1
    [info] using mirror: https://gitlab.intra.infineon.com/public-mirror/github.com/zephyrproject-rtos/zephyr
--- Cloning manifest repository from https://gitlab.intra.infineon.com/public-mirror/github.com/zephyrproject-rtos/zephyr
```

```
=== updating hal_infineon (modules/hal/infineon):
--- hal_infineon: initializing
Initialized empty Git repository in /builds/RaeSteven/test_c1/modules/hal/infineon/.git/
--- hal_infineon: fetching, need revision 13e4b3cd70fb2d862a37e55665f03af09fe74201
    [info] using mirror: https://gitlab.intra.infineon.com/public-mirror/github.com/zephyrproject-rtos/hal_infineon
warning: redirecting to https://gitlab.intra.infineon.com/public-mirror/github.com/zephyrproject-rtos/hal_infineon.git/
From https://gitlab.intra.infineon.com/public-mirror/github.com/zephyrproject-rtos/hal_infineon    
```

```
=== updating hal_ti (modules/hal/ti):
--- hal_ti: initializing
Initialized empty Git repository in /builds/RaeSteven/test_c1/modules/hal/ti/.git/
--- hal_ti: fetching, need revision 2e6514333cdb59e44a7635fb5852035c325f4c11
    [debug] mirror was specified, but "https://gitlab.intra.infineon.com/public-mirror/github.com/zephyrproject-rtos/hal_ti" was not found...
From https://github.com/zephyrproject-rtos/hal_ti
```

src/west/app/project.py
steverae-work: in our company, the mirrors are maintained by a separate team, so if they chose to take a month to create a (new) mirror, then this job should simply report that the expected mirror is still missing, and allow the fall-back to be used.

src/west/app/project.py
steverae-work: OK -- but this feature intentionally reads them in order, so the order of the configuration does matter ( if multiple items are specified )

steverae-work:
this change switches to JSON formatted configuration
- west config --global url.insteadof '{ "remote1" : "mirror1", "remote2" : "mirror2" }'

steverae-work:
this change:
- renames all tests to "test_url_insteadof_....()"

src/west/app/project.py
marc-hb: I guess what I'm asking is why do these options matter to merely test whether some server location is valid and up.

src/west/app/project.py
marc-hb: Fair enough.

So my JSON recommendation should be a list and not a dictionary. EDIT: Python guarantees ordered dictionaries since 3.7 but JSON apparently not.

src/west/app/project.py
marc-hb: I'm not sure what and where but something is wrong.

When I run `west -vvv list`, the DEBUG prefix is added automatically. If you look at `git grep self.dbg`, you can see there is never a need to manually prefix with the log level. This is obviously something that every logging system does automatically for you.

Now what most logging systems also offer is customization, including the ability to drop log level prefixes if you desire. Has this prefix been somehow turned off on your system?

Weird.

src/west/app/project.py
marc-hb: Then the comment above is confusing and should probably be adjusted?

I'm not familiar with the caching implementation but it's pretty obvious that caching and mirroring can potentially interact in "interesting" ways so we need to be extra careful and clear there. Ideally, some test coverage using both at the same time could help everyone sleep better.

marc-hb (commented):
Thanks for the prompt answers!

steverae-work:
> So my JSON recommendation should be a list and not a dictionary.

you are asking for this, correct?
```
west config --global url.insteadof '[ {"remote1": "mirror1"}, {"remote2": "mirror2"} ]'
```

src/west/app/project.py
```diff
@@ -207,6 +208,50 @@ def delay(func, project):
...
+        # attribute. Falls back to the original 'url' if no mapping
+        # matches, the configuration is invalid, or no mirror is reachable.
+        url_insteadof_config = self.config.get('url.insteadof')
+        if not url_insteadof_config:
+            return url
+
+        try:
+            url_mappings = json.loads(url_insteadof_config)
+        except json.JSONDecodeError:
+            self.dbg(f'    [debug] ignoring invalid url.insteadof JSON: {url_insteadof_config!r}')
```
marc-hb: - Errors should not be logged at the `dbg` level
- There is no reason to keep going, west should abort. Asking the user to provide a valid JSON list is not a demanding requirement at all and should not be optional. You can even drop the `try/except` here if the default JSON error message is good enough.

src/west/app/project.py
```diff
@@ -207,6 +208,50 @@ def delay(func, project):
...
+            return url
+
+        if not isinstance(url_mappings, dict):
+            self.dbg('    [debug] ignoring non-object url.insteadof configuration')
+            return url
+
+        for remote, mirror in url_mappings.items():
+            if not isinstance(remote, str) or not isinstance(mirror, str):
+                self.dbg('    [debug] ignoring non-string url.insteadof mapping entry')
```
marc-hb: See above. The whole point of using a library is not to have to write all this error-handling code. If the default messages are good enough then you can drop this.

src/west/app/project.py
marc-hb: If the mirrors do not exist yet, then why does anyone try to use them? ~This does not make sense sorry.~

Anyway a warning will not stop anything, the fallback will still work. It will only highlight the misconfiguration.

marc-hb (commented):
> you are asking for this, correct?
> `west config --global url.insteadof '[ {"remote1": "mirror1"}, {"remote2": "mirror2"} ]'`

I didn't think that far, sorry :-( I was only asking for something ordered in JSON (because you want ordering to be a feature) and JSON order requires a list. That would work and I'm OK with it but others might object with the number of quotes and delimiters.

This middle-ground could be an alternative:
```
west config -- global url.insteadof '[ "remote1>mirror1", "remote2>mirror2" ]'
```
Note a colon `:` would not work because someone may want to do this: `[ "https://external_server>http://internal_mirror" ]`

`>` is explictly forbidden in URIs:
https://datatracker.ietf.org/doc/html/rfc2396#section-2.4.3
`%` and `#` are also safe.

I don't have a strong opinion as long as some existing parser (and its built-in error messages) is leveraged.

src/west/app/project.py
steverae-work: yes - correct

steverae-work:
this change:
- uses: ``` west config --global url.insteadof '[ "remote1=mirror1", "remote2=mirror2" ]' ```
- updates a misleading comment
- errors and fails on invalid json configuration
- adds a debug log to the tests/* scripts

src/west/app/project.py
steverae-work: according to Copilot:
```
What json.loads handles:

Syntax parsing only, at [project.py:226](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20Visual%20Studio%20Code/e4c7e7b1d6/resources/app/out/vs/code/electron-browser/workbench/workbench.html).
It can tell you whether the text is valid JSON.

What lines 233-255 are doing:

Enforcing your custom schema and semantics, at [project.py:233](vscode-file://vscode-app/c:/Program%20Files/Microsoft%20Visual%20Studio%20Code/e4c7e7b1d6/resources/app/out/vs/code/electron-browser/workbench/workbench.html).
Validating type is array of strings.
Validating each string contains = and non-empty remote/mirror parts.

A JSON library call by itself cannot validate remote=mirror semantics inside each string. You would still need custom checks unless you introduce a schema/validation layer (for example jsonschema or pydantic), which is likely overkill here.
```

steverae-work:
> If the mirrors do not exist yet, then why does anyone try to use them? This does not make sense sorry.

In order to handle the case were _some_ mirrors exist, but not _all_ mirrors exist.
( if all mirrors exist, then just use the "git insteadof" feature ... )

see: https://github.com/zephyrproject-rtos/west/pull/994#issuecomment-5454745459

src/west/app/project.py
steverae-work: done in the current commit

steverae-work:
> I'm not familiar with the caching implementation but it's pretty obvious that caching and mirroring can potentially interact in "interesting" ways so we need to be extra careful and clear there. Ideally, some test coverage using both at the same time could help everyone sleep better.

all three caches essentially work the same way
1. code tries to access the data from the cache,
2. if the data does not exist, then the data is seeded into the cache
3. code accesses the data in the cache

So, this feature only affects step 2, by using either the mirror.url or falling back to the original url

src/west/app/project.py
steverae-work: done - fixed the comment

steverae-work:
regarding the debug logs:
- it seems that the "_logger.debug()" prepends content like: "west.manifest: DEBUG: "
- but the self.dbg() does not update the message

see: ./src/west/app/main.py
```
 182 class LogFormatter(logging.Formatter):
 183     def __init__(self):
 184         super().__init__(fmt='%(name)s: %(levelname)s: %(message)s')
 ```

see: ./src/west/commands.py
```
433     def dbg(self, *args, level: Verbosity = Verbosity.DBG, end: str = '\n'):
434         '''Print a verbose debug message.
435
436         The message is only printed if *self.verbosity* is at least *level*.
437
438         :param args: sequence of arguments to print
439         :param level: verbosity level of the message
440         '''
441         if self.verbosity < level:
442             return
443         print(*args, end=end)
```

**Comment by steverae-work** [comment]
@ALL 
It seems like we are "peeling an onion": we are spending a lot of effort and seemingly not making very much progress....
I trust that we will soon be able to get on the same page and to get this implemented.

It is unclear to me exactly what the remaining issues are;
so I would request those who commented to "hide" their comments if they are outdated or resolved,
( because I don't have permission to do that... )
so that the outstanding issues can be addressed.

- Thanks in advance, Steve

**Comment by marc-hb** [comment]
> It is unclear to me exactly what the remaining issues are;
> so I would request those who commented to "hide" their comments if they are outdated or resolved,
> ( because I don't have permission to do that... )
> so that the outstanding issues can be addressed.

I am usually closing resolved comments and I'll try to be more thorough but it's preferable to leave the more complex ones open so later reviewers do not restart the same conversations.

Looking at the "Files" Tab right now, I see only 4 discussions still open. That's a very small number. I admit there are more loose threads in the main "Conversation" tab, hopefully we can converge on those soon.

You can also help by not leaving questions hanging. Back to the "Files" tab for instance, you did not explain yet why we need to preserve all git options (and a lambda argument!) to merely test that a git remote is available and up. A basic, universal git command should be enough? If not then that I think deserves a comment why in the source.

**Comment by marc-hb** [comment]
> so I would request those who commented to "hide" their comments if they are outdated or resolved,
> ( because I don't have permission to do that... )

https://github.com/zephyrproject-rtos/west/pull/994/changes#r3909576885

Screenshot 2026-09-02 at 08 49 57

**Comment by marc-hb** [comment]
> It is unclear to me exactly what the remaining issues are;
so I would request those who commented to "hide" their comments if they are outdated or resolved,
> [...]
> regarding the debug logs: ...

https://github.com/zephyrproject-rtos/west/pull/994/changes#r3906762787

Another way to help: avoid splitting conversations between the "Conversation" and "Files" tab for no obvious reason.

Stones / glass houses...

File: src/west/app/project.py (hunk 8 above)
Comment (marc-hb) [reply src/west/app/project.py]: (trying to move https://github.com/zephyrproject-rtos/west/pull/994#issuecomment-5502668402 back to the "Files" tab)

> regarding the debug logs:

> it seems that the "_logger.debug()" prepends content like: "west.manifest: DEBUG: "
> but the self.dbg() does not update the message
>
> ...

Thanks for investigating this. Please file a new issue if needed and drop your manual `[debug]` and `[info]` prefixes. If there is a bug in the logging code then it obviously must be fixed once for all in those functions themselves, not worked around in all the places invoking it, in a "whack-a-mole" and inconsistent style. No one should ever write `debug("DEBUG ...")`

File: src/west/app/project.py (hunk 7 above)
Comment (marc-hb) [reply src/west/app/project.py]: Trying to move https://github.com/zephyrproject-rtos/west/pull/994#issuecomment-5502185152 back to the "Files" tab

> In order to handle the case were some mirrors exist, but not all mirrors exist.
> ( if all mirrors exist, then just use the "git insteadof" feature ... )

Apologies for missing this.

File: src/west/app/project.py (hunk 8 above)
Comment (steverae-work) [reply src/west/app/project.py]: created issue:
https://github.com/zephyrproject-rtos/west/issues/999

... but I am opposed to dropping the ```[debug]``` and ```[info]``` until it is resolved

**Comment by steverae-work** [comment]
this change:
- replaced the "project.git()" calls to "git ls-remote" with "subprocess.run()" calls, which then allowed them to be common
in the _resolve_mirror_url() function
- added "tests" to validate src/west/app/project.py lines 231 to 253

File: src/west/app/project.py (hunk 9 above)
Comment (steverae-work) [reply src/west/app/project.py]: replaced the project.git() calls to "git ls-remote ..." to subprocess.run() calls
then combined them into the _resolve_mirror_url() function

the "'git -c core.askPass=true ls-remote --exit-code mirror_url" call
is a best-practice, standard call to ensure that
- the remote repo exists,
- prevent prompting for "username and password" if it does not exist
- and to return a exit code: 0 if it exists and is reachable, non-zero if it does not, or requires authorization, etc.

**Comment by steverae-work** [comment]
I believe that I have done everything requested, except:
- src/west/app/project.py:269:                self.inf(f"    [info] using mirror: {mirror_url}")
- src/west/app/project.py:272:                self.dbg(f'    [debug] mirror was specified, but "{mirror_url}" was not found...')

I believe that this solves a real-life problem, and I believe that this is ready to merge...

Please trigger the required tests, approve, and merge this PR

Thanks in advance, Steve

**Comment by steverae-work** [comment]
@pdgendt @marc-hb 
... please

File: src/west/app/project.py [hunk 13]
```diff
@@ -207,6 +208,71 @@ def delay(func, project):
...
+        url_insteadof_config = self.config.get('url.insteadof')
+        if not url_insteadof_config:
+            return url
+
+        try:
+            url_mappings = json.loads(url_insteadof_config)
+        except json.JSONDecodeError:
+            self.die(
+                f'invalid url.insteadof configuration: expected JSON array of "remote=mirror" '
+                f'strings, got {url_insteadof_config!r}'
```
marc-hb, line 230: As asked before in https://github.com/zephyrproject-rtos/west/pull/994#discussion_r3909054698 (that you closed as "Resolved" yourself), is this custom error message significantly better compared to the standard error message from the built-in json module? How does the built-in error message look like?

The value of re-using libraries is not having to re-invent stuff like this. Writing such a simple JSON list based on an example in the documentation should not require that much hand holding. Typos do happen but I'm surprised the default error message is not enough when they do. Please share the value of this custom error message with some typo example or just drop it.

src/west/app/project.py
```diff
@@ -207,6 +208,71 @@ def delay(func, project):
...
+            )
+
+        if not isinstance(url_mappings, list):
+            self.die(
+                'invalid url.insteadof configuration: expected JSON array of '
+                '"remote=mirror" strings'
```
marc-hb: OK, unlike the previous one, I suspect this custom error message (and the next one) can replace rather cryptic "index error" messages. Fair enough.

src/west/app/project.py
marc-hb: > Re-parsing every time for every url is a waste; it should be done either lazily or just once at startup.

Not resolved.

src/west/app/project.py
marc-hb: Please correct me but I don't think we want to "cascade" mirroring and run that logic AFTER finding a cache. That does not make sense to me. Granted, it seems unlikely to actually happen in practice but it theoretically could and that makes the logic here look puzzling.

In other words, I don't think mirroring should be involved at all when a cache is found. Should it?

src/west/app/project.py
```diff
@@ -207,6 +208,71 @@ def delay(func, project):
...
+            )
+
+        for mapping in url_mappings:
+            if not isinstance(mapping, str):
+                self.die(
+                    'invalid url.insteadof configuration: each entry must be a '
+                    '"remote=mirror" string'
+                )
+
+            remote, sep, mirror = mapping.partition('=')
```
```suggestion
            remote, sep, mirror = mapping.partition('>')
```
It's unlikely but `=` is allowed in URIs whereas `>` is not, see link to RFC and previous comment at https://github.com/zephyrproject-rtos/west/pull/994#pullrequestreview-5083756329

Also, the "direction" of the change is clearer with `>` than `=`. In this current `=` implementation, the mirror comes last whereas it comes first with `git` insteadOf

```
[url "https://company_mirror.com/"]
    insteadOf = "https://github.com/"
```

src/west/app/project.py
```diff
@@ -207,6 +208,71 @@ def delay(func, project):
...
+
+            remote = remote.strip()
+            mirror = mirror.strip()
+            if not remote or not mirror:
+                self.die(
+                    f'invalid url.insteadof mapping entry {mapping!r}: empty remote or mirror value'
+                )
+
+            remote = remote.rstrip('/') + '/'
+            mirror = mirror.rstrip('/') + '/'
```
marc-hb: This `/` trick look like it may cause more confusion that it solves problems. What specific problem(s) does it solve? Please share an example.

src/west/app/project.py
```diff
@@ -1958,8 +2025,9 @@ def handle_auto_cache(self, project):
             cache_dir_parent = Path(cache_dir).parent
             cache_dir_parent.mkdir(parents=True, exist_ok=True)
             self.dbg(f'{project.name}: create auto-cache for {project.url} in {cache_dir}')
+            _url = self._resolve_mirror_url(project.url, cwd=cache_dir_parent)
             project.git(
-                ['clone', '--mirror', '--', project.url, os.fspath(cache_dir)], cwd=cache_dir_parent
+                ['clone', '--mirror', '--', _url, os.fspath(cache_dir)], cwd=cache_dir_parent
```
marc-hb, line 2046: Mirrors are not used when updating an already existing auto cache below. Is this by design and why?

src/west/app/project.py
marc-hb: > done - fixed the comment

The comment still does not seem to match the code.

I don't think we want to "cascade" mirroring and run any mirroring logic AFTER finding an auto_cache. That does not make sense to me. Granted, it seems unlikely to actually happen in practice but it theoretically could and that makes the logic here look quite puzzling (and not matching the comment)

In other words, I don't think mirroring should be involved at all when an auto_cache was found.

src/west/app/project.py
marc-hb: > ... but I am opposed to dropping the [debug] and [info] until it is resolved

Sorry but I really can't see reason why your log statements should be treated differently from any other log statement.

Also no idea why you marked again this comment as "resolved" - in any case this problem is clearly not resolved. This really does not help with tracking issues - what you complained about first.

marc-hb (commented):
> I believe that this solves a real-life problem, 

Yes!

> and I believe that this is ready to merge...

Don't "believe" but instead stop trying to paper over some comments and issues.

> Please trigger the required tests, approve, and merge this PR

This type of "command" is really not useful. Quite counter-productive actually.

------

Forgot to mention this in the previous "bookkeeping" thread.

https://docs.zephyrproject.org/latest/contribute/contributor_expectations.html#workflow-suggestions-that-help-reviewers

> Unless they applied the reviewer’s recommendation exactly, authors must not resolve and hide comments, they must let the initial reviewer do it. The Zephyr project does not require all comments to be resolved before merge. Leaving some completed discussions open can sometimes be useful to understand the greater picture.

src/west/app/project.py
```diff
@@ -207,6 +208,71 @@ def delay(func, project):
...
+                )
+
+            remote = remote.rstrip('/') + '/'
+            mirror = mirror.rstrip('/') + '/'
+            if not url.startswith(remote):
+                continue
+
+            mirror_url = mirror + url.removeprefix(remote)
+            result = subprocess.run(
+                ['git', '-c', 'core.askPass=true', 'ls-remote', '--exit-code', mirror_url],
```
marc-hb: > core.askPass=true

Is there a `true` command on Windows?

tests/test_url_insteadof.py
```diff
@@ -0,0 +1,548 @@
...
+import subprocess
+
+import pytest
+from conftest import (
+    cmd,
+    cmd_raises,
+)
+
+
+def dump_url_insteadof_config(cwd=None):
```
marc-hb:
```suggestion
def log_config(cwd=None):
```

The name of this function does not match what it does. It does nothing specific to "url.insteadof". It just happens to be only used in such tests - for now.

tests/test_url_insteadof.py
```diff
@@ -0,0 +1,548 @@
...
+            mapping_entry(f'{remotes}/', f'{mirrors2}/'),
+        ]
+    )
+    cmd(['config', 'url.insteadof', mirror_config], cwd=workspace)
+    dump_url_insteadof_config(cwd=workspace)
+
+    output = cmd('update', cwd=workspace)
+
+    # First mirror should be used (normalize paths for comparison)
+    normalized_output = output.replace('\\', '/')
```
marc-hb, line 320:
```suggestion
    normalized_output = output.replace(r'\', r'/')
```

tests/test_url_insteadof.py
```diff
@@ -0,0 +1,548 @@
...
+    manifest = remotes / 'zephyr'
+
+    cmd(['init', '-m', str(manifest), str(workspace)], env={'ZEPHYR_BASE': None})
+
+    # Don't set url.insteadof at all
+    output = cmd('update', cwd=workspace)
+
+    # Should work normally
+    assert (workspace / 'net-tools' / '.git').check(dir=1)
+    assert 'using mirror:' not in output
```
marc-hb:
```suggestion
    assert 'irror' not in output
```
More robust when the code changes.

tests/test_url_insteadof.py
```diff
@@ -0,0 +1,548 @@
...
+
+
+def test_url_insteadof_invalid_json_fails(repos_tmpdir):
+    """Test that malformed url.insteadof JSON fails the command."""
```
marc-hb: This is merely testing a standard library, you can drop this test. There is no chance an invalid JSON input will magically work. That's the value of code re-use: less test code too.

tests/test_url_insteadof.py
```diff
@@ -0,0 +1,548 @@
...
+
+    # No mirror messages should appear
+    assert 'using mirror:' not in output
+
+    # Projects should still be cloned from original
+    assert (workspace / 'net-tools' / '.git').check(dir=1)
+
+
+def test_url_insteadof_with_trailing_slashes(repos_tmpdir):
+    """Test that url.insteadof handles trailing slashes correctly."""
```
marc-hb: A I wrote before I don't think this feature should try to be "smart" with slashes. At its core this merely performs a dead-simple text substitution and the simpler, the better. If a user gets their slashes wrong then too bad. Please prove me wrong with a convincing example.

tests/test_url_insteadof.py
```diff
@@ -0,0 +1,548 @@
...
+        ],
+        cwd=workspace,
+    )
+    dump_url_insteadof_config(cwd=workspace)
+
+    # Run update - should use original URLs since prefix doesn't match
+    output = cmd('update', cwd=workspace)
+
+    # No mirror messages should appear
+    assert 'using mirror:' not in output
```
marc-hb:
```suggestion
    assert 'irror' not in output
```
More robust when the code changes.

marc-hb (commented):
I only started looking at the test code now because it's long and because looking at test code when the corresponding production code is still evolving could be a waste of time.

I was a bit scared by its length but I actually found it well organized and easy to read - thank you! Some comments below.

marc-hb:
Now that I understand this PR much better, I think that `url.mirrors` is better than `url.insteadof`. One important reason: on the surface, this feature looks similar to `insteadof` in git but  there are huge differences:
- git uses the longest matching string and does not care about the order
- git tries only one "mirror"
- git does not fallback. If the mirror is down, it fails.

However please do not go and rename everything yet, let's wait for other opinions first.

https://github.com/zephyrproject-rtos/west/pull/994#issuecomment-5458494165
> Suggestion: url.mirror (or url.mirrors if you want to signal it's a list)
>
> Rationale:
> 
> - No existing url.* section in west config today, so it doesn't collide with manifest.*, update.*, build.*, grep.*, zephyr.*.
> - It's short, and mirrors the naming/spirit of Git's own url..insteadOf, which anyone familiar with Git config will recognize immediately.
> - It doesn't imply blob-only or init/update-only scope — accurately reflects that it's consulted by Init, Update, and west blobs alike.
> ... I think I like url.mirrors the best
> 
> @pdgendt - what do you think?

src/west/app/project.py
steverae-work: west config --global url.insteadof "garbage"

```
        try:
            url_mappings = json.loads(url_insteadof_config)
        except json.JSONDecodeError as e:
            self.die(e)
```

reports:
```
FATAL ERROR: Expecting value: line 1 column 1 (char 0)
```

src/west/app/project.py
steverae-work: according to AI
```
On Windows this can fail if true.exe is not discoverable on PATH.
```

src/west/app/project.py
marc-hb: I stand corrected, thanks for the example.

src/west/app/project.py
steverae-work: it guarantees that the string will end with a single '/'
.... but I will delete it...

steverae-work:
```suggestion
```

src/west/app/project.py
```diff
@@ -207,6 +208,71 @@ def delay(func, project):
...
+                self.die(f'invalid url.insteadof mapping entry {mapping!r}: missing "=" separator')
+
+            remote = remote.strip()
+            mirror = mirror.strip()
+            if not remote or not mirror:
+                self.die(
+                    f'invalid url.insteadof mapping entry {mapping!r}: empty remote or mirror value'
+                )
+
+            remote = remote.rstrip('/') + '/'
```
steverae-work:
```suggestion
```

tests/test_url_insteadof.py
```diff
@@ -0,0 +1,548 @@
...
+    mirrors_url = 'file://' + str(mirrors).replace('\\', '/')
+    cmd(
+        [
+            'config',
+            '--global',
+            'url.insteadof',
+            json.dumps([mapping_entry(remotes_url, mirrors_url)]),
+        ]
+    )
+    dump_url_insteadof_config()
```
steverae-work:
```suggestion
    log_config()
```

tests/test_url_insteadof.py
```diff
@@ -0,0 +1,548 @@
...
+    # Configure mirror pointing to nonexistent HTTPS repo (would normally prompt)
+    cmd(
+        [
+            'config',
+            'url.insteadof',
+            json.dumps([mapping_entry(str(remotes), 'https://example.com/nonexistent')]),
+        ],
+        cwd=workspace,
+    )
+    dump_url_insteadof_config(cwd=workspace)
```
steverae-work:
```suggestion
    log_config(cwd=workspace)
```

tests/test_url_insteadof.py
```diff
@@ -0,0 +1,548 @@
...
+
+    # Initialize workspace
+    cmd(['init', '-m', str(manifest), str(workspace)], env={'ZEPHYR_BASE': None})
+
+    # Configure url.insteadof to redirect from remotes to mirrors
+    cmd(
+        ['config', 'url.insteadof', json.dumps([mapping_entry(str(remotes), str(mirrors))])],
+        cwd=workspace,
+    )
+    dump_url_insteadof_config(cwd=workspace)
```
steverae-work:
```suggestion
    log_config(cwd=workspace)
```

tests/test_url_insteadof.py
```diff
@@ -0,0 +1,548 @@
...
+            json.dumps(
+                [
+                    mapping_entry('https://example.invalid/', 'https://mirror.invalid/'),
+                    mapping_entry(str(remotes), str(mirrors)),
+                ]
+            ),
+        ],
+        cwd=workspace,
+    )
+    dump_url_insteadof_config(cwd=workspace)
```
steverae-work:
```suggestion
    log_config(cwd=workspace)
```

tests/test_url_insteadof.py
```diff
@@ -0,0 +1,548 @@
...
+    # Initialize workspace
+    cmd(['init', '-m', str(manifest), str(workspace)], env={'ZEPHYR_BASE': None})
+
+    # Configure url.insteadof with nonexistent mirror path
+    nonexistent = repos_tmpdir / 'does-not-exist'
+    cmd(
+        ['config', 'url.insteadof', json.dumps([mapping_entry(str(remotes), str(nonexistent))])],
+        cwd=workspace,
+    )
+    dump_url_insteadof_config(cwd=workspace)
```
steverae-work:
```suggestion
    log_config(cwd=workspace)
```

tests/test_url_insteadof.py
```diff
@@ -0,0 +1,548 @@
...
+
+    mirror_config = [
+        mapping_entry(net_tools_url, mirror1_net_tools_url),
+        mapping_entry(kconfiglib_url, mirror2_kconfiglib_url),
+    ]
+    cmd(
+        ['config', 'url.insteadof', json.dumps(mirror_config)],
+        cwd=workspace,
+    )
+    dump_url_insteadof_config(cwd=workspace)
```
steverae-work:
```suggestion
    log_config(cwd=workspace)
```

tests/test_url_insteadof.py
```diff
@@ -0,0 +1,548 @@
...
+    # Configure url.insteadof with unrelated prefix
+    cmd(
+        [
+            'config',
+            'url.insteadof',
+            json.dumps([mapping_entry('https://github.com/', 'https://mirror.example.com/')]),
+        ],
+        cwd=workspace,
+    )
+    dump_url_insteadof_config(cwd=workspace)
```
steverae-work:
```suggestion
    log_config(cwd=workspace)
```

tests/test_url_insteadof.py
```diff
@@ -0,0 +1,548 @@
...
+    manifest = remotes / 'zephyr'
+
+    cmd(['init', '-m', str(manifest), str(workspace)], env={'ZEPHYR_BASE': None})
+
+    # Test with and without trailing slashes (should be normalized)
+    cmd(
+        ['config', 'url.insteadof', json.dumps([mapping_entry(f'{remotes}/', f'{mirrors}/')])],
+        cwd=workspace,
+    )
+    dump_url_insteadof_config(cwd=workspace)
```
steverae-work:
```suggestion
    log_config(cwd=workspace)
```

tests/test_url_insteadof.py
```diff
@@ -0,0 +1,548 @@
...
+
+    # Configure two mappings - first one should win
+    mirror_config = json.dumps(
+        [
+            mapping_entry(str(remotes), str(mirrors1)),
+            mapping_entry(f'{remotes}/', f'{mirrors2}/'),
+        ]
+    )
+    cmd(['config', 'url.insteadof', mirror_config], cwd=workspace)
+    dump_url_insteadof_config(cwd=workspace)
```
steverae-work:
```suggestion
    log_config(cwd=workspace)
```

tests/test_url_insteadof.py
```diff
@@ -0,0 +1,548 @@
...
+def test_url_insteadof_invalid_json_fails(repos_tmpdir):
+    """Test that malformed url.insteadof JSON fails the command."""
+    remotes = repos_tmpdir / 'repos'
+    workspace = repos_tmpdir / 'workspace'
+    manifest = remotes / 'zephyr'
+
+    cmd(['init', '-m', str(manifest), str(workspace)], env={'ZEPHYR_BASE': None})
+
+    cmd(['config', 'url.insteadof', '["missing-end"'], cwd=workspace)
+    dump_url_insteadof_config(cwd=workspace)
```
steverae-work:
```suggestion
    log_config(cwd=workspace)
```

tests/test_url_insteadof.py
```diff
@@ -0,0 +1,548 @@
...
+    # Configure both auto-cache and url.insteadof
+    cmd(['config', 'update.auto-cache', str(cache_dir)], cwd=workspace)
+    # Use file:// URL pattern to match what's in the manifest
+    remotes_url = 'file://' + str(remotes).replace('\\', '/')
+    mirrors_url = 'file://' + str(mirrors).replace('\\', '/')
+    cmd(
+        ['config', 'url.insteadof', json.dumps([mapping_entry(remotes_url, mirrors_url)])],
+        cwd=workspace,
+    )
+    dump_url_insteadof_config(cwd=workspace)
```
steverae-work:
```suggestion
    log_config(cwd=workspace)
```

tests/test_url_insteadof.py
```diff
@@ -0,0 +1,548 @@
...
+    cmd(['init', '-m', str(manifest), str(workspace)], env={'ZEPHYR_BASE': None})
+
+    # Configure url.insteadof to use mirrors for any clones
+    remotes_url = 'file://' + str(remotes).replace('\\', '/')
+    mirrors_url = 'file://' + str(mirrors).replace('\\', '/')
+    cmd(
+        ['config', 'url.insteadof', json.dumps([mapping_entry(remotes_url, mirrors_url)])],
+        cwd=workspace,
+    )
+    dump_url_insteadof_config(cwd=workspace)
```
steverae-work:
```suggestion
    log_config(cwd=workspace)
```

tests/test_url_insteadof.py
```diff
@@ -0,0 +1,548 @@
...
```
steverae-work:
```suggestion
    log_config(cwd=workspace)
```

tests/test_url_insteadof.py
steverae-work: done - in the next commit

tests/test_url_insteadof.py
steverae-work: done - in the next commit

src/west/app/project.py
steverae-work: "git remote update...." essentially performs a "git fetch" for _all_ of the remotes configured for this repo....

Since there is no URL passed into this command, attempting to temporarily "rewrite" all of the URL's in the repos ".git/config" file was considered to risky - and since the major use-case of this feature is most likely CI Pipelines ( where the auto-cache was just created a few minutes ago ) the chances of needing to update this auto-cache are very low.

If this is required in the future, then replacing a single "git remote update" with a loop that processes each individual repo ( including the url.insteadof feature ) could be used....

tests/test_url_insteadof.py
```diff
@@ -0,0 +1,548 @@
...
+            mapping_entry(f'{remotes}/', f'{mirrors2}/'),
+        ]
+    )
+    cmd(['config', 'url.insteadof', mirror_config], cwd=workspace)
+    log_config(cwd=workspace)
+
+    output = cmd('update', cwd=workspace)
+
+    # First mirror should be used (normalize paths for comparison)
+    normalized_output = output.replace(r'\', r'/')
```
steverae-work: @marc-hb 
this is now reported as a syntax error:
```
E     File "C:\Users\raesteven\work\UPSTREAM\west\tests\test_url_insteadof.py", line 319
E       normalized_output = output.replace(r'\', r'/')
E                                                   ^
E   SyntaxError: unterminated string literal (detected at line 319)
```

src/west/app/project.py
steverae-work: changed to "GIT_TERMINAL_PROMPT=0" is this commit

steverae-work:
> Sorry but I really can't see reason why your log statements should be treated differently from any other log statement.

OK - I updated them....

> Also no idea why you marked again this comment as "resolved" - in any case this problem is clearly not resolved. This really does not help with tracking issues - what you complained about first.

( ... trying to see what I still need to work on .... )

steverae-work:
>> Re-parsing every time for every url is a waste; it should be done either lazily or just once at startup.

> Not resolved.

done in this commit

steverae-work:
> The comment still does not seem to match the code.

> I don't think we want to "cascade" mirroring and run any mirroring logic AFTER finding an auto_cache. That does not make sense to me. Granted, it seems unlikely to actually happen in practice but it theoretically could and that makes the logic here look quite puzzling (and not matching the comment)

> In other words, I don't think mirroring should be involved at all when an auto_cache was found.

done - fixed in this commit

src/west/app/project.py
```diff
@@ -207,6 +212,83 @@ def delay(func, project):
 
         self.inf(result, colorize=False)
 
+    def _parsed_url_insteadof_mappings(self):
```
pdgendt, line 215: If we're going to put JSON inside the config files (which I'm not sure yet how I feel about that), we shouldn't be turning it into a micro-parser mechanism.

JSON has an object type with key -> value mapping already, why do you need to parse `remote>mirror` like strings?

```json
{
  "remote": "mirror"
}
```

pdgendt:
All-in-all I'm still uncertain we'd want this feature:

- The non-fallback part already exists using `git`'s native `url.<mirror>.insteadof`, which also works for submodules (this feature does not).
- The fallback part _is_ the feature, but it's also a new cost (`git ls-remote`), for each project on `west update`, _forever_. And the `git ls-remote` might not even provide a guaranteed result if the fetch/clone would succeed.
- We already have cache support for high-availability.
- The feature is a few lines of scripting for CI pipelines (if you have manifests without imports!):
```bash
# UNTESTED
west init -m "$MANIFEST_MIRROR_URL" .
west list -f '{url}' | sort -u | while read -r url; do
    mirror="https://gitlab.intra/mirror/${url#https://github.com/}"
    if GIT_TERMINAL_PROMPT=0 git ls-remote --exit-code "$mirror" >/dev/null 2>&1; then
        git config --global "url.$mirror.insteadOf" "$url"
    fi
done
west update
```

marc-hb:
On the review comments bookkeeping topic again: can you please use the green "Start Review" button? I just received almost 20 one-line Github notifications with only "done" or "log_config" in each. Not convenient.

See https://docs.zephyrproject.org/latest/contribute/contributor_expectations.html#workflow-suggestions-that-help-reviewers (please read the entire page while there)

src/west/app/project.py
marc-hb: I was the one who suggested `"remote > mirror"` to reduce the number of quotes. No strong opinion, I'm fine with `{ "remote" : "mirror" }` too.

I was also the one recommending above to "embed" JSON to get rid of a fair amount of manual and brittle parsing code that was here before and to avoid INI issues (e.g. #996 , https://jsontotable.io/blog/ini-parser-dialects,...) without breaking backwards compatibility. It seems to do the job nicely - at least in simple cases like this. I am NOT suggesting to generalize it.

marc-hb:
> See https://docs.zephyrproject.org/latest/contribute/contributor_expectations.html#workflow-suggestions-that-help-reviewers (please read the entire page while there)

Also, you need to squash fixup commits: https://docs.zephyrproject.org/latest/contribute/guidelines.html#contribution-workflow

For this PR you probably want 2 commits: one for the feature and one for the tests. Or just one.

steverae-work:
I am withdrawing this request
- let's discuss the bigger issue and the solution in https://github.com/zephyrproject-rtos/west/issues/1003 first

Issue #608, opened 2022-10-20
Provide a command line option for skipping unwanted portions of manifest


omkar3141:
Doing a west update can consume a significant amount of time if there are many dependencies specified in the west.yml file. It will be really great to have a command line option to skip updates of certain repositories.

For example:
`west update --skip "matter,homekit,liblc3codec,hal_st,other-unwanted-stuff"`

Such an option should apply the filter to all nested manifest files (if any) and not just the top-level west.yml file.

Such commands can then be converted to shorthand aliases to quickly swap your checked-out repos depending on the projects being worked on. This can also help save some amount of disk space.

marc-hb:
Feels like a duplicate of
- https://github.com/zephyrproject-rtos/zephyr/issues/46801

See also
- #519

mbolivar-nordic:
> TL;DR the whole thing would need to be thought over very carefully and a lot of tests would need to be written to validate the design

... or we would need to limit the scope of the feature, e.g. "you can't skip a project update if the project has imports", something like that. Even then I'm not totally sure it's trivial.

PR #995, opened 2026-08-29
Support `remapping` of west project url

thorsten-klein:
Proposal for #615 (Reopened #869 with no conversation)

CC: @tobiaskaestner

## Problem

West manifests can import other manifests, and those can import further manifests — chains that often span organizational boundaries (e.g. app → internal platform manifest → Zephyr's own manifest). Every project URL in that chain is fixed by whoever wrote the manifest it came from; upstream has no way to know about a downstream consumer's specific setup or needs — most commonly, using mirrored repositories:

- **Mirrored/cached repositories.** Many organizations mirror upstream repos internally (GitHub Enterprise, Gitea, Artifactory, etc.). Today, using those mirrors means either forking and hand-editing every upstream manifest that references the original URLs, or maintaining a fully rewritten manifest tree in parallel — in both cases you lose upstream's revision pins and any future updates unless you keep re-merging them by hand.
- **Fixed URL layouts from an intermediate manifest.** A downstream consumer of a shared "category" manifest doesn't get to choose where the referenced repos live — the URLs are baked in by whoever authored that manifest. There is currently no supported way for a manifest further downstream to change *only* the URL while leaving everything else (revision, path, groups, submodules, ...) untouched.

Manifest imports intentionally don't let you override individual project fields, and duplicating a project entry to change its URL breaks the import mechanism. There's currently no way to point an inherited project at a different remote without listing the whole project metadata in your manifest again.

## Solution

A new, optional `remapping` section, under a `url` key, lets any manifest declare plain search-and-replace patterns (`old` → `new`, literal substrings, not regex) that get applied to every project URL resolved during import — including URLs pulled in from imported/nested manifests, recursively.

```yaml
manifest:
  remapping:
    url:
      - old: github.com/zephyr-rtos
        new: github.company.com/mirrors
```

With this, all URLs your workspace resolves under `github.com/zephyr-rtos/...` — whether the project is declared directly in your manifest or pulled in through several levels of imports — get rewritten to your mirror, while everything else about the project (`revision`, `path`, `groups`, `west-commands`, ...) still comes from upstream untouched. Updating upstream's manifest (e.g. bumping a revision) keeps working with zero merge conflicts, because you never touched upstream's file.

Multiple, more targeted patterns are also supported, applied in the order listed within a manifest, so you can mirror different repos to different places:

```yaml
manifest:
  remapping:
    url:
      - old: github.com/foo/libfoo
        new: github.company.com/foo/mirrored-libfoo
      - old: github.com/foo/libbar
        new: github.othercompany.com/forked/libbar
```

## How resolution order works across imports

Remapping rules accumulate as the manifest tree is walked and are applied **innermost-first, outermost-last**: a rule declared by a manifest deep in the import chain (e.g. Zephyr's own manifest, or an intermediate platform manifest) is applied before a rule declared by the manifest closer to your actual workspace. This means the manifest you control directly — the one at the root of your workspace — always has the last word on a project's final URL, even if several layers of imported manifests each rewrote it along the way. Each manifest's own `remapping` only affects itself and anything it imports; it has no effect on manifests that import *it*.

## Why not just use `west config` (or `git config url.<base>.insteadOf`)?

This is the proposal done in https://github.com/zephyrproject-rtos/west/pull/994.

West already has a config mechanism, and Git has its own native URL rewriting via `url.<base>.insteadOf`. Both look like they'd solve this, but neither actually does, for the same underlying reason: **they live outside the manifest, so they don't travel with it.**

- **They're local machine/user state, not manifest state.** `westconfig`/`.westconfig` and `.west/config` are never part of the manifest repository — they live on whatever machine runs west. The same is true of `git config`, whether set globally on a workstation or per-repo. None of that is checked in, versioned alongside the manifest, or distributed to anyone who clones the workspace or imports the manifest.
- **Every project and every user has to redo it.** Because the config isn't attached to the manifest, each downstream project, each contributor's laptop, and each CI runner/container needs its own out-of-band setup step to apply the same mirror mapping — typically some onboarding doc or provisioning script that has to be written, kept up to date by hand, and run *before* `west init`/`west update` even works. Miss that step (a new contributor, a fresh CI image, a new downstream fork) and you silently fall back to the un-mirrored upstream URLs, or `west update` fails outright in an environment that can't reach them.
- **It isn't inherited through imports.** The entire point of this feature is that a project several import-levels downstream (e.g. app → internal platform manifest → Zephyr) should transparently get the mirrored URLs without needing to know the mirror even exists. A config value set on one machine has no relationship to the manifest-import graph — it can't be "inherited" by whoever imports your manifest the way a `remapping` entry declared in the manifest itself automatically is.
- **No single point of truth.** With config-based rewriting, the actual mapping in effect on any given machine is invisible from the manifest itself — you have to go inspect that machine's config to know what URLs will really be used. With `remapping` in the manifest, the mapping is declarative, versioned, code-reviewed, and identical for every consumer by construction.

In short: config-based URL rewriting can work for a single person's single workspace, but it doesn't compose with west's import model at all — it has to be manually re-applied at every level (every project, every user, every CI job) instead of being defined once, upstream or downstream, and automatically honored by everything that imports that manifest.

## Notes

- Fully opt-in: manifests without a `remapping` section behave exactly as before.
- Plain, sequential substring replacement — no regex, no partial-match ambiguity to reason about beyond "does this substring appear."
- Covered by new tests in `tests/test_project.py`, including a multi-level import chain (app → middle → base → zephyr remote) verifying that a mirror URL set only in the top-level manifest correctly overrides remote URLs several import-levels deep, plus a case where the remapping is picked up through a manifest's own `self.import`.

### DIFF
--- a/tests/test_project.py
+++ b/tests/test_project.py
@@ -112,6 +112,183 @@ def test_workspace(west_update_tmpdir):
     assert wct.join('zephyr', 'subsys', 'bluetooth', 'code.c').check(file=1)
 
 
+def test_workspace_remap_url(tmpdir, repos_tmpdir):
+    remotes_dir = repos_tmpdir / 'repos'
+    workspace_dir = tmpdir / 'workspace'
+    workspace_dir.mkdir()
+
+    # use remote zephyr
+    remote_zephyr = tmpdir / 'repos' / 'zephyr'
+
+    # create a local base project with a west.yml
+    project_base = remotes_dir / 'base'
+    create_repo(project_base)
+    add_commit(
+        project_base,
+        'manifest commit',
+        # zephyr revision is implicitly master, see #821
+        files={
+            'west.yml': textwrap.dedent('''
+            manifest:
+              remapping:
+                url:
+                - old: xxx
+                  new: yyy
+              remotes:
+                - name: upstream
+                  url-base: xxx
+              projects:
+              - name: zephyr
+                remote: upstream
+                path: zephyr-rtos
+                import: True
+              ''')
+        },
+    )
+
+    # create another project with another west.yml (stacked on base)
+    project_middle = remotes_dir / 'middle'
+    create_repo(project_middle)
+    add_commit(
+        project_middle,
+        'manifest commit',
+        # zephyr revision is implicitly master:
+        files={
+            'west.yml': f'''
+                      manifest:
+                        remapping:
+                          url:
+                          - old: yyy
+                            new: zzz
+                        projects:
+                        - name: base
+                          url: {project_base}
+                          import: True
+                      '''
+        },
+    )
+
+    # create an app that uses middle project
+    project_app = workspace_dir / 'app'
+    project_app.mkdir()
+    with open(project_app / 'west.yml', 'w') as f:
+        f.write(
+            textwrap.dedent(f'''\
+            manifest:
+              remapping:
+                url:
+                  - old: zzz
+                    new: {os.path.dirname(remote_zephyr)}
+              projects:
+                - name: middle
+                  url: {project_middle}
+                  import: True
+        ''')
+        )
+
+    # init workspace in projects_dir (project_app's parent)
+    cmd(['init', '-l', project_app])
+
+    # update workspace in projects_dir
+    cmd('update', cwd=workspace_dir)
+
+    # zephyr projects from base are cloned
+    for project_subdir in [
+        Path('subdir') / 'Kconfiglib',
+        'tagged_repo',
+        'net-tools',
+        'zephyr-rtos',
+    ]:
+        assert (workspace_dir / project_subdir).check(dir=1)
+        assert (workspace_dir / project_subdir / '.git').check(dir=1)
+
+
+def test_workspace_remap_url_from_self_import(repos_tmpdir):
+    remote_zephyr = repos_tmpdir / 'repos' / 'zephyr'
+    projects_dir = repos_tmpdir / 'projects'
+    projects_dir.mkdir()
+
+    # create a local base project with a west.yml
+    project_base = projects_dir / 'base'
+    project_base.mkdir()
+    with open(project_base / 'west.yml', 'w') as f:
+        f.write(
+            textwrap.dedent('''\
+            manifest:
+              remotes:
+                - name: upstream
+                  url-base: nonexistent
+              projects:
+              - name: zephyr
+                remote: upstream
+                path: zephyr-rtos
+                import: True
+        ''')
+        )
+
+    # create another project with another west.yml (stacked on base)
+    project_middle = projects_dir / 'middle'
+    project_middle.mkdir()
+    with open(project_middle / 'west.yml', 'w') as f:
+        f.write(
+            textwrap.dedent('''\
+            manifest:
+              self:
+                import: ../base
+        ''')
+        )
+
+    # create another project with another west.yml (stacked on base)
+    project_another = projects_dir / 'another'
+    project_another.mkdir()
+    with open(project_another / 'west.yml', 'w') as f:
+        f.write(
+            textwrap.dedent('''\
+            manifest:
+              # this should not have any effect since there are no imports
+              remapping:
+                url:
+                  - old: nonexistent
+                    new: from-another
+        ''')
+        )
+
+    # create another project with another west.yml (stacked on base)
+    project_app = projects_dir / 'app'
+    project_app.mkdir()
+    with open(project_app / 'west.yml', 'w') as f:
+        f.write(
+            textwrap.dedent(f'''\
+            manifest:
+              remapping:
+                url:
+                  - old: nonexistent
+                    new: {os.path.dirname(remote_zephyr)}
+              self:
+                import:
+                - ../another
+                - ../middle
+        ''')
+        )
+
+    # init workspace in projects_dir (project_app's parent)
+    cmd(['init', '-l', project_app])
+
+    # update workspace in projects_dir
+    cmd('update', cwd=projects_dir)
+
+    ws = projects_dir
+    # zephyr projects from base are cloned
+    for project_subdir in [
+        Path('subdir') / 'Kconfiglib',
+        'tagged_repo',
+        'net-tools',
+        'zephyr-rtos',
+    ]:
+        assert (ws / project_subdir).check(dir=1)
+        assert (ws / project_subdir / '.git').check(dir=1)
+
+
 def test_list(west_update_tmpdir):
     # Projects shall be listed in the order they appear in the manifest.
     # Check the behavior for some format arguments of interest as well.
--- a/src/west/manifest-schema.yml
+++ b/src/west/manifest-schema.yml
@@ -66,6 +66,25 @@ mapping:
             required: true
             type: str
 
+  # allow some remapping of values by downstream projects
+  remapping:
+    required: false
+    type: map
+    mapping:
+      # search-replace within URLs (e.g. to use mirror URLs)
+      url:
+        required: false
+        type: seq
+        sequence:
+          - type: map
+            mapping:
+              old:
+                required: true
+                type: str
+              new:
+                required: true
+                type: str
+
   # The "projects" key specifies a sequence of "projects", each of which has a
   # remote, and may specify additional configuration.
   #
--- a/src/west/manifest.py
+++ b/src/west/manifest.py
@@ -7,6 +7,7 @@
 Parser and abstract data types for west manifests.
 '''
 
+import copy
 import enum
 import errno
 import logging
@@ -181,6 +182,34 @@ ManifestDataType = str | dict
 # Logging
 
 _logger = logging.getLogger(__name__)
+
+
+# Representation of remapping
+
+
+class ImportRemapping:
+    """Represents `remapping` within a manifest."""
+
+    def __init__(self, manifest_data: dict | None = None):
+        """Initialize a new ImportRemapping instance."""
+        self.url_replaces: list[tuple[str, str]] = []
+        self.append(manifest_data or {})
+
+    def append(self, manifest_data: dict):
+        """Append values from a manifest data (dictionary) to this instance."""
+        for kind, values in manifest_data.get('remapping', {}).items():
+            if kind == 'url':
+                self.url_replaces += [(v['old'], v['new']) for v in values]
+
+    def merge(self, other):
+        """Merge another ImportRemapping instance into this one."""
+        if not isinstance(other, ImportRemapping):
+            raise TypeError(f"Unsupported type'{type(other).__name__}'")
+        self.url_replaces += other.url_replaces
+
+    def copy(self):
+        """Return a deep copy of this instance."""
+        return copy.deepcopy(self)
 
 
 # Type for the submodule value passed through the manifest file.
@@ -456,6 +485,9 @@ class _import_ctx(NamedTuple):
 
     # Bit vector of flags that modify import behavior.
     import_flags: 'ImportFlag'
+
+    # remapping
+    remapping: ImportRemapping
 
 
 def _imap_filter_allows(imap_filter: ImapFilterFnType, project: 'Project') -> bool:
@@ -2070,6 +2102,7 @@ class Manifest:
             current_repo_abspath=current_repo_abspath,
             project_importer=project_importer,
             import_flags=import_flags,
+            remapping=ImportRemapping(),
         )
 
     def _recursive_init(self, ctx: _import_ctx):
@@ -2091,6 +2124,10 @@ class Manifest:
         _logger.debug('loading %s', loading_what)
 
         manifest_data = self._ctx.current_data['manifest']
+
+        # append values from resolved manifest_data to current context
+        new_remapping = ImportRemapping(manifest_data)
+        self._ctx.remapping.merge(new_remapping)
 
         schema_version = str(manifest_data.get('version', SCHEMA_VERSION))
 
@@ -2340,6 +2377,7 @@ class Manifest:
             current_abspath=pathobj_abs,
             current_relpath=pathobj,
             current_data=pathobj_abs.read_text(encoding=Manifest.encoding),
+            remapping=self._ctx.remapping.copy(),
         )
         try:
             Manifest(topdir=self.topdir, internal_import_ctx=child_ctx)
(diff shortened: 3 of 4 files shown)


marc-hb (changes requested):
I'm replying to the above but also to the comment in https://github.com/zephyrproject-rtos/west/pull/994#issuecomment-5463334518 because it's more readable to keep each proposal in a single place.


> With the west config approach, each user would have to configure the option separately.
> ...
> We specify the mirrors in our category manifest, and all downstream projects inherit the URLs automatically. There is no need for downstream users to configure anything in Git or west themselves.

Saving one `west config ...` command once in a while does not really matter and it is certainly not enough of a justification to make a change to how manifests work.

1. If you look at https://docs.zephyrproject.org/latest/develop/getting_started/index.html, you can see that workspace initialization requires many other actions, like getting a toolchain/SDK, installing and configuring other dependencies, virtualenv(s) etc. People and CIs routinely automate that sort of setup with site-specific and/or project-specific configuration scripts and files. Adding a simple `west config ...` command in those scripts costs nothing at all and it's the most logical and correct place for it.

2. Manifest imports is by far the most complex topic in `west`. Marti spent weeks trying to get them right and he's still not 100% sure he did. And he's gone now.
See for instance https://github.com/zephyrproject-rtos/west/issues/608 and
https://github.com/zephyrproject-rtos/west/issues?q=label%3A%22Partial%20imports%22
Fun fact: try `west update some_imported_project`. Investigate why it does not work.


1+2. Any change to how manifest imports work must be REALLY worth it and saving an infrequent `west config ...` command is really not worth it. That bar is much higher.


> they live outside the manifest, so they don't travel with it.

Yes and decoupling WHAT you get versus HOW you get it is indeed an important feature. It's the core feature of git and many other tools (like CDNs). This was already discussed in #615 among other places.

Consider a team spread across the globe. Why would different team members use different manifests to build the same thing? When they want to ensure perfect build reproduction to reproduce some elusive race condition, why would _mirroring_ pollute their manifest comparison? That makes no sense.


> *No single point of truth*. With config-based rewriting, the actual mapping in effect on any given machine is invisible from the manifest itself — you have to go inspect that machine's config to know what URLs will really be used. 

- If rewrites are performed for mirroring the same content / SHA1s, then "URL truth" does not matter because it's the same content. This is in fact the very essence of git. So you don't need to "inspect" anything. (For investigating performance issues and rate limits, you need detailed logs and local access anyway).

- If you are performing rewrites on branches then all bets are off anyway because branches MOVE by definition. So inspecting any solution will never prove any "truth". Even worse: branch differences can affect the imports themselves. I bet (and hope) this is not an issue in your particular case but it definitely is in the general case.

Also: don't forget users can ALREADY use git's `insteadof` and for sure some already do for... mirroring. How do you make _that_ visible?


> This is why we have been using the approach from #869 for the past nine months, and it has been working well for us.

It's fantastic that you are ready to fork `west` in production and share your experience. I bet few users have the time and expertise to do the same, so please keep doing that; it's invaluable. I mean it. However this does not mean your goals, priorities and test coverage align with the majority of west users.

Sometimes these goals even look extraordinary like for instance #852 and #872 that insisted for a pretty long time on adding to a multi-repo tool (west) support for... monorepos?!

thorsten-klein:
The use case here is different: the URL mapping is part of the **manifest/import definition** itself. For example, a company manifest can define that repositories from an imported manifest should be resolved through an internal mirror, without requiring every consumer to configure this separately. 

So I see these as complementary:

- `git insteadOf`: local Git configuration
- `west config`: workspace/user configuration
- manifest remapping: policy defined by the manifest/import hierarchy

I agree that import handling should remain simple, and I'm happy to refine the implementation/semantics if needed. But I think the manifest-level use case is distinct enough to justify the feature.

marc-hb:
> The use case here is different: the URL mapping is part of the manifest/import definition itself. For example, a company manifest can define that repositories from an imported manifest should be resolved through an internal mirror,...

Do you mean _intentionally_ changing the content by pointing at branch revisions[*] located somewhere else? This would indeed sound like something very different from the usual meaning of "mirroring", can you please avoid the word "mirroring" in that context?

That does not feel like a good idea either but (if I understand correctly) I'd need more time to think about it and I agree it's quite different.

[*] obviously can't do that with SHA1s.

tests/test_project.py
```diff
@@ -112,6 +112,183 @@ def test_workspace(west_update_tmpdir):
...
+    # use remote zephyr
+    remote_zephyr = tmpdir / 'repos' / 'zephyr'
+
+    # create a local base project with a west.yml
+    project_base = remotes_dir / 'base'
+    create_repo(project_base)
+    add_commit(
+        project_base,
+        'manifest commit',
+        # zephyr revision is implicitly master:
```
marc-hb:
```suggestion
        # zephyr revision is implicitly master, see #821
```

thorsten-klein:
> Do you mean _intentionally_ changing the content by pointing at branch revisions located somewhere else? This would indeed sound like something very different from the usual meaning of "mirroring", can you please avoid the word "mirroring" in that context?

Short answer: no — in our case it really is mirroring in the traditional sense. **The revision never changes**, only the host/URL does. Every project keeps its exact `name`, `path`, and `revision` (tag, or a pinned SHA1); only the `url` prefix is rewritten to point at our internal mirror of the same commit. That's the whole feature, as far as we need it: it does *not* let a manifest redirect to a different branch or a different piece of history — it only relocates where an identical, already-pinned revision is fetched from.

---

Thanks a lot for taking time to think about this again and understanding our use case. I really appreciate it!

Maybe some more concrete information which might help understanding:

### The Setup

We have a company-wide platform manifest (`company-platform.git`). Our products rely on `company-platform` (see later).

`company-platform` relies on deliveries from other companies, which add some 3rd party projects. They declare external dependencies pointing straight at their upstream hosts.

In `company-platform` our `west list` looks like this:

| Project     | Path                      | Revision                                    | URL |
|-------------|---------------------------|----------------------------------------------|-----|
| googletest  | `modules/lib/googletest`  | `v1.17.0`                                     | `https://github.com/google/googletest.git` |
| rapidjson   | `modules/lib/rapidjson`   | `24b5e7a8b27f42fa16b96fc70aade9106cf7102f`    | `https://github.com/Tencent/rapidjson.git` |
| cmsis       | `modules/hal/cmsis`       | `512cc7e895e8491696b61f7ba8066b4a182569b8`    | `https://github.com/zephyrproject-rtos/cmsis` |
| …and a dozen more zephyrproject-rtos modules (cmsis-dsp, hostap, littlefs, lvgl, mbedtls, mcuboot, nanopb, net-tools, openthread, segger, zcbor, …) |

Instead, we want every one of those external projects fetched from our internal mirror instead of directly from `github.com` — same tag, same commit, same bytes, just a different, internally-reachable URL.

Of course the external company will not change the URL to our mirrors. Instead we must do the URL remapping on our own and we want to do it in our downstream manifest (company-platform).

| Project     | Revision (unchanged)                          | Mirrored URL |
|-------------|-------------------------------------------------|--------------|
| googletest  | `v1.17.0`                                         | `https://github.company.com/mirrors/google-googletest.git` |
| rapidjson   | `24b5e7a8b27f42fa16b96fc70aade9106cf7102f`        | `https://github.company.com/mirrors/Tencent-rapidjson.git` |
| cmsis       | `512cc7e895e8491696b61f7ba8066b4a182569b8`        | `https://github.company.com/mirrors/zephyr-cmsis` |
| …           | …                                                 | `https://github.company.com/mirrors/zephyr-*` |

Nothing about *what* gets checked out changes — only *where it's fetched from*. That is exactly the property `remapping` is meant to preserve: it rewrites a URL prefix, it does not touch `revision`, and (per the PR's own note) it plainly can't be used to redirect a SHA1-pinned project to different content, because the SHA1 travels with it unchanged.

## The `remapping` Rules

`company-platform.git`'s `west.yml` declares according rewrite rules (in the exact same manifest where the 3rd party delivery is integrated!):

```yaml
manifest:
  remapping:
    url:
      - old: github.com/google/
        new: github.company.com/mirrors/google-
      - old: github.com/Tencent/
        new: github.company.com/mirrors/Tencent-
      - old: github.com/zephyrproject-rtos/
        new: github.company.com/mirrors/zephyr-
```

Each rule is a simple search-replace pattern which matches exactly the transform shown in the table above. The last rule alone accounts for the dozen `zephyrproject-rtos` modules (cmsis, cmsis-dsp, hostap, littlefs, lvgl, mbedtls, mcuboot, nanopb, net-tools, openthread, segger, zcbor, ...).

Three rules cover all ~18 affected projects, instead of editing the `url:` of each one individually — and, more importantly, instead of requiring every *downstream consumer* to know about the substitution at all (see below).

### Why Not Just Use `west config` / Git `insteadOf`?

This is the part that's specific to our situation, and the actual reason we want this handled at the *manifest* level rather than per workspace: the teams consuming these dependencies have no information about our mirrors at all — and neither side wants that to change. `company-platform` wants every product to simply use whatever `company-platform` has configured; product teams, in turn, don't want to maintain any configuration of their own just to build on top of it. Every product repository's manifest reflects that: it does nothing but import the platform —

```yaml
manifest:
 projects:
    - name: company-platform
      url: https://github.company.com/team-ghi/company-platform.git
      path: company-platform
```

— and yet ends up fetching every dependency from the correct mirror, with no configuration of its own.

`git config --global url."...".insteadOf`, or an equivalent `west config` setting, can't give us that: it would have to be set up in *every* product team's environment — every developer machine, every CI runner — before `west update` ever runs, and kept in sync by hand every time we add or move a mirror. That's exactly the out-of-band, per-workspace state we're trying to avoid, for something that's a property of the dependency, not of the workspace.

By putting the rewrite rules in `company-platform.git`'s own manifest instead, they travel with the platform: any product manifest that imports `company-platform` gets the mirrored URLs automatically, with no workspace setup and no need for the product team to even know a mirror is involved. That's the whole motivation for expressing this as manifest `remapping` rather than local git/west configuration — not to redirect to different content, but to keep "fetch this exact, unchanged revision from our mirror instead of upstream" a property of the dependency declaration itself, inherited automatically by everyone downstream.

marc-hb:
> it would have to be set up in every product team's environment — every developer machine, every CI runner — before west update ever runs, 

No, not "every" developer and CI runner: only the heavy hitters running into rate limits. Mirroring is only an optimization, it's not a hard requirement. Also, mirror configuration is something that needs to be (re-)defined rarely, **much** more rarely than running `west update`

Because `west` supports not just fixed SHA1 revisions but also moving branches, this would be introducing interesting "bootstrap" issues where `west update` can self-modify the locations where it's downloading from. Marti already hit some of these issues, ran away scared and turn some of that off (did you try that `west update imported_project` command above?). So we definitely do not want to add fuel to that fire.
 
> That's exactly the out-of-band, per-workspace state we're trying to avoid, 

So you have a configuration deployment problem. That's a very common issue. What I don't understand is why you keep describing this problem like manifests and `west update` is the only solution available in that space, like there is nothing else available. I bet your CIs and developers do not download SDKs, PyPI and other dependencies using `west update`, do they? So how do they do that instead?


> and kept in sync by hand every time we add or move a mirror.

On one hand you seem to have a fairly simplistic, "one-size-fits-all-users" mirror configuration that does not care about where you download from. Yet it's high maintenance and requires frequent changes?


> for something that's a property of the dependency, not of the workspace.

No: mirror location is _definitely_ not a property of the content. It's an optimization that is by design meant to be close to where the workspace is located.

Thanks for sharing the additional details but I'm afraid they have not changed my opinion that you are simply using the wrong tool for the job. That is a pretty common thing to do in IT and it can work well in specific cases. But that does not mean everyone should do the same.

tobiaskaestner:
Chiming in from the side here, as @thorsten-klein mentioned me and we had some prior chat about this, too. 

Actually, I can very well see a use case for this feature as we have had deployed various "work-arounds" to cope with downstream mirrors of upstream repos in a lot of (if not all) our client's project. Note: I always recommend to setup mirrors of the repos a product maker depends on into their own infrastructure. As much as we may wish for it, there is simply no guarantue that the project a product maker depends on still exists tomorrow (or in 5-10 years from now).

The url.insteadOf is not very appealing as it breaks the abstraction that west actually brings. The config path is still manual work so what we end up doing is either a) avoid importing west projects from other west project manifest files by explicitly pulling in all required projects into the outmost manifest or b) patch for instance zephyr's west.yml to do the rewrite of the url in-file. 

If I understand this PR correctly, then both ways could be eliminated by placing the proposed remapping into the outer-most downstream manifest. So, no existing manifest file would break today. AFAIK, this is strictly added functionality (that most of us won't, yet some of us will make use of - which is true for 80% of any feature set of any open source).

And since manifest files already name git repo URLs that's also no violation of abstraction. 

And if west is lacking maintainers, that's another problem and needs addressing anyways.

marc-hb:
> As much as we may wish for it, there is simply no guarantue that the project a product maker depends on still exists tomorrow (or in 5-10 years from now).

Agreed: and there can be other reasons where the original software source cannot be relied upon. In _that_ distrust case, then the recommendation is very simple: _maintain your own manifests_. One should not `import` some flaky and/or unreliable and untrusted manifest while dynamically patching it up on every `west update` invocation. That is still flaky and unreliable.

If you do not fully trust some upstream repo(s) for whatever reason, then this type of "mirroring" is not a download performance optimization anymore, now I'm afraid we've been confusing those two use cases most of the time. In that other case, maintaining your own manifests is a requirement and the correct way to guarantee a "single source of truth".

BTW this is why some (large) companies like Google use Zephyr without... using `west` _at all_! Because 1) they do not consider anything external reliable or secure as a rule, and 2) they already have their own "mirroring" and manifesting infrastructure (mutualized across many different projects and not just Zephyr)

> Actually, I can very well see a use case for this feature as we have had deployed various "work-arounds" to cope with downstream mirrors of upstream repos in a lot of (if not all) our client's project

Do not deploy SBOM "work-arounds": take actual control and maintain your own manifests. SBOMs are not a place for quick and dirty rewrite hacks.

> a) avoid importing west projects from other west project manifest files by explicitly pulling in all required projects into the outmost manifest or b) patch for instance zephyr's west.yml to do the rewrite of the url in-file.

These look like steps into the right direction: "explicit" looks good. Maintaining your own manifests does not mean completely ignoring external manifests, far from it. It only means keeping a careful, static record of your SBOMs etc. That record can be derived from external, not completely trusted manifests BUT more statically, slowly, carefully and safely than with a (too) dynamic `import` and some rewriting hacks automagically patching things up on the fly at `west update` time.

> And since manifest files already name git repo URLs that's also no violation of abstraction.

I did not get this sorry. What abstraction?

--------

BTW there is another good reason NOT to directly import the upstream Zephyr manifest  that you indirectly alluded to: it's bloated. This bloat has been discussed many times in many places and it keeps coming up again and again:
- https://github.com/zephyrproject-rtos/zephyr/issues/108783#issuecomment-4526275800

PR #996, opened 2026-09-01
configuration: do not interpolate configuration values

pdgendt:
The internal `ConfigParser` was created with configparser's default
`BasicInterpolation`, which makes '%' a metacharacter in values. West's
documentation never mentions interpolation, but it makes a '%' in any
configuration file fatal for every west command:

```
  $ cat .west/config
  [manifest]
  path = zephyr
  [build]
  cmake-args = -DFOO=%PATH%
  $ west topdir
  configparser.InterpolationSyntaxError: '%' must be followed by '%' or
  '(', found: '%PATH%'
```

Before dispatching a command, `WestApp.run()` iterates over every option
in every configuration file twice: once to populate the deprecated
`west.configuration.config` object, and once to load aliases. The first
of the two raises, which is why the failure has nothing to do with the
command being run or with the option holding the value.

Escaping as '%%' does not help either: `_copy_to_configparser()` re-sets
the already interpolated value into a second parser, which then rejects
the '%' it just produced. Both paths date back to v0.13.0, which
introduced the Configuration class and the compatibility bridge.

Values that need a '%' are not exotic: percent-encoded URLs, paths that
contain one, and cmake or grep arguments all run into this.

Pass `interpolation=None` so that values are stored, written and returned
verbatim. Two configparser behaviours go away with it, neither of them
documented by west: '%%' is no longer unescaped to '%', and
'%(other-option)s' no longer expands to another value from the same
section.

### DIFF
--- a/src/west/configuration.py
+++ b/src/west/configuration.py
@@ -33,6 +33,9 @@ You can override these files' locations with the ``WEST_CONFIG_SYSTEM``,
 Configuration values from later configuration files override configuration
 from earlier ones. Local values have highest precedence, and system values
 lowest.
+
+Values are stored and returned verbatim: no variable interpolation is
+performed, and no character needs escaping.
 '''
 
 import configparser
@@ -52,7 +55,9 @@ class MalformedConfig(Exception):
 
 
 def _configparser():  # for internal use
-    return configparser.ConfigParser(allow_no_value=True)
+    # Interpolation is disabled: configuration values are stored and
+    # returned verbatim, so that '%' is an ordinary character.
+    return configparser.ConfigParser(allow_no_value=True, interpolation=None)
 
 
 class _InternalCF:
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -910,3 +910,33 @@ def test_list():
 def test_round_trip():
     cmd('config pytest.foo bar,baz')
     assert cmd('config pytest.foo').strip() == 'bar,baz'
+
+
+def test_round_trip_percent():
+    # A '%' is an ordinary character in a value: it is not escaped on the
+    # way in, and not interpolated on the way out.
+    url = 'https://example.com/some%20path'
+
+    cmd(['config', 'pytest.url', url])
+
+    assert cmd('config pytest.url').strip() == url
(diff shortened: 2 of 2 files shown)


marc-hb (commented):
Looks good but what about Hyrum's law?

pdgendt:
> Looks good but what about Hyrum's law?

And what about a feature like https://github.com/zephyrproject-rtos/west/pull/994, where URL-encoded strings are likely to contain '%'-characters?

kartben (approved):

marc-hb:
> Looks good but what about Hyrum's law?

Sorry for being terse and tentatively funny. There is a serious concern here: I just tried interpolation and (at least in simple cases) it works. So, this will break any user who was using that interpolation.

But more importantly: this will break such a user **and not offer any alternative**. Requesting some adjustments in users configuration files would not be very demanding but it's not just that: it's removing that interpolation feature completely.

Also, keep in mind `west config` is very frequently used by west extensions - which could be literally anything.

> West's documentation never mentions interpolation, ...

West's documentation does not, but https://docs.python.org/3/library/configparser.html#configparser.BasicInterpolation takes very little effort to find.

> but it makes a '%' in any configuration file fatal for every west command: [...]
> Values that need a '%' are not exotic: percent-encoded URLs, paths that
> contain one, and cmake or grep arguments all run into this.

Yet this issue was never reported before? So not exotic but apparently very rare (here I'm referring to literal '%' signs, NOT to interpolation).

How about this?

```ini
[west]
ini_interpolation = False # The default. "True" is unsupported, avoid.
```

When True, the config is re-loaded a second time with interpolation ON and a deprecation warning. After a year or two without any report, INI interpolation is finally removed for good.

PS: In any case, the commit message with the change should have... a literal '%' sign in the title for much better visibility. I mean as simple as: "configuration: do not interpolate configuration values (with %*"

pdgendt:
Alternative PR here: https://github.com/zephyrproject-rtos/west/pull/1000

pdgendt:
Superseded by https://github.com/zephyrproject-rtos/west/pull/1000

Issue #779, opened 2025-01-13
`west config` corner cases when setting option `foo.bar=baz`


mbolivar:
This command:

```
$ west config foo.bar=baz qux
```

should, I think, either error out, or set an option named `foo.bar=baz` to `qux`.

Instead, it does some mix of the two, because afterwards, running:

```
$ west config foo.bar=baz
```

results in west exiting 1 with the option unset , but

```
$ west config -l | grep foo
foo.bar=baz = qux
```

something is wrong here, likely related to how configparser deals with `=` signs

d3zd3z:
Perhaps consider just forbidding equal signs in the names of configs? Config files are stored in a init-style format, and although have spaces, I suspect don't parse. It doesn't seem the library is expecting these names, and may very well have other characters that aren't workable.

yaml will allow almost anything, if quoted is only going to include alphanumeric, digits, '-', '_',. Dot is allowed but it gets weirdly ambiguous if we are parsing dots. Most punctuation is not allowed, without quoting.

d3zd3z:
Oh, and yaml will do really weird things if the key can be interpreted as a boolean. It works fine, but the key in the dict is actually the boolean, not a string. Likewise the same will happen with valid numbers.

marc-hb:
> likely related to how configparser deals with = signs

Yes: first '=' sign wins. configparser happily accepts equal signs in keys, writes them to the .ini file without quoting or escaping, and finally interprets the first `=` sign it finds on the line. I verified this with a debugger.

Simplest demo:

```
west config --global sect.one=two   three
west config -l
                  => sect.one=two = three
grep -B1 two .west/config
                  =>      [sect]
                  =>      one=two = three
west config sect.one
                  =>          two = three
```

It's confusing because the way `configparser` write the .ini file makes you think it cares about spaces. But not really.

> Perhaps consider just forbidding equal signs in the names of configs? 

Yes; with a pointer to the corresponding `configparser`  bug.

marc-hb:
https://docs.python.org/3/library/configparser.html#customizing-parser-behaviour states:

> delimiters, default value: ('=', ':')

> Delimiters are substrings that delimit keys from values within a section. The first occurrence of a delimiting substring on a line is considered a delimiter. This means values (but not keys) can contain the delimiters.

... but `configparser` does not check whether a key contains a delimiter on input and just writes it as is in the file. I verified this without west. Same bug with `:`.

marc-hb:
[...]

PR #1000, opened 2026-09-07
configuration: fix the '%%' escape in configuration values

pdgendt:
configparser interpolates values when they are read, so a literal '%'
has to be written as '%%'. That escape has not worked since v0.13.0:
`_copy_to_configparser()` re-sets the already interpolated value into a
second interpolating parser, which rejects the '%' it just produced.
Because `WestApp.run()` copies every option before dispatching a command,
a single escaped value makes every west command fail:

```shell
$ cat .west/config
[manifest]
path = zephyr
[build]
cmake-args = -DFOO=100%%

$ west topdir
ValueError: invalid interpolation syntax in '-DFOO=100%' at position 12
```

Escape the value again on the way into the compatibility object, which
interpolates it back to what `Configuration.get()` returns.

With the escape working, make it unnecessary to know about: 'west config'
now escapes what it writes, so a value containing a '%' can be set from
the command line and is read back verbatim. Only hand-written files need
the escape, which is what the module documentation now says.

An unescaped '%' remains an error, but no longer a configparser traceback
from an unrelated command: interpolation failures are reported as
`MalformedConfig` naming the option and how to fix it, like the other
malformed configuration errors.

Interpolation itself keeps working, so '%(other-option)s' references in
existing configuration files are unaffected.

### DIFF
--- a/src/west/configuration.py
+++ b/src/west/configuration.py
@@ -33,6 +33,18 @@ You can override these files' locations with the ``WEST_CONFIG_SYSTEM``,
 Configuration values from later configuration files override configuration
 from earlier ones. Local values have highest precedence, and system values
 lowest.
+
+Values are interpolated when they are read, by configparser from the
+Python standard library: a literal ``%`` must be written as ``%%``, and
+``%(other-option)s`` expands to the value of another option in the same
+section. The exact rules, which depend on the Python version in use, are
+documented at
+https://docs.python.org/3/library/configparser.html#interpolation-of-values
+
+``west config`` and ``Configuration.set()`` escape what they write, so a
+literal ``%`` needs no care there, and an interpolation reference can only
+be added by editing a configuration file. The deprecated
+``update_config()`` function does not escape and is the exception to both.
 '''
 
 import configparser
@@ -53,6 +65,13 @@ class MalformedConfig(Exception):
 
 def _configparser():  # for internal use
     return configparser.ConfigParser(allow_no_value=True)
+
+
+def _escape_percent(value):  # for internal use
+    # Configuration values are interpolated when they are read, so a
+    # literal '%' has to be doubled on the way in. Non-string values are
+    # left alone; configparser rejects them on its own.
+    return value.replace('%', '%%') if isinstance(value, str) else value
 
 
 class _InternalCF:
@@ -112,6 +131,20 @@ class _InternalCF:
     def getfloat(self, option: str):
         return self._get(option, self.cp.getfloat)
 
+    def _malformed_value(self, section: str, key: str) -> MalformedConfig:
+        readable = [str(p) for p in self.paths]
+        return MalformedConfig(
+            f"invalid value for '{section}.{key}' in one of '{readable}': "
+            "a literal '%' must be written as '%%'"
+        )
+
+    def value(self, section: str, key: str):
+        # The interpolated value of an option, by section and key.
+        try:
+            return self.cp.get(section, key)
+        except configparser.InterpolationError as err:
+            raise self._malformed_value(section, key) from err
+
     def _get(self, option, getter):
         section, key = _InternalCF.parse_key(option)
 
@@ -119,6 +152,8 @@ class _InternalCF:
             return getter(section, key)
         except (configparser.NoOptionError, configparser.NoSectionError) as err:
             raise KeyError(option) from err
+        except configparser.InterpolationError as err:
+            raise self._malformed_value(section, key) from err
 
     def set(self, option: str, value: Any):
         self._check_single_config()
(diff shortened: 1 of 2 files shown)


tests/test_config.py
```diff
@@ -910,3 +910,53 @@ def sorted_list(other_args=''):
 def test_round_trip():
     cmd('config pytest.foo bar,baz')
     assert cmd('config pytest.foo').strip() == 'bar,baz'
+
+
+def test_round_trip_percent():
+    # A '%' does not have to be avoided on the command line: it is
```
```suggestion
    # A '%' does not have to be manually escaped on the command line: it is automatically
```

src/west/configuration.py
```diff
@@ -398,8 +426,10 @@ def load(cf: _InternalCF):
                     continue
                 if section not in cp:
                     cp.add_section(section)
-                for key, value in contents.items():
-                    cp[section][key] = value
+                for key in contents:
+                    # 'cp' interpolates when it is read from, so the
+                    # value goes in escaped: it is interpolated already.
```
```suggestion
                    # value must be re-escaped.
```

src/west/configuration.py
```diff
@@ -55,6 +60,13 @@ def _configparser():  # for internal use
     return configparser.ConfigParser(allow_no_value=True)
 
 
+def _escape(value):  # for internal use
```
```suggestion
def _escape_percent(value):  # for internal use
```

src/west/configuration.py
```diff
@@ -112,13 +124,29 @@ def getint(self, option: str):
...
         return self._get(option, self.cp.getfloat)
 
+    def value(self, section: str, key: str):
+        # The interpolated value of an option, by section and key.
+        try:
+            return self.cp.get(section, key)
+        except configparser.InterpolationError as err:
+            raise self._malformed_value(section, key) from err
+
+    def _malformed_value(self, section: str, key: str) -> MalformedConfig:
```
marc-hb, line 134: Move this before first use?

I generally don't care much about the order in which functions are defined but here it looks weird for this to be used before AND after its definition; I feel like this is crossing a line. Before or after: pick one? :-)

marc-hb (commented):
Generally happy with the approach but to be sure this is sound I would very much like to see test coverage for more edge cases with triple and quadruple `%`, both from the command line and from the API.

src/west/configuration.py
```diff
@@ -33,6 +33,11 @@
 Configuration values from later configuration files override configuration
 from earlier ones. Local values have highest precedence, and system values
 lowest.
+
+Values are interpolated when they are read: a literal ``%`` must be
```
```suggestion
https://docs.python.org/3/library/configparser.html interpolates values when they are read
```
I meant to and forgot to ask for this in the last round, sorry.

Please phrase it better and the exact URL may not even be necessary but I think it's important to mention that we defer to this standard Python module so:
- There is a pointer to the exhaustive documentation in case anyone needs it.
- In case of issues with the configparser itself like #779 (now fixed in configparser), people realize that the `configparser` version / Python version matters.

src/west/configuration.py
```diff
@@ -33,6 +33,11 @@
...
+written as ``%%``, and ``%(other-option)s`` expands to the value of
+another option from the same section. Values written by ``west config``
+are escaped automatically, so only hand-written files need care.
```
marc-hb: I think we should add that interpolation features cannot be used from the command line, nor from the West APIs, correct? It requires manual config file edition or the Python configparser API.

Fine by me, it just needs to be stated.

tests/test_config.py
```diff
@@ -910,3 +910,89 @@ def sorted_list(other_args=''):
...
+    cmd(['config', 'pytest.v', value])
+
+    assert cmd('config pytest.v').strip() == value
+    escaped = value.replace('%', '%%')
+    assert f'v = {escaped}\n' in local_config().read_text()
+
+
+@pytest.mark.parametrize('value', PERCENT_VALUES)
+def test_round_trip_percent_api(value):
+    # Same through the API, which is what extension commands use.
```
```suggestion
    # Same through the west API, which is what extension commands use.
```
... as opposed to the configparser API. Admittedly low risk of confusion but this is just one extra word.

marc-hb (commented):
Just a few, small doc and comments changes. Everything else LGTM!

src/west/configuration.py
marc-hb: > interpolation features cannot be used from the command line, nor from the West APIs, correct? 

You could argue that this is very strictly speaking a regression...

But I think it's fine considering:
- The interpolation feature was never advertised.
- It's becoming even more of a stretch to imagine someone using interpolation AND scripting config changes at the same time!
- Unlike the abandoned #996, the interpolation feature is still available and usable. It only requires some code changes to keep using it.

marc-hb (approved):

kartben (approved):

PR #1002, opened 2026-09-08
test_config.py: rename and comment cryptic west_env

marc-hb:
Rename and clarify the somewhat cryptic `west_env` array.

Use uppercase because these are (very!) constant. Add highly relevant yet missing pointer to important conftest.py fixture. Add `_NAME` suffixes to remove the confusion with their value. Drop `west_` prefix in the name that did not add any information.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -20,13 +20,17 @@ GLOBAL = wconfig.ConfigFile.GLOBAL
 LOCAL = wconfig.ConfigFile.LOCAL
 ALL = wconfig.ConfigFile.ALL
 
-west_env = {
+# An auto fixture in conftest.py provides default values located in a
+# temporary, test-specific directory. The files are not created by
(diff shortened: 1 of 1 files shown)


marc-hb:
Spotted while reviewing #1000. Minor conflict with it, easy to solve. Either could go first.

pdgendt (commented):
Why not do the same for `west_flag` then?

pdgendt (approved):

kartben (approved):

pdgendt:
> Spotted while reviewing #1000. Minor conflict with it, easy to solve. Either could go first.

You even warned me 🫣 I was too fast, hotfix: https://github.com/zephyrproject-rtos/west/pull/1004

marc-hb:
I wasn't loud about this conflict because I assumed it would include a _git_ conflict blocking the second merge... It did not :-(

PR #1005, opened 2026-09-10
Fix incorrect f-string

yvesll:
Fixed a missing f prefix that left {project.abspath} uninterpolated in this error message.

While checking when this could fire, I noticed the surrounding `isinstance(project, ManifestProject)` branch looks unreachable with the current manifest-import code, not sure if it's worth deleting as dead code or better left alone.

### DIFF
--- a/src/west/app/project.py
+++ b/src/west/app/project.py
@@ -1616,7 +1616,7 @@ class Update(_ProjectCommand):
     def update_importer(self, project, path):
         if isinstance(project, ManifestProject):
             if not project.is_cloned():
-                self.die("manifest repository {project.abspath} was deleted")
+                self.die(f"manifest repository {project.abspath} was deleted")
         else:
             # There's no need to call self.project_is_active(),
             # because the Manifest API guarantees that 'groups' cannot


pdgendt (approved):

marc-hb (changes requested):
> While checking when this could fire, I noticed the surrounding isinstance(project, ManifestProject) branch looks unreachable with the current manifest-import code, not sure if it's worth deleting as dead code or better left alone.

I think it depends. If it's really obvious that this could never ever happens then maybe it could be removed. In all other cases it's just safer and really can't hurt. I haven't looked much but I would say just leave it for now. It will go away anyway if/when #327 gets done.

So you saw this bug only by reading the code then? Worth making even clearer in the commit message. You wrote an excellent commit message but it can only be found in the PR description, the actual git commit message of https://github.com/zephyrproject-rtos/west/pull/1005/changes/24d9f21fb5bfcc65154953130c68042323d9a615 is empty.

Approved except for the empty commit message!

PS @pdgendt: maybe we could use a PR check strips all Signed-off: and other tags and warn when the message is empty. I suspect it's a very common mistake for "regular" GitHub users. Maybe the main repo has that already?

yvesll:
> So you saw this bug only by reading the code then? Worth making even clearer in the commit message. You wrote an excellent commit but it can only be found in the PR description, the actual git commit message of [24d9f21](https://github.com/zephyrproject-rtos/west/commit/24d9f21fb5bfcc65154953130c68042323d9a615) is empty.
> 
> Approved except for the empty commit message!

Thanks for catching the empty message. It was scanned by VSCode's syntax highlighting feature when I read the code. Amended the commit with a proper message now.

marc-hb (approved):

PR #1006, opened 2026-09-21
manifest: add a WestCommands class for west-commands entries

### LINKED ISSUE #959: Improvement: WestCommandsType should be a class

Zzz210s:
Fixes #959.

`_west_commands_list()`, `_west_commands_maybe_delist()` and
`_west_commands_merge()` are replaced by a `WestCommands` class that also owns the
manifest directory of each entry. That directory used to live in a private attribute
of `Project` and `ManifestProject`, kept in sync with the list by hand.

Two choices worth explaining:

- **List subclass, not dict.** `west_commands` is documented and tested as a list of
  paths (`tests/test_manifest.py` asserts the import order), so a dict would change a
  public type. A list subclass leaves everything outside `manifest.py` untouched.
  `as_manifest_value()` still returns a plain `str`/`list`, so manifests serialize as
  before.
- **`merge()`, not `__iadd__`.** On a list `+=` means "extend, duplicates included".
  Filtering duplicates there would be surprising; `merge()` says what it does.

A falsey `west-commands:` value is treated as "no west commands" exactly as before
(the merge stays inside the truthiness check), with a regression test.


Zzz210s:
Fixed in 0234523 (force-pushed): the merge is back inside the truthiness check, so
`self: {west-commands: ''}` gives `west_commands == []` again.
`test_west_commands_falsey_value_in_self_is_ignored` covers it.

A falsey value on a project is unchanged from before this PR: the old
`_west_commands_list('')` also returned `['']`. Normalizing that would be a separate
change, and I'd rather not fold it in here.

On metadata for list mutations: an entry added with `append()` has no recorded
directory and resolves relative to the project root, which is the documented behavior
for entries without a directory. Removing an entry leaves an unreachable entry in
`manifest_dirs`, and lookups go through the list itself (`_ext_specs()` in
commands.py), so it cannot be picked up.

Zzz210s:
Heads-up for reviewers, and my mistake: I opened this without noticing that #966 is an open implementation of the same issue #959, so there are now two approaches in flight.

The scope difference, as I understand it:

* this PR is internal only: the parallel `_west_commands_manifest_dirs` dict moves into a `WestCommands` list subclass. No schema change, no change to the public `west_commands` attribute, no behavior change.
* #966 is user visible: per-entry `WestCommandsEntry(path, base_dir)` records, a new canonical `west_commands_entries` attribute, `west_commands` deprecated, plus a manifest schema change so a resolved manifest round-trips through `as_yaml()` without losing `base_dir`.

They overlap in the same code and are probably not worth landing both. I have no attachment to this one: if the direction is #966, I will close it and help there instead (rebase, extra tests). Could a maintainer say which way to go? I would rather not add review load with a second PR pointing in a different direction.

(Also force-pushed `cf03d9e`, which drops a constructor parameter that no caller used - thank you @Copilot for the nudge to look at the API surface.)

Issue #999, opened 2026-09-02
self.dbg() and self.inf() do not prepend the level to the message


steverae-work:
while debugging issues in https://github.com/zephyrproject-rtos/west/pull/994
( https://github.com/zephyrproject-rtos/west/pull/994#issuecomment-5502668402 )

it was discovered that self.dbg() does not prepend "[DEBUG]" to the message, so
the current code is performing:
``` self.dbg(f'    [debug] ...') ```
which is deemed as not acceptable by the reviewers
- see function at:  
src/west/commands.py:433:    def dbg(self, *args, level: Verbosity = Verbosity.DBG, end: str = '\n'):

additionally, self.inf() has the same issue,
the current code is performing:
``` self.inf(f'    [info] ...') ```
which is also deemed as not acceptable by the reviewers
- see function at:  
src/west/commands.py:445:    def inf(self, *args, colorize: bool = False, end: str = '\n'):

**Comment by marc-hb** [comment]
I did some git archeology and found that "it's always been inconsistent like this": ERR and WRN always had prefixes, INF and DBG never had... @mbolivar , any rationale?

It started with the "global" and now deprecated `/log.py` in the very first west commit c662ddf8a675.

And it stayed like this when migrating to per-command logging to fix #149, see commit b4d0b38c0b740c `commands: add output helpers to WestCommand`

But they have been on-and-off exceptions like when `manifest.py` started using the standard Python `logging` library as a shim in 71aaf57045c (in Jan 2020 with `logger = logging.getLogger('west.manifest')`), then reworked in 92c18ac55ac (Feb 2020) and 23db6e1f3cc7fb (2023) with various `logging.Formatter` configurations.

As of today's commit e560a33a46e3 and probably since 23db6e1f3cc7fb (2023), the current `logging.Formatter` gives `manifest.py` DEBUG and INFO `levelname` prefixes. Example:

```
west -vv status
west.manifest: DEBUG: "git status" exit code...
```

The same inconsistency could also be found in Zephyr extensions  where statements using an identical Python logging indirection (long story in #952) get DEBUG and INFO prefixes, while those using the `WestCommand.dbg()` API directly do not. Before zephyr commit [dcb1f490](https://github.com/zephyrproject-rtos/zephyr/commit/dcb1f490) (May 2026), this inconsistency could currently be observed in the output of this single command: the second log line (`zcmake.py`) had a DEBUG prefix (and more) whereas the first line (`class Sdk(WestCommand)`) had not:

[...]

PR #1007, opened 2026-09-21
commands: add DEBUG: and TRACE: log prefixes 

marc-hb:
See #999 and commit messages for details.


src/west/commands.py
```diff
@@ -115,6 +115,19 @@ class Verbosity(IntEnum):
...
+    Verbosity.WRN: "WARNING",
+    # No Verbosity.INF because inf() has been abused for regular output, see #999
+    Verbosity.DBG: "DEBUG",
+    Verbosity.DBG_MORE: "FINE",
+    Verbosity.DBG_EXTREME: "FINER",
+}
+
+# Append colon+space. Do not use a @property decorator for performance
+for lvl, prefix in _VERBOSITY_PREFIXES.items():
+    lvl.line_prefix = prefix + ': '  # type: ignore[attr-defined]
```
marc-hb: `Verbosity` is part of the public API so should probably be
```suggestion
    lvl._line_prefix = prefix + ': '  # type: ignore[attr-defined]
```

src/west/commands.py
```diff
@@ -115,6 +115,19 @@ class Verbosity(IntEnum):
     DBG_EXTREME = 6
 
 
+_VERBOSITY_PREFIXES = {
+    Verbosity.ERR: "ERROR",
+    Verbosity.WRN: "WARNING",
+    # No Verbosity.INF because inf() has been abused for regular output, see #999
+    Verbosity.DBG: "DEBUG",
+    Verbosity.DBG_MORE: "FINE",
+    Verbosity.DBG_EXTREME: "FINER",
```
pdgendt: `FINE` and `FINER` are odd to read, why not stick with `DEBUG` as the prefix?

marc-hb: I would like the different levels to be shown differently if possible. I took these names from java.util.logging because DEBUG_EXTREME is too long. I don't care much how they look. Any other idea?

marc-hb: TRACE is very popular, in fact FINE is a shorthand for "FINE trace". So this could be TRACE1 and TRACE2 or just TRACE for both. The latter hides the difference between the DBG_MORE and DBG_EXTREME levels but it's at least different from DEBUG so maybe this is a good trade-off.

marc-hb:
As reported privately by @pdgendt , this (makes no sense but) crashes: `dbg('quiet level', level=Verbosity.QUIET)`. I will fix.

Note "QUIET" is even more critical than ERROR :-)

src/west/commands.py
```diff
@@ -115,6 +115,23 @@ class Verbosity(IntEnum):
...
+    Verbosity.INF: "INFO",
+    Verbosity.DBG: "DEBUG",
+    Verbosity.DBG_MORE: "TRACE",
+    Verbosity.DBG_EXTREME: "TRACE",
+}
+
+# Append colon+space. Do not use a @property decorator for performance
+for lvl, prefix in _VERBOSITY_PREFIXES.items():
+    lvl._line_prefix = prefix + ': '  # type: ignore[attr-defined]
+Verbosity.QUIET._line_prefix = ''  # type: ignore[attr-defined]
```
pdgendt, line 132: I wouldn't attach the prefix to the enum machinery, but rather use a simple helper function. This also avoid the `type: ignore` stuff maybe?

```python
_VERBOSITY_PREFIXES = {
    Verbosity.ERR: 'ERROR: ',
    Verbosity.WRN: 'WARNING: ',
    Verbosity.DBG: 'DEBUG: ',
    Verbosity.DBG_MORE: 'TRACE: ',
    Verbosity.DBG_EXTREME: 'TRACE: ',
}

def _line_prefix(level: Verbosity) -> str:
    return _VERBOSITY_PREFIXES.get(level, '')
```

marc-hb: > Do not use a @property decorator for performance

You are not concerned about the cost?

> This also avoid the type: ignore stuff maybe?

Yes indeed.

pdgendt: > > Do not use a @Property decorator for performance
> 
> You are not concerned about the cost?

We're talking about Python, right? 😆 Joking aside, I think I favor readability in this case?

marc-hb: Claude found that I can just add a field to the Enum to avoid the "dynamic" property and all the mypy silencers. I wrongly assumed this was not possible because Enum were special or something... So I dropped all the `mypy` silencers without adding one extra function call on every log statement.

Good for you?

pdgendt (approved):

dkalowsk (approved):

